PR #513, opened 2016-05-04
Add initializer parameter to the factory

drungrin:
The factory should have a initializer parameter to have a class initializer-like usage.


fealebenpae:
@drungrin I'm closing this as it seems to be stale now. Please feel free to open it again at a later date.

PR #515, opened 2016-05-05
Better rhs handling. More Methods.

drungrin:
- Implemented default behavior of SingleOrDefault and FirstOrDefault. Still need to implement DefaultIfEmpty to add custom behavior
- Better rhs handling on BinaryExpression

### DIFF
--- a/Realm.Shared/linq/RealmResultsVisitor.cs
+++ b/Realm.Shared/linq/RealmResultsVisitor.cs
@@ -7,6 +7,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 using System.Runtime.CompilerServices;
 
 using LazyMethod = System.Lazy<System.Reflection.MethodInfo>;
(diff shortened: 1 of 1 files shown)


Realm.Shared/linq/RealmResultsVisitor.cs
```diff
@@ -254,20 +261,13 @@ protected void VisitCombination(BinaryExpression b,  Action<QueryHandle> combine
 
         internal static object ExtractConstantValue(Expression expr)
         {
-            var constant = expr as ConstantExpression;
-            if (constant != null)
-            {
-                return constant.Value;
-            }
+            var objectMember = Expression.Convert(expr, typeof(object));
```
kristiandupont: Instead of replacing this completely, could we use `LambdaExpression.Compile` as a third option if the first two conditions fail, just so we don't have to rely on it when the faster options are available?

kristiandupont:
Hi @drungrin, apologies for the delay. We were busy launching, so that captured most of our mental resources for a bit :-)

If you could sign our CLA, that will allow us to accept PR's from you: https://docs.google.com/a/realm.io/forms/d/1bVp-Wp5nmNFz9Nx-ngTmYBVWVdwTyKj4T0WtfVm0Ozs/viewform?fbzx=4154977190905366979

I've made a comment on one of the changes. Also, we will need some unit tests before we can merge this. Unfortunately I can't push to your branch. Can I ask you to add some?

And finally, thank you for making this effort, it's greatly appreciated!

drungrin:
@kristiandupont There is something weird going on, maybe because of the migration.

I created another pull request, #533 

We can continue on it, and I had put the Compile as a *fourth* option.

PR #531, opened 2016-05-11
Change ios package in Xamarin Component to unified

fealebenpae:
Otherwise the component shows up as using the legacy 32bit API on Xamarin.iOS.


kristiandupont:
:+1:

PR #547, opened 2016-05-17
Add a INotifyCollectionChanged adapter around RealmResults

fealebenpae:


Realm.Shared/extensions/RealmResultsCollectionChanged.cs
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        /// <summary>
+        /// Wraps a <see cref="RealmResults{T}" /> in an implementation of <see cref="INotifyCollectionChanged" /> so that it may be used in MVVM databinding.
+        /// </summary>
+        /// <param name="results">The <see cref="RealmResults{T}"/ > to observe for changes.</param>
+        /// <param name="errorCallback">An error callback that will be invoked if the observing thread raises an error.</param>
+        /// <param name="coalesceMultipleChangesIntoReset">
+        /// When a lot of items have been added or removed at once it is more efficient to raise <see cref="INotifyCollectionChanged.CollectionChanged" /> once
+        /// with <see cref="NotifyCollectionChangedAction.Reset" /> instead of multiple times for every single change. Pass <c>true</c> to opt-in to this behavior.
```
AndyDentFree: Why is this called `Reset`? Is that a convention for this kind of messaging? It doesn't make sense to me but I'm far from being a _Reactive Thinker_

fealebenpae: I'm not sure what you mean - `NotifyCollectionChangedAction` is a framework enum.

AndyDentFree: The naming as **Reset** sounds weird, found a reference which explains it _"The content of the collection changed dramatically"._ and I accept it's MS weirdness rather than yours, so no problem.

Realm.Shared/extensions/RealmResultsCollectionChanged.cs
```diff
@@ -0,0 +1,150 @@
...
+            readonly IDisposable _token;
+            readonly Action<Exception> _errorCallback;
+            readonly bool _coalesceMultipleChangesIntoReset;
+
+            private bool _suspendNotifications;
+
+            internal Adapter(RealmResults<T> results, Action<Exception> errorCallback, bool coalesceMultipleChangesIntoReset) : base(results)
+            {
+                _results = results;
+                _errorCallback = errorCallback;
```
AndyDentFree: Is `errorCallback` able to be null? There are no null checks later. Could we default it to an empty lambda?

Realm.Shared/extensions/RealmResultsCollectionChanged.cs
```diff
@@ -0,0 +1,150 @@
...
+                else if (change != null)
+                {
+                    _suspendNotifications = _coalesceMultipleChangesIntoReset && change.InsertedIndices.Length + change.DeletedIndices.Length > 1;
+
+                    foreach (var removed in change.DeletedIndices.Reverse())
+                    {
+                        // the row has been deleted, we need to recreate the adapter's contents
+                        if (!this[removed].RowHandle.IsAttached)
+                        {
+                            Recreate();
```
AndyDentFree: Isn't this a nested loop recreating the lot for each change?

Realm.Shared/extensions/RealmResultsCollectionChanged.cs
fealebenpae: That's a good point - I'll add a null check.

Realm.Shared/extensions/RealmResultsCollectionChanged.cs
fealebenpae: Only if the current batch contains an object that has been deleted from the Realm. Such an object is invalid so we can't remove it from the observable collection because doing so would raise `CollectionChanged` with the `Remove` action and pass an invalid object to subscribers. So it's much simpler to just rebuild the observable collection and raise `CollectionChanged` with `Reset` instead.

AndyDentFree: My point was that we are calling `Recreate` for **each** of the deleted items? Shouldn't it only be called once per `OnChange` invocation?

fealebenpae: Well, there is a `return` statement right after `Recreate`.

AndyDentFree: Noted, (now) but that raises a question as to **why?** The structure of the rest of the method suggests this is a rare condition so could we please have a comment to that effect. What happens if there are more than one such deleted object? Is this code assuming there can only be one anomaly? Or, does it expect to be called back again?

fealebenpae: If there is at least one deleted row we need to recreate the collection. There could be twenty deleted rows in the changeset, but it's enough for us to detect the first one to know we have to throw away the whole thing.

nirinchev:
While I can see the reasoning behind resetting the collection on remove, it appears to be kind of restricting. In a hypothetical scenario, where I have bound a UITableView to the observable collection, to remove a row with a nice animation, only the index of the deleted item would suffice. With the current implementation, I would imagine the removing will be instantaneous at best and flickering at worst.

Obviously this will break the expectations of classic CollectionChanged subscribers, so it could be an opt-in behaviour, similar to coalesceMultipleChangesIntoReset.

fealebenpae:
@nirinchev oddly enough the Xamarin.Forms ListView does not (noticably) repaint when a `CollectionChanged` event is raised with a `Reset` action.

Right now the Xamarin.Forms ListView is the only consumer of `INotifyCollectionChanged` that is likely to use this adapter, so its functionality is sort of tailor-fit. However, in the `UITableView` scenario you should use the `RealmResults<T>.SubscribeForNotifications` API directly which gives you a changeset with the indices that have been added and/or removed which is exactly what `UITableView` needs.

I'm wary of making `ToNotifyCollectionChanged` too configurable. I believe `SubscribeForNotifications` is what developers should use in scenarios outside Xamarin.Forms.

nirinchev:
Makes sense. I was thinking about MVVMLight's [ObservableTableViewController](https://mvvmlight.codeplex.com/SourceControl/latest#GalaSoft.MvvmLight/GalaSoft.MvvmLight.Platform%20(iOS)/Helpers/ObservableTableViewController.cs) - if Realm's adapter were to emit delete events, consumers of the MVVMLight toolkit would be able to bind directly to it. As for `SubscribeForNotifications`, it is aligned with the classic Obj-C way of doing things, but I don't think it fits very well with MVVM where the iOS ViewControllers are dumbed down to views (although I don't know if that's the cool way of doing Xamarin.iOS/Android development).

Anyway, if you're primarily targeting Forms currently, it doesn't make a lot of sense to go out of your way and pollute the API to support a scenario of questionable business value, just wanted to give you some food for thought for future developments :)

fealebenpae:
@nirinchev thanks, your input is much appreciated.

Personally I hope that the base API provides the tools others would need to come up with specialized adapters for various scenarios, such as when targetting MVVMLight or RxUI.

kristiandupont:
:+1:

Issue #541, opened 2016-05-13
Xamarin android Crash


jpsim:
Cross-posted from realm/realm-cocoa#3574:

> Hi Team !

> I install Realm to project via nuget to test this.
> When it run to
> 
> var freshRealm = Realm.GetInstance(); --> crash here
> Below is the exception:

```
{System.NullReferenceException: Object reference not set to an instance of an object
at Realms.Realm.CreateRealmObjectMetadata (System.Type realmObjectType) [0x00019] in :0 
at System.Linq.Enumerable.ToDictionaryType,Type,Metadata [0x0002f] in /Users/builder/data/lanes/2185/53fce373/source/mono/mcs/class/System.Core/System.Linq/Enumerable.cs:2924 
at System.Linq.Enumerable.ToDictionaryType,Type,Metadata [0x00000] in /Users/builder/data/lanes/2185/53fce373/source/mono/mcs/class/System.Core/System.Linq/Enumerable.cs:2911 
at Realms.Realm..ctor (Realms.SharedRealmHandle sharedRealmHandle, Realms.RealmConfiguration config) [0x00037] in :0 
at Realms.Realm.GetInstance (Realms.RealmConfiguration config) [0x00171] in :0 
at SaveCall.Repositories.RealmRepository.GetData () [0x0000e] in D:\Projects\SaveCall\SaveCall\Repositories\RealmRepository.cs:30 }
```

fealebenpae:
Hi @jerryno6, it would really help us diagnose this crash if you could give us some more information such as the Xamarin.Android version you are running and the version of Xamarin Studio or Visual Studio you're running.

Also, it would be great if you could post a gist with your `RealmObject` classes, or better yet if you could prepare a small repro project that demonstrates this issue.

Thanks!

YordanYanakiev:
I have just the same issue.
While in **debug mode everything is working fine, but once in release mode - it crashing just there** on GetInstance( Realms.RealmConfiguration.DefaultConfiguration );

AndyDentFree:
@YordanYanakiev please give us more information or sample code. I just built a tiny Android sample from scratch to re-check the NuGet and it works fine in Release mode. 

```csharp
using Android.App;
using Android.Widget;
using Android.OS;

using Realms;

namespace TestRealmRelease0
{

    class RT : RealmObject {
        public string Name {get;set;}
        public int Age { get; set; }
    };


[...]

PR #550, opened 2016-05-19
Ad/bug/getinstance crash #541

AndyDentFree:
Throws an exception where it can be seen by user rather than in Debug Console.


fealebenpae:
:+1:

PR #551, opened 2016-05-20
Ensure thread safety in the Android WeakRealmNotifier

fealebenpae:
WeakRealmNotifier's move assignment operator could be invoked on another thread so we need to make sure to remove ourselves from the looper always on the looper thread.


kristiandupont:
:+1:

PR #556, opened 2016-05-23
Remove REALM_PLATFORM_ANDROID define

fealebenpae:
and replace it with REALM_ANDROID, which is defined by core in realm/util/features.h


wrappers/src/object-store/src/impl/external_commit_helper.hpp
```diff
@@ -23,7 +23,7 @@
 
 #if REALM_PLATFORM_APPLE
```
AndyDentFree: For consistency we should have a similar naming pattern for other platforms.

fealebenpae: Well, there's already `REALM_IOS` and `REALM_TVOS` and suchlike. `REALM_PLATFORM_APPLE` is like an umbrella definition, from what I can tell. I imagine that `REALM_PLATFORM_LINUX` would be the equivalent, except core treats it as the default so there isn't a separate define for it 😁

Please open an issue on core if you're interested so we can discuss this with everyone.

kristiandupont:
:+1:

PR #557, opened 2016-05-23
WeakRealmNotifier should use POSIX write correctly

fealebenpae:
We need to call write(2) with the address of the buffer, not the buffer itself, d'oh


kristiandupont:
:+1: :+1: :+1:

Issue #510, opened 2016-05-03
Binary data properties


kristiandupont:
The obvious choice is to go with `byte[]`. However, those are mutable and there is no way to get around that, so we will probably go with [System.Data.Linq.Binary](https://msdn.microsoft.com/en-us/library/system.data.linq.binary(v=vs.110).aspx) instead. It's unfortunate because it requires referencing a non-default assembly but at least it's a framework assembly.

flagbug:
I don't think `System.Data.Linq.Binary` is available on any other platform than the Full .NET Framework and Windows Phone Silverlight, so using it is problematic

kristiandupont:
@flagbug you are right (http://stackoverflow.com/a/15753777/1417293). Back to the drawing board then. Any suggestions?

flagbug:
You could take something out of [System.Collections.Immutable](https://www.nuget.org/packages/System.Collections.Immutable/1.2.0-rc2-24027) maybe? 

Help me understand what exactly the problem with the mutability of `byte[]` is?

kristiandupont:
I think we might end up going with pure `byte[]` after all. The problem is that we cannot detect mutations (`myObject.BinData[0] = 0;`, etc.). So contrary to what you would expect, such changes will not be persisted.

flagbug:
>  So contrary to what you would expect, such changes will not be persisted.

Generally in .NET you wouldn't expect change notifications for collections, unless they implement [INotifyCollectionChanged](https://msdn.microsoft.com/en-us/library/system.collections.specialized.inotifycollectionchanged(v=vs.110).aspx), so I don't think that would be the case, but maybe that's just me :smile:

fealebenpae:
@flagbug The issue I'm worried about is that changing the byte array you get from a Realm object will not change the same value in the database. `myObject.BinData[0] = 0` might seem like it's a change that will be persisted by Realm, but it actually won't be. That's why we wanted to go with something immutable.

flagbug:
@kristiandupont @fealebenpae So how about `IReadOnlyList<byte>`?

fealebenpae:
Implemented in #567

PR #567, opened 2016-05-27
Add support for byte[] properties

fealebenpae:
Implements #510

### DIFF
--- a/Realm.Shared/linq/RealmResultsVisitor.cs
+++ b/Realm.Shared/linq/RealmResultsVisitor.cs
@@ -366,6 +366,13 @@ namespace Realms
             else if (valueType == typeof(byte[]))
             {
                 var buffer = (byte[])value;
+                if (buffer.Length == 0)
+                {
+                    // see RealmObject.SetByteArrayValue
+                    NativeQuery.binary_equal(queryHandle, columnIndex, (IntPtr)0x1, IntPtr.Zero);
+                    return;
+                }
+
                 unsafe
                 {
                     fixed (byte* bufferPtr = (byte[])value)
@@ -401,6 +408,13 @@ namespace Realms
             else if (valueType == typeof(byte[]))
             {
                 var buffer = (byte[])value;
+                if (buffer.Length == 0)
+                {
+                    // see RealmObject.SetByteArrayValue
+                    NativeQuery.binary_not_equal(queryHandle, columnIndex, (IntPtr)0x1, IntPtr.Zero);
+                    return;
+                }
+
                 unsafe
                 {
                     fixed (byte* bufferPtr = (byte[])value)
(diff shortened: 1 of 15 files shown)


Tests/IntegrationTests.Shared/AccessTests.cs
```diff
@@ -71,5 +95,19 @@ public void SetValueAndReplaceWithNull(string propertyName, object propertyValue
 
             Assert.That(TestHelpers.GetPropertyValue(ato, propertyName), Is.EqualTo(null));
         }
+
+        public IEnumerable<object[]> SetAndReplaceWithNullCases()
+        {
+            yield return new object[] { "NullableCharProperty", '0' };
```
AndyDentFree: Looks like copy and paste error - shouldn't these all be null?

Tests/IntegrationTests.Shared/AccessTests.cs
```diff
@@ -41,15 +44,36 @@ public void TearDown()
...
+        {
+            yield return new object[] { "CharProperty", '0' };
+            yield return new object[] { "ByteProperty", (byte)100 };
+            yield return new object[] { "Int16Property", (short)100 };
+            yield return new object[] { "Int32Property", 100 };
+            yield return new object[] { "Int64Property", 100L };
+            yield return new object[] { "SingleProperty", 123.123f };
+            yield return new object[] { "DoubleProperty", 123.123 };
+            yield return new object[] { "BooleanProperty", true };
+            yield return new object[] { "ByteArrayProperty", new byte[] { 0xde, 0xad, 0xbe, 0xef } };
```
AndyDentFree: Also would like a case to clarify that an array of length zero is not null.

Tests/IntegrationTests.Shared/AccessTests.cs
fealebenpae: This test first sets the property to a non-null value and asserts the getter returns the value, and then sets the property to null and asserts that the getter returns null.

AndyDentFree: OK, yup noted it now.

AndyDentFree:
👍

Tests/IntegrationTests.Shared/AccessTests.cs
fealebenpae: Good catch, that needed additional work - see latest commit.

Realm.Shared/linq/RealmResultsVisitor.cs
```diff
@@ -363,6 +363,24 @@ private static void AddQueryEqual(QueryHandle queryHandle, string columnName, ob
                 NativeQuery.double_equal((QueryHandle)queryHandle, columnIndex, (double)value);
             else if (valueType == typeof(DateTimeOffset))
                 NativeQuery.datetime_seconds_equal(queryHandle, columnIndex, ((DateTimeOffset)value).ToUnixTimeSeconds());
+            else if (valueType == typeof(byte[]))
+            {
+                var buffer = (byte[])value;
+                if (buffer.Length == 0)
+                {
+                    // see RealmObject.SetByteArrayValue
+                    NativeQuery.binary_equal(queryHandle, columnIndex, (IntPtr)0x1, IntPtr.Zero);
```
AndyDentFree: I'm bothered by this `0x1` magic value being repeated as a literal. Does it have significance and, if so, could it be a constant?

fealebenpae: No, it just has to be anything but zero, as long as the length is zero. Seeing as how it's a dummy value I wouldn't bother with a constant.

AndyDentFree:
👍 with the added check for empty byte arrays!

kristiandupont:
:+1:

PR #568, opened 2016-05-27
Use more relative paths in NuGet targets

### LINKED ISSUE #561: nuget.config is not obeyed
#### Goal

Get future Realm dotnet packages to obey the directions in a nuget.config file

#### Expected Results

I have a nuget.config file which places any nuget packages in the same folder hierarchy . See below. The expected result is that the Realm dotnet package would look for the  RealmWeaver.Fody.dll in the following directory Root\packages\Realm0.74.1\tools

Root Folder  contains nuget.config file
   -> packages - containing all the nuget packages for all the child solutions 
   -> Child folder 1  contains child1.sln
   -> Child folder 2 contains child2.sln

#### Actual Results

What actually happened is that the RealmWeaver.Fody.dll is expected to be in a child directory of the solution e.g. Root\ChildSolution1\packages\Realm0.74.1\tools

Add a nuget.config file which specifies where the packges should be copied to e.g

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
	<config>     
		<add key="repositorypath" value="..\packages" /> 
	</config>
</configuration>
```

#### Code Sample

See nuget.config file above

```

Realm version(s): 0.74.1

Xamarin/Visual Studio version: VS2015 Update 2

Which operating system version and device: Windows 7

fealebenpae:
Fixes #561 

/cc @AndyDentFree

### DIFF
--- a/NuGet/NuGet.Library/Realm.targets
+++ b/NuGet/NuGet.Library/Realm.targets
@@ -3,24 +3,24 @@
 <!-- Puts a copy of RealmWeaver.Fody.dll where it will be found by Fody -->
   <Target Name="CopyRealmWeaver" BeforeTargets="FodyTarget">
     <Message Text="CopyRealmWeaver" />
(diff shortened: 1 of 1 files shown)


AndyDentFree:
👍

Issue #461, opened 2016-04-08
Support new DateTime format


AndyDentFree:
Mainly means removing code truncating to seconds and expanding unit tests to ensure cover the nanosecond resolution.

fealebenpae:
Implemented in #572

PR #572, opened 2016-05-30
Ad/feature/core  1.0 new timestamp #461

AndyDentFree:
This is based on my diariised previous changes to cope with the new timestamp and minor string changes to move to core 0.99.0 onwards.

I have not reviewed it against more recent objectstore changes - we can do that as a separate PR or include in this, but I thought it important to move forward.


fealebenpae:
:+1:

Issue #253, opened 2015-12-03
Add PropertyChanged function to our weaver


AndyDentFree:
Include the features of the popular [PropertyChanged](https://github.com/Fody/PropertyChanged) weaver as an option in our weaving. Weaving our properties will directly conflict with this weaver and the feature it provides is very useful for GUI development and aligned with our philosophy of clean, simple code.

PR #573, opened 2016-05-30
Add automatic support for INotifyPropertyChanged via weaver.

jdbrock:
Hi folks,

I've added basic support for INotifyPropertyChanged, having checked out and experimented with Realm, I was hoping for something I could drop in next to my UI code and see live updates. This seems to do the trick.

I had initially tried to use PropertyChanged.Fody in conjunction with Realm, but once an object became managed, the early return short-circuited raising the property changed at the end of the setter.

This seems to be related to the following issue: https://github.com/realm/realm-dotnet/issues/253

Cheers,
Joe

### DIFF
--- a/RealmWeaver/ModuleWeaver.cs
+++ b/RealmWeaver/ModuleWeaver.cs
@@ -40,16 +40,21 @@ public class ModuleWeaver
     // An instance of Mono.Cecil.ModuleDefinition for processing
     public ModuleDefinition ModuleDefinition { get; set; }
 
-    TypeSystem typeSystem;
+    private AssemblyDefinition RealmAssembly;
+    private TypeDefinition RealmObject;
+    private MethodReference RealmObjectIsManagedGetter;
 
-    AssemblyDefinition RealmAssembly;
-    TypeDefinition RealmObject;
-    MethodReference realmObjectIsManagedGetter;
+    private AssemblyDefinition CorLib;
+    private TypeDefinition System_Object;
+    private TypeDefinition System_Boolean;
+    private TypeReference System_String;
+    private TypeReference System_Type;
 
-    AssemblyDefinition CorLib;
-    TypeDefinition System_Object;
-    TypeReference System_String;
-    TypeReference System_Type;
+    private MethodReference PropChangedEventArgsConstructor;
+    private MethodReference PropChangedEventHandlerInvokeReference;
+    private TypeReference PropChangedEventHandlerReference;
+
+    private MethodReference PropChangedDoNotNotifyAttributeConstructorDefinition;
 
     // Init logging delegates to make testing easier
     public ModuleWeaver()
@@ -84,20 +89,6 @@ public class ModuleWeaver
         return matches.Count() == 1;
     }
 
-
-    internal MethodReference MethodNamed(TypeDefinition assemblyType, string name)
-    {
-        try
-        {
-            return ModuleDefinition.ImportReference(assemblyType.Methods.First(x => x.Name == name));
-        }
-        catch (InvalidOperationException e)
-        {
-            throw new InvalidOperationException("Trying to find method '" + name + "' failed.", e);
-        }
-    }
-
-
     public void Execute()
     {
         // UNCOMMENT THIS DEBUGGER LAUNCH TO BE ABLE TO RUN A SEPARATE VS INSTANCE TO DEBUG WEAVING WHILST BUILDING
@@ -112,12 +103,10 @@ public class ModuleWeaver
             }
         });
 
-        typeSystem = ModuleDefinition.TypeSystem;
-
         RealmAssembly = ModuleDefinition.AssemblyResolver.Resolve("Realm");  // Note that the assembly is Realm but the namespace Realms with the s
 
         RealmObject = RealmAssembly.MainModule.GetTypes().First(x => x.Name == "RealmObject");
-        realmObjectIsManagedGetter = ModuleDefinition.ImportReference(RealmObject.Properties.Single(x => x.Name == "IsManaged").GetMethod);
+        RealmObjectIsManagedGetter = ModuleDefinition.ImportReference(RealmObject.Properties.Single(x => x.Name == "IsManaged").GetMethod);
         
         var typeTable = new Dictionary<string, string>()
         {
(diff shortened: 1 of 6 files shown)


kristiandupont:
Excellent!

A couple of things before we can merge it.

First of all we need a bit of test coverage. Are you able to run the WeaverTests?

Secondly, this implementation would have a performance impact on every property setter which is a very performance sensitive bit of code. We would like it changed so that rather than having the `RealmObject` class implement `INotifyPropertyChanged`, you would implement it on your model class. The weaver should then check to see whether this is the case for a given class and only then add the event triggering. This is probably best done by moving the code from your `RaisePropertyChanged` method directly into the property setter. Would you be able to make this change?

Thank you for contributing, this is a very nice addition!

drungrin:
You should use the weaver to add the INotifyPropertyChanged. I don’t think this is realm related.

 

It is a info that can be in the wiki / readme to help the newcomer, like “You can use the weaver to add the INotifyPropertyChanged, just add the attribute [ImplementPropertyChanged]

 after installing the nuget PropertyChanged.Fody”

 

https://github.com/Fody/PropertyChanged

 

De: Kristian Dupont <notifications@github.com>
Responder para: realm/realm-dotnet <reply@reply.github.com>
Data: terça-feira, 31 de maio de 2016 08:17
Para: realm/realm-dotnet <realm-dotnet@noreply.github.com>
Assunto: Re: [realm/realm-dotnet] Add automatic support for INotifyPropertyChanged via weaver. (#573)

 

Excellent!

A couple of things before we can merge it.

First of all we need a bit of test coverage. Are you able to run the WeaverTests?

Secondly, this implementation would have a performance impact on every property setter which is a very performance sensitive bit of code. We would like it changed so that rather than having the RealmObject class implement INotifyPropertyChanged, you would implement it on your model class. The weaver should then check to see whether this is the case for a given class and only then add the event triggering. This is probably best done by moving the code from your RaisePropertyChanged method directly into the property setter. Would you be able to make this change?

Thank you for contributing, this is a very nice addition!

kristiandupont:
@drungrin as @jdbrock found out, the two weavers are in fact incompatible, unfortunately. And you are right, we should probably add something about this in the documentation.

drungrin:
I answered that from e-mail client :D and didn't read his post. 

Maybe you can use an attribute to control when you need that, and a [IgnorePropertyChanged] to methods to ignore, to aid on your concern about the performance impact.

kristiandupont:
I suppose that would be an alternative, yes. But I would rather go with the option to check for the interface to keep the API as small and simple as possible.

drungrin:
Nice. I didn't know this about the Weaver.

I thought that it would annoy me with the IDE error about the interface not implemented, but good to know about this 👍

jdbrock:
That all sounds great, thanks for the feedback both!

@kristiandupont Happy to make the change you outlined - I agree that there's a bit of a performance penalty. Should have a chance to push this in the next evening or two. Will try to slip in some test coverage as well.

jdbrock:
I've been playing around with this for a few hours. One of the issues I'm facing (with the proposed approach of weaving in the PropertyChanged code simply by detecting the interface) is that in my example project, I have PropertyChanged.Fody in there as well (to be used on classes other than RealmObjects, e.g. some view models). 

Unfortunately, PropertyChanged.Fody *also* detects the interface implementation and tries to do the weaving automatically itself. I was banging my head off the desk for a while before I realised that was why my IL wasn't valid!

I suppose this leaves the following options:

- Don't support PropertyChanged.Fody being used in the solution (not ideal).
- Have a custom attribute which adds PropertyChanged support, instead of detecting the INotifyPropertyChanged interface, e.g. [RealmPropertyChanged].
- All RealmObjects implement INotifyPropertyChanged aware by default (not ideal). Note that the RaisePropertyChanged method is likely to be inlined by the jitter (and can be forced to do so with an aggressive inlining attribute) so the performance penalty should be reduced to loading the event field and doing a null check.
- Mystery fourth alternative option?

Any feedback would be much appreciated!

AndyDentFree:
Weaver detection and ordering was [proposed and dismissed](https://github.com/Fody/Fody/issues/74)  back in 2013 :-(

I was thinking we could both check for `INotifyPropertyChanged` and fallback to a `RealmPropertyChanged`

jdbrock:
@AndyDentFree That's a shame - would've been very useful here!

Good news though - I've managed to re-implement this in a way where we're unaffected by PropertyChanged.Fody (which I've now pushed to the branch).

For starters, I've removed `INotifyPropertyChanged` from `RealmObject` and I'm now weaving the event raising code directly in the setter when `INotifyPropertyChanged` is implemented on the user class (as suggested above).

In order to work around `<PropertyChanged />` being specified _before us_ in `FodyWeavers.xml`, I'm clearing instructions from the setter of the auto-property.

In order to work around `<PropertyChanged />` being specified _after us_ in `FodyWeavers.xml`, I'm adding a `[DoNotNotifyAttribute]` onto the woven property. The result being that we end up with exactly the same setter body regardless of whether PropertyChanged is being used.

I haven't added a RealmPropertyChanged as it looks like we may not need it in the end - let me know if you'd like this though.

Test coverage is still to be done, I'm hoping I'll get a little more time this week to hook that up. Thought it would be best to push this beforehand for a final review.

fealebenpae:
This looks good to me. Thank you for the contribution, @jdbrock!

Just a couple of cosmetic issues: `RealmObject.cs` needs to be reverted, IMO, as with the latest version of this PR we shouldn't need to touch it; also, some indentation looks wrong in `ModuleWeaver.cs` (you can see that in the diff here in GitHub).

jdbrock:
Good catch, thanks!

RealmObject reverted and ModuleWeaver whitespace fixed (I blame Xamarin Studio.)

fealebenpae:
Could you please rebase this against latest master, as there seem to be merge conflicts.

jdbrock:
@fealebenpae: Done!

jdbrock:
I've added the following test to `WeaverTests`:

* Check that a `RealmObject` which implements `INotifyPropertyChanged` is raising `PropertyChanged` events.

I'm also running all tests under the following conditions:

* `PropertyChanged.Fody` is not loaded.
* `PropertyChanged.Fody` is loaded, `RealmWeaver.Fody` is ran first, then `PropertyChanged.Fody`. 
* `PropertyChanged.Fody` is loaded, `PropertyChanged.Fody` is ran first, then `RealmWeaver.Fody`.

I've also fixed up a couple of existing tests which were failing out of the box with invalid test case parameters.

The tests are running successfully here. Although I've noticed that your CI build is failing, let me know if there's anything I can do to help there!

kristiandupont:
The build error seems to stem from the reference to the `System.ObjectModel` assembly which is not found when building for iOS. If I remove the reference (ModuleWeaver.cs:171) and find `PropertyChangedEvent` and `PropertyChangedEventArgs` in `System`, it seems to work for both platforms. Could you try that?

jdbrock:
Good suggestion. Although it looks like loading only from System seems to fail when weaving a PCL (Profile78) on my sample app (with the PropertyChangedEventArgs type not being found).

I've made a TryResolveAssembly(...) method which should hopefully resolve the issue on iOS/Android and keep it functional for PCLs.

kristiandupont:
Looks good, except now will it not simply fail if the assembly isn't found? It doesn't seem to fall back to searching for `PropertyChangedEvent` et al in `System`?

jdbrock:
It should failover properly for `PropertyChangedEventArgs` but I did miss `PropertyChangedEventHandler`. Will fix that and rebase against master shortly.

Sorry that it's going back and forth so much, I had managed to convince myself that this was a straightforward change :-)

AndyDentFree:
>I had managed to convince myself that this was a straightforward change 

(Andy falls off ball laughing)

We **never** do that of course!

(thinking of my long weekend spent demolishing and rebuilding a ceiling, with my wife's comments on the 3rd day "you didn't say it was going to be this much work")

Seriously, your professionalism and attention to detail on the weaver conflict has been very much appreciated!

jdbrock:
Haha - thanks, I appreciate that Andy! Hope your ceiling is on the mend. I was going to make a pun, but I thought it might go over everyone's heads ;-)

kristiandupont:
:+1: This looks good to me. Solid work, @jdbrock!

PR #596, opened 2016-06-07
Standalone objects should not crash on equals

### LINKED ISSUE #587: RealmObject.Equals throws NullReferenceException
#### Goal

For some reason, I need in generic method to check objects equality thanks to EqualityComparer<T>.Default.Equals(obj1, obj2).

#### Expected Results

I expected EqualityComparer<T>.Default.Equals to return true or false.

#### Actual Results

It throws a NullReferenceException. obj1 and obj2 are not null.

System.NullReferenceException: Object reference not set to an instance of an object
  at Realms.RealmObject.Equals (System.Object obj) [0x00045]
  at System.Collections.Generic.ObjectEqualityComparer`1[T].Equals (System.Collections.Generic.T x, System.Collections.Generic.T y) [0x00016]
  at MyProject.Core.BaseNotifyPropertyChanged.SetField[T] (MyTennisPlayer.Core.T& field, MyProject.Core.T value, System.String propertyName) [0x0000d] 

Create a classe that inherit from RealmObject.
Create two objects of that class, without adding them to a Realm or loading them from a Realm.
Compare equality.

```csharp

public class MyObject : RealmObject
{
	[ObjectId]
	public int Id { get; set; }
}

...

var obj1 = new MyObject () {
	Id = 1,
	Name = "My object 1"
};
var obj2 = new MyObject () {
	Id = 2,
	Name = "My object 2"
};

bool areEquals = EqualityComparer<MyObject>.Default.Equals(obj1, obj2);

```
#### Code Sample

```csharp

protected bool SetField<T>(ref T field, T value, [CallerMemberName]string propertyName = null)
{
    if (EqualityComparer<T>.Default.Equals(field, value)) return false;
    field = value;
    OnPropertyChanged(propertyName);
    return true;
}

```

Realm version(s): 0.74.1

Xamarin/Visual Studio version: Xamarin Studio 5.10

Which operating system version and device: iPhone Simulator / iOS 9.2

fealebenpae:
Fix #587

### DIFF
--- a/Realm.Shared/RealmObject.cs
+++ b/Realm.Shared/RealmObject.cs
@@ -785,6 +785,10 @@ namespace Realms
             if (this.GetType() != obj.GetType())
                 return false;
 
+            // standalone objects cannot participate in the same store check
+            if (!IsManaged)
+                return false;
+
             // Return true if the fields match. 
             // Note that the base class is not invoked because it is 
             // System.Object, which defines Equals as reference equality. 
--- a/Tests/IntegrationTests.Shared/StandAloneObjectTests.cs
+++ b/Tests/IntegrationTests.Shared/StandAloneObjectTests.cs
@@ -77,5 +77,13 @@ namespace IntegrationTests.Shared
                 Assert.That(p.IsInteresting);
             }
         }
+
+        [Test]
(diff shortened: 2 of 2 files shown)


kristiandupont:
:+1:

PR #597, opened 2016-06-07
Make wrappers build on Windows again

kristiandupont:
This allows us to build wrappers with VS again. Note that the external commit helper is broken, but we're not using it on Windows anywhere so it doesn't break anything at the moment.


fealebenpae:
👍

PR #605, opened 2016-06-08
Fix woven code so that the Xamarin linker likes it

### LINKED ISSUE #577: Fix Fody issues with Linking
When you turn on the misnamed _linking_ (actually more like _linker stripping_ ) feature of Xamarin Studio, Fody seems to have problems. In at least one case we have found it only working for **nested** classes.

fealebenpae:
The Xamarin linker would mess up the parameter of the `WovenAttribute` constructor because the woven helper type had a namespace.

Fixes #577

### DIFF
--- a/Realm.Shared/Attributes.cs
+++ b/Realm.Shared/Attributes.cs
@@ -42,9 +42,9 @@ namespace Realms
         }
     }
 
-    internal class WovenAttribute : Attribute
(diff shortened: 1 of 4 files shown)


kristiandupont:
:+1:

AndyDentFree:
👍  🎉

PR #608, opened 2016-06-08
Raise an error when weaving indirect descendants of RealmObject

### LINKED ISSUE #603: Detect and warn about second-level descendants from RealmObject
Until we support proper inheritance, we should warn about second-level descendants of `RealmObject`. We could warn about this during weaving or at least when attempting to `Manage()` an object.

fealebenpae:
Fixes #603


Tests/WeaverTests/RealmWeaver.Tests/WeaverTests.cs
```diff
@@ -174,7 +170,7 @@ public void SetValueUnmanagedShouldSetBackingField(string typeName, object prope
         [TestCase("Int32", 100)]
         [TestCase("Int64", 100L)]
         [TestCase("Single", 123.123f)]
-        [TestCase("Double", 123,123)]
+        [TestCase("Double", 123.123)]
```
kristiandupont: :-O

kristiandupont:
:+1:

PR #623, opened 2016-06-13
Ignore types from dynamic assemblies

fealebenpae:
Calling `Assembly.GetTypes()` on an assembly that is a `System.Reflection.Emit.AssemblyBuilder` throws an exception if any of the types in the assembly aren't fully built yet.

We don't care about dynamic assemblies because there's no way for them to be woven anyway.


kristiandupont:
:+1:

PR #624, opened 2016-06-13
Refactor and extract AssemblyInfo.cs files

fealebenpae:
Extract all the common parts of all the various AssemblyInfo files in a central one. This makes it easier to stamp versions in CI.

### DIFF
--- a/Realm.PCL/Properties/AssemblyInfo.cs
+++ b/Realm.PCL/Properties/AssemblyInfo.cs
@@ -1,28 +1,21 @@
-﻿using System.Reflection;
-using System.Runtime.CompilerServices;
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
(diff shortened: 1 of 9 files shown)


kristiandupont:
:+1:

AndyDentFree:
Where does the assemblyVersion now come from? Does this mean the source lacks it altogether?
I would have expected it to be in `realm-dotnet/RealmAssemblyInfo.cs`

fealebenpae:
The CI job that will build the nuget package accepts a version string as a parameter and appends the `AssemblyVersion` and `AssemblyFileVersion` attributes to `RealmAssemblyInfo.cs` along with the build number.

AndyDentFree:
Yeah, I don't like that - it means there's no record in source control of what version the assembly was tagged with. That also seems contrary to the other products - Cocoa has a `Realm-info.plist` and Java has `version.txt`

fealebenpae:
That's cool, we can still have the version in the source.

PR #626, opened 2016-06-13
Prepare weaver tests for CI

fealebenpae:
This changes the `RealmWeaver.Tests` project so that it builds from the OS X Terminal. The goal is to build these from the command-line and run them with the NUnit Console Runner as part of CI.

### DIFF
--- a/Tests/WeaverTests/RealmWeaver.Tests/WeaverTests.cs
+++ b/Tests/WeaverTests/RealmWeaver.Tests/WeaverTests.cs
@@ -106,11 +106,6 @@ namespace Tests
             _sourceAssemblyPath = typeof(AssemblyToProcess.AllTypesObject).Assembly.Location;
             _targetAssemblyPath = _sourceAssemblyPath.Replace(".dll", $".{_weaverOptions}.dll");
 
-            if (File.Exists(_targetAssemblyPath))
-            {
-                File.Delete(_targetAssemblyPath);
-            }
-
             var moduleDefinition = ModuleDefinition.ReadModule(_sourceAssemblyPath);
             (moduleDefinition.AssemblyResolver as DefaultAssemblyResolver).AddSearchDirectory(Path.GetDirectoryName(_sourceAssemblyPath));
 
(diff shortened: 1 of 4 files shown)


Tests/WeaverTests/RealmWeaver.Tests/WeaverTests.cs
```diff
@@ -16,6 +16,9 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+extern alias realm;
```
kristiandupont: Is there some convention that I am unaware of stating that extern aliases are named with lowercase?

Tests/WeaverTests/RealmWeaver.Tests/WeaverTests.cs
```diff
@@ -135,42 +93,50 @@ private void WeavePropertyChanged(string targetAssemblyPath)
...
         public void FixtureSetup()
         {
-            var assemblyDir = GetAssemblyDir();
-            TryCleanupOldAssemblies(assemblyDir);
+            _sourceAssemblyPath = typeof(AssemblyToProcess.AllTypesObject).Assembly.Location;
+            _targetAssemblyPath = _sourceAssemblyPath.Replace(".dll", $".{_weaverOptions}.dll");
 
-            _sourceAssemblyPath = GetSourceAssemblyPath();
-            _targetAssemblyPath = CopyAssemblyToNextUniquePath(_sourceAssemblyPath);
+            if (File.Exists(_targetAssemblyPath))
```
kristiandupont: I believe this check is unnecessary?

kristiandupont:
:+1:

Tests/WeaverTests/RealmWeaver.Tests/WeaverTests.cs
fealebenpae: Well, `global`, the default alias, is lowercase, but I don't know of a convention.

Tests/WeaverTests/RealmWeaver.Tests/WeaverTests.cs
fealebenpae: Yeah, I believe we can do without. It was left from an earlier iteration I had on this, but is now superfluous.

PR #631, opened 2016-06-14
Make Android tests debuggable always

fealebenpae:
That way we can run headless unit tests in CI when building for Release.


kristiandupont:
:+1:

PR #632, opened 2016-06-14
Disable debugging symbols for iOS wrappers in Release

### LINKED ISSUE #628: Slim down NuGet IOS component
Our NuGet package as of 0.76.0 is 178MB of which 131MB is the IOS10 Realm.dll.

We are starting with a 26MB `librealm-ios-no-bitcode.a` and building a 137MB `libwrappers.a` so it seems something is being builked out in building wrappers.

fealebenpae:
Fixes #628


kristiandupont:
:+1:

AndyDentFree:
👍

PR #634, opened 2016-06-14
Trivial test change to run unit tests with NuGet

AndyDentFree:
DateTimeTests.cs
- SortingFinelyDifferentDateTimes use ToUnixTimeMilliseconds instead of our private offset variation.


fealebenpae:
:+1:

Issue #635, opened 2016-06-14
Fody Unhandled Exception: Sequence contains no matching elements


heiser:
I installed Realm for Xamarin (0.76) into my PCL, iOS and Android projects. Everything compiles fine.

Then I created an object derived from RealmObject like so:

`    public class HistoryProperty : RealmObject
    {
        public string Name { get; set; }
        public int Order { get; set; }
    }
`

This results in the following exception (if I remove RealmObject, no error; if I derive from it, then I get the exception):

1>    Fody: Fody (version 1.29.4.0) Executing
1>MSBUILD : error : Fody: An unhandled exception occurred:
1>MSBUILD : error : Exception:
1>MSBUILD : error : Sequence contains no matching element
1>MSBUILD : error : StackTrace:
1>MSBUILD : error :    at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
1>MSBUILD : error :    at ModuleWeaver.WeaveRealmObjectHelper(TypeDefinition realmObjectType)
1>MSBUILD : error :    at ModuleWeaver.Execute()
1>MSBUILD : error :    at lambda_method(Closure , Object )
1>MSBUILD : error :    at InnerWeaver.ExecuteWeavers() in C:\projects\fody\FodyIsolated\InnerWeaver.cs:line 161
1>MSBUILD : error :    at InnerWeaver.Execute() in C:\projects\fody\FodyIsolated\InnerWeaver.cs:line 82
1>MSBUILD : error : Source:
1>MSBUILD : error : System.Core
1>MSBUILD : error : TargetSite:
1>MSBUILD : error : TSource Single[TSource](System.Collections.Generic.IEnumerable`1[TSource], System.Func`2[TSource,System.Boolean])
1>MSBUILD : error : 

The Realm assembly and FodyWeaver.xml files are present in all three projects.

AndyDentFree:
Do you have _Linker Behaviour_ set to _Link All_ in the Build settings (Build - IOS Build)? That would explain this issue - #615 schedules adding full linking support.

Also, are you compiling for IOS or Android or experiencing this on both platforms?

heiser:
Currently, my iOS build setting for Linker Behaviour is set to Link SDKs.

But I'm experiencing this problem when attempting to compile for the Android platform.

AndyDentFree:
What is your Android linking setting - is that also Link SDKs or full linking?

heiser:
[...]

PR #636, opened 2016-06-15
Weaver error when RealmObject class is missing default constructor

kristiandupont:
This might help diagnose #635

### DIFF
--- a/RealmWeaver/ModuleWeaver.cs
+++ b/RealmWeaver/ModuleWeaver.cs
@@ -230,8 +230,18 @@ public class ModuleWeaver
             }
         }
 
+        var objectConstructor = type.GetConstructors()
+            .SingleOrDefault(c => c.Parameters.Count == 0 && c.IsPublic && !c.IsStatic);
+        if (objectConstructor == null)
+        {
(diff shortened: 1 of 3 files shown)


RealmWeaver/ModuleWeaver.cs
```diff
@@ -230,8 +230,18 @@ private void WeaveType(TypeDefinition type, Dictionary<string, Tuple<MethodRefer
             }
         }
 
+        var objectConstructor = type.GetConstructors()
+            .SingleOrDefault(c => c.Parameters.Count == 0 && c.IsPublic && !c.IsStatic);
+        if (objectConstructor == null)
+        {
+            var nonDefaultConstructor = type.GetConstructors().First();
```
kristiandupont: Is there a scenario where this will fail? Should I fall back on `LogError`?  I can't think of a situation, except some weird case where another weaver created the non-default constructor which would leave us with not sequence point below, but in that case maybe the general purpose error handler is good enough?

fealebenpae: This will never fail unless someone royally messes up the IL so I wouldn't worry.

fealebenpae:
👍

PR #637, opened 2016-06-15
Doc and test updates to match 0.76.1 changes

AndyDentFree:
SortingTests.cs
- added SortsAccentedAndSpecialCorrectly (will fail on previous builds)


kristiandupont:
:+1:

Issue #639, opened 2016-06-15
How to best remove managed child lists


AndyDentFree:
Document and respond to [SO question](http://stackoverflow.com/questions/37819634/best-method-to-remove-managed-child-lists-one-to-many-parent-child-relationsh)

PR #640, opened 2016-06-15
#639 Sample code to verify approach for cascading delete workaround

AndyDentFree:
RelationshipTests.cs
- TestDeleteChildren added with Product and Report class definitions


kristiandupont:
:+1:

PR #642, opened 2016-06-16
Manual migrations API

### LINKED ISSUE #545: Migrations API

fealebenpae:
Fixes #545


Realm.Shared/native/Configuration.cs
```diff
@@ -0,0 +1,63 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Realms.Native
```
kristiandupont: Should we introduce this namespace in general? I don't like the name too much, but it would match the folder structure.

fealebenpae: I think it makes sense, especially as a holder for structs we marshal whose names might otherwise conflict with types in the `Realms` namespace. I'm not attached to the name, though - my main concern is to group together those types that exist solely for marshalling.

kristiandupont:
:+1:

kristiandupont:
..just missing a CHANGELOG update!

kristiandupont:
Recreated in #812

PR #658, opened 2016-06-23
Schema API

### LINKED ISSUE #645: Add a Schema API
Expose the Object Store Schema APIs in C#, much like Cocoa's [RLMSchema](https://realm.io/docs/objc/1.0.1/api/Classes/RLMSchema.html)

fealebenpae:
Fixes #645


Realm.Shared/Realm.cs
```diff
@@ -420,10 +333,10 @@ static public void DeleteRealm(RealmConfiguration configuration)
 
 
         [ReliabilityContract(Consistency.WillNotCorruptState, Cer.Success)]
-        private TableHandle GetTable(Type realmType)
+        private TableHandle GetTable(Schema.Object schema)
         {
             var result = new TableHandle();
-            var tableName = "class_" + realmType.Name;
+            var tableName = "class_" + schema.Name;
```
bdash: This should eventually use `ObjectStore::table_name_for_object_type` rather than constructing table names itself.

Realm.Shared/Schema/PropertyType.cs
```diff
@@ -0,0 +1,103 @@
...
+    internal enum PropertyType : byte
+    {
+        Int = 0,
+        Bool = 1,
+        Float = 9,
+        Double = 10,
+        String = 2,
+        Data = 4,
+        Any = 6,
+        Date = 8,
```
kristiandupont: Shouldn't this be either `DateTimeOffset` or `TimeStamp`?

kristiandupont:
:+1: whew!

Realm.Shared/Schema/PropertyType.cs
fealebenpae: Well, I went and copied the names from the Object Store `PropertyType` enum.

Issue #287, opened 2015-12-11
IList declarations need more work to enable Standalone objects with relationships


AndyDentFree:
This is a concern as I thought they were working and heavily tested.

If you replace RealmList with IList in the declarations in `RelationshipTests.cs` then all the tests apart from the _TopDog_ tests will fail. I'm not sure if this is a problem with Fody or the RealmList classes.

CrimsonKyle:
Being able to define an IList of realm objects with a setter in a model is essential to being able to deserialize a root level Json doc into more than 2 levels deep i.e. (Parent < Child).  Do you know where ILists are currently on the priority list?

AndyDentFree:
@CrimsonKyle is the concern over having a setter or using `IList` in the declaration? The setter may come earlier as improving initialisation syntax is a passionate goal of mine too. The JSON decoding is a great scenario to help bump the priority of this task.

Shazbot89:
Not certain, but we think the Newtonsoft deserializer is unable to map the child RealmLists without a set method. @CrimsonKyle was able to verify non-Realm objects using IList properties with { set; } were able to deserialize the entire tree of objects while Realm inherited objects with RealmList stop deserializing at the first level.

CrimsonKyle:
#492

AndyDentFree:
Whilst this is working with the current commit, based on a discussion with @fealebenpae I am going to rewrite it to avoid generating a return `RealmList` all the time and allow init with simple braced expressions even if you just declare the `get` without `set`.

PR #659, opened 2016-06-24
Ad/feature/ilist declarations #287

AndyDentFree:
Includes changing all samples to use IList declarations and minor wrappers change so we can optimally check if a relationship is currently null.


AndyDentFree:
@fealebenpae Please have a look at this and what I am trying to weave - it crashes on the standalone tests.

AndyDentFree:
retest this please

AndyDentFree:
retest this please

AndyDentFree:
retest this please

Tests/IntegrationTests.Shared/RelationshipTests.cs
```diff
@@ -360,25 +360,43 @@ public void TestAddingStandAloneObjectToToManyRelationship()
...
+
+         A null property has to be initialised with a full object:
+         Friends = new List<Person> { new Person {  FullName = "Sally"} }
+
+         We have two approaches under consideration which would allow the simple braced init:
+         1) automatically weave in a clause in the ctor initialising the list properties, 
+            which adds overhead to creating an object if they never use the list.
+         2) If the user chooses to init a list in the ctor, preserve the ctor init (which is a 
+            requested feature anyway in issue 109)
+        */
```
kristiandupont: Seems to me this comment should be an issue instead?

fealebenpae: On the topic of the issue described in the comment I think a good compromise will be to weave the list getter so that it lazily instantiates the list in the unmanaged case:
```
public IList<Person> Friends
{
    get
    {
        if (!base.IsManaged)
        {
             if (_friends == null)
                 _friends = new List<Person>();

             return _friends;
        }

        return base.GetListValue<Person>(...);
    }
}
```

RealmWeaver/ModuleWeaver.cs
```diff
@@ -303,29 +301,31 @@ private void WeaveType(TypeDefinition type, Dictionary<string, Tuple<MethodRefer
...
-            // RealmList allows people to declare lists only of _realmObject due to the class definition
-            if (!prop.IsAutomatic())
-            {
-                LogErrorPoint(
-                    $"{type.Name}.{columnName} is not an automatic property but its type is a RealmList which normally indicates a relationship",
-                    sequencePoint);
+            var elementType = ((GenericInstanceType)prop.PropertyType).GenericArguments.Single();
+            // IList or RealmList allows people to declare lists only of _realmObject due to the class definition
+            if (!elementType.Resolve().BaseType.IsSameAs(_realmObject)) {
+                LogWarningPoint(
```
kristiandupont: I can't decide if I think this should be a warning or an error.
In favor of it being an error: you can add `[Ignore]` to explicitly state that you know what you are doing, but you might miss the warning. On the other hand, it's more in line with our decision to implicitly ignore properties with non-automatic getters or setters.

fealebenpae:
`ICopyValuesFrom` should be an internal interface and as such doesn't need to be in the PCL.

kristiandupont:
:+1:, apart from @fealebenpae's comment.

RealmWeaver/ModuleWeaver.cs
fealebenpae: I think we should look into creating a Roslyn Code Analyzer plugin which will be emitting warnings instead of the weaver. Because it ties into the IDEs people will be able to silence those. It's nice for us to be verbose about what we don't persist in the weaver with warnings and all, but right now there's no way to turn off these warnings so more experienced developers are going to be annoyed if we issue too many warnings from the weaver.

fealebenpae:
👍 from me as well. The `ICopyValuesFrom` thing is mostly cosmetic as it really won't be visible in the PCL because it's internal, but it's best if it was out IMO. Also, I'd suggest declaring it in the same file as `RealmList` because they're tightly coupled (see `RealmResults` where I employ the same internal interface trick).

fealebenpae:
@AndyDentFree: I'm not sure that implementation of internal interfaces is visible to other assemblies referencing the class, but like I said I think having the interface in the PCL isn't a big deal, so if removing it prevents bait'n'switch from working then let's keep it.

Realm.Shared/RealmList.cs
```diff
@@ -32,14 +32,14 @@ namespace Realms
...
     /// 
     /// <typeparam name="T">Type of the RealmObject which is the target of the relationship.</typeparam>
-    public class RealmList<T> : IList<T> where T : RealmObject
+    public class RealmList<T> : IList<T>, ICopyValuesFrom where T : RealmObject
     {
         private class RealmListEnumerator : IEnumerator<T> 
         {
             private int index;
-            private RealmList<T> enumerating;
+            private IList<T> enumerating;
```
kristiandupont: This will always be a `RealmList<T>` still, won't it?

Realm.Shared/RealmObject.cs
```diff
@@ -698,7 +701,8 @@ protected void SetNullableDateTimeOffsetValue(string propertyName, DateTimeOffse
                 NativeTable.set_null(_metadata.Table, _metadata.ColumnIndices[propertyName], (IntPtr)rowIndex);
         }
 
-        // TODO make not generic
+
+        // This is a generic fallback, we normally handle with woven type-specific setters above
```
kristiandupont: We do? This is for `RealmObject`, I don't think there is anything else for those?

RealmWeaver/ModuleWeaver.cs
```diff
@@ -258,7 +260,12 @@ private void WeaveType(TypeDefinition type, Dictionary<string, Tuple<MethodRefer
         if (mapToAttribute != null)
             columnName = ((string) mapToAttribute.ConstructorArguments[0].Value);
 
+     //   Debugger.Launch();
+
         var backingField = GetBackingField(prop);
+        if (backingField == null) {
+            LogErrorPoint($"Property {prop.Name} has no backing field", sequencePoint);
```
kristiandupont: Can this happen?

AndyDentFree: Ahhh, good catch, that was a debugging addition when I was trying to decipher the weird error messages from my IL being wrong.

Realm.Shared/RealmList.cs
AndyDentFree: We don't *require* it to be a `RealmList` though, but yeah it could be reverted and avoids dispatch through the interface.

RealmWeaver/ModuleWeaver.cs
```diff
@@ -304,28 +302,55 @@ private void WeaveType(TypeDefinition type, Dictionary<string, Tuple<MethodRefer
...
+                backingDef.Attributes &= ~FieldAttributes.InitOnly;  // without a set; auto property has this flag we must clear
+            }
+            ReplaceListGetter(prop, backingField, columnName,
+                new GenericInstanceMethod(_genericGetListValueReference) { GenericArguments = { elementType } }, elementType,
+                             ModuleDefinition.ImportReference(concreteListConstructor) );
+        } else if (prop.PropertyType.Name == "RealmList`1" && prop.PropertyType.Namespace == "Realms") {
+            var elementType = ((GenericInstanceType)prop.PropertyType).GenericArguments.Single();
+            if (!elementType.Resolve().BaseType.IsSameAs(_realmObject)) {
+                LogWarningPoint(
+  $"SKIPPING {type.Name}.{columnName} because it is a RealmList but its generic type is not a RealmObject subclass, so will not persist",
```
kristiandupont: Creating a `RealmList` of something that is not a `RealmObject` should be an error, shouldn't it? You can't use it for anything.

kristiandupont: Ah, but the compiler will have complained already, so it doesn't really make a difference.

Tests/WeaverTests/RealmWeaver.Tests/WeaverTests.cs
```diff
@@ -29,6 +29,7 @@
 using NUnit.Framework;
 using Realms;
 using System.ComponentModel;
+using AssemblyToProcess;
```
kristiandupont: Does this work? That will reference the pre-weaving assembly.

RealmWeaver/ModuleWeaver.cs
AndyDentFree: Compiler will allow a further descendant of `RealmObject` - there is no way to constrain to just **direct** descendants.

kristiandupont: Right, but if we encounter such a type we will have already created an error message.

RealmWeaver/ModuleWeaver.cs
```diff
@@ -439,22 +448,116 @@ void ReplaceGetter(PropertyDefinition prop, string columnName, MethodReference g
...
+    // if you don't have an automatic set; on the property, it shows in the debugger with
+    //         Attributes    Private | InitOnly    Mono.Cecil.FieldAttributes
+    void ReplaceListGetter(PropertyDefinition prop, FieldReference backingField, string columnName, MethodReference getListValueReference, TypeReference elementType, MethodReference listConstructor)
+    {
+        /// A synthesized property getter looks like this:
+        ///   0: ldarg.0  // load the this pointer
+        ///   1: ldfld <backingField>
+        ///   2: ret
+        /// We want to change it so it looks somewhat like this, from IL pane in Linqpad 5:
+        /*
```
fealebenpae: Can you change the IL snippet here to match what LINQPad outputs when it compiles with query optimization (`/optimize+`)? The one here has a bunch of debug stuff we don't need/want and it might confuse someone not used to reading IL.

AndyDentFree: It would make more sense to just remove the LINQpad reference as I rewrote it anyway. I am not working from LINQpad output any more.

fealebenpae: That works too 👍

Realm.Shared/RealmObject.cs
```diff
@@ -64,27 +64,30 @@ internal void _CopyDataFromBackingFieldsToRow()
...
+            foreach (var prop in wovenPropertiesWithBacking)
             {
                 var value = prop.Field.GetValue(this);
-                if (prop.Info.PropertyType.IsGenericType)
-                {
-                    var genericType = prop.Info.PropertyType.GetGenericTypeDefinition();
-                    if (genericType == typeof(RealmList<>))
+                if (value != null) {
+                    var listValue = value as IEnumerable<RealmObject>;
+                    if (listValue != null)  // assume it is NOT a RealmList so need to wipe afer copy
```
fealebenpae: We can use `prop.Type == Realms.Schema.PropertyType.Array` to assume list properties more correctly.

fealebenpae: Ah, apologies, I mistakenly though this was rebased against the Schema API changed in master.

fealebenpae:
This looks okay so far, but I'd like to see it rebased against master before 👍-ing it.

fealebenpae:
👍 👍 👍 

There are still some merge conflicts, but I was mainly interested in how the schema stuff would be reconciled. 

IMO I think we're not far from needing to weave the implementation of `_CopyDataFromBackingFieldsToRow`, but that's an entirely separate point.

PR #668, opened 2016-06-29
Row validation

### LINKED ISSUE #662: Row validation on get/set
As we are not relying on object store accessors, we need to add row validation ourselves.

kristiandupont:
This fixes #662.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ### Minor Changes
 * Fixed a crash on iOS when creating many short-lived realms very rapidly in parallel (#653)
+* `RealmObject.IsValid` lets you know if a managed object has been deleted
+* Accessing properties on invalid objects will throw an exception rather than crash with a segfault (#662)
 
 0.76.1 (2016-06-15)
 -------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ### Minor Changes
 * Fixed a crash on iOS when creating many short-lived realms very rapidly in parallel (#653)
-* `RealmObject.IsValid` lets you know if a managed object has been deleted
+* `RealmObject.IsValid` can be called to check if a managed object has been deleted
 * Accessing properties on invalid objects will throw an exception rather than crash with a segfault (#662)
 
 0.76.1 (2016-06-15)
--- a/Realm.Shared/RealmObject.cs
+++ b/Realm.Shared/RealmObject.cs
@@ -49,7 +49,7 @@ namespace Realms
         /// will throw an exception.
         /// Unmanaged objects are always considered valid.
         /// </summary>
-        public bool IsValid => _rowHandle == null || _rowHandle.IsAttached;
+        public bool IsValid => _rowHandle?.IsAttached != false;
 
         internal void _Manage(Realm realm, RowHandle rowHandle)
         {
--- a/Realm.Shared/RealmObject.cs
+++ b/Realm.Shared/RealmObject.cs
@@ -47,8 +47,9 @@ namespace Realms
         /// Returns true if this object is managed and represents a row in the database.
         /// If a managed object has been removed from the Realm, it is no longer valid and accessing properties on it
         /// will throw an exception.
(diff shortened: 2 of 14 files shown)


Realm.Shared/RealmObject.cs
```diff
@@ -43,6 +43,13 @@ public class RealmObject
         /// </summary>
         public bool IsManaged => _realm != null;
 
+        /// <summary>
+        /// Returns true if this object is managed and represents a row in the database.
+        /// If the object has been deleted (say, on another thread), it is no longer valid and accessing properties on it
+        /// will throw an exception.
+        /// </summary>
+        public bool IsValid => _rowHandle != null &&_rowHandle.IsAttached;
```
fealebenpae: How about `_rowHandle?.IsAttached`? 😛

fealebenpae:
👍

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 
 ### Minor Changes
 * Fixed a crash on iOS when creating many short-lived realms very rapidly in parallel (#653)
+* `RealmObject.IsValid` lets you know if a managed object has been deleted
```
AndyDentFree: I think the comment _lets you know_ sounds a bit too active - how about _can be called to check_?

Realm.Shared/RealmObject.cs
```diff
@@ -43,6 +43,13 @@ public class RealmObject
         /// </summary>
         public bool IsManaged => _realm != null;
 
+        /// <summary>
+        /// Returns true if this object is managed and represents a row in the database.
+        /// If the object has been deleted (say, on another thread), it is no longer valid and accessing properties on it
```
AndyDentFree: Maybe needs to say _ If a **managed** object has been deleted_ ?

Tests/IntegrationTests.Shared/ObjectIntegrationTests.cs
```diff
@@ -247,6 +247,23 @@ public void IteratePeople()
...
+            _realm.Write(() =>
+            {
+                var p1 = _realm.CreateObject<Person>();
+                Assert.That(p1.IsValid);
+
+                _realm.Remove(p1);
+                Assert.That(p1.IsValid, Is.False);
+            });
+
+            // IsValid should always return false for standalone objects
```
AndyDentFree: This **really** bothers me because the semantics are just wrong - a standalone object **is** _valid_

bdash: It doesn't match the semantics of Cocoa's equivalent either. `-[RLMObject isInvalidated]` returns true only if the object is managed and its row is detached. An unmanaged object is always valid.

Realm.Shared/RealmObject.cs
bdash: The aside that mentions another thread is a little weird too. It may give users the impression that objects can be invalidated at any time, when in fact they'll only be invalidated by another thread when the Realm they belong to is advanced to a newer version.

Tests/IntegrationTests.Shared/ObjectIntegrationTests.cs
cmelchior: Hmm, seems that Java got these semantics wrong then. We only return `true` for `isValid` if it is managed and not deleted. I could see the sense in changing it to return true for unmanaged objects as well since data is actually readable.

Realm.Shared/RealmObject.cs
```diff
@@ -47,8 +47,9 @@ public class RealmObject
         /// Returns true if this object is managed and represents a row in the database.
         /// If a managed object has been removed from the Realm, it is no longer valid and accessing properties on it
         /// will throw an exception.
+        /// Unmanaged objects are always considered valid.
         /// </summary>
-        public bool IsValid => _rowHandle?.IsAttached == true;
+        public bool IsValid => _rowHandle == null || _rowHandle.IsAttached;
```
nirinchev: You could keep the null propagation by `_rowHandle?.IsAttached != false` 😛

AndyDentFree:
👍

PR #675, opened 2016-07-04
Implement dynamic API

### LINKED ISSUE #646: Add a Dynamic API
Implement [`IDynamicMetaObjectProvider`](https://msdn.microsoft.com/en-us/library/system.dynamic.idynamicmetaobjectprovider(v=vs.110).aspx). Add stringly-typed overloads of `Realm.All<T>()`, `Realm.CreateObject<T>()`, and `Realm.RemoveAll<T>()` that expose the `dynamic` API.

Needs #645

fealebenpae:
Fixes #646 

This still needs tests for results, lists and relationships, but the implementation is there.

- [x] Tests for getting and setting properties dynamically
- [x] Tests for getting and setting relationship properties
- [x] Tests for using RealmResults of dynamic objects

### DIFF
--- a/Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
+++ b/Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using Realms.Weaving;
+
+namespace Realms.Dynamic
+{
+    internal class DynamicRealmObjectHelper : IRealmObjectHelper
+    {
+        internal static readonly DynamicRealmObjectHelper Instance = new DynamicRealmObjectHelper();
+
+        public RealmObject CreateInstance()
+        {
+            return new DynamicRealmObject();
+        }
+    }
+}
+
--- a/Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
+++ b/Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
@@ -1,4 +1,22 @@
-﻿using System;
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+using System;
 using Realms.Weaving;
 
 namespace Realms.Dynamic
--- a/Realm.Shared/Dynamic/MetaRealmObject.cs
+++ b/Realm.Shared/Dynamic/MetaRealmObject.cs
@@ -1,4 +1,22 @@
-﻿using System;
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+using System;
 using System.Collections.Generic;
 using System.Dynamic;
 using System.Linq.Expressions;
--- a/Realm.Shared/Dynamic/MetaRealmObject.cs
+++ b/Realm.Shared/Dynamic/MetaRealmObject.cs
@@ -29,16 +29,10 @@ namespace Realms.Dynamic
                 return base.BindGetMember(binder);
             }
 
-            var convertedExpression = Expression;
-            if (convertedExpression.Type != LimitType)
-            {
-                convertedExpression = Expression.Convert(convertedExpression, LimitType);
-            }
-
             var arguments = new List<Expression>
             {
                 WeakConstant(_metadata.Table),
-                Expression.Field(convertedExpression, RealmObjectRowHandleField),
+                Expression.Field(GetLimitedSelf(), RealmObjectRowHandleField),
                 Expression.Constant(_metadata.ColumnIndices[property.Name])
             };
 
@@ -71,7 +65,7 @@ namespace Realms.Dynamic
                         realmObjectOpsMethod = typeof(RealmObjectOps).GetMethod(nameof(RealmObjectOps.GetDoubleValue));
                     break;
                 case Schema.PropertyType.String:
-                    arguments.Insert(0, Expression.Field(convertedExpression, RealmObjectRealmField));
+                    arguments.Insert(0, Expression.Field(GetLimitedSelf(), RealmObjectRealmField));
                     realmObjectOpsMethod = typeof(RealmObjectOps).GetMethod(nameof(RealmObjectOps.GetStringValue));
                     break;
                 case Schema.PropertyType.Data:
@@ -84,12 +78,12 @@ namespace Realms.Dynamic
                         realmObjectOpsMethod = typeof(RealmObjectOps).GetMethod(nameof(RealmObjectOps.GetDateTimeOffsetValue));
                     break;
                 case Schema.PropertyType.Object:
-                    arguments.Insert(0, Expression.Field(convertedExpression, RealmObjectRealmField));
+                    arguments.Insert(0, Expression.Field(GetLimitedSelf(), RealmObjectRealmField));
                     arguments.Add(Expression.Constant(property.ObjectType));
                     realmObjectOpsMethod = typeof(RealmObjectOps).GetMethod(nameof(RealmObjectOps.GetObjectValue)).MakeGenericMethod(typeof(DynamicRealmObject));
                     break;
                 case Schema.PropertyType.Array:
-                    arguments.Insert(0, Expression.Field(convertedExpression, RealmObjectRealmField));
+                    arguments.Insert(0, Expression.Field(GetLimitedSelf(), RealmObjectRealmField));
                     arguments.Add(Expression.Constant(property.ObjectType));
                     realmObjectOpsMethod = typeof(RealmObjectOps).GetMethod(nameof(RealmObjectOps.GetListValue)).MakeGenericMethod(typeof(DynamicRealmObject));
                     break;
@@ -102,7 +96,7 @@ namespace Realms.Dynamic
             }
 
             var argumentShouldBeDynamicRealmObject = BindingRestrictions.GetTypeRestriction(Expression, typeof(DynamicRealmObject));
-            var argumentShouldBeInTheSameRealm = BindingRestrictions.GetInstanceRestriction(Expression.Field(convertedExpression, RealmObjectRealmField), _realm);
+            var argumentShouldBeInTheSameRealm = BindingRestrictions.GetInstanceRestriction(Expression.Field(GetLimitedSelf(), RealmObjectRealmField), _realm);
             return new DynamicMetaObject(expression, argumentShouldBeDynamicRealmObject.Merge(argumentShouldBeInTheSameRealm));
         }
 
@@ -114,12 +108,10 @@ namespace Realms.Dynamic
                 return base.BindSetMember(binder, value);
             }
 
-            var convertedExpression = Expression.Convert(Expression, typeof(DynamicRealmObject));
-
             var arguments = new List<Expression>
             {
                 WeakConstant(_metadata.Table),
-                Expression.Field(convertedExpression, RealmObjectRowHandleField),
+                Expression.Field(GetLimitedSelf(), RealmObjectRowHandleField),
                 Expression.Constant(_metadata.ColumnIndices[property.Name]),
             };
 
@@ -178,7 +170,7 @@ namespace Realms.Dynamic
                     break;
                 case Schema.PropertyType.Object:
                     argumentType = typeof(RealmObject);
-                    arguments.Insert(0, Expression.Field(convertedExpression, RealmObjectRealmField));
+                    arguments.Insert(0, Expression.Field(GetLimitedSelf(), RealmObjectRealmField));
                     realmObjectOpsMethod = typeof(RealmObjectOps).GetMethod(nameof(RealmObjectOps.SetObjectValue)).MakeGenericMethod(typeof(RealmObject));
                     break;
             }
@@ -199,7 +191,7 @@ namespace Realms.Dynamic
             var expression = Expression.Block(Expression.Call(realmObjectOpsMethod, arguments), Expression.Default(binder.ReturnType));
 
             var argumentShouldBeDynamicRealmObject = BindingRestrictions.GetTypeRestriction(Expression, typeof(DynamicRealmObject));
-            var argumentShouldBeInTheSameRealm = BindingRestrictions.GetInstanceRestriction(Expression.Field(convertedExpression, RealmObjectRealmField), _realm);
+            var argumentShouldBeInTheSameRealm = BindingRestrictions.GetInstanceRestriction(Expression.Field(GetLimitedSelf(), RealmObjectRealmField), _realm);
             return new DynamicMetaObject(expression, argumentShouldBeDynamicRealmObject.Merge(argumentShouldBeInTheSameRealm));
         }
 
@@ -216,9 +208,15 @@ namespace Realms.Dynamic
             return Expression.Convert(property, typeof(T));
         }
 
-        private static MethodInfo GetOpsMethod(string methodName)
+        private Expression GetLimitedSelf()
         {
-            return typeof(RealmObjectOps).GetMethod(methodName, BindingFlags.Static | BindingFlags.NonPublic);
+            var convertedExpression = Expression;
+            if (convertedExpression.Type != LimitType)
+            {
+                convertedExpression = Expression.Convert(convertedExpression, LimitType);
+            }
+
+            return convertedExpression;
         }
     }
 }
--- a/Realm.Shared/Dynamic/MetaRealmObject.cs
+++ b/Realm.Shared/Dynamic/MetaRealmObject.cs
@@ -1,0 +1,225 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Dynamic;
+using System.Linq.Expressions;
+using System.Reflection;
+
+namespace Realms.Dynamic
+{
+    internal class MetaRealmObject : DynamicMetaObject
+    {
+        private readonly Realm _realm;
+        private readonly RealmObject.Metadata _metadata;
+
+        private static readonly FieldInfo RealmObjectRealmField = typeof(RealmObject).GetTypeInfo().GetField("_realm", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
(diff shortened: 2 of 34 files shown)


fealebenpae:
This still needs tests for results, lists and relationships, but the implementation is there.

Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using Realms.Weaving;
+
+namespace Realms.Dynamic
+{
+    internal class DynamicRealmObjectHelper : IRealmObjectHelper
```
nirinchev: This could use a private ctor 😉

nirinchev: Also, I'm not what's the convention here, but I'd expose the `Instance` as a property backed by a `Lazy` field.

Realm.Shared/Dynamic/MetaRealmObject.cs
```diff
@@ -0,0 +1,225 @@
...
+            }
+
+            var arguments = new List<Expression>
+            {
+                WeakConstant(_metadata.Table),
+                Expression.Field(convertedExpression, RealmObjectRowHandleField),
+                Expression.Constant(_metadata.ColumnIndices[property.Name])
+            };
+
+            MethodInfo realmObjectOpsMethod = null;
```
nirinchev: I think you can avoid some code duplication by inferring just the method name rather than the whole info (as all of them seem to get the info from `typeof(RealmObjectOps).GetMethod({name})`

Realm.Shared/Dynamic/MetaRealmObject.cs
```diff
@@ -0,0 +1,225 @@
...
+
+        private static Expression WeakConstant<T>(T value) where T : class
+        {
+            var weakReference = new WeakReference(value);
+            var constant = Expression.Constant(weakReference);
+            var property = Expression.Property(constant, nameof(weakReference.Target));
+            return Expression.Convert(property, typeof(T));
+        }
+
+        private static MethodInfo GetOpsMethod(string methodName)
```
nirinchev: This seems unused

Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
fealebenpae: This object is so small and inexpensive to create that `Lazy`-fying it would be overkill, IMO.

nirinchev: Makes sense, how about a c# 6 auto property then? 😸 
```csharp
internal static DynamicRealmObjectHelper Instance { get; } = new DynamicRealmObjectHelper();
```

Realm.Shared/Dynamic/MetaRealmObject.cs
```diff
@@ -0,0 +1,225 @@
...
+using System.Reflection;
+
+namespace Realms.Dynamic
+{
+    internal class MetaRealmObject : DynamicMetaObject
+    {
+        private readonly Realm _realm;
+        private readonly RealmObject.Metadata _metadata;
+
+        private static readonly FieldInfo RealmObjectRealmField = typeof(RealmObject).GetTypeInfo().GetField("_realm", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
```
kristiandupont: `_realm` is not public?

fealebenpae: Ugh, don't ask - this is the only way I could get reflection to find it. Doesn't make sense, but...

Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
fealebenpae: I suppose that'd work just as well, although recently I've begun favoring fields instead of auto properties for pure storage, especially when it's internal to the assembly.

@kristiansupont @AndyDentFree, what do you think?

Realm.Shared/PreserveAttribute.cs
```diff
@@ -15,16 +15,27 @@
...
-        public bool Conditional { get; set; }
+
+        public bool AllMembers;
+        public bool Conditional;
+
+        public PreserveAttribute(string reason = null)
+        {
+        }
+
+        public PreserveAttribute(Type type)
```
kristiandupont: What's the purpose of this constructor? Do the linkers respect this signature? It conflicts with the changes in #674 which match the official Mono implementation.

Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
kristiandupont: I have no problems with fields. Is there any advantage to a property here besides from binary signature compatibility in case we would ever need to switch to a property (in which case, all uses will be recompiled as well, so not really an issue)?

AndyDentFree: Fields are faster- quite happy if we don't have any other reason to use properties

Realm.Shared/RealmObjectOps.cs
```diff
@@ -0,0 +1,393 @@
+﻿using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Realms
+{
+    [Preserve(AllMembers = true)]
+    internal static class RealmObjectOps
```
kristiandupont: Ops as in "operations"? Why are all these duplicated methods necessary, is there really no way to avoid that?

Realm.Shared/PreserveAttribute.cs
fealebenpae: I actually copied this from the `Foundation.PreserveAttribute` signature. I'll remove the changes here once I've rebased #674 onto this branch.

Realm.Shared/RealmObjectOps.cs
fealebenpae: I've modelled this class after [`System.Runtime.CompilerServices.RuntimeOps`](http://referencesource.microsoft.com/System.Core/Microsoft/Scripting/Actions/ExpandoObject.cs.html#35ec732587a0a4c3). The `DynamicMetaObject` for `System.Dynamic.ExpandoObject` emits expressions that call the methods on this class, much like the `MetaRealmObject` does here. I liked that pattern and I thought I'd copy it.

Another reason for this class to exist separately is that the methods here take a column index directly, because it can be cached in the dynamic callsite. It's also a cop-out, because the expression to look up the column index would be more complex. In any case, I hope that soon we start weaving the methods here and remove the protected methods on `RealmObject`.

bmunkholm:
Please add remainin sub tasks in your first comment e.g.:
[x] This is done
[ ] Testing
[ ] Release note
[ ] Whatever else is missing

Realm.Shared/RealmObjectOps.cs
```diff
@@ -0,0 +1,393 @@
...
+
+        public static void SetNullableInt64Value(TableHandle table, RowHandle row, IntPtr columnIndex, long? value)
+        {
+            if (value.HasValue)
+                NativeTable.set_int64(table, columnIndex, row.RowIndex, value.Value);
+            else
+                NativeTable.set_null(table, columnIndex, row.RowIndex);
+        }
+
+        public static void SetSingleValue(TableHandle table, RowHandle row, IntPtr columnIndex, float value)
```
bmunkholm: "Single" value? Where does that term come from :-) How about "Float" value?

fealebenpae: In .net `float` is aliased to `System.Single`.

bmunkholm:
@fealebenpae You may want to have @cmelchior to review the API of this since he made it for Java.

Realm.Shared/Dynamic/MetaRealmObject.cs
```diff
@@ -0,0 +1,241 @@
...
+                Expression.Constant(_metadata.ColumnIndices[property.Name])
+            };
+
+            MethodInfo realmObjectOpsMethod = null;
+
+            switch (property.Type)
+            {
+                case Schema.PropertyType.Int:
+                    if (property.IsNullable)
+                        realmObjectOpsMethod = typeof(RealmObjectOps).GetMethod(nameof(RealmObjectOps.GetNullableInt64Value));
```
AndyDentFree: These repeated expressions bother me. Apart from the opportunity to simplify the repeated `typeof(RealmObjectOps)` would it be possible to look them up **once** and stash in static vars so they are not being called per-property?

Realm.Shared/Dynamic/MetaRealmObject.cs
```diff
@@ -0,0 +1,241 @@
...
+
+        private static Expression WeakConstant<T>(T value) where T : class
+        {
+            var weakReference = new WeakReference(value);
+            var constant = Expression.Constant(weakReference);
+            var property = Expression.Property(constant, nameof(weakReference.Target));
+            return Expression.Convert(property, typeof(T));
+        }
+
+        private Expression GetLimitedSelf()
```
AndyDentFree: Why is this necessary? Could we please have more comments explaining the rationale behind the code.

Realm.Shared/Realm.cs
```diff
@@ -357,55 +365,77 @@ private TableHandle GetTable(Schema.Object schema)
         /// <exception cref="RealmOutsideTransactionException">If you invoke this when there is no write Transaction active on the realm.</exception>
         public T CreateObject<T>() where T : RealmObject, new()
         {
+            RealmObject.Metadata metadata;
```
AndyDentFree: Why does this variant of the ctor no longer test if in a transaction

AndyDentFree: Sorry, forget about it, just realised it is forwarding.

Realm.Shared/Realm.cs
```diff
@@ -357,55 +365,77 @@ private TableHandle GetTable(Schema.Object schema)
...
+        {
             if (!IsInTransaction)
                 throw new RealmOutsideTransactionException("Cannot create Realm object outside write transactions");
 
-            var objectType = typeof(T);
-            if (Config.ObjectClasses != null && !Config.ObjectClasses.Contains(objectType))
-                throw new ArgumentException($"The class {objectType.Name} is not in the limited set of classes for this realm");
-
-            var metadata = Metadata[typeof(T)];
+            if (!Metadata.TryGetValue(className, out metadata))
```
AndyDentFree: Not happy about this change. Previously this was guarded for the vast majority of cases where people had not specified an `ObjectClasses`. Now it _appears_ we have added a dictionary lookup on **every object creation**?

AndyDentFree: OK, we had indexing on type before with `var metadata = Metadata[typeof(T)];` but I was hoping we could move to something more efficient. Is the string key faster and could we, with weaving, simplify this further? Can we add an issue to further refine this down to a simple indexed array rather than dictionary search? We can certainly weave in an ordinal unique class int and use that to build up the lookup table **once**.

Realm.Shared/Dynamic/MetaRealmObject.cs
fealebenpae: The way `dynamic` works in C# already stashes the result of this computation in a static field. Whenever the compiler encounters a dynamic expression, it creates a `CallSite` object and stores it in a static field. The various `CallSiteBinder` objects such as `GetMemberBinder` and `SetMemberBinder` describe the operation and a series of `DynamicMetaObject`objects produce an expression tree to execute the operation. The `CallSite` compiles this expression tree into a lambda and caches it.

In essence, in
```csharp
for (var i = 1; i <= 1000; i++)
{
    var person = realm.CreateObject("Person");
    person.Name = $"Person {i}";
}
```
`MetaRealmObject.BindSetMember` will be called only once. This is why I think we won't gain much by caching here. Alternatively, we could use something like [`InfoOf.Fody`](https://github.com/Fody/InfoOf), but I didn't feel like pulling in the dependency just for that.

AndyDentFree: Thanks for the detailed explanation. Please add a copy of that as a comment so people viewing the complex code will be similarly enlightened ;-)

Realm.Shared/Dynamic/MetaRealmObject.cs
fealebenpae: I suggest going through [Bill Wagner's "Implementing Dynamic Interfaces"](http://download.microsoft.com/download/5/4/B/54B83DFE-D7AA-4155-9687-B0CF58FF65D7/implementing-dynamic-interfaces.pdf) if you're interested in what this does and why, as I'm afraid I can't explain it properly without coming up with a massive infodump.

Realm.Shared/Realm.cs
fealebenpae: You're right, performance-wise a dictionary lookup would probably be an overkill and I suppose we can improve that with binary search over a sorted list. As you've noted, we've always done a dictionary lookup (which was an improvement over the previous reflection-based approach) so I think further optimizing this is beyond the scope of the PR.

Unfortunately I don't think weaving is a solution to this problem as the exact schema is only known at runtime.

Realm.Shared/Dynamic/MetaRealmObject.cs
nirinchev: Sorry to hijack, but this is quite interesting topic :) If the cache is per `CallSite`, wouldn't that mean that the following (pretty ugly) code snippet:
```csharp
var person = realm.CreateObject("Person");
person.Name = "Peter";
person.Name += " ";
person.Name += "Johnson";
```
will actually call `BindSetMember` 3 times and `BindGetMember` twice? If that is the case, the built-in caching will only be useful when properties are updated in a loop, but not when they are accessed on different lines?

Realm.Shared/Realm.cs
AndyDentFree: >Unfortunately I don't think weaving is a solution to this problem as the exact schema is only known at runtime.

I was musing, in a vague _go to bed it's 2:30am_ manner... 

1. the weaver sees the set of all classes from all assemblies
2. say a unique ascending index **for the current build** is generated and woven into each `RealmObject` subclass
3. Metadata lookup can then be constructed as a static array indexed by those RealmObject subclass indexes - just because a given Realm may have a limited schema doesn't stop it from indexing a giant central array of all metadata.

Realm.Shared/RealmList.cs
```diff
@@ -32,48 +33,35 @@ namespace Realms
...
 
             /// <summary>
             ///  Move the iterator to the next related object, starting "before" the first object.
             /// </summary>
             /// <returns>True only if can advance.</returns>
             public bool MoveNext()
             {
-                index++;
-                if (index >= enumerating.Count)
+                if (++_index >= _enumerating.Count)
```
AndyDentFree: Ugg, I never got around to investigating this but you sparked a nasty wraparound thought here on reviewing this code. I know it is a weird edge case but if someone calls `MoveNext` a stupid number of times, this implementation would eventually wrap around and get back to `_index==0`.

Fringe case if someone has a wild loop but suspect it **does** break the formal semantics of `MoveNext`.

Realm.Shared/Realm.cs
fealebenpae: We can also now define the schema dynamically at runtime and operate in environments where there are no actual woven classes (e.g. migrations).

With that it mind, [I've compared retrieval from a `Dictionary`, `SortedDictionary`, and `SortedList`](https://gist.github.com/fealebenpae/8e2959ac9c38bc52e20e738ebac2b133) and funnily enough `Dictionary` seems to be a clear winner in a dataset similar to our schema and metadata cache.

Realm.Shared/RealmList.cs
fealebenpae: I don't fully understand how this is functionally different, but I can revert this change.

AndyDentFree: It is not functionally different - the (extreme) wraparound bug is originally mine :-( I am not sure if it is worth adding a check for and refusing to increment `_index`.

Realm.Shared/Dynamic/MetaRealmObject.cs
fealebenpae: > will actually call BindSetMember 3 times and BindGetMember twice?

I believe so, yes.

Our use-case is for [properties to be retrieved in a loop during migration](https://realm.io/docs/swift/latest/#updating-values) so there should be few callsites. Besides, the compiler might be able to optimize in this scenario, but I'm not sure.

Of course, if the compiler doesn't optimize, this'll be slower even when using the statically-typed API because you'll perform two more stores and retrievals than necessary.

Depending on the code our users write we can always optimize if necessary using [`InfoOf.Fody`](https://github.com/Fody/InfoOf).

Realm.Shared/Realm.cs
nirinchev: I can't say I'm surprised that a O(1) dictionary lookup is faster than the O(ln(n)) lookups for the sorted structures :) Compared to the previous implementation, I think the performance difference will be based on the speed of `GetHashCode` for `Type` and `string` which should be negligible. Additionally I'd be surprised if an array lookup significantly outperforms the dictionary, but that might be worth benchmarking.

Realm.Shared/RealmObjectOps.cs
```diff
@@ -0,0 +1,393 @@
...
+            {
+                unsafe
+                {
+                    fixed (byte* buffer = value)
+                    {
+                        NativeTable.set_binary(table, columnIndex, row.RowIndex, (IntPtr)buffer, (IntPtr)value.LongLength);
+                    }
+                }
+            }
+        }
```
cmelchior: The Java Dynamic API also have a lot more methods that can be useful when working with data dynamically. Like 
```
boolean hasField(String field)
RealmFieldType getFieldType(String field)
String[] getFieldNames()
```

This is useful if your data is truly dynamic, like e.g reading field names from one file while reading data from another file.

https://realm.io/docs/java/latest/api/io/realm/DynamicRealmObject.html

cmelchior: Also there doesn't seem to be a way to set a List?

fealebenpae: > Also there doesn't seem to be a way to set a List?

Can you elaborate?

fealebenpae: > This is useful if your data is truly dynamic, like e.g reading field names from one file while reading data from another file.

I suppose we can make the [Schema API](https://github.com/realm/realm-dotnet/tree/master/Realm.Shared/Schema) public so that you'll be able to introspect a dynamic object by its schema.

AndyDentFree: Yeah I like making Schema API public. We should also consider the needs of people writing browsers, not just our primary need of migration.

cmelchior: If a Person object has a RealmList, I don't see a way for replacing that list?

Regarding the Schema API. In Java the Schema API is public as well as it is used by the migrations, but it also means that migrations are just one use case for it. You can always access it and modify it if need be (and some do).

https://realm.io/docs/java/latest/api/io/realm/RealmSchema.html 
https://realm.io/docs/java/latest/api/io/realm/RealmObjectSchema.html

fealebenpae: Our List properties are read-only by design.

Re. schema: do you think exposing a `Schema` property of type `Schema.Object` on a `RealmObject` would be a good equivalent of the Java APIs with regards to introspection of an object?

cmelchior: Not an easy question and probably mostly boils down to the feel of the rest of the API and language. 

In Java `DynamicRealmObject` is a first class object that is promoted as dynamic. In that sense it make sense to have first level methods for inspecting it's state which include fields names and types. Having a `Schema` reference make sense for our code base, but I am not sure it help the users write better code as it adds an extra level of indirection. Added to that the Schema object probably also contains methods for manipulating the schema like deleting fields. 

If you are working with dynamic data I would probably expect these methods to be directly available.

fealebenpae: I don't think adding a lot of APIs on dynamic objects is a good idea because as soon as the C# `dynamic` keyword comes into play you lose all code-completion and it'll be hard to discover those members.

I think there's a sort of a fundamental difference between how dynamic is implemented in the Java binding where you get an object with a dictionary-like API and how dynamic is usually done in C# with the `dynamic` keyword.

fealebenpae: I'm more concerned whether the proposed implementation of dynamic offers equivalent functionality to what's in Java and Cocoa as opposed to whether the implementation itself is equivalent. Obviously introspection is a big chunk of functionality missing from the current implementation and rectifying that would mean making the Schema API publicly accessible.

kristiandupont:
:+1:

PR #678, opened 2016-07-05
Use unsigned char for PropertyType enum

kristiandupont:
..also some small fixes to make it compile on Windows

### DIFF
--- a/wrappers/src/schema_cs.cpp
+++ b/wrappers/src/schema_cs.cpp
@@ -62,7 +62,7 @@ REALM_EXPORT Schema* schema_create(SchemaObject* objects, int objects_length, Sc
                 Property p;
                 p.name = property.name;
                 p.type = property.type;
-                p.object_type = property.object_type ?: "";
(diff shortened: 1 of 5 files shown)


fealebenpae:
👍

wrappers/src/schema_cs.cpp
```diff
@@ -62,7 +62,7 @@ REALM_EXPORT Schema* schema_create(SchemaObject* objects, int objects_length, Sc
                 Property p;
                 p.name = property.name;
                 p.type = property.type;
-                p.object_type = property.object_type ?: "";
+                p.object_type = property.object_type ? property.object_type : "";
```
AndyDentFree: Was this just an aesthetic change?

fealebenpae: Apparently MSVC doesn't like the other syntax. Functionally it's equivalent, though.

tgoyne: `?:` is unfortunately a (rather useful) GNU extension.

PR #681, opened 2016-07-07
Enhanced error handling

### LINKED ISSUE #679: Fix native errorhandling
..so it doesn't rely on the exception callback. This will help #503 and work on Windows

kristiandupont:
This fixes #679.


Realm.Shared/Dynamic/MetaRealmObject.cs
```diff
@@ -50,64 +51,63 @@ public override DynamicMetaObject BindGetMember(GetMemberBinder binder)
             var arguments = new List<Expression>
             {
                 WeakConstant(_metadata.Table),
-                Expression.Field(GetLimitedSelf(), RealmObjectRowHandleField),
-                Expression.Constant(_metadata.ColumnIndices[property.Name])
+                Expression.Constant(_metadata.ColumnIndices[property.Name]),
```
AndyDentFree: Could we start documenting **why** we do some of these things with comments?
This is complex stuff for somebody to come back to or new to the team to learn.

fealebenpae: This needs to build an expression tree that describes a call to any of the `Get*Value/Set*Value` methods and because the parameters in those methods changed, the list of arguments in the expression needs to change as well.

AndyDentFree: I mean in comments that stay in code - review stuff is useless later

AndyDentFree: Is this a criticism of the code or an explanation, which should be saved in code comments rather than here?

fealebenpae: You can say I'm criticizing it through explaining it and thus reinforcing your assertion it needs explaining? 😁

Realm.Shared/Realm.cs
```diff
@@ -432,27 +429,27 @@ internal RealmObject MakeObjectForRow(string className, RowHandle rowHandle)
...
             return CreateResultsHandle(resultsPtr);
         }
 
 
         internal SortOrderHandle MakeSortOrderForTable(string className)
         {
-            var tableHandle = Metadata[className].Table;
-            IntPtr sortOrderPtr = NativeSortOrder.create_for_table(tableHandle);
-            return CreateSortOrderHandle(sortOrderPtr);
+            var result = new SortOrderHandle();
```
AndyDentFree: Why not just have a factory function - no more `new`?

kristiandupont: Very true. There is an additional refactoring that we could do regarding the construction of handles all over. They are wrapped in `ConstrainedExecutionRegion` and generally aren't too intuitive. I just felt that it might be too radical a change in one go, so I kept the old style more or less.

AndyDentFree:
Would you consider a kater performance review on the exception handling pasttern? 
So that we maybe don't pass back the structure on every call that pass a slight earthquake and then call natve to get the exception structure filled. I'm a little worried about the overhead but I don't think it's worth rewriting now for a potentially small gain.

wrappers/src/table_cs.cpp
```diff
@@ -66,33 +60,33 @@ REALM_EXPORT Row* table_get_link(Table* table_ptr, size_t column_ndx, size_t row
...
-REALM_EXPORT SharedLinkViewRef* table_get_linklist(Table* table_ptr, size_t column_ndx, size_t row_ndx)
+REALM_EXPORT SharedLinkViewRef* table_get_linklist(Table* table_ptr, size_t column_ndx, size_t row_ndx, NativeException::Marshallable& ex)
 {
-    return handle_errors([&]() -> SharedLinkViewRef* {
-        SharedLinkViewRef sr = std::make_shared<LinkViewRef>(table_ptr->get_linklist(column_ndx, row_ndx));
-        return new SharedLinkViewRef{ sr };  // weird double-layering necessary to get a raw pointer to a shared_ptr
-    });
+  return handle_errors(ex, [&]() -> SharedLinkViewRef* {
+    SharedLinkViewRef sr = std::make_shared<LinkViewRef>(table_ptr->get_linklist(column_ndx, row_ndx));
+    return new SharedLinkViewRef{ sr };  // weird double-layering necessary to get a raw pointer to a shared_ptr
```
bdash: Moving `sr` into the `SharedLinkViewRef` constructor will avoid an unnecessary increment / decrement of the `shared_ptr`'s refcount.

wrappers/src/table_cs.cpp
```diff
@@ -198,130 +192,145 @@ REALM_EXPORT size_t table_get_nullable_timestamp_milliseconds(const Table* table
...
+{
+    return handle_errors(ex, [&]() {
+        return new Results(*realm, *object_schema, *table_ptr);
+    });
+}
+
+REALM_EXPORT Results* table_create_sorted_results(Table* table_ptr, SharedRealm* realm, ObjectSchema* object_schema, SortOrderWrapper* sortorder_ptr, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        return new Results(*realm, *object_schema, Query(table_ptr->where()), sortorder_ptr->sort_order);
```
bdash: Is the explicit call to the `Query` constructor here necessary? `Table::where` returns a `Query` so it looks redundant to me.

Realm.Shared/handles/LinkListHandle.cs
```diff
@@ -15,19 +15,100 @@
...
+        }
+
         internal LinkListHandle(RealmHandle root) : base(root)
         {
         }
 
         protected override void Unbind()
         {
-            NativeLinkList.destroy(handle);
+            NativeMethods.destroy(handle);
```
AndyDentFree: Can we get an exception here?
I am a bit suspicious of any native calls now which do not have the exception handling, as implicitly they all had it by the previous callback mechanism. It seems to me that because we're destroying reference counted pointers, this could trigger quite a lot of work in core in some cases

kristiandupont: It boils down to `delete liklist_ptr;`, so no exceptions should be thrown here.

Realm.Shared/linq/RealmResults.cs
```diff
@@ -295,16 +296,16 @@ internal static class RealmResultsNativeHelper
...
         #if __IOS__
-        [ObjCRuntime.MonoPInvokeCallback(typeof(NativeResults.NotificationCallback))]
+        [ObjCRuntime.MonoPInvokeCallback(typeof(ResultsHandle.NotificationCallback))]
         #endif
-        internal static void NotificationCallback(IntPtr managedResultsHandle, PtrTo<NativeResults.CollectionChangeSet> changes, PtrTo<NativeException> exception)
+        internal static void NotificationCallback(IntPtr managedResultsHandle, IntPtr changes, IntPtr exception)
         {
             var results = (Interface)GCHandle.FromIntPtr(managedResultsHandle).Target;
-            results.NotifyCallbacks(changes.Value, exception.Value);
+            results.NotifyCallbacks(new PtrTo<ResultsHandle.CollectionChangeSet>(changes).Value, new PtrTo<NativeException>(exception).Value);
```
AndyDentFree: This `NotificationCallback ` bothers me a little bit I am wondering if it is robust if the callbacks triggered anything asynchronous because `new PtrTo<ResultsHandle.CollectionChangeSet` Will now go out of scope at the end of the notify callbacks call. I am not certain if this is any more dangerous than what we had before, however.

Realm.Shared/linq/RealmResultsVisitor.cs
```diff
@@ -389,41 +389,40 @@ private static void AddQueryEqual(QueryHandle queryHandle, string columnName, ob
 
         private static void AddQueryNotEqual(QueryHandle queryHandle, string columnName, object value)
         {
-            var columnIndex = NativeQuery.get_column_index((QueryHandle)queryHandle, columnName, (IntPtr)columnName.Length);
+            var columnIndex = queryHandle.GetColumnIndex(columnName);
 
             var valueType = value.GetType();
```
AndyDentFree: This `GetType` could be moved underneath the string test because it's only used for the other cases

Realm.Shared/native/MarshaledVector.cs
```diff
@@ -0,0 +1,45 @@
...
+        IntPtr count;
+
+        internal IEnumerable<T> AsEnumerable()
+        {
+            return Enumerable.Range(0, (int)count).Select(MarshalElement);
+        }
+
+        unsafe T MarshalElement(int elementIndex)
+        {
+            var @struct = default(T);
```
AndyDentFree: Could we have an exception throw if out of range here on the index?

Tests/IntegrationTests.Shared/PerformanceTests.cs
```diff
@@ -133,6 +133,37 @@ public void BindingSetValuePerformanceTest(int count)
             Console.WriteLine("Kilo-iterations per second: {0:0.00}", ((count/1000) / sw.Elapsed.TotalSeconds));
         }
 
+        string benchmarkString;
+
+        [Test, Explicit("Temporary benchmarker")]
```
AndyDentFree: How about a quick comment saying why this is here and what it was being used to compare against?

AndyDentFree:
👍  would like couple of minor nitpicks from me addressed but that's optional. Huge job very well done!

Tests/IntegrationTests.Shared/PerformanceTests.cs
kristiandupont: That test should probably be removed. It was mostly to get a feel for the change, but it's not a very reliable performance test in general..

PR #699, opened 2016-07-14
Report error when a RealmObject class has no persisted properties

### LINKED ISSUE #689: Error when a RealmObject class has no persisted properties

kristiandupont:
This fixes #689.


fealebenpae:
👍

Tests/WeaverTests/AssemblyToProcess/FaultyClasses.cs
```diff
@@ -18,13 +18,16 @@
...
+using System.Configuration;
 using Realms;
 
 namespace AssemblyToProcess
 {
     public class RealmListWithSetter : RealmObject
     {
         public IList<Person> People { get; set; } 
+
+        public int PropertyToEnsureOtherwiseHealthyClass { get; set; }
```
AndyDentFree: This should not be necessary - `IList` alone is legal

Tests/WeaverTests/AssemblyToProcess/RemappedClass.cs
```diff
@@ -28,5 +28,6 @@ namespace AssemblyToProcess
     [MapTo("RemappedTable")]
```
AndyDentFree: are we dropping `MapTo`? If that is the case then we could  drop this entire test

Tests/WeaverTests/AssemblyToProcess/FaultyClasses.cs
kristiandupont: Not when the `IList` has a `set` method.

Tests/WeaverTests/AssemblyToProcess/RemappedClass.cs
kristiandupont: We never really concluded anything, so I decided to leave it for now.

Tests/WeaverTests/AssemblyToProcess/FaultyClasses.cs
AndyDentFree: doh!

kristiandupont:
Retest this please

PR #724, opened 2016-07-21
Specific error messages for detached rows

kristiandupont:


Realm.Shared/native/NativeException.cs
```diff
@@ -29,22 +29,22 @@ internal unsafe struct NativeException
         public sbyte* messageBytes;
         public IntPtr messageLength;
 
-        internal Exception Convert()
+        internal Exception Convert(Func<RealmExceptionCodes, Exception> overrider)
```
AndyDentFree: This could also default to null? Or do you prefer explicitly requiring people pass in a null `overrider`?

kristiandupont: I am not sure exactly why I felt it was different here. It was a conscious, though hesitant, choice. But since I don't really have a reason, I'll change it :-)

kristiandupont:
Retest this please

AndyDentFree:
👍  **assuming passes CI**

Issue #715, opened 2016-07-18
Weaver fails when weaving a PCL


kristiandupont:

AndyDentFree:
This was **not** fixed in 0.77.0 release

PR #730, opened 2016-07-25
Ad/bug/fix not weaving pcl #715

AndyDentFree:
I think this is a robust fix based on building the PCL project now in the Realm.sln

### DIFF
--- a/RealmWeaver/ModuleWeaver.cs
+++ b/RealmWeaver/ModuleWeaver.cs
@@ -130,7 +130,7 @@ public class ModuleWeaver
 
         Debug.WriteLine("Weaving file: " + ModuleDefinition.FullyQualifiedName);
 
-        var submitAnalytics = System.Threading.Tasks.Task.Factory.StartNew (() => 
+        var submitAnalytics = System.Threading.Tasks.Task.Factory.StartNew(() =>
         {
             var analytics = new RealmWeaver.Analytics(ModuleDefinition);
             try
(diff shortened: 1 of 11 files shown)


kristiandupont:
:+1:

RealmWeaver/ModuleWeaver.cs
```diff
@@ -178,11 +178,12 @@ public void Execute()
         var listTypeDefinition = _corLib.MainModule.GetType("System.Collections.Generic.List`1");
         if (listTypeDefinition == null)
         {
-            var collectionsAssembly = AssemblyResolver.Resolve("System.Collections");
-            listTypeDefinition = collectionsAssembly.MainModule.ExportedTypes.FirstOrDefault(x => x.FullName == "System.Collections.Generic.List`1").Resolve();
-            
+            System_IList = _realmAssembly.MainModule.ImportReference(typeof (System.Collections.Generic.List<>));
```
fealebenpae: Shouldn't this be importing the type reference in the `ModuleDefinition` of the assembly we're currently weaving, as opposed to the Realm assembly?

AndyDentFree: It is a fallback when we couldn't get the main definition so is only done in the PCL - empirically it seems to work.

fealebenpae: I see - it works because we `Resolve()` this `TypeReference` to get its `TypeDefinition` so it doesn't really matter what module imports the reference. Still, I think this could cause subtle bugs later, because this is `IList<>` as referenced from `Realm.dll`, not the woven assembly, and if we ever end up trying to use this reference on the woven assembly things might go wrong.

I strongly recommend replacing this with `ModuleDefinition.ImportReference(typeof(IList<>))` if the latter works.

AndyDentFree: Noted, merged that into the release branch after doing local testing. Rebuilding the NuGet with it now.

PR #735, opened 2016-07-27
Add PCL built schemas into unit tests

AndyDentFree:
Simple tests added on top of the earlier PCL fix to see if could trigger bug seen in `QuickJournal`. however, these work nicely.

ObjectIntegrationTests.cs
- CreateObjectFromPCLTest added

InstanceTests.cs
- GetInstanceFromPCLTest added

IntegrationTests.XamarinAndroid
IntegrationTests.XamarinIOS
- added reference to PurePCLBuildableTest

PurePCLBuildableTest/Class1.cs
- renamed to ObjectInPCL
- added properties of each type

PurePCLBuildableTest
- TestBuildingRealmFromPCL.MakeARealmWithPCL added


Tests/IntegrationTests.Shared/InstanceTests.cs
```diff
@@ -62,6 +62,16 @@ public void GetInstanceWithJustFilenameTest()
             // Arrange, act and "assert" that no exception is thrown, using default location + unique name
             Realm.GetInstance(specialRealmName).Close();
         }
+        
+        [Test]
```
kristiandupont: Might we not just as well move this (and the `ObjectIntegrationTests.CreateObjectFromPCLTest` test below) into the PCL themselves? Seems to me that they belong in there.

Tests/PurePCLBuildableTest/ObjectInPCL.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
```
kristiandupont: Copyright header missing

Tests/IntegrationTests.Shared/InstanceTests.cs
AndyDentFree: Can we put the unit tests inside the PCL? I figured that opened a bigger can of worms plus I wanted to explicitly test calling functions **from** native code, that 

1. use a PCL object on a realm in the native app and
2. go through the whole Realm creation **in** the PCL

kristiandupont:
:+1:

Issue #737, opened 2016-07-28
Document RealmSchema family


AndyDentFree:
RealmSchema was introduced in 0.76 but lacks XML documentation tags so fails to show in IDE or Doxygenerated API

See also ObjectSchema.cs

PR #738, opened 2016-07-29
Fix comment errors

AndyDentFree:
very trivial changes. Substantially more needs to be written for #737

### DIFF
--- a/Realm.Shared/RealmList.cs
+++ b/Realm.Shared/RealmList.cs
@@ -26,10 +26,14 @@ using System.Dynamic;
 namespace Realms
 {
     /// <summary>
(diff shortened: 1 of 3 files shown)


Realm.Shared/RealmList.cs
```diff
@@ -26,11 +26,15 @@
 namespace Realms
 {
     /// <summary>
-    /// Used to declare to-many relationships and as the return type when you access such a relationship.
+    /// Return type for a managed object property when you declare a to-many relationships with IList.
```
kristiandupont: Typo: either singular or plural.

kristiandupont:
:+1:

PR #742, opened 2016-08-01
Ad/feature/refactor schema logic

AndyDentFree:
This is a refactor that happened on the way to another (simpler) refactor, as I wanted to remove just one common dictionary lookup and it cascaded a little.

So I thought it worth putting up for review and hopefully merging by itself.

### DIFF
--- a/Realm.PCL/RealmListPCL.cs
+++ b/Realm.PCL/RealmListPCL.cs
@@ -222,8 +222,6 @@ namespace Realms
 
         LinkListHandle IRealmList.Handle { get; }
 
-        string IRealmList.SchemaClassName { get; }
-
         #endregion
 
     }
@@ -238,7 +236,6 @@ namespace Realms
     {
         Realm Realm { get; }
         LinkListHandle Handle { get; }
-        string SchemaClassName { get; }
     }
 
 }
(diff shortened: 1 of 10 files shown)


fealebenpae:
Why not just refactor so that we pass `RealmObject.Metadata` everywhere? Be it for lists, results, or whatever, at some point we have the metadata object on hand - either for the table handle or the schema handle. We don't have any paths where we get names from `typeof(T).Name` or the like anymore, so maybe we can just trickle down the metadata object everywhere.

AndyDentFree:
I was mulling over taking it that far but wanted to get feedback and see if this is an acceptable PR by itself with that as a later cleanup. I also wasn't sure if some of the dynamic stuff you are doing might conflict.

fealebenpae:
Dynamic is pretty much done now. You only need to worry about it if you change any of the `RealmObjectOps` method signatures anyway.

Personally I don't think those dictionary lookups have a measurable impact, but if you are up for reducing them I'm all for it. My only concern is that we might be better off going with a single argument type across the number of `Create*` methods we have internally. Right now it's a mix between `string`, `Schema.Object` and `RealmObject.Metadata` so I think it would be good for our sanity if we just went with `RealmObject.Metadata` across the board.

kristiandupont:
:+1:

PR #751, opened 2016-08-04
Split up close and dispose

### LINKED ISSUE #746: Mysterious detached-row exceptions
A number of users have been experiencing detached row exceptions in data binding scenarios. This started appearing after the release of 0.77.0/0.77.1.

kristiandupont:
This fixes #746, hopefully..


Realm.Shared/Realm.cs
```diff
@@ -246,36 +246,50 @@ private void NotifyChanged(EventArgs e)
...
+        {
+            GC.SuppressFinalize(this);
+            Dispose(true);
+        }
 
         /// <summary>
         ///  Dispose automatically closes the Realm if not already closed.
         /// </summary>
-        public void Dispose()
+        public void Dispose(bool disposing)
```
nirinchev: This should probably be private, as it will be confusing for end-users to have multiple dispose methods.

kristiandupont: Good point! That was a slip.

Tests/IntegrationTests.Shared/LifetimeTests.cs
```diff
@@ -0,0 +1,55 @@
...
+            GC.WaitForPendingFinalizers();
+
+            // Assert
+            Assert.That(realm.IsAlive);
+            Assert.That(((Realm)realm.Target).IsClosed, Is.False);
+            Assert.That(person.IsValid);
+        }
+
+        [Test]
+        public void FinalizedRealmsShouldNotInvalidateSiblingRealms()
```
AndyDentFree: Should we allow people to have a realm instance in a `using` statement and another realm on the same file, in the same thread, and expect the second to remain open?

AndyDentFree:
👍  noting the `using` scenario [I commented on ](https://github.com/realm/realm-dotnet/pull/751#discussion_r73534398) is being split off to be tacked separately.

Tests/IntegrationTests.Shared/LifetimeTests.cs
kristiandupont: You are right. Registered as #759.

UKDeveloper99:
So if I get this in the next release it should all work out of the box? What about getting Realm instances, with GetInstance("SomeRealm"). What's the best practice for doing that now in a PCL project (for example with MvvmCross like I was using). Do I still want a singleton to access the Realm objects?

```
 private static Realm _realm;
        public static Realm db {
            get {
                if (_realm == null)
                    _realm = Realm.GetInstance("KittensRealm.realm");
                return _realm;
            }
        }
```

kristiandupont:
@UKDeveloper99 It should, yes. It's something we will monitor further, especially because disposing a `Realm` instance will close the database, effectively invalidating other instances in the same thread. There is an issue registered for this: #759. 
But your specific problem should be resolved with this change and hopefully make things a bit easier to reason about.

PR #760, opened 2016-08-08
Ignore android resource.designer.cs

kristiandupont:


fealebenpae:
👍

.gitignore
```diff
@@ -243,3 +243,5 @@ RealmWeaver.Fody.dll
 wrappers/libs-dbg
 
 *.xam
+
+Tests/IntegrationTests.XamarinAndroid/Resources/Resource.designer.cs
```
AndyDentFree: Yes in spirit but after discussion on Slack, would rather we ignore it and remove it everywhere.

PR #762, opened 2016-08-08
Pull in re-linq and use its partial expression evaluation

fealebenpae:


Realm.XamarinAndroid/packages.config
```diff
@@ -1,4 +1,16 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="DotNetCross.Memory.Unsafe" version="0.2.2" targetFramework="monoandroid60" />
+  <package id="Remotion.Linq" version="2.1.1" targetFramework="MonoAndroid60" />
```
AndyDentFree: Should we be using `targetFramework="MonoAndroid44"` instead?

Tests/IntegrationTests.XamarinAndroid/packages.config
```diff
@@ -2,4 +2,16 @@
 <packages>
   <package id="DotNetCross.Memory.Unsafe" version="0.2.2" targetFramework="monoandroid51" />
   <package id="Fody" version="1.29.4" targetFramework="MonoAndroid44" developmentDependency="true" />
+  <package id="Remotion.Linq" version="2.1.1" targetFramework="MonoAndroid51" />
```
AndyDentFree: This is another `MonoAndroid` level inconsistent with `Realm.XamarinAndroid/packages.config`

kristiandupont:
Sweet!

Realm.XamarinAndroid/packages.config
fealebenpae: I think that's due to the value of the `TargetFrameworkVersion` property. In `Realm.XamarinAndroid` it's set to Latest, whereas in `IntegrationTests.XamarinAndroid` it's set to an older version.

Tests/IntegrationTests.XamarinAndroid/packages.config
fealebenpae: See comment above.

Realm.XamarinAndroid/packages.config
fealebenpae: We actually get a warning about this when building, but it's sort of benign, seeing as how we don't actually use any Android APIs.

kristiandupont:
:+1:

PR #763, opened 2016-08-08
Ad/fody log more

AndyDentFree:
Logging change only for when you are running RealmWeaver

### DIFF
--- a/RealmWeaver/ModuleWeaver.cs
+++ b/RealmWeaver/ModuleWeaver.cs
@@ -26,9 +26,9 @@ using System.Diagnostics;
 
 public class ModuleWeaver
 {
-    // Will log an informational message to MSBuild
-    public Action<string> LogDebug { get; set; } = m => { };
-    public Action<string> LogInfo { get; set; } = m => { };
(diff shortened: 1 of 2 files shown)


RealmWeaver/ModuleWeaver.cs
```diff
@@ -311,7 +309,13 @@ private void WeaveType(TypeDefinition type, Dictionary<string, Tuple<MethodRefer
             return false;
         }
 
-        if (!prop.IsAutomatic())
+        if (prop.IsAutomatic()) 
+        {
+            var objectIdMsg = isObjectId ? "[ObjectId]" : "";
+            var indexedMsg = isIndexed ? "[Indexed]" : "";
+            LogDebug($"Woven {type.Name}.{prop.Name} as a {prop.PropertyType.FullName} {objectIdMsg} {indexedMsg}");
```
kristiandupont: Shouldn't this log statement be at the bottom of `WeaveProperty`? There are conditions below that might cause the property not to be woven, like `IList` with setter, etc..

kristiandupont:
:+1:

Issue #739, opened 2016-07-29
Xamarin Studio Only Error - Cannot build an empty ObjectSchema with encryption


m-mccool:
#### Goal

To be able to run an iOS application from Xamarin Studio using Realm 0.77.1

#### Expected Results

The app runs

#### Actual Results

An exception is raised when attempting to retrieve an encrypted realm

System.InvalidOperationException Cannot build an empty ObjectSchema at Realms.Schema.ObjectSchema
This works on an application built using Visual Studio, it only seems to happen with an application built using Xamarin studio

#### Steps & Code to Reproduce

> Describe your current debugging efforts.

#### Code Sample

The code I am running is similar to that below

```
var config = new RealmConfiguration("MyEncrypted.realm");
                var encryptionKey = new byte[64] // key MUST be exactly this size
                    {
                  0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
                  0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
                  0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
                  0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
                 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
                  0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
                  0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
                  0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
                    };

                config.EncryptionKey = encryptionKey;
                var realm = Realm.GetInstance(config);
                return realm;
```

It throws the exception on the Realm.GetInstance(config) line

#### Version of Realm and tooling
Realm 0.77.1

Visual Studio 2015 Update 3 v14.0.25424.00 with
Xamrin 4.1.1.3
Xamarin Android 6.1.1.1
Xamarin iOS 9.8.1.4

Xamarin Studio
=== Xamarin Studio Community ===

Version 6.0.1 (build 8)
Installation UUID: c0a4c8dc-0061-4b22-a3ad-b4e6bc68b1db
Runtime:
	Mono 4.4.1 (mono-4.4.0-branch-c7sr0/4747417) (64-bit)
	GTK+ 2.24.23 (Raleigh theme)

[...]

PR #764, opened 2016-08-09
#739 Better empty schema exception messages

AndyDentFree:
ObjectSchema.cs
- Builder.Build - throw message identifying bad schema name with URL for docs

RealmSchema.cs
- Builder.Build - throw message saying no RealmObjects, with URL for docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 0.77.x (TBD)
 ### Enhancements
 * Setting your **Build Verbosity** to `Detailed` or `Normal` will now display a message for every property woven, which can be useful if you suspect errors with Fody weaving.
+* Better exception messages will helo diagnose _EmptySchema_ problems (#739)
 
 ### Bug fixes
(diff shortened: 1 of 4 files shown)


fealebenpae:
Is the second `throw new InvalidOperationException("Cannot build an empty ObjectSchema")` necessary?

AndyDentFree:
There was already an existing exception there, and yes it is necessary. One is triggered if the properties are missing, the other if no RealmObjects - the latter can happen also if there is a Fody error and the fact that they were both the same message previously was confusing the diagnosis of what was going wrong.

kristiandupont:
I think what @fealebenpae is pointing out is that there seems to be two consecutive throw statements in both situations. I think you forgot to delete the old ones?

AndyDentFree:
Sigh, sorry, no I did delete them but obviously re-applying a stash to a new branch didn't quite merge the way I thought it would.

fealebenpae:
There's another leftover unstashing cruft in `RealmSchema.cs`.

kristiandupont:
:+1:

PR #766, opened 2016-08-09
Null comparisons in LINQ

### LINKED ISSUE #517: IsNull Support
The NativeQuery doesn't have a IsNull method.

This is very important to query optional properties such as `int?`

It should have a 

`internal static extern void is_null(QueryHandle queryPtr, IntPtr columnIndex);`

and then we can complement the https://github.com/realm/realm-dotnet/pull/515

```
private static void AddQueryEqual(QueryHandle queryHandle, string columnName, object value)
		{
			var columnIndex = NativeQuery.get_column_index((QueryHandle)queryHandle, columnName, (IntPtr)columnName.Length);

			var valueType = value?.GetType();
			if (value == null)
				NativeQuery.is_null ((QueryHandle)queryHandle, columnIndex);
			else if (value is string)
			{
				string valueStr = (string)value;
				NativeQuery.string_equal((QueryHandle)queryHandle, columnIndex, valueStr, (IntPtr)valueStr.Length);
			}
			else if (valueType == typeof(bool))
				NativeQuery.bool_equal((QueryHandle)queryHandle, columnIndex, MarshalHelpers.BoolToIntPtr((bool)value));
			else if (valueType == typeof(int))
				NativeQuery.int_equal((QueryHandle)queryHandle, columnIndex, (IntPtr)((int)value));
			else if (valueType == typeof(float))
				NativeQuery.float_equal((QueryHandle)queryHandle, columnIndex, (float)value);
			else if (valueType == typeof(double))
				NativeQuery.double_equal((QueryHandle)queryHandle, columnIndex, (double)value);
			else if (valueType == typeof(DateTimeOffset))
				NativeQuery.datetime_seconds_equal(queryHandle, columnIndex, ((DateTimeOffset)value).ToUnixTimeSeconds());
			else
				throw new NotImplementedException();
		}
``` 

I  cann't to that myself because I don't know where to start :D

### LINKED ISSUE #570: Support null comparison in LINQ queries
Similar to #517 

However, `Query::is_null` works only for nullable columns. `byte[]` and perhaps `string` properties will need additional work.

fealebenpae:
Fixes #570 
Closes #517

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 ### Enhancements
 * Setting your **Build Verbosity** to `Detailed` or `Normal` will now display a message for every property woven, which can be useful if you suspect errors with Fody weaving.
 * Better exception messages will helo diagnose _EmptySchema_ problems (#739)
+* Support for LINQ queries that check for `null` against `string`, `byte[]` and `Nullable<T>` properties.
+* Support for `string.IsNullOrEmpty` on persisted properties in LINQ queries.
 
 ### Bug fixes
 * `RealmResults<T>` should implement `IQueryable.Provider` implicitly (#752)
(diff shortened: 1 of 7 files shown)


kristiandupont:
:+1:

AndyDentFree:
👍

PR #767, opened 2016-08-09
Update the changelog for #762

### LINKED ISSUE #525: LINQ variable expressions not fully supported
Benchmarkr uses a constant  in `examples/Benchmarkr/Realm/Benchmark.cs` the query in `ConvertQuery` has a constant and comment that variable argument is still not supported.

@fealebenpae said _it still doesn’t work, because we don’t do recursive constant folding_

### LINKED ISSUE #755: Cannot use another linq parameter (The rhs of the binary operator 'Equal' should be a constant or closure variable expression.)
#### Goal
I want to manipulate a variable coming from another linq query. I currently have an exception (The rhs of the binary operator 'Equal' should be a constant or closure variable expression.) when doing this use case.

```csharp
var myList = new List<MyObject>();
var myRealmList = realm.All<MyRealmObject>();

// this does not work
myList.Where(mo => myRealmList.Any(mro => mro.Field == mo.Field));

// this works (my current workaround)
myList.Where(mo => 
{
   var field = mo.Field;
   return myRealmList.Any(mro => mro.Field == field);
})
```

Realm version: 0.77.1
Xamarin/Visual Studio version: VS Enterprise 2015

Which operating system version and device: ?
iOS (iPhone or Simulator), dev on Windows and Visual studio

#### Conclusion
Hoping this is not a duplicate (I did not find anything coming from this exact use case).
Thanks again :)

fealebenpae:
Fixes #755
Closes #525

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ### Enhancements
 * Setting your **Build Verbosity** to `Detailed` or `Normal` will now display a message for every property woven, which can be useful if you suspect errors with Fody weaving.
 * Better exception messages will helo diagnose _EmptySchema_ problems (#739)
+* Partial evaluation of LINQ expressions means more expressions types are supported as operands in binary expressions (#755)
 * Support for LINQ queries that check for `null` against `string`, `byte[]` and `Nullable<T>` properties.
 * Support for `string.IsNullOrEmpty` on persisted properties in LINQ queries.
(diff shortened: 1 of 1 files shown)


AndyDentFree:
👍

kristiandupont:
:+1:

PaulARoy:
:+1:

Issue #518, opened 2016-05-08
Explicit schema versions


kristiandupont:

AndyDentFree:
Is this needed to make migrations robust and especially fix issues with trivial migration?

fealebenpae:
The migration callbacks implement their business logic conditionally based on the schema version numbers. We already have the API needed for users to set the schema version, we should just encourage its use, IMO.

AndyDentFree:
We have already, and recently, confirmed these work for some SO questions and help requests. I propose a quick check to confirm initial value and then add some main documentation will be enough.

AndyDentFree:
We set our initial version in `RealmConfiguration`:

```
// to match ObjectStore::NotVersioned
 public static UInt64 NotVersioned => UInt64.MaxValue;

...
public UInt64 SchemaVersion { get; set;} = RealmConfiguration.NotVersioned;
```

However, that was a misunderstanding of how  Cocoa uses the ObjectStore `std::numeric_limits<uint64_t>::max`. We should be using zero to match Cocoa and Java.

fealebenpae:
Implemented in #768

PR #768, opened 2016-08-09
#518 Start Schema Version at zero

AndyDentFree:
Relatively trivial change but needed a bit of test updating including of course our inbuilt realm, so it now has a zero version.
@kristiandupont or @fealebenpae  to review.

RealmConfiguration.cs
- NotVersioned removed
- SchemaVersion now initialised to zero

RealmConfigurationPCL.cs
- NotVersioned removed

ObjectIntegrationTests.cs
- TriggerMigrationBySchemaVersion expect zero initial version

ForMigrationsToCopyAndMigrate.realm
- updated standard database with initial (zero) schema version

RealmMigrationTests.cs
- refactored out from ObjectIntegrationTests.cs
- MigrationTriggersDelete - disabled as it is not triggered due issue 507


AndyDentFree:
retest this please

kristiandupont:
:+1:

m-mccool:
Hi, @AndyDentFree, What happens if I am attempting to upgrade an app. V0 uses realm 0.77.1 and V1 uses V0.88.1. At the moment the V1 app does not seem to detect that a migration is required. It throws an exception, instead of entering the migration callback. Should this work? Is there anyway of forcing a migration?

AndyDentFree:
@m-mccool depending on the exception, there may be other issues.

Have you increased the [SchemaVersion](https://realm.io/docs/xamarin/0.82.0/api/reference/Realms.RealmConfigurationBase.html#Realms_RealmConfigurationBase_SchemaVersion) number in the new app? If your changes are additive (new classes or properties) it should be able to migrate automatically with a new number.

m-mccool:
I think the exception was my issue when experimenting.

Under more controlled circumstances I ran it this morning it upgraded without an exception but did not call the handler. So I experimented and removed a field from a realm object class in the new version, It did not throw an exception but the data in that field would have been lost. What I would have wanted to do was be able to access that data from the deleted field and maybe append it to another field.

Is this an issue because my old app I want to upgrade uses a version of realm before the migration feature was added? Or should it call the migration callback

AndyDentFree:
@m-mccool see #1186 - this needed its own issue.

Issue #507, opened 2016-05-03
RealmMigrationNeededException no longer thrown


AndyDentFree:
This means that our flag `ShouldDeleteIfMigrationNeeded` is meaningless. See test `RealmMigrationTests.MigrationTriggersDelete`

AndyDentFree:
Note that this changed (again?) and as of our released 0.74.1 is throwing a `RealmMismatchedConfigException`

AndyDentFree:
As per Slack discussion this is possibly fixed by newer cores 0.100.0 onwards

AndyDentFree:
As of core 1.4.0 and our current master (release 0.77.2 plus goodies) we are getting a general `RealmException` with a message such as _Migration is required due to the following errors: - Property 'Silly' has been added to latest object model_

PR #771, opened 2016-08-11
Ad/bug migration needed exception #507

AndyDentFree:
Minor change just to catch the exception currently being thrown from ObjectStore and map it to our desire C# exception.

@fealebenpae or @kristiandupont please review

**NOTE** this does **not** fix our `MigrationTriggersDelete` test as that crashes due to schema issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,12 @@
 0.77.x (TBD)
+-------------------
+### Bug fixes
+* Now throws a RealmMigrationSchemaNeededException if you have changed a `RealmObject` subclass declaration and not incremented the `SchemaVersion` (#518)
+
+Uses core 1.4.2
+
+0.77.2 (2016-08-11)
+-------------------
 ### Enhancements
 * Setting your **Build Verbosity** to `Detailed` or `Normal` will now display a message for every property woven, which can be useful if you suspect errors with Fody weaving.
 * Better exception messages will helo diagnose _EmptySchema_ problems (#739)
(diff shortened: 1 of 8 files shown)


fealebenpae:
I think we need a new error type for this: `FileFormatUpgradeRequired ` will be thrown by core when the realm uses a deprecated file format in need of upgrade and automatic updates have been explicitly disabled, and it will be translated by object store to `FormatUpgradeRequired`. It has nothing to do with the schema in object store.

AndyDentFree:
Well, this is a relatively quick fix for a common user mistake and to align with what is being thrown by ObjectStore right now. I can dig deeper and push for a commonly accepted OS fix to do a unique exception but that is (literally) a separate issue.

fealebenpae:
I don't propose changing object store - just add a new member to the enum in `realm_error_type.hpp` and use that. OS already has distinct exception types - we're the ones that conflate them under a single error type when marshalling to c#.

AndyDentFree:
> I don't propose changing object store

The point is that objectStore is throwing the _wrong_ exception in the first place.

fealebenpae:
> objectStore is throwing the wrong exception in the first place

Oh, I was under the impression that `SchemaMismatchException` is the intended exception type.

AndyDentFree:
Sorry, you are right - I was looking at old notes - the thing which **is** wrong is that we are overloading `FileFormatUpgradeRequired` because previously, that **was** the _wrongly returned_ error.

fealebenpae:
👍

kristiandupont:
:+1:

PR #772, opened 2016-08-11
Core 1.5.0 with matching ObjectStore and SortDescriptor

### LINKED ISSUE #750: Use Object Store as a git submodule

kristiandupont:
~~This resolves (well, it used to) #750. Core is bumped to 1.5.0 as well, as the current version of object store requires it.~~
This is now solely to bring us up to core 1.5.0 and the matching object store. I will make a new PR for the submodule transition.


Realm.Shared/handles/QueryHandle.cs
```diff
@@ -182,10 +183,13 @@ private static class NativeMethods
             public static extern IntPtr count(QueryHandle QueryHandle, out NativeException ex);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "query_create_results", CallingConvention = CallingConvention.Cdecl)]
-            public static extern IntPtr create_results(QueryHandle queryPtr, SharedRealmHandle sharedRealm, IntPtr objectSchema, out NativeException ex);
+            public static extern IntPtr create_results(QueryHandle queryPtr, SharedRealmHandle sharedRealm, out NativeException ex);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "query_create_sorted_results", CallingConvention = CallingConvention.Cdecl)]
-            public static extern IntPtr create_sorted_results(QueryHandle queryPtr, SharedRealmHandle sharedRealm, IntPtr objectSchema, SortOrderHandle sortOrderHandle, out NativeException ex);
+            public static extern IntPtr create_sorted_results(QueryHandle queryPtr, SharedRealmHandle sharedRealm, IntPtr tablePtr,
```
fealebenpae: You can use `TableHandle` here instead of `IntPtr` - handle types are only limited when used as members of a type, but are fine as parameters.

kristiandupont: Good point, I just moved it from the struct but obviously I can use the handle here.

kristiandupont:
@AndyDentFree and/or @fealebenpae please review this.

Realm.Shared/native/SortDescriptorBuilder.cs
```diff
@@ -0,0 +1,87 @@
...
+        private readonly RealmObject.Metadata _metadata;
+
+        public TableHandle TableHandle => _metadata.Table;
+
+        public SortDescriptorBuilder(RealmObject.Metadata metadata)
+        {
+            _metadata = metadata;
+        }
+
+        // This method adds a direct sort clause. To add clauses that follow links,
```
fealebenpae: What does this comment mean?
If the `ColumnIndexChain` needs to be further mutated, the `AddClause` method should return the new `Clause` instance so that consumers of the API can mutate its `ColumnIndexChain` property?

Realm.Shared/linq/RealmResults.cs
```diff
@@ -124,12 +124,15 @@ private ResultsHandle CreateResultsHandle()
             {
                 return _realm.MakeResultsForTable(_targetMetadata);
             }
-            // do all the LINQ expression evaluation to build a query
-            var qv = _provider.MakeVisitor();
-            qv.Visit(Expression);
-            var queryHandle = qv._coreQueryHandle; // grab out the built query definition
-            var sortHandle = qv._optionalSortOrderHandle;
-            return _realm.MakeResultsForQuery(ObjectSchema, queryHandle, sortHandle);
+            else
```
AndyDentFree: This  doesn't need to be an else block unless for some reason you prefer them aesthetically. When I see a return and then an else I wonder if there was a mistake and look twice at the code.

Realm.Shared/native/SortDescriptorBuilder.cs
kristiandupont: Yeah, that *is* unclear, true. It's just to say that this method currently doesn't support creating an array of indices but just one level. Once we want to support sorting via links, we will have to add a method that takes a collection of indices instead of just one. I'll change it.

Realm.Shared/linq/RealmResults.cs
kristiandupont: Right, that was the result of a conflict merge. I'll remove the else statement.

fealebenpae:
👍

Realm.Shared/native/SortDescriptorBuilder.cs
```diff
@@ -0,0 +1,87 @@
...
+            _clauses.Add(new Clause {ColumnIndexChain = new List<IntPtr> { columnIndex }, Ascending = @ascending});
+        }
+
+        /// <summary>
+        /// Create a flattened array of all the clauses by concatenating the lists.
+        /// </summary>
+        /// <returns>
+        /// A tuple with two elements. Item1 is the concatenated list of indices. Item2 is the list of clauses,
+        /// indicating the offset and count, as well as the ascending flag for each clause.
+        /// </returns>
```
AndyDentFree: This comment explains what the code does without really explaining **why** you want to flatten things. It needs to say something like:

_When you have more than one indice, all but the last one are link fields, so following the chain to sort by_ `dog.owner.name` _would need two link field indices then the index of_ `name` _in_ `Person`.

wrappers/src/marshalable_sort_clause.hpp
```diff
@@ -0,0 +1,40 @@
...
+struct MarshalableSortClause {
+    size_t offset;
+    size_t count;
+    bool ascending;
+};
+
+inline void unflatten_sort_clauses(MarshalableSortClause* sort_clauses, size_t clause_count, size_t* flattened_column_indices, 
+    std::vector<std::vector<size_t>>& column_indices, std::vector<bool>& ascending)
+{
+    for(auto i = 0; i < clause_count; ++i) {
```
AndyDentFree: Being nit-picking, `ascending` could reserve space equal to `clause_count`

wrappers/src/object-store/src/binding_context.hpp
```diff
@@ -1,6 +1,6 @@
 ////////////////////////////////////////////////////////////////////////////
 //
-// Copyright 2016 Realm Inc.
+// Copyright 2015 Realm Inc.
```
AndyDentFree: Are we supposed to be Copyright 2015 or 2016?

wrappers/src/object-store/src/impl/android/weak_realm_notifier.cpp
```diff
@@ -110,16 +110,13 @@ void WeakRealmNotifier::notify()
...
         // this is a pointer to a heap-allocated weak Realm pointer created by the notifiying thread.
         // the actual address to the pointer is communicated over a pipe.
         // we have to delete it so as to not leak, using the same memory allocation facilities it was allocated with.
         std::weak_ptr<Realm>* realm_ptr = nullptr;
         while (read(fd, &realm_ptr, sizeof(realm_ptr)) == sizeof(realm_ptr)) {
             if (auto realm = realm_ptr->lock()) {
-                if (!realm->is_closed()) {
-                    realm->notify();
-                }
+                realm->notify();
```
AndyDentFree: Unconditional notification seems a bit of a step backwards - do our notifications handle being notified by a closed realm now?

AndyDentFree: [Yes they do](https://github.com/realm/realm-object-store/pull/105)

wrappers/src/shared_realm_cs.cpp
```diff
@@ -68,7 +68,7 @@ REALM_EXPORT SharedRealm* shared_realm_open(Schema* schema, uint16_t* path, size
 
         Realm::Config config;
         config.path = pathStr.to_string();
-        config.read_only = read_only;
+        //config.read_only = read_only;
```
AndyDentFree: If this was intentionally commented out, please add a note explaining why.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -77,7 +77,11 @@ REALM_EXPORT SharedRealm* shared_realm_open(Schema* schema, uint16_t* path, size
         else
           config.encryption_key = std::vector<char>(encryption_key, encryption_key+64);
 
-        config.schema.reset(schema);
+        if (read_only) {
+            config.schema_mode = SchemaMode::ReadOnly;
+        }
+        
+        config.schema.emplace(*schema); // TODO: This copies the schema, so the handle kept in .net is wrong.
```
AndyDentFree: Please add an issue to ensure this TODO is followed up

fealebenpae: https://github.com/realm/realm-dotnet/issues/789

AndyDentFree:
A lot of `std::runtime_exception` throws have been replaced with `std:: logic_error` but neither is handled **specifically** by `error_handling.cpp`. They will be picked up by the generic default for `std::exception`. So not really a problem but wonder if we should add a clause for them.

wrappers/src/shared_realm_cs.cpp
AndyDentFree: I see replacement code a few lines below so this looks like cruft that should have just been deleted.

AndyDentFree:
👍 with a bit of cleanup as noted above

wrappers/src/object-store/src/binding_context.hpp
kristiandupont: This just brings us in sync with object store master.

PR #773, opened 2016-08-12
Remove ToList from expressions mostly in RelationshipTests

AndyDentFree:
so we don't give people the idea it is needed.

Reformatted to undo the effects of Xamarin Studio silly settings at a previous edit.

@kristiandupont please review, just test changes, no functional changes so no changelog entry.

### DIFF
--- a/Tests/IntegrationTests.Shared/RelationshipTests.cs
+++ b/Tests/IntegrationTests.Shared/RelationshipTests.cs
@@ -15,7 +15,7 @@
 // limitations under the License.
 //
 ////////////////////////////////////////////////////////////////////////////
- 
+
 using NUnit.Framework;
 using Realms;
 using System;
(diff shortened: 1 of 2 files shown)


Tests/IntegrationTests.Shared/RelationshipTests.cs
```diff
@@ -295,56 +302,56 @@ public void TestExceptionsFromEmptyListOutOfRange()
...
+
             using (var trans = realm.BeginWrite())
             {
                 owner.TopDog = dog;
                 trans.Commit();
             }
 
-            var dogAgain = realm.All<Dog>().Where(d => d.Name == "Astro").ToList().SingleOrDefault();
+            var dogAgain = realm.All<Dog>().Single(d => d.Name == "Astro");
             Assert.That(dogAgain, Is.Not.Null);
```
kristiandupont: Technically these assertions are now redundant as `Single` cannot return `null`. But it's not a big deal..

kristiandupont:
:+1:

PR #775, opened 2016-08-15
Refactoring MakeObject logic to assist ObjectID work and reduce lookups

AndyDentFree:
adding back in after was removed prior to 0.77.1, plus some fixes
@fealebenpae and @kristiandupont this is essentially what was reviewed and passed as #742 but that had a nasty recursive bug which has required a few minor changes in this PR.

Realm.cs
- MakeObjectForRow overload added to take a RealmObject.Metadata instead of doing lookup, and simple rowPtr
- MakeObjectForRow original taking string, changed to take rowPtr for second arg and forward to new overload
- MakeResultsForTable take metadata instead of class name
- MakeSortOrderForTable take metadata instead of class name
- All(both overloads) - lookup metadata for RealmResults

RealmList.cs
- RealmList
  - added field _targetMetadata
  - use new MakeObjectForRow
  - IRealmList.SchemaClassName removed
  - _schemaClassName removed
  - ctor take a metadata param instead of className and set _targetMetadata
- IRealmList - removed SchemaClassName

RealmResults.cs
- added field _targetMetadata
- this[int index]
  - init _targetMetadata if null
  - use new MakeObjectForRow
- CreateResultsHandle
  - init _targetMetadata if null
- MakeResultsForTable take metadata instead of name
- RemoveAll - lookup metadata for MakeResultsForTable
- ctor take Metadata instead of schema

RealmResultsProvider.cs
- _schema replaced by _metadata
- ctor take Metadatata instead of schema

RealmObjectOps.cs
-  GetObjectValue<T>(string propertyName) pass only rowPtr to MakeObjectForRow
- GetList lookup metadata for RealmList ctor

RealmObject.cs
-  GetObjectValue<T>(string propertyName) pass only rowPtr to MakeObjectForRow
- GetListValue<T>(string propertyName) lookup metadata for RealmList ctor

RealmResultsEnumerator.cs
- MoveNext
  - quicker check for failure knowing a null rowPtr returned

RealmResultsVisitor.cs
- _schema removed
- _metadata added
- VisitMethodCall
  - quicker check for failure knowing a null rowPtr returned

RealmListPCL.cs
- removed IRealmList.SchemaClassName

### DIFF
--- a/Realm.Shared/Realm.cs
+++ b/Realm.Shared/Realm.cs
@@ -431,18 +431,36 @@ namespace Realms
             return result;
         }
 
-        internal RealmObject MakeObjectForRow(string className, RowHandle rowHandle)
+        internal RealmObject MakeObjectForRow(RealmObject.Metadata metadata, IntPtr rowPtr)
         {
-            var metadata = Metadata[className];
             RealmObject ret = metadata.Helper.CreateInstance();
-            ret._Manage(this, rowHandle, metadata);
+            ret._Manage(this, CreateRowHandle(rowPtr, SharedRealmHandle), metadata);
+            return ret;
+        }
+
+        internal RealmObject MakeObjectForRow(string className, IntPtr rowPtr)
+        {
+            return MakeObjectForRow(Metadata[className], rowPtr);
+        }
+
+
+        // extra version retained for callers which already have a RowHandle
+        internal RealmObject MakeObjectForRow(string className, RowHandle row)
+        {
+            return MakeObjectForRow(Metadata[className], row);
+        }
+
+
+        internal RealmObject MakeObjectForRow(RealmObject.Metadata metadata, RowHandle row)
+        {
+            RealmObject ret = metadata.Helper.CreateInstance();
+            ret._Manage(this, row, metadata);
             return ret;
         }
 
 
-        internal ResultsHandle MakeResultsForTable(string className)
+        internal ResultsHandle MakeResultsForTable(RealmObject.Metadata metadata)
         {
-            var metadata = Metadata[className];
             var resultsPtr = NativeTable.CreateResults(metadata.Table, SharedRealmHandle, metadata.Schema.Handle);
             return CreateResultsHandle(resultsPtr);
         }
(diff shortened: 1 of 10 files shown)


Realm.Shared/Realm.cs
```diff
@@ -718,7 +736,7 @@ public void RemoveAll()
 
             foreach (var @object in Schema)
```
fealebenpae: This could be rewritten to foreach the metadata dictionary, which would make it more efficient.

Realm.Shared/Realm.cs
```diff
@@ -642,7 +660,7 @@ public bool Refresh()
             if (schema?.Type != type)
                 throw new ArgumentException($"The class {type.Name} is not in the limited set of classes for this realm");
             
-            return new RealmResults<T>(this, schema, true);
+            return new RealmResults<T>(this, Metadata[typeof(T).Name], true);
```
fealebenpae: Since we need the Metadata object here, this could perform a single Metadata lookup:
```csharp
RealmObject.Metadata metadata;
if (!Metadata.TryGetValue(type.Name, out metadata) || metadata.Schema.Type != type)
{
    throw new ArgumentException(...);
}

return new RealmResults<T>(this, metadata, true);
```

Realm.Shared/Realm.cs
```diff
@@ -656,7 +674,7 @@ public RealmResults<dynamic> All(string className)
             if (schema == null)
                 throw new ArgumentException($"The class {className} is not in the limited set of classes for this realm");
 
-            return new RealmResults<dynamic>(this, schema, true);
+            return new RealmResults<dynamic>(this, Metadata[className], true);
```
fealebenpae: Same as above.

Realm.Shared/Realm.cs
```diff
@@ -431,18 +431,36 @@ private RealmObject CreateObject(string className, out RealmObject.Metadata meta
             return result;
         }
 
-        internal RealmObject MakeObjectForRow(string className, RowHandle rowHandle)
+        internal RealmObject MakeObjectForRow(RealmObject.Metadata metadata, IntPtr rowPtr)
```
fealebenpae: I think this method can be refactored to call `MakeObjectForRow(metadata, CreateRowHandle(rowPtr, SharedRealmHandle))` instead.

AndyDentFree: Fair point - the overload taking a metadata and rowHandle was just added in this merge as part of fixing a bug in RealmResults.

fealebenpae:
👍

Issue #402, opened 2016-02-19
Ensure we can optimally retrieve by PrimaryKey


AndyDentFree:
#To use objects in different threads we need the commnon approach of being able to pass the ~ObjectId~ PrimaryKey to another thread and very quickly get the object back by that id. This is also important to _CreateOrUpdate_ scenarios.

See the [Swift update by PrimaryKey](https://realm.io/docs/swift/latest/#creating-and-updating-objects-with-primary-keys)

AndyDentFree:
Moved back from review because approach I took was rejected. Now waiting for #750 so I can use a new exception type from ObjectStore rather than adding one.

AndyDentFree:
Think this is blocked by #789 now other bugs are fixed, as it is crashing because the schema we pass in has bogus column indexes.

AndyDentFree:
Should post an update on the [Akavache branch](https://github.com/akavache/Akavache/pull/292#issuecomment-242545105) when this ships.

PR #776, opened 2016-08-16
Ad/feature/get by id #402

### LINKED ISSUE #737: Document RealmSchema family

AndyDentFree:
Mostly adding the ById searches, also includes some Doxygen comment updates so closes #737 

@kristiandupont and @fealebenpae to review please


Realm.Shared/Realm.cs
```diff
@@ -194,12 +194,22 @@ private RealmObject.Metadata CreateRealmObjectMetadata(Schema.ObjectSchema schem
                 helper = Dynamic.DynamicRealmObjectHelper.Instance;
             }
 
+            var colMap = new Dictionary<string, IntPtr>();
```
nirinchev: Do we need to shorten those? Even though it's kind of obvious what it means, I find full names more readable. Additionally, it's inconsistent with other names, e.g. `Metadata.ColumnIndices`.

Realm.Shared/Realm.cs
```diff
@@ -676,6 +687,56 @@ public RealmResults<dynamic> All(string className)
             return new RealmResults<dynamic>(this, metadata, true);
         }
 
+
+        #region ById
+
+        private RealmObject.Metadata PrepById<T>(out IntPtr columnIndex)
```
nirinchev: What does `Prep` stand for here?

nirinchev: In the case where you'll need both the `columnIndex` and the `metadata`, I find the `out`-syntax cumbersome. Either return a tuple or simply return the metadata since it already contains the `ObjectIdColIndex` property and we can use that.

Realm.Shared/Schema/ObjectSchema.cs
```diff
@@ -27,6 +27,9 @@
 
 namespace Realms.Schema
 {
+    /// <summary>
+    /// Public descriptio of a class stored in a Realm, as a collection of managed Propertys.
```
nirinchev: Typos here - `descriptio`and `Propertys`

Realm.Shared/Schema/RealmSchema.cs
```diff
@@ -27,12 +27,24 @@
 
 namespace Realms
 {
+    /// <summary>
+    /// Describe the complete set of classes which may be stored in a Realm, either from assembly declarations or dynamically by evaluating a Realm from disk.
```
nirinchev: Missing comma around `dynamically`

Realm.Shared/Schema/RealmSchema.cs
```diff
@@ -114,8 +136,16 @@ private static RealmSchema BuildDefaultSchema()
...
+        /// <summary>
+        /// Helper class used to construct a RealmSchema.
+        /// </summary>
         public class Builder : List<ObjectSchema>
         {
+            /// <summary>
+            /// Build the RealmSchema to include all ObjectSchema added to this Builder.
+            /// </summary>
+            /// <exception cref="InvalidOperationException">Thrown if the Builder is empty.</exception>
+            /// <returns>A completed RealmSchema suitable for creating a new Realm.</returns>
```
nirinchev: comma before `suitable`

Realm.Shared/linq/RealmResults.cs
```diff
@@ -148,11 +148,12 @@ IEnumerator IEnumerable.GetEnumerator()
 
 
         /// <summary>
-        /// Fast count all objects of a given class.
+        /// Fast count all objects of a given class or invoked instead of LINQ in some casting scenarios.
         /// </summary>
         /// <remarks>
         /// Resolves to this method instead of the LINQ static extension <c>Count&lt;T&gt;(this IEnumerable&lt;T&gt;)</c>, when used directly on Realm.All.
         /// </remarks>
+        /// <returns>A zero-based count of all objects in a class or in the results of a search, without instantiating them.</returns>
```
nirinchev: What does `zero-based` count mean? Aren't all counts `zero-based`?

Realm.Shared/Realm.cs
```diff
@@ -676,6 +687,56 @@ public RealmResults<dynamic> All(string className)
...
+                throw new RealmClassLacksObjectIdException($"Class {typeof(T).Name} does not have a property marked as ObjectId");
+            columnIndex = (IntPtr)metadata.ObjectIdColIndex;
+            return metadata;
+        }
+
+
+        /// <summary>
+        /// Fast lookup of an object from a class which has an ObjectId property.
+        /// </summary>
+        /// <typeparam name="T">The Type T must not only be a RealmObject but also have been processd by the Fody weaver, so it has persistent properties.</typeparam>
```
nirinchev: Typo in `processd`

Realm.Shared/Realm.cs
```diff
@@ -676,6 +687,56 @@ public RealmResults<dynamic> All(string className)
...
+        /// <param name="id">Id to be matched exactly, same as an == search. Int64 argument works for all integer properties supported as ObjectId.</param>
+        /// <returns>Null or an object matdhing the id.</returns>
```
nirinchev: Typo in `matdhing`

Realm.Shared/Realm.cs
```diff
@@ -676,6 +687,56 @@ public RealmResults<dynamic> All(string className)
...
+        /// <exception cref="RealmClassLacksObjectIdException">If the RealmObject class T lacks an [ObjectId].</exception>
+        public T ById<T>(Int64 id) where T : RealmObject
```
nirinchev: I'm not convinced non-verbial method names are very .NET-friendly. Unless you've chosen it for consistency with other languages, I'd suggest going for `Find`/`GetById`.

fealebenpae:
I think we should be using the `Table::find_first_string` and `Table::find_first_int` methods from core like Object Store does. I'm not sure whether a `Query` does the same thing.

fealebenpae:
Technically we don't need to store additional metadata for the primary key columns - all that's already in the OS schema. All we need is to pass the `ObjectSchema` handle along with a string or a long to native, and we already have the `ObjectSchema` in `RealmObject.Metadata`.

I don't like the approach of caching column indexes in more and more places as those caches would need invalidation if we start using the new additive schema mode that allows new columns to be inserted in the schema in runtime.

Realm.Shared/Realm.cs
```diff
@@ -676,6 +687,56 @@ public RealmResults<dynamic> All(string className)
...
+        {
+            IntPtr columnIndex;
+            var metadata = PrepById<T>(out columnIndex);
+            NativeException nativeException;
+            var rowPtr = NativeTable.row_for_int_id(metadata.Table, columnIndex, id, out nativeException);
```
fealebenpae: `NativeTable` should expose a higher-level method such as `RowForIntId` or something that encapsulated the P/Invoked one and deals with exception handling internally.

Realm.Shared/Realm.cs
```diff
@@ -676,6 +687,56 @@ public RealmResults<dynamic> All(string className)
...
+        /// <typeparam name="T">The Type T must not only be a RealmObject but also have been processd by the Fody weaver, so it has persistent properties.</typeparam>
+        /// <param name="id">Id to be matched exactly, same as an == search.</param>
+        /// <returns>Null or an object matdhing the id.</returns>
+        /// <exception cref="RealmClassLacksObjectIdException">If the RealmObject class T lacks an [ObjectId].</exception>
+        public T ById<T>(string id) where T : RealmObject
+        {
+            IntPtr columnIndex;
+            var metadata = PrepById<T>(out columnIndex);
+            NativeException nativeException;
+            var rowPtr = NativeTable.row_for_string_id(metadata.Table, columnIndex, id, (IntPtr)id.Length, out nativeException);
```
fealebenpae: Same as above.

AndyDentFree:
The OS primary key column just stores the name of the column, which adds one more dictionary lookup to something which is supposed to be a fast operation.

> I don't like the approach of caching column indexes in more and more places

There is only one kind of object storing the index, the `RealmObject.Metadata` and we already need to propagate a rebuild of those. (That is why I picked that to hold the `ObjectIdColIndex`.

nirinchev:
IMO a dictionary lookup *is* a fast operation and I'd be surprised to see a meaningful performance difference between the current implementation and the iOS one. But to be able to put substance to one's words, I'd advise to setup a few load test scenarios which can be used as a tie-breaker in such cases. Additionally, having performance targets will help a lot when evaluating different approaches.

AndyDentFree:
I was wrong, our column name lookup is not a dictionary, it's an `ArrayString::find_first` called via `Spec::get_column_index`. I wouldn't have **started** using a cached column index if it was the first occasion but as the type managing it already has an array of them, it seemed a trivial extension which kept the knowledge on the C# side. That also means the check to see if there is an ObjectId for the column is also very quick (it has to be performed in each call to `ObjectById`).

@nirinchev the overhead of calling native from C# I am fairly sure is considerably higher than calling C++ code from Objective-C. So minimising calls to native is a design goal.

fealebenpae:
We already have the knowledge for this in Object Store, and looking up the column index there would be pretty fast, so I propose we move the bulk of the implementation to C++ instead of C#.

AndyDentFree:
Does that mean you want me to strip out the C# `Metadata. ObjectIdColIndex` and also have the exception throwing (for no ObjectId) moved into C++?

fealebenpae:
@AndyDentFree: yes, I think it'll be best for us to handle this in c++ as much as possible.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
-0.77.x (TBD)
+0.78.0 (TBD)
 -------------------
+### Enhancements
+* You can retrieve single objects quickly using `Realm.ObjectById()` if they have an `[ObjectOd]` column specified.
```
bdash: "ObjectOd"

wrappers/src/query_cs.cpp
```diff
@@ -391,4 +391,33 @@ REALM_EXPORT Results* query_create_sorted_results(Query * query_ptr, SharedRealm
     });
 }
 
+
+#pragma mark  ObjectId Searches
+// These use the faster calls such as find_first_int rather than having to build and use a query in one go.
+// Note to bypass the overhead of looking up the PrimaryKey field by name,
```
bdash: This seems like a weird place to put this comment. The behavior it is describing lives in the callers of these functions.

I also question the value of having a comment describing Cocoa's current behavior since it's very likely to get out of date, and unlikely to be updated when that happens.

Tests/IntegrationTests.Shared/ObjectIdTests.cs
```diff
@@ -0,0 +1,201 @@
...
+                var obj = _realm.CreateObject<ObjectIdCharObject>();
+                obj.CharProperty = 'x';
+            });
+
+            var foundObj = _realm.ObjectById<ObjectIdCharObject>('x');
+            Assert.IsNotNull(foundObj);
+            Assert.That(foundObj.CharProperty, Is.EqualTo('x'));
+        }
+
+        [Test]
```
kristiandupont: Maybe we should refactor out `AccessTests.SetAndGetValueCases` into a `TestHelper` so we could reuse them for testcases here? They also seem related to the weaver tests `RandomAndDefaultValues`..

Realm.Shared/linq/RealmResults.cs
```diff
@@ -148,11 +148,12 @@ IEnumerator IEnumerable.GetEnumerator()
 
 
         /// <summary>
-        /// Fast count all objects of a given class.
+        /// Fast count all objects of a given class or invoked instead of LINQ in some casting scenarios.
```
kristiandupont: Invoked instead of LINQ? What does that mean?

kristiandupont:
>also includes some Doxygen comment updates

I think we should try to stick with one PR per issue as much as possible. Not a big deal so we can diverge this time but I think it would make our history a bit easier to work with if we try to stick to it.

wrappers/src/object-store/src/object_store.hpp
```diff
@@ -253,6 +253,11 @@ namespace realm {
         };
         InvalidLinkingObjectsPropertyException(Type error_type, std::string const& object_type, Property const& property);
     };
+
+    class PrimaryKeyNotDeclaredException : public ObjectSchemaValidationException {
```
bdash: This seems like it's duplicating `MissingPrimaryKeyException`.

wrappers/src/query_cs.cpp
```diff
@@ -391,4 +392,38 @@ REALM_EXPORT Results* query_create_sorted_results(Query * query_ptr, SharedRealm
...
+
+#pragma mark  PrimaryKey Searches
+Row* row_for_primarykey(Table* table_ptr, ObjectSchema* object_schema, std::function<size_t(Table*, size_t columnIndex)> finder, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        size_t columnIndex = realm::npos;
+        if (object_schema->primary_key.size() > 0)
+            columnIndex = table_ptr->get_column_index(object_schema->primary_key);
+        if (columnIndex == realm::npos)
+            throw PrimaryKeyNotDeclaredException(object_schema->name);
```
bdash: These few lines could be written as:
```c++
auto primary_prop = object_schema.primary_key_property();
if (!primary_prop) {
    throw …;
}
```

Then you'd go on to use `primary_prop->table_column` instead of `columnIndex`.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
-0.77.x (TBD)
+0.78.0 (TBD)
 -------------------
+### Breaking Changes
+* The term `ObjectId` has been replaced with `PrimaryKey` in order to align with the other SDKs. This affects the `[ObjectId]` attribute used to decorate a property.
```
bdash: Seems confusing to include this rename as part of adding a new feature.

wrappers/src/object-store/src/object_store.hpp
AndyDentFree: Yes, that has been added to ObjectStore since the level we are currently using in dotnet (waiting on moving to  using it as a submodule.)

CHANGELOG.md
AndyDentFree: Normally yes, but most of the new feature was using the ObjectId nomenclature so introducing it with naming we decided to replace didn't make sense.

wrappers/src/query_cs.cpp
AndyDentFree: I saw `primary_key_property` is implemented calling `property_for_name` and if the user hasn't specified one, it will loop through all defined properties before failing. Quickly checking if the name is blank seemed a better way to go. If you insist on doing the alternative, how about enhancing `property_for_name` to check for empty names?

bdash: The motivation of my suggestion is to use the object schema's cached version of the column index rather than retrieving it from the table.

It's programmer error to try to look up an object by primary key when the class has no primary key defined. I don't see why it's useful to optimize for that case?

AndyDentFree:
Too much has been refactored out into other PRs - will do a merge and submit new one.

PR #781, opened 2016-08-18
#737 Isolated RealmSchema and couple of other minor doc changes out

AndyDentFree:
This reduces the noise in the other PR - should be JUST comment changes. Will use this to simplify other primary key PR

@fealebenpae or @kristiandupont to approve

### DIFF
--- a/Realm.PCL/Schema/ObjectSchemaPCL.cs
+++ b/Realm.PCL/Schema/ObjectSchemaPCL.cs
@@ -22,10 +22,19 @@ using System.Collections.Generic;
 
 namespace Realms.Schema
 {
+    /// <summary>
+    /// Public description of a class stored in a Realm, as a collection of managed Property objects.
+    /// </summary>
     public class ObjectSchema : IReadOnlyCollection<Property>
     {
+        /// <summary>
+        /// Name of the original class declaration from which the schema was built.
+        /// </summary>
         public string Name { get; }
 
+        /// <summary>
+        /// Number of properties in the schema, which is the persistent properties from the original class.
+        /// </summary>
         public int Count { get; }
 
         private ObjectSchema()
@@ -33,6 +42,12 @@ namespace Realms.Schema
             RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
         }
 
+        /// <summary>
+        /// Looks for a Property by Name. Failure to find means it is not regarded as a property to persist in a Realm.
+        /// </summary>
+        /// <returns><c>true</c>, if a property was found matching Name, <c>false</c> otherwise.</returns>
+        /// <param name="name">Name of the Property to match exactly.</param>
+        /// <param name="property">Property returned only if found matching Name.</param>
         public bool TryFindProperty(string name, out Property property)
         {
             RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
@@ -40,6 +55,10 @@ namespace Realms.Schema
             return false;
         }
 
+        /// <summary>
+        /// Property enumerator factory for an iterator to be called explicitly or used in a foreach loop.
+        /// </summary>
+        /// <returns>An enumerator over the list of Property instances described in the schema.</returns>
         public IEnumerator<Property> GetEnumerator()
         {
             RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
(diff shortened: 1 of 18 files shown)


Realm.Shared/RealmObject.cs
```diff
@@ -567,7 +567,7 @@ protected void SetByteArrayValue(string propertyName, byte[] value)
         /// Compare objects with identity query for persistent objects.
         /// </summary>
         /// <remarks>Persisted RealmObjects map their properties directly to the realm with no caching so multiple instances of a given object always refer to the same store.</remarks>
-        /// <param name="obj"></param>
+        /// <param name="obj">Object being compared against to see if is same C# object or maps to same managed object in Realm.</param>
```
nirinchev: should be "...see if *it* is *the* same C# object or maps to *the* same..."

Realm.Shared/Schema/Property.cs
```diff
@@ -21,6 +21,9 @@
 
 namespace Realms.Schema
 {
+    /// <summary>
+    /// Describe a single property of a class stored in a Realm.
```
nirinchev: Probably "Describe*s*"

Realm.Shared/Schema/RealmSchema.cs
```diff
@@ -44,6 +56,12 @@ private RealmSchema(SchemaHandle handle, IEnumerable<ObjectSchema> objects)
             Handle = handle;
         }
 
+        /// <summary>
+        /// Find the definition of a class in this schema.
```
nirinchev: Not sure what the convention at Realm is, but the .NET documentation usually describes what the method does from a 3rd person point of view, e.g. "[It] Finds the definition of a class in this schema."

Realm.Shared/Schema/RealmSchema.cs
```diff
@@ -44,6 +56,12 @@ private RealmSchema(SchemaHandle handle, IEnumerable<ObjectSchema> objects)
...
+        /// </summary>
+        /// <param name="name">A valid class name which may be in this schema.</param>
+        /// <exception cref="ArgumentException">Thrown if an empty name is supplied.</exception>
```
nirinchev: Should this be "null or empty name"?

Realm.Shared/linq/RealmResults.cs
```diff
@@ -148,11 +148,16 @@ IEnumerator IEnumerable.GetEnumerator()
 
 
         /// <summary>
-        /// Fast count all objects of a given class.
+        /// Fast count all objects of a given class, or in a RealmResults after casting.
         /// </summary>
         /// <remarks>
         /// Resolves to this method instead of the LINQ static extension <c>Count&lt;T&gt;(this IEnumerable&lt;T&gt;)</c>, when used directly on Realm.All.
+		/// <br>
```
nirinchev: Indentation issue

Realm.Shared/linq/RealmResults.cs
```diff
@@ -148,11 +148,16 @@ IEnumerator IEnumerable.GetEnumerator()
...
+		/// if someone CASTS a RealmResults<blah> variable from a Where call to
```
nirinchev: Should this be `RealmResults&lt;T&gt;`? Or does the `remarks` tag properly handle those? Even if that's the case, we probably should replace `<blah>` with `<T>`

AndyDentFree:
retest this

AndyDentFree:
retest this please

Realm.PCL/Schema/ObjectSchemaPCL.cs
```diff
@@ -48,21 +67,38 @@ public IEnumerator<Property> GetEnumerator()
 
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
 
+        /// <summary>
+        /// Creates a schema describing .
```
kristiandupont: Is there a word missing here?

Realm.PCL/Schema/RealmSchemaPCL.cs
```diff
@@ -22,21 +22,43 @@
 
 namespace Realms
 {
+    /// <summary>
+    /// Describes the complete set of classes which may be stored in a Realm, either from assembly declarations or, dynamically, by evaluating a Realm from disk.
+    /// </summary>
+    /// <remarks>
+    /// By default this will be all the RealmObjects in all your assemblies unless you restrict with RealmConfiguration.ObjectClasses. 
+    /// Just because a given class <em>may</em> be stored in a Realm doesn't imply much overhead. There will be a small amount of metadata
```
kristiandupont: This doesn't seem like the right place for this bit of information to me. We could add it to the FAQ if necessary, or in some general documentation somewhere.

kristiandupont:
:+1:

Realm.PCL/Schema/RealmSchemaPCL.cs
AndyDentFree: Our docs seem very lean compared to the other SDKs

kristiandupont: Sure, I just mean that this is not a place where you would look for this information. The overhead concern is not limited to people who define specific schemas and you are unlikely to encounter this class if you don't (even then it's unlikely as you just need `RealmConfiguration`).

PR #783, opened 2016-08-19
General search and replace ObjectId with PrimaryKey

AndyDentFree:
@fealebenpae if you have a minute, easy review
@kristiandupont please review if sits there

Is just a search and replace of terms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
-0.77.x (TBD)
+0.78.0 (TBD)
 -------------------
+### Breaking Changes
+* The term `ObjectId` has been replaced with `PrimaryKey` in order to align with the other SDKs. This affects the `[ObjectId]` attribute used to decorate a property.
+
 ### Bug fixes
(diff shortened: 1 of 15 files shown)


kristiandupont:
There seems to be a couple of replaces missing in the weaver tests, looking at CI. Otherwise it looks good

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
-0.77.x (TBD)
+0.78.0 (TBD)
```
kristiandupont: Very minor nit, but shouldn't we do these in the release branch? Otherwise it's not clear who should do it, and it's not really relevant to any specific feature/bugfix.

AndyDentFree: My thinking is that the person responsible for adding an entry in changelog is making an assertion that the entry is sufficiently different to bump the version. Sorting out a precise minor number would be done in the release branch if necessary.

kristiandupont:
:+1:

PR #798, opened 2016-08-26
Update RealmResultsCollectionChanged.cs

promontis:
I'd like to return a different collection than ReadOnlyObservableCollection from the extension method to support incremental loading. This collection still adheres to ObservableCollection, except it implements an extra interface with incremental loading support. 

However, Adapter is not publicly available. This PR is making the Adapter public.

### DIFF
--- a/Realm.Shared/extensions/RealmResultsCollectionChanged.cs
+++ b/Realm.Shared/extensions/RealmResultsCollectionChanged.cs
@@ -69,7 +69,7 @@ namespace Realms
             return new ReadOnlyObservableCollection<T>(new Adapter<T>((RealmResults<T>)results, errorCallback, coalesceMultipleChangesIntoReset));
         }
 
-        sealed class Adapter<T> : ObservableCollection<T> where T : RealmObject
+        public sealed class Adapter<T> : ObservableCollection<T> where T : RealmObject
         {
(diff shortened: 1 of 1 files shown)


AndyDentFree:
If we are making this class public, should it be renamed something more specific such as `Adapter`?

nirinchev:
Or even more specific - `RealmResultsObservableAdapter`? :) Anyway, I agree with the change - internals were actually the bane of Xamarin.Forms, so it would be unwise to make the same mistakes.

fealebenpae:
👍

fealebenpae:
Although if you mean to instantiate `Adapter` yourself you might need to make its constructor public as well.

promontis:
Ah... I went the lazy route and edited everything on Github. It should indeed be in a separate file named something like RealmResultsObservableAdapter and the ctor should also be public in order to actually use it lol. Will do a proper clone and re-PR.

kristiandupont:
I merged this one already. I think we can keep the class as an embedded class, but if you want to make the constructor public, you can create a new PR for that.

promontis:
@kristiandupont I kept the class as an embedded and did a PR for the public ctor at #800. Does Realm provide nightly or ci builds so that I can start using this change before an official release?

PR #797, opened 2016-08-26
Refactor RealmSchema to remove the native handles

fealebenpae:
Closes #789

PR #799, opened 2016-08-29
Ad/feature/get by primary key #402

AndyDentFree:
This has been through a couple of iterations with removal of the column index approach to use a schema, then the schema information all being stripped out by #797 so we cannot easily pass the native schema back down. In the interests of expediency and efficiency, I have gone back to a column index approach, because the source of the column index has the other column indexes. So if there is ever any live change to the schema that resets them, they are managed in one place in C#. However, I have kept more of the logic over on the C++ side.

@fealebenpae and @kristiandupont to review. If you have an alternative suggestion for passing schema stuff in to then do a lookup, please go ahead.


Realm.PCL/RealmPCL.cs
```diff
@@ -326,6 +326,62 @@ public RealmResults<dynamic> All(string className)
             return null;
         }
 
+
+        /// <summary>
+        /// Fast lookup of an object from a class which has a PrimaryKey property.
+        /// </summary>
+        /// <typeparam name="T">The Type T must not only be a RealmObject but also have been processd by the Fody weaver, so it has persistent properties.</typeparam>
```
kristiandupont: Typo: processd. But I would be tempted to just remove the note about Fody as I think it's a given -- you can't do any of the other operations with non-woven classes either.

Tests/IntegrationTests.Shared/PrimaryKeyTests.cs
```diff
@@ -0,0 +1,201 @@
...
+        public void FindByCharPrimaryKey()
+        {
+            _realm.Write(() => {
+                var obj = _realm.CreateObject<PrimaryKeyCharObject>();
+                obj.CharProperty = 'x';
+            });
+
+            var foundObj = _realm.ObjectForPrimaryKey<PrimaryKeyCharObject>('x');
+            Assert.IsNotNull(foundObj);
+            Assert.That(foundObj.CharProperty, Is.EqualTo('x'));
```
kristiandupont: If `obj` was declared in the outer scope, we could assert that `foundObj` and `obj` are equal instead.

wrappers/src/error_handling.cpp
```diff
@@ -90,6 +91,9 @@ namespace realm {
         catch (const RowDetachedException& e) {
             return { RealmErrorType::RealmRowDetached, e.what() };
         }
+        catch (const MissingPrimaryKeyException& e) {
+            return { RealmErrorType::RealmTableHasNoPrimaryKey, e.what() };
```
kristiandupont: This isn't specific to this PR, but I think it's unfortunate that we are operating with three different naming schemes for these errors. Maybe we could rename our `RealmErrorType` entries to match core exceptions at some point..

Tests/IntegrationTests.Shared/PrimaryKeyTests.cs
AndyDentFree: We only have referential equality and I think the current test emphasises a new, matching, object is retrieved.

kristiandupont:
:+1:

PR #800, opened 2016-08-29
Changed RealmResultsCollectionChanged ctor

promontis:
Adapter ctor is now public... also see #798

### DIFF
--- a/Realm.Shared/extensions/RealmResultsCollectionChanged.cs
+++ b/Realm.Shared/extensions/RealmResultsCollectionChanged.cs
@@ -78,7 +78,7 @@ namespace Realms
 
             private bool _suspendNotifications;
 
-            internal Adapter(RealmResults<T> results, Action<Exception> errorCallback, bool coalesceMultipleChangesIntoReset) : base(results)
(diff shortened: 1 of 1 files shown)


fealebenpae:
I suppose you might as well unseal the class. Does `Adapter`'s current shape enable you to implement the incremental loading interface you mentioned?

promontis:
The sealed attribute is not really in my way... perhaps I even kinda think it should be there, just to indicate that we probably shouldn't be using this class as a base, but instead plug it into the ctor of a derivative of ObservableCollection. 

I will be doing this as well for the incremental loading support. The main goal of the PR is to have the Adapter class available to plug into a derivative of ObservableCollection. This class will inherit both from OberservableCollection and IIncrementalLoadingSupport which has a Load method which is triggered after scrolling a list for a certain amount of cells. This Load method is implemented as a lambda and will in this case fetch data from a REST service and add it to Realm. If I understand Realm correctly, the adapter will correctly translate these additions into a collection changed event from the ObservableCollection so that the UI will be updated :) So yes, I think the current shape is sufficient.

Issue #802, opened 2016-08-29
Remaining Element Operators


kristiandupont:
Missing:
 - [x] FirstOrDefault
 - [x] ElementAt
 - [x] ElementAtOrDefault
 - [x] SingleOrDefault
 - [x] Last
 - [x] LastOrDefault
 - [ ] DefaultIfEmpty

promontis:
+1 for this. This should be low hanging fruit right?

kristiandupont:
Hopefully :-)

AndyDentFree:
The `ElementAt` appears to be implemented in #802 but needs unit tests.

AndyDentFree:
Dropping `DefaultIfEmpty` from this and moving to lower priority #816 as it seems very complex to implement with our current architecture and of little appeal.

PR #815, opened 2016-09-01
Ad/linq/feature/more elements #802

AndyDentFree:
This is mostly done except for the `DefaultIfEmpty` behaviour.

@fealebenpae if you could review the rest, and have any suggestions on how that should work, would appreciate. thanks


Realm.Shared/linq/RealmResultsVisitor.cs
```diff
@@ -179,44 +216,60 @@ internal override Expression VisitMethodCall(MethodCallExpression m)
...
                     }
-                    if (firstRowPtr == IntPtr.Zero)
-                        throw new InvalidOperationException("Sequence contains no matching element");
-                    return Expression.Constant(_realm.MakeObjectForRow(_metadata, firstRowPtr));
+                    if (firstRowPtr != IntPtr.Zero)
+                        return Expression.Constant(_realm.MakeObjectForRow(_metadata, firstRowPtr));
+                    if (m.Method.Name == "First")
+                            throw new InvalidOperationException("Sequence contains no matching element");
+                    Debug.Assert (m.Method.Name == "FirstorDefault");
+                    return MakeDefaultWithTransactionIfNeeded(_metadata);
```
nirinchev: Not sure if I understand it correctly, but it seems here, we'll create a new object instead of returning null or am I missing something?

Realm.Shared/linq/RealmResultsVisitor.cs
```diff
@@ -179,44 +216,60 @@ internal override Expression VisitMethodCall(MethodCallExpression m)
...
-                        using (ResultsHandle rh = _realm.MakeResultsForQuery(_coreQueryHandle, OptionalSortDescriptorBuilder))
-                        {
-                            var rowPtr = rh.GetRow(index);
-                            row = Realm.CreateRowHandle(rowPtr, _realm.SharedRealmHandle);
-                        }
+                        var lastIndex = rh.Count() - 1;
+                        if (lastIndex >= 0)
+                            lastRowPtr = rh.GetRow(lastIndex);
                     }
+                    if (lastRowPtr != IntPtr.Zero)
```
nirinchev: With the above methods, we first evaluate the `rowPtr == IntPtr.Zero` case and then the `!= IntPtr.Zero` case, so it's a little inconsistent here.

Realm.Shared/linq/RealmResultsVisitor.cs
```diff
@@ -179,44 +216,60 @@ internal override Expression VisitMethodCall(MethodCallExpression m)
...
+                    return MakeDefaultWithTransactionIfNeeded(_metadata);
+                }
+                if (m.Method.Name == nameof(Queryable.ElementAt))
+                {
+                    var row = VisitElementAt(m);
+                    if (row == null || row.IsInvalid)
+                        throw new ArgumentOutOfRangeException();
+                    return Expression.Constant(_realm.MakeObjectForRow(_metadata, row));
+                }
+                if (m.Method.Name == nameof(Queryable.ElementAtOrDefault))
```
nirinchev: Why is this different from the rest? It seems it's a matter of an if-case inside the `if (row == null || row.IsInvalid)`

nirinchev:
After reviewing the tests, it seems you've decided on an unconventional `OrDefault` behavior. Generally, I would expect to receive `null` instead of writing a new object to the Realm. If this is intentional and not a mistake, then it should be **very** well documented.

Realm.Shared/linq/RealmResultsVisitor.cs
```diff
@@ -219,9 +206,19 @@ internal override Expression VisitMethodCall(MethodCallExpression m)
                     if (firstRowPtr != IntPtr.Zero)
                         return Expression.Constant(_realm.MakeObjectForRow(_metadata, firstRowPtr));
                     if (m.Method.Name == "First")
-                            throw new InvalidOperationException("Sequence contains no matching element");
+                        throw new InvalidOperationException("Sequence contains no matching element");
                     Debug.Assert (m.Method.Name == "FirstorDefault");
```
fealebenpae: There's a typo in the method name here. Use `nameof(Queryable._methodName_)` to avoid this.

Realm.Shared/linq/RealmResultsVisitor.cs
```diff
@@ -179,44 +203,70 @@ internal override Expression VisitMethodCall(MethodCallExpression m)
...
+                    if (firstRowPtr != IntPtr.Zero)
+                        return Expression.Constant(_realm.MakeObjectForRow(_metadata, firstRowPtr));
+                    if (m.Method.Name == "First")
                         throw new InvalidOperationException("Sequence contains no matching element");
-                    return Expression.Constant(_realm.MakeObjectForRow(_metadata, firstRowPtr));
+                    Debug.Assert (m.Method.Name == "FirstorDefault");
+                    return Expression.Constant(null);
                 }
-                if (m.Method.Name == "Single")  // same as unsorted First with extra checks
+                if (m.Method.Name == "DefaultIfEmpty")
```
fealebenpae: I don't think this is how `DefaultIfEmpty` is supposed to behave - http://stackoverflow.com/questions/8854181/defaultifempty-in-linq

Tests/IntegrationTests.Shared/RealmResults/SortingTests.cs
```diff
@@ -223,6 +223,25 @@ public void FirstIsDifferentSorted()
             Assert.That(sortedFirst.Email, Is.EqualTo("john@doe.com"));
         }
 
+
+        [Test]
+        public void LastIsDifferentSorted()
```
nirinchev: Not sure what is the purpose of this test. What are we validating with it?

Tests/IntegrationTests.Shared/RealmResults/SortingTests.cs
```diff
@@ -223,6 +223,25 @@ public void FirstIsDifferentSorted()
...
+            foreach(var p in _realm.All<Person>().OrderByDescending(p => p.Score))
+                System.Diagnostics.Debug.WriteLine($"Sorted Person {p.FullName} Score {p.Score}");
+            var lowestScore = _realm.All<Person>().OrderByDescending(p => p.Score).Last();
+            Assert.That(lowestScore.Email, Is.EqualTo("john@smith.com"));
+
+            var sortedLastInteresting = _realm.All<Person>().OrderByDescending(p => p.LastName).Last(p => p.IsInteresting);
+            Assert.That(sortedLastInteresting.Email, Is.EqualTo("peter@jameson.net"));
+
+            var sortedLast = _realm.All<Person>().
+                Where(p => p.FirstName=="John").
```
nirinchev: The dots should be at the beginning of the line

Realm.Shared/linq/RealmResultsVisitor.cs
nirinchev: Also, after you merge them, you can inline `VisitElementAt`

Realm.Shared/linq/RealmResultsVisitor.cs
```diff
@@ -117,6 +118,29 @@ private void AddSort(LambdaExpression lambda, bool isStarting, bool ascending)
         }
 
 
+        private RowHandle VisitElementAt(MethodCallExpression m)
+        {
+            Visit(m.Arguments.First());
+            var index = (int)ExtractConstantValue(m.Arguments.Last());
+
+            RowHandle row = null;
```
nirinchev: No need to set to null here.

Tests/IntegrationTests.Shared/RealmResults/SimpleLINQtests.cs
```diff
@@ -361,6 +373,91 @@ public void FirstWorks()
...
+
+            expectedDef = _realm.All<Person>().Where(p => p.FirstName == "Just Some Guy").FirstOrDefault();
+            Assert.That(expectedDef, Is.Null);
+        }
+
+
+        [Test]
+        public void LastFailsToFind()
+        {
+            Assert.Throws<InvalidOperationException>(() => _realm.All<Person>().Last(p => p.Latitude > 100));
```
nirinchev: I don't think we need 4 test cases for that - it will either fail with the first two or it will not :)

Tests/IntegrationTests.Shared/RealmResults/SimpleLINQtests.cs
```diff
@@ -361,6 +373,91 @@ public void FirstWorks()
...
+
+            expectedDef = _realm.All<Person>().Where(p => p.FirstName == "Just Some Guy").LastOrDefault();
+            Assert.That(expectedDef, Is.Null);
+        }
+
+
+        [Test]
+        public void ElementAtOutOfRange()
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => _realm.All<Person>().Where(p => p.Latitude > 140).ElementAt(0));
```
nirinchev: We could add two more tests without the `.Where` - `.ElementAt(-1)` and `.ElementAt(100000)`

Tests/IntegrationTests.Shared/RealmResults/SimpleLINQtests.cs
```diff
@@ -361,6 +373,91 @@ public void FirstWorks()
...
+            Assert.Throws<ArgumentOutOfRangeException>(() => _realm.All<Person>().Where(p => p.FirstName == "Samantha").ElementAt(0));
+            Assert.Throws<ArgumentOutOfRangeException>(() => _realm.All<Person>().Where(p => p.FirstName == "Peter").ElementAt(10));
+        }
+
+
+        [Test]
+        public void ElementAtInRange()
+        {
+            var s0 = _realm.All<Person>().Where(p => p.Longitude < -70.0 && p.Longitude > -90.0).ElementAt(0);
```
nirinchev: Need test without a `.Where`

nirinchev: Also, a test with `.OrderBy().ElementAt()`

nirinchev:
I would add tests to validate that `.XXX()` and `.XXXOrDefault()` return the same element in the non-empty/default case.

Realm.Shared/linq/RealmResultsVisitor.cs
AndyDentFree: I am pretty sure the compiler was (wrongly) complaining about using an uninitialised variable.

Realm.Shared/linq/RealmResultsVisitor.cs
AndyDentFree: As I said in the original PR comment, that is not yet done.

Tests/IntegrationTests.Shared/RealmResults/SortingTests.cs
AndyDentFree: It validates that sorting affects which item shows up Last.

Realm.Shared/linq/RealmResultsVisitor.cs
nirinchev: I'd guess it was before you assigned it in one of the branches, or it could be a XS issue - their static analysis is not super-duper stable 😜

Tests/IntegrationTests.Shared/RealmResults/SortingTests.cs
nirinchev: Then perhaps we could look for a more suitable name - e.g. `LastWorksWithOrderBy`? Also, since we have such test for `.Last`, we should probably do it for `.First`, since it can also be different based on sorting.

AndyDentFree: The test above is `FirstIsDifferentSorted` and was there already.

nirinchev: My bad, I was only looking going through the diff and didn't notice it 👍

Realm.Shared/linq/RealmResultsVisitor.cs
```diff
@@ -179,47 +203,73 @@ internal override Expression VisitMethodCall(MethodCallExpression m)
                             firstRowPtr = rh.GetRow(0);
                         }
                     }
-                    if (firstRowPtr == IntPtr.Zero)
+                    if (firstRowPtr != IntPtr.Zero)
+                        return Expression.Constant(_realm.MakeObjectForRow(_metadata, firstRowPtr));
+                    if (m.Method.Name ==nameof(Queryable.First))
```
fealebenpae: formatting - space before `nameof`

Realm.Shared/linq/RealmResultsVisitor.cs
```diff
@@ -179,47 +203,73 @@ internal override Expression VisitMethodCall(MethodCallExpression m)
...
                         throw new InvalidOperationException("Sequence contains no matching element");
-                    return Expression.Constant(_realm.MakeObjectForRow(_metadata, firstRowPtr));
+                    Debug.Assert (m.Method.Name == nameof(Queryable.FirstOrDefault));
+                    return Expression.Constant(null);
+                }
+                if (m.Method.Name == nameof(Queryable.DefaultIfEmpty))
```
fealebenpae: I thought we were dropping `DefaultIfEmpty`?

AndyDentFree: I left the implementation in there with the FIXME because otherwise I think it might be too hard to come back to and dig it out from git. I added #816 to come back to this.

fealebenpae:
I think this is good. We might want to add a few more test cases for `ElementAdd` that @nirinchev pointed out, and I think we should get rid of `DefaultIfEmpty` entirely for now if we're unsure of how we're gonna tie its semantics to Realm.

You might also want to run a `Format Document` on a few files to get rid of formatting inconsistencies, but I wouldn't make merging the PR contingent on this. 👍 from me.

Realm.Shared/linq/RealmResultsVisitor.cs
fealebenpae: Oh, makes sense. But maybe just comment out the whole block so that it's still there in source form, but if anybody calls `DefaultIfEmpty` they would get a not supported exception, instead of undefined behavior.

PR #819, opened 2016-09-06
Ad/linq/feature/char queries #708

AndyDentFree:
Fairly mechanical extension except ran into a weird LINQ behaviour at the top level - the lambda converts a literal char on the RHS into an `Int32` them wraps the character property on the LHS in a `Convert` which needed unpacking as a `UnaryExpression`. A nice side-effect is that also fixes errors if people use a different int type from their property, which has already caused at least one user bug report.

Branched off the `more-elements` branch rather than master as I expect to do a bunch of LINQ work before people have time to review, with absences this week.

@fealebenpae please review if you have time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,19 @@
 * The term `ObjectId` has been replaced with `PrimaryKey` in order to align with the other SDKs. This affects the `[ObjectId]` attribute used to decorate a property.
 
 ### Enhancements
-* You can retrieve single objects quickly using `Realm.ObjectForPrimaryKey()` if they have a `[PrimaryKey]` property specified.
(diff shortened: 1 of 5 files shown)


fealebenpae:
Shouldn't this PR be against `master` instead of your LINQ branch?

Other than that, 👍

fealebenpae:
please retest this

fealebenpae:
CI seems to choke on this. I think it's the tweak to architectures in the iOS test app project. Could you revert that change?

fealebenpae:
CI seems to choke on this. I think it's the tweak to architectures in the iOS test app project. Could you revert that change?

AndyDentFree:
I did this as a branch off my LINQ branch so the PR reflects the branching structure. Otherwise, merging would be an annoying set of clashes.

PR #821, opened 2016-09-06
Decorate the managed migration callback with MonoPInvokeCallback

### LINKED ISSUE #820: Migration callbacks broken on iDevice
The test `MigrationTests.ExceptionInMigrationCallback` fails when run on an iOS device:

```
Expected: <System.AggregateException>
  But was:  <System.ExecutionEngineException> (Attempting to JIT compile method '(wrapper native-to-managed) Realms.Migration:MigrationCallback (intptr,intptr,Realms.Native.Schema,ulong,intptr)' while running with --aot-only
```

and similar for `TriggerMigrationBySchemaEditing` which makes me think the unit tests also need to clear a migration callback in the `[TearDown]` method.

fealebenpae:
Fixes #820 

/cc @AndyDentFree

### DIFF
--- a/Realm.Shared/Migration.cs
+++ b/Realm.Shared/Migration.cs
@@ -79,6 +79,9 @@ namespace Realms
             return true;
         }
 
+#if __IOS__
+        [ObjCRuntime.MonoPInvokeCallback(typeof(Native.MigrationCallback))]
+#endif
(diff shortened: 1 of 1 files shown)


AndyDentFree:
👍

PR #835, opened 2016-09-13
Add migration experiment

kristiandupont:
This code has to be commented out because the duplicate names will break all tests that don't supply `ObjectClasses` in their configuration. So it might not be suites to keep this in the main repo. Let me know what you think.


kristiandupont:
I'm closing this for now. It might be something to investigate again at some point but probably not like this.

PR #843, opened 2016-09-19
Ad/weaver write persist attribute#822

AndyDentFree:
@fealebenpae please review.
Tried adding the equivalent of `[Persist(allMembers=true)]` to type but that seems insufficient so ended up adding a simple `[Persist]` on all woven members.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+0.7x.x (TBD)
+-------------------
+### Enhancements
+* IOS Linking all should work - we now add a [Preserve] attribue to all woven members of your `RealmObject` subclasses so you do not need to manually add `[Preserve(allMembers=true)]`  (#822)
+
+
 0.78.1 (2016-09-15)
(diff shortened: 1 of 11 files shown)


nirinchev:
It seems counter-intuitive that we add a `[Preserve]` to all properties and the default ctor AND a `[Preserve(AllMembers = true)]` to the type.

You say that `Preserve(true)` is insufficient - does it fail on iOS/Android/both?

AndyDentFree:
It has only ever failed on IOS. Yes it seems counter-intuitive and I could strip it back a bit and just put the simple `[Preserve]` on the type but thought I may as well leave that in there in case it helps - I don't think it can cause any harm.

I have spent far more time on this particular issue than I am happy with, and the brute force approach of putting it on all properties is the only thing I could get to work with full linking.

AndyDentFree:
retest this please

AndyDentFree:
retest this please

PR #847, opened 2016-09-28
Add test showing how you can get a PK name as is from iterating schema

AndyDentFree:
I thought I might need a code change but ended up just adding this sample as a unit test to prove our existing schema info allows you to iterate classes and determine which have primary keys.

The need came from someone doing a complicated abstraction layer who wanted to use the dynamic form of `ObjectForPrimaryKey`

### DIFF
--- a/Tests/IntegrationTests.Shared/PrimaryKeyTests.cs
+++ b/Tests/IntegrationTests.Shared/PrimaryKeyTests.cs
@@ -290,11 +290,11 @@ namespace IntegrationTests.Shared
                 else
                 {
                     classesWithPK++;
-                    Debug.WriteLine($"PK test - {objSchema.Name} has a PK {primaryKeyProp.Name}");
+                    // uncomment the next line to see output list all primary keys in all classes in all unit tests
+                    // Debug.WriteLine($"PK test - {objSchema.Name} has a PK {primaryKeyProp.Name}");
                 }
             }
-            Assert.That(classesWitoutPK > classesWithPK);  // slightly crap assertion to ensure test works
-
+            Assert.That(classesWitoutPK > classesWithPK);  // simple assertion likely to remain valid regardless how many classes in test suite overall
         }
             
     }
--- a/Tests/IntegrationTests.Shared/PrimaryKeyTests.cs
+++ b/Tests/IntegrationTests.Shared/PrimaryKeyTests.cs
@@ -18,7 +18,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Text;
+using System.Diagnostics;
 using System.Linq;
 using NUnit.Framework;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


Tests/IntegrationTests.Shared/PrimaryKeyTests.cs
```diff
@@ -275,6 +275,27 @@ public void PrimaryKeyFailsIfClassNotinRealm()
...
+                var primaryKeyProp = objSchema.Where(prop => prop.IsPrimaryKey).FirstOrDefault();
+                if (string.IsNullOrEmpty(primaryKeyProp.Name))
+                    classesWitoutPK++;
+                else
+                {
+                    classesWithPK++;
+                    Debug.WriteLine($"PK test - {objSchema.Name} has a PK {primaryKeyProp.Name}");
+                }
+            }
+            Assert.That(classesWitoutPK > classesWithPK);  // slightly crap assertion to ensure test works
```
kristiandupont: Couldn't we assert that the schema of a class with a known primary key returns the expected result?

AndyDentFree: Yes, that might be a nicer unit test but I wanted this to be fairly independent of changes to all our other class declarations and, most importantly, illustrate the point of iterating through a bunch of classes and pulling out those with primary keys.

It would get larger and noisier to add the PK/class matches.

Tests/IntegrationTests.Shared/PrimaryKeyTests.cs
```diff
@@ -275,6 +275,27 @@ public void PrimaryKeyFailsIfClassNotinRealm()
...
+            int classesWitoutPK = 0;
+            foreach (var objSchema in _realm.Schema)
+            {
+                var primaryKeyProp = objSchema.Where(prop => prop.IsPrimaryKey).FirstOrDefault();
+                if (string.IsNullOrEmpty(primaryKeyProp.Name))
+                    classesWitoutPK++;
+                else
+                {
+                    classesWithPK++;
+                    Debug.WriteLine($"PK test - {objSchema.Name} has a PK {primaryKeyProp.Name}");
```
fealebenpae: Can we comment this out or something? Logging to the console in unit tests makes it harder to reason about their output when running them headless or just looking at the application output.

Tests/IntegrationTests.Shared/PrimaryKeyTests.cs
```diff
@@ -275,6 +275,27 @@ public void PrimaryKeyFailsIfClassNotinRealm()
                 var obj = skinny.ObjectForPrimaryKey<PrimaryKeyInt64Object>(42);
             });
         }
+
+        // mostly a little test to show iterating schema and getting primary key names
+        [Test]
+        public void CanGetNameOfPrimaryKeyFromSchema()
+        {
+            int classesWithPK = 0;
```
nirinchev: These should be `var`s.

Tests/IntegrationTests.Shared/PrimaryKeyTests.cs
```diff
@@ -275,6 +275,27 @@ public void PrimaryKeyFailsIfClassNotinRealm()
...
+            int classesWitoutPK = 0;
+            foreach (var objSchema in _realm.Schema)
+            {
+                var primaryKeyProp = objSchema.Where(prop => prop.IsPrimaryKey).FirstOrDefault();
```
nirinchev: We could simplify by using `FirstOrDefault(prop => prop.IsPrimaryKey)`, no need for a `.Where`.

Tests/IntegrationTests.Shared/PrimaryKeyTests.cs
```diff
@@ -275,6 +275,27 @@ public void PrimaryKeyFailsIfClassNotinRealm()
...
+                if (string.IsNullOrEmpty(primaryKeyProp.Name))
+                    classesWitoutPK++;
```
nirinchev: A typo here - missing `h` in `without`.

Issue #808, opened 2016-08-30
Update object by PrimaryKey


kristiandupont:
Something functionally equivalent to the Java [copyToRealmOrUpdate](https://realm.io/docs/java/1.2.0/api/io/realm/Realm.html#copyToRealmOrUpdate-E-) or Swift's [add(object, true)](https://realm.io/docs/swift/1.0.2/api/Classes/Realm.html#/s:FC10RealmSwift5Realm3addFTCS_6Object6updateSb_T_)

AndyDentFree:
Consider whilst naming and designing this, keeping consistency amongst PrimaryKey-based methods:

* potential change to add the overload `CreateObject(primaryKey)`
* not just passing in one object by primary key but also a future update on a `RealmResults` that might set a bunch of properties from a dictionary, or lambda, like a SQL `update` 
* adding an overload `RemoveForPrimaryKey` for efficient removal

This is related to #661 but is more constrained as it requires a PrimaryKey

CrimsonKyle:
@AndyDentFree Is Realm Xamarin looking to support functionality similar to Realm Swift to allow updateing by Primary Key if Object with Primary Key Exists during a write transaction instead of throwing the current RealmDuplicatePrimaryKeyValueException just overwrite the object?
[https://realm.io/docs/swift/latest/#creating-and-updating-objects-with-primary-keys]

PR #857, opened 2016-10-06
Speed-up calls to Realm.Manage

nirinchev:
This is a change that aims two things
1) Bring the performance of `Realm.Manage` closer to `Realm.CreateObject`
2) Pave the way for `AddOrUpdate` (#808)

The essence here is weaving `CopyToRealm` in `IRealmObjectHelper` implementors that replaces the reflection that we used before. Here's the performance comparison

|         | CreateObject | Manage (woven) | Manage (reflection) |
|---------|--------------|----------------|---------------------|
| Sparse  | 1235         | ~~2931~~ 1523      | 11291               |
| Dense | 827          | 951            | 2527                |

Times are in `ms` and the test was performed with `100000` objects.
`Sparse` refers to objects with a lot of properties, where only a few are set.
`Dense` refers to objects where all properties are set.

The current implementation guards against hitting ObjectStore by doing
```csharp
if (field != default(fieldType))
{
    Property = field
}
```

**only for** value types, objects, and lists. It doesn't support nullable value types because the IL is getting out of hand as is and didn't seem like it would add a lot of value.


Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
```diff
@@ -25,6 +27,34 @@ internal class DynamicRealmObjectHelper : IRealmObjectHelper
     {
         internal static readonly DynamicRealmObjectHelper Instance = new DynamicRealmObjectHelper();
 
+        public void CopyToRealm(RealmObject instance)
```
fealebenpae: Objects whose helper is `DynamicRealmObjectHelper` cannot exist in an unmanaged state - they always belong to a realm. I think it's safe to make this method throw an exception instead.

RealmWeaver/ModuleWeaver.cs
```diff
@@ -806,6 +826,74 @@ private TypeDefinition WeaveRealmObjectHelper(TypeDefinition realmObjectType, Me
...
+            /*
+				var casted = (ObjectType)instance;
+				
+				*foreach* non-list woven property in casted's schema
+				casted.Property = casted.Field;
+
+				*foreach* list woven property in casted's schema
+				var list = casted.field;
+				casted.field = null;
+				((ICopyValuesFrom)casted.Property).CopyValuesFrom(list);
```
fealebenpae: We should get rid of `ICopyValuesFrom` instead of making it public. Let's add a public `AddRange(IEnumerable<T>)` to `RealmList<T>` and call that instead.

nirinchev:
I figured it's better to split it in two, so that we can brag about "major speed improvements to Manage" in our release notes. Then after a release or two, we could add the default check and brag about speed improvements again 😆 If we make it ultra-fast in one go, people will think we're not improving anymore!

fealebenpae:
Oh haha, fair point. We can save some performance improvements for the next release.

nirinchev:
Ping @fealebenpae care to take a second look?

RealmWeaver/ModuleWeaver.cs
nirinchev: That was a bit of a pain, but I got rid of `ICopyValuesFrom`. I'll leave internalizing `RealmList` to #858.

Tests/IntegrationTests.Shared/StandAloneObjectTests.cs
```diff
@@ -85,5 +82,90 @@ public void RealmObject_WhenStandalone_ShouldHaveDefaultEqualsImplementation()
 
             Assert.DoesNotThrow(() => _person.Equals(otherPerson));
         }
+
+        [Test]
+        public void RealmObject_WhenManaged_ShouldNotThrow()
```
nirinchev: Sure, I will take a look at the WeaverTests infrastructures and try to come up with one.

RealmWeaver/ModuleWeaver.cs
```diff
@@ -810,6 +809,83 @@ private static bool IsRealmList(PropertyDefinition property)
         return property.PropertyType.Name == "RealmList`1" && property.PropertyType.Namespace == "Realms";
     }
 
+    private static bool IsDateTimeOffset(PropertyDefinition property)
```
kristiandupont: I know we've been using strings already, but would it be nicer to create something like
```
private static bool Is<T>(PropertyDefinition property)
{
    return property.PropertyType.Name == typeof(T).Name && ...
}
```
?

RealmWeaver/ModuleWeaver.cs
kristiandupont: Very nice!!

RealmWeaver/ModuleWeaver.cs
nirinchev: Yes it would, but I didn't want to diverge too much from the coding style. Also, when we add StyleCop, we'll probably have to reorder/reformat the whole file and it'll complain about constants/nameof/typeof :)

fealebenpae: It bugs me that we do string comparison in the first place - I'd much rather we compare `TypeDefinition` instances.

fealebenpae:
I believe that you should remove the `BackingFieldName` property of `WovenPropertyAttribute` - it shouldn't be used anymore.

nirinchev:
Do we need the `WovenPropertyAttribute` in that case?

fealebenpae:
We could get rid of `WovenPropertyAttribute`, but right now it's a very useful marker for properties that are persisted. Prior to that we had duplicate code in the library and the weaver.

👍 from me.

kristiandupont:
:+1: from me as well. Very nice work!

Issue #759, opened 2016-08-08
Don't close in Dispose


kristiandupont:
Currently, Dispose closes the database. This causes other instances to close as well which is wrong. However, just relying on the reference count seems not to work on Windows so that needs to be resolved.

petersondrew:
So is there currently a workaround to this behavior? Also, when *should* we call close? I had been considering using `ThreadLocal<>` to manage my Realm instances across threads but decided it was easier to call `GetInstance` wrapped in a using statement whenever I need a Realm instance. This bug would mean disposing would actually create cross-thread issues correct?

petersondrew:
@kristiandupont Any advice for the above questions? Thanks.

kristiandupont:
@petersondrew my apologies, this flew off my radar. Calling `GetInstance` in a using statement will work fine because only other instances on the same thread will be closed along with the one being disposed. We are going to remove the `Close` method altogether because it's quite confusing.

nirinchev:
This has been superseded by the `RealmState` approach.

PR #858, opened 2016-10-06
Rename public API to adhere to .NET standards

### LINKED ISSUE #853: Revisit the public API and align it with .NET standards
Currently there are some places where our public API is confusing and not aligned with .NET naming conventions.

Todo:
- [x] Rename `Realm.All` to `GetAll`
- [x] Rename `Realm.ObjectForPrimaryKey` to `Find`
- [x] Rename `Realm.Manage` to `Add`
- [ ] Rename `Realm.RealmChanged` to something
- [x] Rename `RealmConfiguration.ReadOnly` to `IsReadOnly`
- [x] Rename `RealmConfiguration.PathToRealm` to `GetPathToRealm`
- [x] Remove `RealmList.ITEM_NOT_FOUND`
- [x] Rename `RealmResults.NotificationCallback` to `NotificationCallbackDelegate`
- [x] Remove `Realm.Close` - related to #759 
- [ ] Make `RealmList` internal

nirinchev:
This PR adjusts some of the public API names to adhere to the [guidelines](https://msdn.microsoft.com/en-us/library/ms229002%28v=vs.110%29.aspx).

Status:
- [x] ~~Rename `Realm.All` to `GetAll`~~
- [x] Rename `Realm.ObjectForPrimaryKey` to `Find`
- [x] Rename `Realm.Manage` to `Add`
- [ ] ~~Rename `Realm.RealmChanged` to something~~
- [x] Rename `RealmConfiguration.ReadOnly` to `IsReadOnly`
- [x] Rename `RealmConfiguration.PathToRealm` to `GetPathToRealm`
- [x] Remove `RealmList.ITEM_NOT_FOUND`
- [x] Rename `RealmResults.NotificationCallback` to `NotificationCallbackDelegate`
- [x] Remove `Realm.Close` #759
- [x] Make `RealmList` internal
- [ ] Run FXCop or something similar to see if we've missed something

**TBD**
@kristiandupont suggested revisiting class names as well to see if we should get rid of the `Realm` prefix some of them have.

Notes:
- I've removed the `ObjectDisposedException` thrown in `Realm.Dispose`. My reasoning is that it is inconsistent with the way other disposables behave - we only throw it in `Dispose`, but not otherwise - e.g. in `Realm.Config`. This creates weird scenarios like in our tests, where we dispose a Realm and then start accessing stuff, which would traditionally throw. Now we could market `Dispose` as a convenience around `Close` that can be used with a `using ()`.
- `RealmChanged` has not been renamed because I'm still figuring out a nice name.
- The docs should be updated to start using the new API.

Fixes #853


Realm.Shared/RealmList.cs
```diff
@@ -306,7 +300,7 @@ public void RemoveAt(int index)
         private void ManageObjectIfNeeded(T obj)
```
kristiandupont: I guess this should be `AddIfNeeded`?

Tests/WeaverTests/AssemblyToProcess/ClassWithStaticProperty.cs
```diff
@@ -18,7 +18,7 @@ public static Dog MyOnlyDog
                 //{
                 //    return myDog;
                 //}
-                //var dogs = Realm.GetInstance().All<Dog>();
+                //var dogs = Realm.GetInstance().etAll<Dog>();
```
kristiandupont: typo: `etAll`

Realm.Shared/RealmList.cs
nirinchev: I was only looking through public members, I'm guessing that's why this one slipped. I'll take a second look for other similar cases - I'd say rename everywhere to avoid confusion among team members as well.

Realm.PCL/RealmObjectPCL.cs
```diff
@@ -36,7 +36,7 @@ public class RealmObject
         private Realm _realm;  // may not be used but wanted it included in definition of IsManaged below.
 
         /// <summary>
-        /// Allows you to check if the object has been associated with a Realm, either at creation or via Realm.Manage.
+        /// Allows you to check if the object has been associated with a Realm, either at creation or via Realm.Add.
         /// </summary>
         public bool IsManaged => _realm != null;
```
nirinchev: In Java it's [`isManaged`](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/RealmObject.java#L263), while the `Add` operation is called [`copyToRealm`](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/Realm.java#L882) so a little inconsistency seems fine (or at least it won't be us to set a precedent). Personally, I think that it's fine to stick with `IsManaged` and the `XXXManagedXXXException`s.

AndyDentFree:
As a key principle, just saying something is _to adhere to the guidelines_ is too vague - for any significant change we need to say **which** guideline.

I really dislike renaming All to GetAll. It is really a property which only is implemented as a member because it is generic. **Nobody** has complained about that name, AFAIK.

I also think renaming ObjectForPrimaryKey to Find is losing syntactic information and will confuse people thinking it is a generic Find. It is a natural complement to All so if GetAll makes sense why not GetObject? The changes are inconsistent.

Most of all, I think we should be advertising this widely to the userbase for comment, unless I have missed something.

sushihangover:
My 2cents as a end-user:

ReadOnly -> IsReadOnly 👍 
All -> GetAll : 👎 

ObjectForPrimaryKey -> Find 👎  **VERY** confusing if I can not use it to "Find" on any object member and need to resort to Linq... `FindKey` `FindByKey`, `ByKey`, etc... would be more appropriate.

Manage -> Add 🎱  If you have a "Add" and ignore the fact that "Add" now has this strange 2nd parameter named `update`, will you move to a CRUD naming style since Realm is a `Transaction`al system thus `CreateObject` should just be `Create`, etc... or move to a **Atomic** Collection/List style API wth implied single object transactions: i.e.:


`CRUD` style: 👍 
* CreateObject -> Create<T>()
* Remove -> Delete<T>()
* RemoveAll -> Clear<T>()
* Manage w/ update: false -> Add<T>()
* Manage w/ update: true -> Update<T>() 
* BeginWrite => BeginTransaction

Or Mutable `Collection` style: 🎱 (but I like it in terms of thinking Realm is object-based, especially if it had atomic-style naming/intents for Transactions (see `BeginWrite` -> `Do` below)
* CreateObject -> Create<T>()
* RemoveAll       -> Clear<T>() 
* Manage -> `Add<T>()` 🎱 👎   All RealmObjects are mutable so does `Add` and/or `Update` follow some mutation pattern: if object has PK and is found, update non-PK members, if non-PK-based does it always Add if the object is not currently Managed?, etc.. `Managing` PK vs non-PK objects and currency managed vs **non-managed** objects can really distort a `Collection` style API...
* BeginWrite -> Do 👍 👍 👍  (C# Atomics)

```
Realm.Do() => {
    obj.PropertyA = value;
    obj.PropertyB = value;
});
```
* ....

Personally `Manage` really makes sense to me if I have an un-manage object and now wish to have Realm manage it... the fact that is handles managed/unmananged/adds/updates seems like functional  overload... but that just me...

IHMO, once you established an API usage pattern, transaction CRUD or atomic Collection based, the API naming falls in to place...

nirinchev:
@sushihangover thank you so much for the detailed feedback. We had a long call to discuss the renaming a few hours ago, so let me try to summarize some of the motivation :) We wanted the API to be loosely based on existing .NET naming conventions, especially with regard to collections. That is the reason for `Add`, `Remove`, `RemoveRange`, `RemoveAll` (taken from `List<T>`). 

Regarding `CreateObject` -> `Create`, I like it and I do feel the `Object` part is somewhat redundant.

`Find` has been probably most extensively discussed - initially, I went with `Find` because Entity Framework has it and does what our `Find`/`ObjectForPrimaryKey` does, so wanted to adhere to a familiar persistence API. On the other hand, it deviates from our other `List`-like API in that, indeed, it does not take a predicate like `List<T>.Find` does.

Regarding `Add<T>(T item, bool update)`, now that you brought it up, I feel that I should have split that into two separate methods - `Add` and `AddOrUpdate`, but at the time an optional argument seemed to work just fine. However, changing that is beyond the scope of this PR because it will require changes to the woven logic, which are not simple renames, and I wanted to avoid hiding meaningful changes in a PR that touches almost all files :) But we could add a separate issue about it.

tl;dr: In general, we aim more towards the Mutable Collection style, borrowing heavily from `List<T>` and EF, but we're also okay with deviating from that when it makes sense.

PS: `Manage<T>`/`Add<T>` does indeed follow some very simple set of rules:
- If managed, return, we're done here.
- If unmanaged, `update = true`, and no PK, adds.
- If unmanaged, `update = true`, and has PK, updates if existing object has the same PK.

For lists/relationships, `Manage` is called with the parent value of the `update` argument and the same rules apply - e.g. if you have an object with PK, that points to an object without, the child object is added, and the parent - updated.

sushihangover:
@nirinchev Thanks for the detailed response... and tracking the API against `List<T>` is 👍 from me...
 
I would *highly* recommend adding those `Manage<T>/Add<T> "set of rules" to the public docs as I figured it out, but actually added a NUnit test to performed those sequences just so I could catch any changes to those sequences as my app is *assuming* that flow otherwise I could be tracing weird data "corruption" issues. I have some Realm-based forward tree chaining time-based sequenced lists that allow *infinite* undo/redo, copy/paste and replay actions that I will be using for real-time collaboration via the Realm Mobile Platform ❤️ (was planing for a custom sync before the RMP beta announcement) and it is highly dependent upon that ordered set of rules...

The only thing that really smells is `Add<T>(T item, bool update)`, that `update:` param just really bothers me some reason ... but as you said, not a part of this PR, maybe a new issue (hopefully before a 1.0 release 👍 ) 

Thanks again.

nirinchev:
Yes - we were holding off the docs update to avoid having to rewrite them when this PR lands, but as soon as it does, we will write a section for `Add` and we also plan to write a detailed blog post about the rationale of the renaming :)

nirinchev:
retest this please

PR #862, opened 2016-10-07
Introduce Realm Object Server API

### LINKED ISSUE #860: ROS api

kristiandupont:
This resolves #860.


Realm.Shared/User.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public static User FromJson()
+        {
+            throw new NotImplementedException();
+        }
+
+        public static User Authenticate(Credentials credentials, string serverUrl)
```
fealebenpae: Unless we really need to have it, I'd suggest nixing the synchronous API.

Realm.Shared/User.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace Realms
+{
+    public class User
+    {
+        public static User GetCurrentUser()
+        {
+            throw new NotImplementedException();
+        }
+
+        public static User FromJson()
```
fealebenpae: I would suggest we have `User` implement `ISerializable` which Json.NET and other serializers support natively.

Realm.Shared/User.cs
```diff
@@ -0,0 +1,62 @@
...
+        public string Identity { get; private set; }
+
+        public bool IsValid { get; private set; }
+
+        public void LogOut()
+        {
+            throw new NotImplementedException();
+        }
+
+        public string ToJson()
```
fealebenpae: Same as the comment I left on `User.FromJson`.

Realm.Shared/Credentials.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Realms
+{
+    public class Credentials
+    {
+        public static Credentials Custom(string identityProvider, string userIdentifier,
+            Dictionary<string, object> userInfo)
```
fealebenpae: I would suggest making this an `IDictionary<string, object>`.

Realm.Shared/Session.cs
```diff
@@ -0,0 +1,40 @@
...
+        Initial,
+        Stopped,
+        Unbound
+    }
+
+    public class Session
+    {
+        public SyncConfiguration Configuration { get; private set; }
+        public Uri ServerUri { get; private set; }
+        public SessionState State { get; private set; }
```
fealebenpae: Can we expose a `StateChanged` event or something similar? I mean - do we get notifications like that from Object Store?

Realm.Shared/User.cs
```diff
@@ -0,0 +1,62 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+using System;
+using System.Threading.Tasks;
+
+namespace Realms
+{
+    public class User
+    {
+        public static User GetCurrentUser()
```
fealebenpae: This might work better as a property. Or field.

Realm.Shared/User.cs
```diff
@@ -0,0 +1,62 @@
...
+            throw new NotImplementedException();
+        }
+
+        public string AccessToken { get; private set; }
+
+        public string Identity { get; private set; }
+
+        public bool IsValid { get; private set; }
+
+        public void LogOut()
```
fealebenpae: If the implementation allows it, this method should return a `Task` as well.

Realm.Shared/User.cs
nirinchev: Property! Fields are evil!

Platform.XamarinAndroid/Realm.Sync.XamarinAndroid/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+[assembly: AssemblyTitle("Realm.Sync.XamarinAndroid")]
+[assembly: InternalsVisibleTo("Tests.XamarinAndroid")]
```
nirinchev: Is there LinkerSafe on Android?

Platform.XamarinIOS/Realm.Sync.XamarinIOS/ApiDefinition.cs
```diff
@@ -0,0 +1,62 @@
+﻿// WARNING even though this file contains apparently generic content with no function
```
nirinchev: Seems like the header is missing

Shared/Realm.Sync.Shared/SyncConfiguration.cs
```diff
@@ -17,6 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 using System;
+using System.Runtime.CompilerServices;
```
nirinchev: why is this needed?

fealebenpae:
retest this please

Shared/Realm.Shared/RealmConfiguration.cs
```diff
@@ -239,6 +235,56 @@ public bool Equals(RealmConfiguration rhs)
...
+        internal virtual void DeleteRealm()
+        {
+            // TODO add cache checking when implemented, https://github.com/realm/realm-dotnet/issues/308
+            // when cache checking, uncomment in IntegrationTests.cs RealmInstanceTests.DeleteRealmFailsIfOpenSameThread and add a variant to test open on different thread
+            var lockOnWhileDeleting = new object();
+            lock (lockOnWhileDeleting)
+            {
+                var fullpath = DatabasePath;
+                File.Delete(fullpath);
+                File.Delete(fullpath + ".log_a");  // eg: name at end of path is EnterTheMagic.realm.log_a
```
AndyDentFree: Is this set of files still accurate in current core? I thought the log had been moved into the Realm?

Shared/Realm.Sync.Shared/Credentials.cs
```diff
@@ -0,0 +1,162 @@
...
+using System.Collections.Generic;
+using System.Json;
+
+namespace Realms.Sync
+{
+    public class Credentials
+    {
+        internal static class Providers
+        {
+            internal const string Debug = "debug";
```
AndyDentFree: Are these strings going to get moved into ObjectStore?

Shared/Realm.Sync.Shared/Native/SyncConfiguration.cs
```diff
@@ -0,0 +1,52 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+
+namespace Realms.Sync.Native
+{
+    [StructLayout(LayoutKind.Sequential)]
+    [SuppressMessage("StyleCop.CSharp.NamingRules", "SA1307:AccessibleFieldsMustBeginWithUpperCaseLetter")]
```
kristiandupont: Is this suppression still necessary? There doesn't seem to be any accessible fields?

Shared/Realm.Sync.Shared/RealmExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+using System;
+
+namespace Realms.Sync
+{
+    public static class RealmExtensions
+    {
+        public static Session GetSession(this Realm @this)
```
kristiandupont: We could remove all things *session* from this PR if we wanted.

Shared/Realm.Sync.Shared/Credentials.cs
fealebenpae: Not sure - they're only used for REST calls to the object server and that's unlikely to make it in object store.

Shared/Realm.Shared/RealmConfiguration.cs
fealebenpae: It might be stale - @kristiandupont suggested moving this functionality to object store going forward might be better in the long run.

Shared/Tests.Sync.Shared/SyncConfigurationTests.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            _realm.Dispose();
+        }
+
+        [Test]
+        public void Foo()
```
kristiandupont: Hehe well I guess we can safely remove this test...

Platform.XamarinIOS/Realm.Sync.XamarinIOS/ApiDefinition.cs
fealebenpae: It's the same as in `Realm.XamarinIOS`. Since it basically comes from the Xamarin Studio template, I don't know if we need or want to have the license header :/

Platform.XamarinAndroid/Realm.Sync.XamarinAndroid/Properties/AssemblyInfo.cs
fealebenpae: Not that I could find.

Shared/Realm.Sync.Shared/RealmExtensions.cs
fealebenpae: Not all, we still need the `SessionHandle` for authentication :P

I could remove this, but I'd only end up adding it again later :/

Shared/Tests.Sync.Shared/SyncConfigurationTests.cs
fealebenpae: Haha, good catch!

Shared/Realm.Sync.Shared/RealmExtensions.cs
kristiandupont: Yeah, let's just keep it then.

AndyDentFree: What is this for? missing comment for API doc.

Shared/Realm.Sync.Shared/SyncConfiguration.cs
```diff
@@ -0,0 +1,55 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+using System;
+using System.Runtime.CompilerServices;
+
+namespace Realms.Sync
+{
+    public class SyncConfiguration : RealmConfiguration
```
AndyDentFree: Needs XML doc.

fealebenpae: I've added an issue to track API docs - https://github.com/realm/realm-dotnet/issues/929.

My hope is to get this merged soon so that we can tackle the issues, including documentation, in parallel.

AndyDentFree:
Lets make documenting the new stuff a separate issue but must be 1.0!

AndyDentFree:
👍

Shared/Realm.Shared/RealmConfiguration.cs
```diff
@@ -239,6 +235,56 @@ public bool Equals(RealmConfiguration rhs)
...
+
+            srHandle.SetHandle(srPtr);
+            return new Realm(srHandle, this, schema);
+        }
+
+        internal virtual void DeleteRealm()
+        {
+            // TODO add cache checking when implemented, https://github.com/realm/realm-dotnet/issues/308
+            // when cache checking, uncomment in IntegrationTests.cs RealmInstanceTests.DeleteRealmFailsIfOpenSameThread and add a variant to test open on different thread
+            var lockOnWhileDeleting = new object();
```
nirinchev: What is this lock doing?

Shared/Realm.Sync.Shared/Credentials.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            internal const string Password = "password";
+
+            internal const string Identity = "identity";
+
+            internal const string IsAdmin = "isAdmin";
+        }
+
+        public static Credentials Custom(string identityProvider, string userIdentifier,
+            IReadOnlyDictionary<string, object> userInfo)
```
nirinchev: Do we want to force `userInfo` to be passed as readonly?

Shared/Realm.Sync.Shared/Credentials.cs
```diff
@@ -0,0 +1,162 @@
...
+
+        public IReadOnlyDictionary<string, object> UserInfo { get; private set; } = new Dictionary<string, object>();
+
+        private Credentials()
+        {
+        }
+
+        internal JsonObject ToJson()
+        {
+            var user_info = new JsonObject();
```
nirinchev: That's not c++ ;)

Shared/Realm.Sync.Shared/SyncConfiguration.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        public Uri ServerUri { get; private set; }
+
+        public User User { get; private set; }
+
+        public bool ShouldDeleteRealmOnLogOut { get; private set; }
+
+        public SyncConfiguration(User user, Uri serverUri)
```
nirinchev: Can we use string instead of an `Uri` here? I've always felt `Uri`-based API are somewhat cumbersome. We could always do a `var uri = new Uri(ServerUri)` whenever we need.

Shared/Realm.Sync.Shared/SyncConfiguration.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            this.User = user;
```
nirinchev: Are we doing `this`-es now? 😱 🙏

Shared/Realm.Sync.Shared/User.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            var json = new JsonObject
+            {
+                ["data"] = RefreshToken,
+                ["path"] = path,
+                ["provider"] = "realm"
+            };
+
+            var result = await MakeAuthRequestAsync(ServerUri, json, TimeSpan.FromSeconds(30)).ConfigureAwait(continueOnCapturedContext: false);
+            var access_token = result["access_token"];
```
nirinchev: The c# variable doesn't have to be named like the json field

wrappers/src/sync_manager_cs.cpp
```diff
@@ -0,0 +1,86 @@
...
+    std::shared_ptr<SyncUser>* user;
+
+    uint16_t* url;
+    size_t url_len;
+};
+
+void (*s_refresh_access_token_callback)(std::shared_ptr<SyncUser>*, std::shared_ptr<SyncSession>*, const char* path, size_t path_len);
+
+extern "C" {
+REALM_EXPORT void realm_initialize_sync(const uint16_t* base_path_buf, size_t base_path_len, decltype(s_refresh_access_token_callback) refresh_callback)
```
nirinchev: Should the other arguments be const as well?

wrappers/src/sync_manager_cs.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+    
+    
+REALM_EXPORT SharedRealm* shared_realm_open_with_sync(Configuration configuration, SyncConfiguration sync_configuration, SchemaObject* objects, int objects_length, SchemaProperty* properties, uint8_t* encryption_key, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        Realm::Config config;
+        config.schema_mode = SchemaMode::Additive;
+
+        // by definition the key is only allowwed to be 64 bytes long, enforced by C# code
```
nirinchev: Typo - `allowwed`

wrappers/src/sync_manager_cs.cpp
```diff
@@ -0,0 +1,86 @@
...
+        if (encryption_key)
+          config.encryption_key = std::vector<char>(encryption_key, encryption_key+64);
```
nirinchev: Missing curly braces + wrong indentation (2 instead of 4 spaces)

Shared/Realm.Sync.Shared/Credentials.cs
fealebenpae: It's not, but it's the name for the field in the JSON dictionary in the authentication protocol.

wrappers/src/sync_manager_cs.cpp
fealebenpae: They're primitives - we can't change their contents the same way as the string buffer so there's no need to denote them as `const`.

Shared/Realm.Sync.Shared/Credentials.cs
fealebenpae: You can still pass a `Dictionary<string, object>` because it, strangely enough, implements the interface.

Shared/Realm.Sync.Shared/SyncConfiguration.cs
fealebenpae: I prefer using `Uri` because this puts the onus on the user of the API to make sure they have a well-formed URI string to begin with.

fealebenpae: Also, Cocoa uses `NSURL` here.

Shared/Realm.Shared/RealmConfiguration.cs
fealebenpae: Sigh, this was just moved here from somewhere else. I'll just undo the refactoring of the deletion logic.

Shared/Realm.Sync.Shared/User.cs
fealebenpae: You're right, but I find using the same name helps me reason about things better. We also use snake_case for identifiers that denote something from our C++ layer like `DllImport`-ed functions or marshaled structures, so I could argue there is precedent for this.

Shared/Realm.Sync.Shared/SyncConfiguration.cs
nirinchev: That's what I dislike ;) In almost all cases when an API requires `Uri`, I'll pass in `new Uri(myStringUri)`, which feels annoying and cumbersome. We can always do an `Uri.TryParse` and if it's not well formed throw an exception to let the user know something's wrong.

Shared/Realm.Sync.Shared/SyncConfiguration.cs
nirinchev: I don't think there's a rule for that. There is one for forcing `this`-es (that's disabled), but it doesn't come with one that forces their absence.

wrappers/src/sync_manager_cs.cpp
nirinchev: But then you would never learn! Which is the point of those reviews 😜

Shared/Realm.Sync.Shared/Credentials.cs
fealebenpae: Yeah, you're right - that would be better. I'll change it.

Shared/Realm.Sync.Shared/SyncConfiguration.cs
fealebenpae: @AndyDentFree @kristiandupont: `string` vs `Uri`?

Shared/Realm.Shared/Realm.cs
```diff
@@ -301,7 +301,19 @@ public override int GetHashCode()
         /// <param name="configuration">A configuration which supplies the realm path.</param>
         public static void DeleteRealm(RealmConfiguration configuration)
         {
-            configuration.DeleteRealm();
+            // TODO add cache checking when implemented, https://github.com/realm/realm-dotnet/issues/308
+            // when cache checking, uncomment in IntegrationTests.cs RealmInstanceTests.DeleteRealmFailsIfOpenSameThread and add a variant to test open on different thread
+            var lockOnWhileDeleting = new object();
+            lock (lockOnWhileDeleting)
```
nirinchev: I know this is just moving things to where they were, but this lock seems wrong, do you feel like refactoring it or adding a TODO to fix it later? :)

Shared/Realm.Sync.Shared/SyncConfiguration.cs
kristiandupont: I am not strongly in one camp or the other. I do think it's nice to emphasize that this is not the path. That would have been a bigger problem if the user parameter wasn't there, though..

PR #865, opened 2016-10-13
.net standard

kristiandupont:
This is, for now, an experiment to see if we can switch to .net standard and have only one library, before introducing sync. This will require a bit of weaver magic for realm.dll, but it might be possible.


RealmWeaver/RealmAssemblyWeaver.cs
```diff
@@ -0,0 +1,92 @@
...
+using Mono.Cecil;
+using Mono.Cecil.Rocks;
+
+internal static class RealmAssemblyWeaver
+{
+    internal static void PrepareRealmAssemblyForIos(AssemblyDefinition realmAssembly, IAssemblyResolver assemblyResolver)
+    {
+        UpdateWrappersReference(realmAssembly, "__Internal");
+
+        var monotouchModule = assemblyResolver.Resolve("Xamarin.iOS").MainModule;
```
kristiandupont: Yes it should. All this code is in a very experimental stage so please don't read too much into it just yet. :-)

kristiandupont:
Putting this on hold for now because it seems impossible to make it build in Xamarin Studio just yet. We can revisit this soon-ish as support seems to be merged into master: https://github.com/mono/monodevelop/pull/1622

PR #866, opened 2016-10-13
Switch to using property indices instead of columns

### LINKED ISSUE #682: Verify thread confinement on property access

### LINKED ISSUE #864: Use Property indices instead of column indices
In case the schema changes dynamically, we will need to keep the column indices up to date. Currently, we store these in `RealmObject._metaData` when the realm is created. We should use property indices instead, and let object store resolve the column index for every property access.

nirinchev:
Should save us a few managed<->unmanaged transitions - we now opportunistically call native setter/getter and verify realm closed/row detached/transaction/threads.

@fealebenpae @kristiandupont please verify this, especially the c++ code since everything's fairly new to me.

Fixes #864 
Fixes #682


Realm.Shared/RealmList.cs
```diff
@@ -162,13 +162,13 @@ public int Count
                     throw new ArgumentOutOfRangeException();
                 }
 
-                var linkedRowPtr = _listHandle.Get((IntPtr)index);
+                var linkedRowPtr = _listHandle.Get((IntPtr)index, _realm.SharedRealmHandle);
```
fealebenpae: Object Store's `List` already has a `SharedRealm`.

Realm.Shared/RealmList.cs
```diff
@@ -183,7 +183,7 @@ public int Count
         public void Add(T item)
         {
             this.ManageObjectIfNeeded(item);
-            var rowIndex = item.RowHandle.RowIndex;
+            var rowIndex = item.ObjectHandle.RowIndex;
             _listHandle.Add(rowIndex);
```
fealebenpae: Consider passing the `ObjectHandle` directly instead of fetching the row index in C# to save one extra native call. You can/should do that in a follow-up PR, though.

Realm.Shared/Realm.cs
```diff
@@ -716,8 +705,9 @@ public RealmResults<dynamic> All(string className)
         /// <exception cref="RealmClassLacksPrimaryKeyException">If the RealmObject class T lacks an [PrimaryKey].</exception>
         public T ObjectForPrimaryKey<T>(long id) where T : RealmObject
         {
+            // TODO
             var metadata = Metadata[typeof(T).Name];
-            var rowPtr = NativeTable.RowForPrimaryKey(metadata.Table, metadata.PrimaryKeyColumnIndex, id);
+            var rowPtr = NativeTable.RowForPrimaryKey(SharedRealmHandle, metadata.Table, metadata.PrimaryKeyColumnIndex, id);
```
fealebenpae: Get rid of `PrimaryKeyColumnIndex` - you can look it up directly in native.

Realm.Shared/RealmObject.cs
```diff
@@ -84,163 +84,162 @@ internal class Metadata
...
 
         #region Getters
 
         protected string GetStringValue(string propertyName)
         {
-            Debug.Assert(_realm != null, "Object is not managed, but managed access was attempted");
+            Debug.Assert(IsManaged, "Object is not managed, but managed access was attempted");
 
-            return NativeTable.GetString(_metadata.Table, _metadata.ColumnIndices[propertyName], _rowHandle.RowIndex);
+            return NativeTable.GetString(_objectHandle, _metadata.PropertyIndices[propertyName]);
```
fealebenpae: I think the `Get`/`Set` family of methods should now be moved to `ObjectHandle`.

Realm.Shared/linq/RealmResults.cs
```diff
@@ -66,7 +66,7 @@ public class RealmResults<T> : IOrderedQueryable<T>, RealmResultsNativeHelper.In
                     throw new ArgumentOutOfRangeException();
                 }
 
-                var rowPtr = ResultsHandle.GetRow(index);
+                var rowPtr = ResultsHandle.GetObject(index);
```
fealebenpae: Rename `rowPtr` to `objectPtr`.

Realm.Shared/native/NativeTable.cs
```diff
@@ -28,104 +28,103 @@ namespace Realms
     [SuppressMessage("StyleCop.CSharp.ReadabilityRules", "SA1121:UseBuiltInTypeAlias")]
     internal static class NativeTable
     {
-        public static IntPtr AddEmptyRow(TableHandle tableHandle)
+        public static IntPtr AddEmptyRow(TableHandle tableHandle, SharedRealmHandle sharedRealm)
```
fealebenpae: Maybe we need to make a new `CreateEmptyObject(TableHandle)` method on the `SharedRealmHandle` class instead.

wrappers/src/marshalable_sort_clause.hpp
```diff
@@ -20,23 +20,33 @@
...
-        column_indices.push_back(std::vector<size_t>(flattened_column_indices + sort_clauses[i].offset, 
-            flattened_column_indices + sort_clauses[i].offset + sort_clauses[i].count));
+        
+        current_indices.clear();
+        current_indices.reserve(sort_clauses[i].count);
+        for(auto j = sort_clauses[i].offset; j < sort_clauses[i].offset + sort_clauses[i].count; ++j) {
+            size_t column_index = properties[flattened_property_indices[j]].table_column;
+            current_indices.push_back(column_index);
+        }
+        column_indices.push_back(current_indices);
```
fealebenpae: Consider `std::move`-ing `current_indices`.

wrappers/src/table_cs.cpp
```diff
@@ -28,144 +28,267 @@
...
+        
+        const size_t column_ndx = object_ptr->get_object_schema().persisted_properties[property_ndx].table_column;
+        const size_t link_row_ndx = object_ptr->row().get_link(column_ndx);
+        if (link_row_ndx == realm::npos)
+            return nullptr;
+        
+        auto target_table_ptr = object_ptr->row().get_table()->get_link_target(column_ndx);
+        const StringData target_name = ObjectStore::object_type_for_table_name(target_table_ptr->get_name());
+        const ObjectSchema& target_schema = *object_ptr->realm()->schema().find(target_name);
+        return new Object(object_ptr->realm(), target_schema, *new Row((*target_table_ptr)[link_row_ndx]));
```
fealebenpae: Drop `new` on `Row` and just initialize it like any other stack-allocated C++ class.

wrappers/src/table_cs.cpp
```diff
@@ -28,144 +28,267 @@
...
+    });
 }
 
-REALM_EXPORT size_t table_linklist_is_empty(Table* table_ptr, size_t column_ndx, size_t row_ndx, NativeException::Marshallable& ex)
+REALM_EXPORT size_t table_linklist_is_empty(const Object* object_ptr, size_t property_ndx, NativeException::Marshallable& ex)
 {
-    return handle_errors(ex, [&]()  {
-        return bool_to_size_t(table_ptr->linklist_is_empty(column_ndx, row_ndx));
+    return handle_errors(ex, [&]() {
+        if (object_ptr->realm()->is_closed())
```
fealebenpae: These checks occur at many places - consider refactoring them.

wrappers/src/table_cs.cpp
```diff
@@ -28,144 +28,267 @@
...
+        object_ptr->realm()->verify_thread();
+        
+        const size_t column_ndx = object_ptr->get_object_schema().persisted_properties[property_ndx].table_column;
+        const size_t link_row_ndx = object_ptr->row().get_link(column_ndx);
+        if (link_row_ndx == realm::npos)
+            return nullptr;
+        
+        auto target_table_ptr = object_ptr->row().get_table()->get_link_target(column_ndx);
+        const StringData target_name = ObjectStore::object_type_for_table_name(target_table_ptr->get_name());
+        const ObjectSchema& target_schema = *object_ptr->realm()->schema().find(target_name);
```
nirinchev: this is inconsistent with e.g. https://github.com/realm/realm-dotnet/pull/866/files#diff-f651c2b696cbccfb7b7a3b23bf46e21cR54 where we use `auto target_schema = ...`

fealebenpae: Yup, the one you linked to needs an `&` to prevent copying.

Realm.Shared/native/NativeTable.cs
kristiandupont: Yes, I think we are more or less ready to ditch this entire class.

wrappers/src/linklist_cs.cpp
```diff
@@ -46,22 +47,24 @@ REALM_EXPORT void linklist_insert(SharedLinkViewRef* linklist_ptr, size_t link_n
...
-    return new Row(rowExpr);
-  });
+    return handle_errors(ex, [&]() -> Object* {
+        const size_t count = (**linklist_ptr)->size();
+        if (link_ndx >= count)
+            throw IndexOutOfRangeException("Get from RealmList", link_ndx, count);
+        auto rowExpr = (**linklist_ptr)->get(link_ndx);
+        StringData object_name = ObjectStore::object_type_for_table_name((**linklist_ptr)->get_target_table().get_name());
+        auto object_schema = realm->get()->schema().find(object_name);
+        return new Object(*realm, *object_schema, *new Row(rowExpr));
```
kristiandupont: Shouldn't it be `Row(rowExpr)` instead of `*new Row(rowExpr)` to avoid leaking it?

Realm.Shared/RealmList.cs
nirinchev: Here we're using `LinkView` though, and I couldn't find a way to extract the realm from that one 😞

nirinchev:
People, take a look and let me know what you think.

What remains though is:
- [ ] Remove `ObjectHandle.RowIndex`
- [ ] Rework Query to not depend on `columnIndex` - currently it calls native's `tableview_get_column_index` before calling query's methods, all of which depend on `columnIndex`

PR #867, opened 2016-10-14
Remove RowIndex from ObjectHandle

nirinchev:
A follow-up to #866 - this removes the now-unnecessary `RowIndex` from the newly created `ObjectHandle`.

### DIFF
--- a/Realm.Shared/Handles/ObjectHandle.cs
+++ b/Realm.Shared/Handles/ObjectHandle.cs
@@ -127,7 +127,12 @@ namespace Realms
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_remove_row", CallingConvention = CallingConvention.Cdecl)]
             public static extern void remove_row(ObjectHandle handle, out NativeException ex);
+
+            [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_equals_object", CallingConvention = CallingConvention.Cdecl)]
+            public static extern IntPtr equals_object(ObjectHandle handle, ObjectHandle otherHandle, out NativeException ex);
         }
+
+        public static readonly ObjectHandle Null = new ObjectHandle(null);
 
         public bool IsValid
         {
@@ -137,18 +142,6 @@ namespace Realms
                 var result = NativeMethods.get_is_valid(this, out nativeException);
                 nativeException.ThrowIfNecessary();
                 return result == (IntPtr)1;  // inline equiv of IntPtrToBool
-            }
-        }
-
-        public IntPtr RowIndex
-        {
-            get
-            {
-                NativeException nativeException;
-                var result = NativeMethods.get_row_index(this, out nativeException);
-                nativeException.ThrowIfNecessary();
-
-                return result;
             }
         }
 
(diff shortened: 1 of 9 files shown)


nirinchev:
Ping @fealebenpae @kristiandupont for a review

wrappers/src/query_cs.cpp
```diff
@@ -55,7 +55,12 @@ REALM_EXPORT Object* query_find(Query* query_ptr, size_t begin_at_table_row, Sha
         return new Object(*realm, object_schema, Row((*query_ptr->get_table())[row_ndx]));
     });
 }
-
+    
+REALM_EXPORT Object* query_find_next(Query* query_ptr, const Object& after_object, SharedRealm* realm, NativeException::Marshallable& ex)
+{
+    return query_find(query_ptr, after_object.row().get_index() + 1, realm, ex);
```
fealebenpae: You can get `realm` from `after_object` here.

Realm.Shared/Handles/ObjectHandle.cs
```diff
@@ -127,8 +127,13 @@ private static class NativeMethods
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_remove_row", CallingConvention = CallingConvention.Cdecl)]
             public static extern void remove_row(ObjectHandle handle, out NativeException ex);
+
+            [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_equals_object", CallingConvention = CallingConvention.Cdecl)]
+            public static extern IntPtr equals_object(ObjectHandle handle, ObjectHandle otherHandle, out NativeException ex);
```
fealebenpae: This should return a `bool` instead of `IntPtr`. However, because the default size of the managed boolean type in .NET is different than the size of the C++ `bool` type, you will need to apply `[return: MarshalAs(UnmanagedType.I1)]`.

Realm.Shared/Handles/ObjectHandle.cs
```diff
@@ -127,8 +127,13 @@ private static class NativeMethods
...
         }
 
+        public static readonly ObjectHandle Null = new ObjectHandle(null);
```
fealebenpae: This looks strange to me. Can you give me and @kristiandupont a bit more background?

wrappers/src/object_cs.cpp
```diff
@@ -219,175 +219,183 @@ extern "C" {
...
     {
         handle_errors(ex, [&]() {
             verify_can_set(object_ptr);
             
-            object_ptr->row().get_table()->move_last_over(object_ptr -> row().get_index());
+            object_ptr.row().get_table()->move_last_over(object_ptr.row().get_index());
+        });
+    }
+    
+    REALM_EXPORT size_t object_equals_object(const Object& object_ptr, const Object& other_object_ptr, NativeException::Marshallable& ex)
```
fealebenpae: This should return `bool`. See comment on the P/Invoke declaration for this function.

Realm.Shared/Handles/ObjectHandle.cs
nirinchev: It's a leftover - will remove

wrappers/src/object_cs.cpp
```diff
@@ -219,175 +219,182 @@ extern "C" {
...
             
-            const size_t column_ndx = get_column_index(object_ptr, property_ndx);
-            object_ptr->row().set_bool(column_ndx, size_t_to_bool(value));
+            const size_t column_ndx = get_column_index(object, property_ndx);
+            object.row().set_bool(column_ndx, size_t_to_bool(value));
         });
     }
     
-    REALM_EXPORT void object_set_int64(const Object* object_ptr, size_t property_ndx, int64_t value, NativeException::Marshallable& ex)
+    REALM_EXPORT void object_set_int64(const Object& object, size_t property_ndx, int64_t value, NativeException::Marshallable& ex)
```
kristiandupont: I also don't understand how these compile with `const` objects, as `set_int` obviously mutates. In any case, it feels wrong to me to declare a parameter that is being manipulated `const`.

PR #880, opened 2016-10-24
Remove IRealmList

### LINKED ISSUE #873: Revisit the need of IRealmList
It seems with recent changes, the `IRealmList` interface is only used in `MetaRealmList`, so maybe we can get rid of it.

nirinchev:
Seems like it's no longer necessary. Also, make `RealmList` internal.

Resolves #873


nirinchev:
@fealebenpae can you verify the dynamic API is fine like that or I'm missing something?

nirinchev:
retest this please

nirinchev:
I think I've cleaned up the leftovers in the weaver, or have you noticed something that I've missed? I'll add an entry to the Changelog along with other changes that have not been added. Finally, while I generally agree with deprecating API gracefully, in this case, I believe a "nuke it" approach might work better due to the following:
- It's a very simple class swap, so even with a large solution, it shouldn't take users more than a few  minutes to fix their build.
- We'll get deprecated warnings when building from source.

fealebenpae:
You're right - I had missed the `ModuleWeaver.cs` changes in this PR, I'm sorry.

I suppose we can make an exception in this case, yes. It's our fault for not having deprecated `RealmList<T>` when we introduced `IList<T>` property support.

PR #882, opened 2016-10-24
Add weaver warning when applying realm attributes to non-persisted properties

### LINKED ISSUE #870: Add a weaver check that we don't have realm attributes on non-persisted properties
Right now, the weaver is ignoring non-persisted properties, which might introduce subtle bugs. E.g.
```csharp
class MyObject : RealmObject
{
    [PrimaryKey]
    public string Id { get; }
}
```

We should error-out at compile time in this case to warn the user to remove the attribute or add a setter.

nirinchev:
@kristiandupont @fealebenpae do you think this should be a warning or an error?

Resolves #870


nirinchev:
Ping @kristiandupont @fealebenpae

PR #885, opened 2016-10-25
Don't crash when binding to removed object with relationship

### LINKED ISSUE #883: Crash when binding to a child object property in Forms
When having the following object graph:
```csharp
class Person : RealmObject
{
    public Dog Dog { get; set; }
}

class Dog : RealmObject
{
}
```

The following databinding to a Person collection will crash when removing items from the collection/Realm:
```xml
<ViewCell>
  <StackLayout>
    <Label Text="{Binding Name}"/>
    <Label Text="{Binding Dog.Name}"/>
  </StackLayout>
</ViewCell>
```

The reason is that when an item is removed from the Realm, it is becomes invalid and querying its persisted properties throws an exception. On the other hand, Forms will interpret the `Dog` object as BindingContext for the `Label` and try to update it when the collection changes. This will hit the `Dog` getter and crash the app.

**Workaround**: For now, one workaround is to flatten the hierarchy using proxy properties e.g.
```csharp
public class Person : RealmObject
{
    /* ... */
    public string DogName => Dog.Name;
}
```
and bind to `Person.DogName` instead of `Person.Dog.Name`.

Unfortunately, this pollutes the objects and is not really intuitive, so we should guard against those scenarios and make them work out of the box.

nirinchev:
The problem stems from the fact that Xamarin.Forms tries to apply the binding, even after an object is removed. This causes our RealmObject to try query core about the relationship, which obviously fails. While our behavior is semantically correct - you shouldn't try to get properties of a detached object, we should play nicer with Forms.

The proposed solution is to implement `IReflectableType` which Forms [respects](https://github.com/xamarin/Xamarin.Forms/blob/f252238fec90d8f71b49e35212275b9e588964e3/Xamarin.Forms.Core/BindingExpression.cs#L125). Then, replace the getter of the property info with a custom one, that will check if the object is valid and if it's not, return the default value.

The methods we should overrider are [`TypeInfo.GetDeclaredProperty`](https://github.com/xamarin/Xamarin.Forms/blob/f252238fec90d8f71b49e35212275b9e588964e3/Xamarin.Forms.Core/BindingExpression.cs#L319), [`PropertyInfo.GetGetMethod`](https://github.com/xamarin/Xamarin.Forms/blob/f252238fec90d8f71b49e35212275b9e588964e3/Xamarin.Forms.Core/BindingExpression.cs#L326), and obviously, [`MethodInfo.Invoke`](https://github.com/xamarin/Xamarin.Forms/blob/f252238fec90d8f71b49e35212275b9e588964e3/Xamarin.Forms.Core/BindingExpression.cs#L555). 

To verify that the solution works, I modified the QuickJournal example, to contain an artificial related object. When referencing the Realm Nuget package, it will crash with an `InvalidObjectException` after deleting an object, but when referencing Realm built from source, it will work as expected.

This resolves #883

### DIFF
--- a/Platform.PCL/Realm.PCL/RealmObjectPCL.cs
+++ b/Platform.PCL/Realm.PCL/RealmObjectPCL.cs
@@ -20,13 +20,14 @@
 
 using System;
 using System.Collections.Generic;
+using System.Reflection;
 
 namespace Realms
 {
     /// <summary>
     /// Base for any object that can be persisted in a Realm.
     /// </summary>
-    public class RealmObject
+    public class RealmObject : IReflectableType
     {
         private Realm _realm;  // may not be used but wanted it included in definition of IsManaged below.
 
@@ -333,6 +334,12 @@ namespace Realms
         {
             RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
             return false;
+        }
+
+        public TypeInfo GetTypeInfo()
+        {
+            RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
+            return null;
         }
     }
 }
--- a/Platform.PCL/Realm.PCL/RealmObjectPCL.cs
+++ b/Platform.PCL/Realm.PCL/RealmObjectPCL.cs
@@ -336,7 +336,7 @@ namespace Realms
             return false;
         }
 
-        public TypeInfo GetTypeInfo()
+        TypeInfo IReflectableType.GetTypeInfo()
         {
             RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
             return null;
--- a/Shared/Realm.Shared/RealmObject.cs
+++ b/Shared/Realm.Shared/RealmObject.cs
@@ -19,6 +19,9 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Reflection;
 
 namespace Realms
 {
@@ -29,7 +32,7 @@ namespace Realms
     /// Has a Preserve attribute to attempt to preserve all subtypes without having to weave.
     /// </remarks>
     [Preserve(AllMembers = true, Conditional = false)]
-    public class RealmObject
+    public class RealmObject : IReflectableType
     {
         private Realm _realm;
         private ObjectHandle _objectHandle;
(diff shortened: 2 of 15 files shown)


nirinchev:
retest this please

Platform.PCL/Realm.PCL/RealmObjectPCL.cs
```diff
@@ -334,5 +335,11 @@ public override bool Equals(object obj)
             RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
             return false;
         }
+
+        public TypeInfo GetTypeInfo()
```
fealebenpae: Can we make this an explicitly-implemented interface member so it doesn't show up in IntelliSense?

Shared/Realm.Shared/RealmObject.cs
```diff
@@ -531,5 +534,10 @@ public override bool Equals(object obj)
             // System.Object, which defines Equals as reference equality. 
             return ObjectHandle.Equals(((RealmObject)obj).ObjectHandle);
         }
+
+        public TypeInfo GetTypeInfo()
```
fealebenpae: Same as above.

Shared/Tests.Shared/ReflectableTypeTests.cs
```diff
@@ -0,0 +1,239 @@
...
+            var owner = AddDogAndOwner();
+            _realm.Write(() =>
+            {
+                _realm.Remove(owner);
+            });
+
+            var typeInfo = owner.GetTypeInfo();
+            var nameSetter = typeInfo.GetDeclaredProperty(nameof(Owner.Name)).SetMethod;
+
+            Assert.Throws<TargetInvocationException>(() =>
```
nirinchev: Instead check for `Throws.InnerException.TypeOf<T>()`

nirinchev:
retest this please

Shared/Realm.Shared/ReflectableType/WovenGetterMethodInfo.cs
```diff
@@ -0,0 +1,81 @@
...
+        public override object[] GetCustomAttributes(Type attributeType, bool inherit) => _mi.GetCustomAttributes(attributeType, inherit);
+
+        public override MethodImplAttributes GetMethodImplementationFlags() => _mi.GetMethodImplementationFlags();
+
+        public override ParameterInfo[] GetParameters() => _mi.GetParameters();
+
+        public override object Invoke(object obj, BindingFlags invokeAttr, Binder binder, object[] parameters, CultureInfo culture)
+        {
+            var ro = obj as RealmObject;
+            if ((ro?.IsValid).GetValueOrDefault(true))
```
kristiandupont: This is clever but wouldn't `ro != null && ro.IsValid` be more readable?

nirinchev:
Ping @kristiandupont

nirinchev:
retest this please

PR #888, opened 2016-10-25
Update the changelog

nirinchev:
Add a few missed items.


nirinchev:
@kristiandupont can you take a look :P

PR #890, opened 2016-10-25
Minor test improvements

### LINKED ISSUE #347: Test improvements
There are a number of test cases missing
 - [x] `DateTimeOffset` and `DateTimeOffset?`
 - [ ] Setting `null` without prior value in nullable columns
 - [x] Reading nullable columns should default to null
 - [x] Assert that non-unique object ids throw exceptions

And other things to do:
 - [x] There are failing tests, some on Windows, some on iOS
 - [x] Fix `TearDown` to delete temp file in remaining test fixtures
 - [x] Rename `Setup` methods to `SetUp`

nirinchev:
- Assert that all column types default to the .NET defaults (except DateTimeOffset)
- Assert that non-unique object ids throw
- Rename Setup to SetUp

Resolves #347

### DIFF
--- a/Platform.Win32/Tests.Win32/RunSpecificTest.cs
+++ b/Platform.Win32/Tests.Win32/RunSpecificTest.cs
@@ -33,7 +33,7 @@ namespace IntegrationTests.Win32
         public void RunTest()
         {
             var testFixture = new NotificationTests();
-            testFixture.Setup();
+            testFixture.SetUp();
 
             testFixture.CollectionChangedAdapter_DeletingItemFromRealm_ShouldRaiseReset();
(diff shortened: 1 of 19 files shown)


nirinchev:
Ping @kristiandupont not sure what

> Setting `null` without prior value in nullable columns

means.

kristiandupont:
This test https://github.com/realm/realm-dotnet/blob/master/Shared/Tests.Shared/AccessTests.cs#L79 sets a property to a value and then replaces it with `null`. So the case where no value was set initially is never tested. I remember thinking we should but now it doesn't seem like a real concern.

PR #891, opened 2016-10-26
Add support for using static fields/properties as rhs in linq searches

### LINKED ISSUE #872: Using `string.Empty` in LINQ expression fails
`realm.All<Person>().Where(p => p.FirstName == string.Empty)` throws `NotSupportedException` -- the rhs of the 'Equal' operator should be a constant or closure variable.

nirinchev:
Improve rhs constant resolution in LINQ search clauses.
- [x] Handle static fields (`string.Empty`)
- [x] Handle instance fields (`this.myField`)
- [x] Handle nested instance fields (`this.myConstantsInstance.fieldValue`)

Applies for properties as well.

Resolves #872


nirinchev:
@kristiandupont I feel that the tests are a bit too much - should we check that it works for `==` only and assume that it will work with the other operands as well?

kristiandupont:
Hm yeah, I agree. We don't want to test every possible permutation, we just want good coverage and test cases for things that we feel are potentially fragile or that have had bugs.

nirinchev:
That was my thinking as well - with the new tests we want to just verify that constants/fields/properties are interpreted correctly, not that searching itself works, so I'll just leave the equality test to make it easier to read.

nirinchev:
@kristiandupont ping

PR #893, opened 2016-10-26
Add support for .Equals/.StartsWith/.EndsWith ignore case queries

### LINKED ISSUE #794: String query ignore case
Is there anyway to do a case insensitive string query?

Things I tried but didn't work:

`realm.All<Item>().Where(x => String.Equals(x.Name, "abc", StringComparison.OrdinalIgnoreCase))`
`realm.All<Item>().Where(x => x.Name.Equals("abc", StringComparison.OrdinalIgnoreCase))`

nirinchev:
Theoretically, we support case-insensitive contains as well, but .NET's `.Contains` doesn't 🙄 Perhaps we could provide an extension method that we can use to proxy for our needs? @kristiandupont 

Resolves #794

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@
     var peter = realm.All<Person>().FirstOrDefault(d => d.Name == "Peter");
     var petersDogs = realm.All<Dog>().Where(d => d.Owner == peter);
     ```
-* Added support for `StarsWith(string, StringComparison)`, `EndsWith(string, StringComparison)`, and `Equals(string, StringComparison)` filtering in LINQ. (#893)
+* Added support for `StartsWith(string, StringComparison)`, `EndsWith(string, StringComparison)`, and `Equals(string, StringComparison)` filtering in LINQ. (#893)
 
     **NOTE**: Currently only `Ordinal` and `OrdinalIgnoreCase` comparisons are supported. Trying to pass in a different one will result in runtime error. If no argument is supplied, `Ordinal` will be used.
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -27,7 +27,9 @@
 * Added non-generic version of `Realm.Manage`. (#871)
 * Added support for nullable integer PrimaryKeys. Now you can have `long?` PrimaryKey property where `null` is a valid unique value. (#877)
 * Added a weaver warning when applying Realm attributes (e.g. `[Indexed]` or `[PrimaryKey]`) on non-persisted properties. (#882)
+* Added support for `StarsWith(string, StringComparison)`, `EndsWith(string, StringComparison)`, and `Equals(string, StringComparison)` filtering in LINQ. (#893)
```
kristiandupont: Typo: StarsWith

Shared/Realm.Shared/Handles/QueryHandle.cs
```diff
@@ -226,38 +226,38 @@ public void BinaryNotEqual(IntPtr columnIndex, IntPtr buffer, IntPtr bufferLengt
             nativeException.ThrowIfNecessary();
         }
 
-        public void StringContains(IntPtr columnIndex, string value)
+        public void StringContains(IntPtr columnIndex, string value, bool caseSensitive = true)
```
kristiandupont: Feels wrong to me to have default parameters in a class like this that is not a public API and where we probably want to be quite explicit..

nirinchev: I wanted to have a default parameter as a hint toward what's the default behavior that we want (e.g. with `==`). Otherwise, every time the user has not explicitly specified case sensitivity, we'll have to figure out what to default to.

PR #895, opened 2016-10-26
Guard and add tests for removing object from different realm

### LINKED ISSUE #316: Managed object delete errors
Test what happens if try to remove a persistent object that is not in active realm - probably need to add and document exceptions

nirinchev:
Resolves #316

### DIFF
--- a/Shared/Realm.Shared/Realm.cs
+++ b/Shared/Realm.Shared/Realm.cs
@@ -831,6 +831,16 @@ namespace Realms
         /// <exception cref="System.ArgumentNullException">If you invoke this with a standalone object.</exception>
         public void Remove(RealmObject obj)
         {
+            if (obj == null)
+            {
+                throw new ArgumentNullException(nameof(obj));
+            }
+
+            if (!this.Equals(obj.Realm))
+            {
+                throw new RealmObjectManagedByAnotherRealmException("This object is managed by another realm. You should call Remove on the other realm.");
+            }
+
             obj.ObjectHandle.RemoveFromRealm();
         }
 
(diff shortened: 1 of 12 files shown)


Shared/Realm.Shared/Realm.cs
```diff
@@ -831,6 +831,16 @@ public RealmObject ObjectForPrimaryKey(string className, string id)
...
         public void Remove(RealmObject obj)
         {
+            if (obj == null)
+            {
+                throw new ArgumentNullException(nameof(obj));
+            }
+
+            if (!this.Equals(obj.Realm))
+            {
+                throw new RealmObjectManagedByAnotherRealmException("This object is managed by another realm. You should call Remove on the other realm.");
```
kristiandupont: It could also be standalone, right? So maybe just "This object is not managed by this realm"

Shared/Tests.Shared/RemoveTests.cs
```diff
@@ -146,5 +146,84 @@ public void RemoveAllObjectsShouldClearTheDatabase()
...
+        private void PerformWithOtherRealm(string path, Action<Realm> action)
+        {
+            Realm otherRealm;
+            using (otherRealm = Realm.GetInstance(path ?? Path.GetTempFileName()))
+            {
+                action(otherRealm);
+            }
+
+            Realm.DeleteRealm(otherRealm.Config);
+        }
```
kristiandupont: Maybe add tests for standalone objects as well

Shared/Realm.Shared/Realm.cs
nirinchev: Do you think we should throw an exception if the object is standalone?

kristiandupont: I would say so, yes. Unless you can think of a scenario where one would want to do that without any hassle. It seems to me that it's a programmer error if you try to remove something from the realm that wasn't in there.

nirinchev:
Ping @kristiandupont how does that look now? Borrowed some error inspiration from [Java](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/RealmObject.java#L94) and [Cocoa](https://github.com/realm/realm-cocoa/blob/9484a03f5355fa00df99b3386c9009335d77a6a1/Realm/RLMObjectStore.mm#L438).

PR #901, opened 2016-10-28
Create a transaction for property setters

### LINKED ISSUE #887: Create "microtransaction" for property setters in data-binding scenarios
With the changes, proposed with #885, we'll now be able to alter the property setters in data-binding scenarios only. Thus, we can enable two-way binding by starting and committing a transaction on each set.

Related to #647

nirinchev:
This will allow data-bindings to work seamlessly without having to worry about transactions and stuff.

Resolves #887


Shared/Realm.Shared/ReflectableType/RealmObjectTypeInfo.cs
```diff
@@ -17,26 +17,42 @@
...
+using System.Collections.Generic;
 using System.Reflection;
 
 namespace Realms
 {
     internal class RealmObjectTypeInfo : TypeDelegator
     {
-        public RealmObjectTypeInfo(Type type) : base(type)
+        private static readonly ConcurrentDictionary<Type, RealmObjectTypeInfo> typeCache = new ConcurrentDictionary<Type, RealmObjectTypeInfo>();
+        private readonly IDictionary<string, PropertyInfo> _cache = new Dictionary<string, PropertyInfo>();
```
nirinchev: Those caches might benefit from better naming and probably a comment

nirinchev:
@fealebenpae @kristiandupont can you take a look and let me know what you think, especially regarding the throttling mechanics - I have mixed feelings about it, even though I was the one to come up with the idea.

Shared/Realm.Shared/ReflectableType/WovenSetterMethodInfo.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            _mi.Invoke(obj, invokeAttr, binder, parameters, culture);
+
+            if (writeTransaction != null)
+            {
+                writeTransaction.Commit();
+                writeTransaction.Dispose();
+            }
+
+            return null;
```
nirinchev: While this is true, I wonder if we should instead return the result of `_mi.Invoke(...)`

nirinchev:
Ping @realm/dotnet -> I seem to like it now, so can you briefly review it.

nirinchev:
retest this please

Issue #886, opened 2016-10-25
PropertyChanged not raised when a change originates from Core


nirinchev:
This is reproducible with the QuickJournal sample - if you edit an item's name and go back (effectively, Rollback-ing the transaction), the updated name will be displayed in the list, instead of the old one. I imagine this will be even more of a problem with RMP, when we have changes pushed from a remote realm.

PR #909, opened 2016-11-08
Handle change notifications originating from Core

nirinchev:
TODO:
- [x] Unregister observed items (event unsubscribe, remove object)
- [x] Unregister observed items on deletion (`invalidated` vector)
- [x] Handle changes for same-thread realms (I'm thinking of adding `NotifyChanges()` at the end of each `set_xxx` method)
- [x] Lookup property index, instead of using the column index
- [x] Decide whether we want to support `[MapTo]`
- [x] Moar unit tests 👺
- [x] Tests for object deletion

Resolves https://github.com/realm/realm-dotnet/issues/886

### DIFF
--- a/Shared/Realm.Shared/RealmObject.cs
+++ b/Shared/Realm.Shared/RealmObject.cs
@@ -53,7 +53,7 @@ namespace Realms
         {
             var gch = GCHandle.FromIntPtr(realmObjectHandle);
             var realmObject = (RealmObject)gch.Target;
-            var propertyName = realmObject.ObjectSchema.ElementAtOrDefault((int)propertyIndex).Name;
+            var propertyName = realmObject.ObjectSchema.ElementAtOrDefault((int)propertyIndex).PropertyInfo.Name;
             realmObject.RaisePropertyChanged(propertyName);
         }
 
@@ -127,16 +127,6 @@ namespace Realms
             {
                 SubscribeForNotifications();
             }
-        }
-
-        internal void RemoveFromRealm(Realm realm)
-        {
-            if (_propertyChanged != null)
-            {
-                UnsubscribeFromNotifications();
-            }
-
-            ObjectHandle.RemoveFromRealm(realm.SharedRealmHandle);
         }
 
         internal class Metadata
--- a/Shared/Realm.Shared/RealmObject.cs
+++ b/Shared/Realm.Shared/RealmObject.cs
@@ -41,15 +41,10 @@ namespace Realms
     {
         #region static
 
-        static RealmObject()
-        {
-            NativeCommon.register_notify_realm_object_changed(NotifyRealmObjectPropertyChanged);
-        }
-
         #if __IOS__
         [MonoPInvokeCallback(typeof(NativeCommon.NotifyRealmCallback))]
         #endif
-        private static void NotifyRealmObjectPropertyChanged(IntPtr realmObjectHandle, IntPtr propertyIndex)
+        public static void NotifyRealmObjectPropertyChanged(IntPtr realmObjectHandle, IntPtr propertyIndex)
         {
             var gch = GCHandle.FromIntPtr(realmObjectHandle);
             var realmObject = (RealmObject)gch.Target;
--- a/Shared/Realm.Shared/RealmObject.cs
+++ b/Shared/Realm.Shared/RealmObject.cs
@@ -62,6 +62,7 @@ namespace Realms
         private Realm _realm;
         private ObjectHandle _objectHandle;
         private Metadata _metadata;
+        private GCHandle? _notificationsHandle;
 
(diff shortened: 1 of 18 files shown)


nirinchev:
retest this please

nirinchev:
retest this please

Shared/Realm.Shared/RealmObject.cs
```diff
@@ -32,11 +37,58 @@ namespace Realms
...
     /// </remarks>
     [Preserve(AllMembers = true, Conditional = false)]
-    public class RealmObject : IReflectableType
+    public class RealmObject : IReflectableType, INotifyPropertyChanged
     {
+        #region static
+
+        static RealmObject()
+        {
+            NativeCommon.register_notify_realm_object_changed(NotifyRealmObjectPropertyChanged);
```
nirinchev: Yeah, I don't have strong feelings about where it should go - as you said, it's slightly more related to object, so I left it here, but I don't mind moving it to `Realm`.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -30,39 +31,131 @@
...
+    {
+        return observed_rows;
+    }
+    
+    void CSharpBindingContext::add_observed_row(const Object& object, void* managed_object_handle)
+    {
+        auto observer_state = BindingContext::ObserverState();
+        observer_state.row_ndx = object.row().get_index();
+        observer_state.table_ndx = object.row().get_table()->get_index_in_group();
+        observer_state.info = new ObservedObjectDetails(object.get_object_schema(), managed_object_handle);
```
fealebenpae: this needs to be `delete`d later.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -30,39 +31,131 @@
...
+        observed_rows.push_back(observer_state);
```
fealebenpae: use `std::move` to prevent a copy.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -30,39 +31,131 @@
...
+                if (get_managed_object_handle(it->info) == managed_object_handle) {
+                    it = observed_rows.erase(it);
+                } else {
+                    ++it;
+                }
+            }
+        }
+    }
+    
+    void* CSharpBindingContext::get_managed_realm_handle()
```
fealebenpae: define in header instead so that the getter can be inlined in other translation units.

wrappers/src/shared_realm_cs.hpp
```diff
@@ -19,11 +19,42 @@
...
+        void did_change(std::vector<CSharpBindingContext::ObserverState> const& observed, std::vector<void*> const& invalidated) override;
+        std::vector<ObserverState> get_observed_rows() override;
+        void add_observed_row(const Object& object, void* managed_object_handle);
+        void remove_observed_row(void* managed_object_handle);
+        void* get_managed_realm_handle();
+        void notify_change(const size_t row_ndx, const size_t table_ndx, const size_t property_index);
+        void notify_removed(const size_t row_ndx, const size_t table_ndx);
+    private:
+        void* m_managed_realm_handle;
+        std::vector<BindingContext::ObserverState> observed_rows;
```
fealebenpae: use the `m_` prefix for fields.

Shared/Realm.Shared/RealmObject.cs
```diff
@@ -41,15 +41,10 @@ public class RealmObject : IReflectableType, INotifyPropertyChanged
...
-        static RealmObject()
-        {
-            NativeCommon.register_notify_realm_object_changed(NotifyRealmObjectPropertyChanged);
-        }
-
         #if __IOS__
         [MonoPInvokeCallback(typeof(NativeCommon.NotifyRealmCallback))]
         #endif
-        private static void NotifyRealmObjectPropertyChanged(IntPtr realmObjectHandle, IntPtr propertyIndex)
+        public static void NotifyRealmObjectPropertyChanged(IntPtr realmObjectHandle, IntPtr propertyIndex)
```
fealebenpae: this should be `internal`

wrappers/src/shared_realm_cs.cpp
```diff
@@ -65,7 +61,7 @@ namespace binding {
 
     CSharpBindingContext::CSharpBindingContext(void* managed_realm_handle) : m_managed_realm_handle(managed_realm_handle)
     {
-        observed_rows = std::vector<ObserverState>();
+        m_observed_rows = std::vector<ObserverState>();
```
fealebenpae: You don't need to initialize this - the compiler will have added the necessary code to invoke the `std::vector` constructor in the `CSharpBindingContext` constructor.

Issue #912, opened 2016-11-08
Validation on [Required] properties


fealebenpae:
Based on #349 

- [x] Validate that only scalar reference type properties are decorated with `[Required]`
- [x] Setters on those properties should throw if the value is null

nirinchev:
Is the first condition worded correctly? I would assume only **nullable** types can be decorated with `[Required]`

fealebenpae:
@nirinchev: good catch, I tried to enhance the wording.

nirinchev:
The second requirement is handled in native.

PR #913, opened 2016-11-08
Add a Required attribute to enable non-nullable properties

### LINKED ISSUE #349: Java-style [Required] attribute needed for strings
Unlike Swift, C# lacks a `String?` type so there's no way to specify that a String property is optional. That implies the corollary - there's no way to specify one is mandatory either. Java has a similar issue so they have a `@required` annotation.

fealebenpae:
Fixes #349, which is required for Sync.

Validation is deferred to #912 so that we don't defer Sync.

### DIFF
--- a/Shared/Realm.Shared/Schema/ObjectSchema.cs
+++ b/Shared/Realm.Shared/Schema/ObjectSchema.cs
@@ -133,6 +133,11 @@ namespace Realms.Schema
                 schemaProperty.ObjectType = innerType?.Name;
                 schemaProperty.IsNullable = isNullable;
 
+                if (property.GetCustomAttribute<RequiredAttribute>() != null)
+                {
+                    schemaProperty.IsNullable = false;
+                }
+
                 builder.Add(schemaProperty);
             }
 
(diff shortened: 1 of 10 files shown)


Shared/Realm.Shared/Schema/ObjectSchema.cs
```diff
@@ -133,6 +133,11 @@ public static ObjectSchema FromType(Type type)
                 schemaProperty.ObjectType = innerType?.Name;
                 schemaProperty.IsNullable = isNullable;
 
+                if (property.GetCustomAttribute<RequiredAttribute>() != null)
```
nirinchev: Fairly minor, but we could short-circuit this by
```csharp
schemaProperty.IsNullable = isNullable && property.GetCustomAttribute<RequiredAttribute>() == null;
```

Shared/Tests.Shared/TestObjects.cs
```diff
@@ -44,6 +44,9 @@ public class AllTypesObject : RealmObject
 
         public DateTimeOffset DateTimeOffsetProperty { get; set; }
 
+        [Required]
+        public string RequiredStringProperty { get; set; }
```
nirinchev: This doesn't seem to be used, is it just to validate that `[Required]` doesn't cause any funny behavior?

fealebenpae: Just to double-check we have a type that is actually passed to a live realm so that the C++ schema code can exercise and validate it. The `ObjectSchemaTests` fixture doesn't actually touch the C++ schema logic.

PR #938, opened 2016-11-22
Refactor results notifycollectionchanged logic

nirinchev:
This paves the way for a unified base class for results and lists

### DIFF
--- a/Shared/Realm.Shared/RealmCollectionBase.cs
+++ b/Shared/Realm.Shared/RealmCollectionBase.cs
@@ -1,0 +1,304 @@
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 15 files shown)


Shared/Realm.Shared/RealmCollectionBase.cs
```diff
@@ -0,0 +1,304 @@
...
+                    if (removedItems != null)
+                    {
+                        RaiseCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Remove, removedItems, removedStartIndex));
+                    }
+
+                    if (addedItems != null)
+                    {
+                        RaiseCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Add, addedItems, addedStartIndex));
+                    }
+                }
```
fealebenpae: We should raise a `Reset` when both `error` and `change` are null.

fealebenpae: The initial notification, which will be fired as soon as the query is run on the worker thread.

Because the previous implementation used a cache that eagerly evaluated the query we didn't care about the initial notification because the intermediate collection had already pulled in everything in the `Results` object at the time of its creation.

In practice, anyone who cares enough to subscribe to `CollectionChanged` is likely to have attempted to enumerate the `Results` object anyway which will eagerly run its query, so we might not care about the initial notification.

Issue #745, opened 2016-08-01
Refine RemoveRange or overload to take IQueryable


AndyDentFree:
A recommended solution to delete a selection of objects is

```
var filteredObjects = realm.All<MyRealmObject>().Where(x => x.FieldOne == false
                        && x.FieldTwo == "someText");
realm.RemoveRange<MyRealmObject>(filteredObjects);
```

However, that doesn't currently compile because `RemoveRange` takes a `RealmResults` and the LINQ expression returns an `IQueryable`.

An overload of `RemoveRange` could take an `IQueryable` parameter and check its actual type matches. It may also be able to avoid using the generic by using IRealmResults interface to get the `ResultsHandle`.

fealebenpae:
We could have a generic constraint like `void RemoveRange<T>(IQueryable<T> items) where T : RealmObject` pretty much like we do for the `ToNotifyCollectionChanged` extension method.

PR #942, opened 2016-11-24
Make RealmResults internal

nirinchev:
It's a cosmetic refactoring that is yet another step towards observable `RealmList`. The main motivation here is that exposing public `RealmResults<T>` forces its base classes to be public, which means that properties that expose internal classes should be internal instead of protected, making for an ugly and polluted base class API.

With this change, we reduce even further the number of custom classes the user has to care about and instead use the basic `IQueryable<T>` + extension method for notification subscribing.

The docs should be updated to reflect the changes.

Among other things, it addresses #745


Platform.PCL/Realm.PCL/RealmPCL.cs
```diff
@@ -399,7 +400,7 @@ public void Remove(RealmObject obj)
         /// </summary>
         /// <typeparam name="T">Type of the objects to remove.</typeparam>
         /// <param name="range">The query to match for.</param>
-        public void RemoveRange<T>(RealmResults<T> range) where T : RealmObject
+        public void RemoveRange<T>(IQueryable<T> range)
```
fealebenpae: Can we add the generic constraint here?

Shared/Realm.Shared/Linq/IRealmResults.cs
```diff
@@ -16,10 +16,94 @@
...
 
+using System;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+
 namespace Realms
 {
     internal interface IRealmResults
     {
         Schema.ObjectSchema ObjectSchema { get; }
```
fealebenpae: I think this should belong to `IRealmCollection<T>`.

Shared/Realm.Shared/Realm.cs
```diff
@@ -808,14 +808,20 @@ public void Remove(RealmObject obj)
         /// </summary>
         /// <typeparam name="T">Type of the objects to remove.</typeparam>
         /// <param name="range">The query to match for.</param>
-        public void RemoveRange<T>(RealmResults<T> range)
+        public void RemoveRange<T>(IQueryable<T> range)
```
fealebenpae: Can we add a generic constraint here?

Platform.XamarinAndroid/Realm.Sync.XamarinAndroid/Realm.Sync.XamarinAndroid.csproj
```diff
@@ -13,7 +13,7 @@
     <AndroidResgenFile>Resources\Resource.Designer.cs</AndroidResgenFile>
     <GenerateSerializationAssemblies>Off</GenerateSerializationAssemblies>
     <AndroidUseLatestPlatformSdk>True</AndroidUseLatestPlatformSdk>
-    <TargetFrameworkVersion>v6.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v7.0</TargetFrameworkVersion>
```
fealebenpae: That'smost probably added by Xamarin Studio automatically.

Platform.XamarinAndroid/Tests.XamarinAndroid/Tests.XamarinAndroid.csproj
```diff
@@ -14,7 +14,7 @@
     <AndroidApplication>True</AndroidApplication>
     <AndroidUseLatestPlatformSdk>true</AndroidUseLatestPlatformSdk>
     <AssemblyName>Tests.XamarinAndroid</AssemblyName>
-    <TargetFrameworkVersion>v5.1</TargetFrameworkVersion>
```
fealebenpae: Ditto.

Shared/Realm.Shared/Linq/IRealmResults.cs
nirinchev: Do we want to expose it publicly? I'm fairly conservative about exposing stuff that people are unlikely to need.

Platform.XamarinAndroid/Realm.Sync.XamarinAndroid/Realm.Sync.XamarinAndroid.csproj
nirinchev: Yes - but since it does it all the time, I figured it would be easier to commit and not have to discard it each time :)

PR #947, opened 2016-11-25
Reuse cached Realm instances

### LINKED ISSUE #933: Investigate potential object store caching issues
It seems OS may cache `SharedRealm` instances, which can cause subtle bugs with the BindingContext not referencing the correct managed instance. One possible solution is to set the binding context in the managed ctor and if we get a `SharedRealm` with non-null binding context, get the managed instance and return that.

nirinchev:
Resolves #933


nirinchev:
Closing for #1063

PR #949, opened 2016-11-25
Handle crashes in Android tests

### LINKED ISSUE #946: Failed Android activity instrumentation should fail CI
```
INSTRUMENTATION_RESULT: shortMsg=Process crashed.
INSTRUMENTATION_CODE: 0
```

fealebenpae:
When the instrumented unit test runner crashes, the exit code of `adb shell` is still zero so the build isn't failed.

This checks the `INSTRUMENTATION_RESULT` output of `am instrument` and fails the build if the instrumentation errors out, while archiving the logcat output in case of error.

Fixes #946

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -249,7 +249,9 @@ def AndroidTest(stashName) {
           }
           archiveLog = false
         } finally {
-          stopLogCatCollector(backgroundPid, archiveLog, stashName)
+          if (backgroundPid != null) {
+            stopLogCatCollector(backgroundPid, archiveLog, stashName)
+          }
         }
       }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -224,11 +224,59 @@ def AndroidTest(stashName) {
     node('android-hub') {
       deleteDir()
       unstash stashName
-      sh 'adb devices'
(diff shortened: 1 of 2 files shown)


Shared/Tests.Shared/TestHelpers.cs
```diff
@@ -86,7 +86,7 @@ public static void RunEventLoop(int milliseconds)
 
         public static void RunEventLoop()
         {
-            RunEventLoop(1);
+            RunEventLoop(100);
```
emanuelez: What's this for?

Jenkinsfile
```diff
@@ -224,13 +224,61 @@ def AndroidTest(stashName) {
...
+          ''', returnStdout: true
+
+          def result = readProperties text: instrumentationOutput.trim().replaceAll(': ', '=')
+          if (result.INSTRUMENTATION_CODE != '-1') {
+            echo instrumentationOutput
+            error result.INSTRUMENTATION_RESULT
+          }
+          archiveLog = false
+        } finally {
+          stopLogCatCollector(backgroundPid, archiveLog, stashName)
```
emanuelez: surround this line with `if (backgroundPid != null)` in case `startLogCatCollector()` fails

Shared/Tests.Shared/TestHelpers.cs
nirinchev: I'm not sure why this is here - it's supposed to be in master already. Anyway, it controls how much time we idle for when waiting for notifications. 1 ms worked well on iOS, but on Android it seems too short and in some of our tests, the notification will not be delivered, thus giving us false negatives.

fealebenpae: I rebased against `master` and got this commit in, somehow. It's unrelated to the PR.

emanuelez:
:shipit:

PR #952, opened 2016-11-28
Add authentication exception handling

### LINKED ISSUE #925: Handle auth errors coming from the Object Server
so that `Realm.Sync.User.LoginAsync` can throw actionable exceptions.

nirinchev:
Resolves #925

### DIFF
--- a/Shared/Realm.Sync.Shared/User.cs
+++ b/Shared/Realm.Sync.Shared/User.cs
@@ -19,10 +19,9 @@
 using System;
 using System.IO;
 using System.Json;
-using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
-using System.Runtime.Serialization;
+using System.Text;
 using System.Threading.Tasks;
 
 namespace Realms.Sync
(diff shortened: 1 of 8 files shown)


nirinchev:
Ping @fealebenpae do you think we should handle exceptions that the HttpClient itself might throw?

fealebenpae:
Not sure. Those could be various depending on the `HttpClientHandler` implementation the app developer chooses. Maybe just let them bubble up?

Shared/Realm.Sync.Shared/User.cs
```diff
@@ -127,25 +128,53 @@ public void LogOut()
...
                         var problem = JsonValue.Load(stream);
-                        throw new Exception("AuthError"); // FIXME: parse the error details and throw a more specific exception
+
+                        ErrorCode code;
+                        if (!Enum.TryParse(problem["code"], out code))
+                        {
+                            code = ErrorCode.Unknown;
+                        }
+
+                        throw new AuthenticationException(code, response.StatusCode, response.ReasonPhrase, problem.ToString(), problem["title"]);
```
fealebenpae: Would it make sense to expose the problem payload on the `AuthenticationException` as a `JsonObject` property so that provider-specific details can be accessed programatically by the developer?

nirinchev: I guess they could deserialize it if they wanted to. And considering it'll be strongly typed, I imagine it won't make too much difference compared to a simple json string that they can either deserialize or process through a prettifier.

PR #968, opened 2016-12-04
Add Realm.Compact

### LINKED ISSUE #967: Expose realm-core compact()
> **Feature Request**: Just fill in the first two sections below.

> What do you want to achieve?

The ability to compact a Realm database via realm-core [compact ](https://github.com/realm/realm-core/blob/9639036845697fb09b6d90a79ad3ce3cf0baecd2/src/realm/group_shared.hpp#L367) like the Java version of `.compactRealmFile`, or RLMRealm.compact, etc... 

#### Actual Results

Currently we are experiencing **huge** Realm files due to writing and refreshing on different threads.
 
We have issues of device/app faults due to DB size on iOS/Android, had a iOS (iPhone 7 256GB) beta app user with a 120**GB** realm file that compressed to under 12**MB**. 

Basic app/realm setup:

```
─Main instance (UI Thread)
├── "Write" instance on a different thread 
│     i.e. a service performing updates to RealmObjects also creating new objects
├── "Read" instance on a different thread 
│    i.e. a read-only instance that is monitoring changes and performing Refresh() to get the latest changes from the "Write" instance to dispatch events to app.

```

Example: A Realm file on iOS after one app session is 1,25,829,120 bytes

Note: This is about avg. after an app session as we are now eventing DB size with app analytics to track it but have seen a few 100+GB files that had less then 20MB of data.

After implementing a dump DB to JSON-based files, delete & recreate realm routine in the app, file size is 327,680 bytes. Same results if you copy the *bloated* db to OS-X and use Realm Browser to load and then "[Export to Compacted Realm](https://github.com/realm/realm-browser-osx/blob/89c1cb27476bd76bd0f0ba7974d3a816617d04f7/RealmBrowser/Controllers/RLMRealmBrowserWindowController.m#L442)".

Realm version(s):  v.0.80.0
Xamarin/Visual Studio version:  VS2015 / XS 6 / VS4M 7
Which operating system version and device: Android and iOS

nirinchev:
Resolves #967 

Waiting for #976 - should be a small fix, so better to have it in.

Docs are updated for Compact.

### DIFF
--- a/Shared/Tests.Shared/InstanceTests.cs
+++ b/Shared/Tests.Shared/InstanceTests.cs
@@ -261,7 +261,7 @@ namespace IntegrationTests
 
             using (var realm = Realm.GetInstance(config))
             {
-                Assert.That(realm.All<IntPrimaryKeyObject>().Count(), Is.EqualTo(populate ? 500 : 0));
+                Assert.That(realm.All<IntPrimaryKeyWithValueObject>().Count(), Is.EqualTo(populate ? 500 : 0));
             }
         }
 
@@ -334,7 +334,7 @@ namespace IntegrationTests
             {
                 realm.Write(() =>
                 {
-                    realm.Add(new IntPrimaryKeyObject
+                    realm.Add(new IntPrimaryKeyWithValueObject
                     {
                         Id = i,
                         StringValue = "Super secret product " + i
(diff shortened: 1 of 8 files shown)


nirinchev:
retest this please

Shared/Tests.Shared/InstanceTests.cs
```diff
@@ -227,5 +228,134 @@ public void RealmObjectClassesOnlyAllowRealmObjects()
...
+            // TODO: enable when we implement instance caching (#947)
+            // This works because of caching of native instances in ObjectStore.
+            // Technically, we get the same native instance, so Compact goes through.
+            // However, this invalidates the opened realm, but we have no way of communicating that.
+            // That is why, things seem fine until we try to run queries on the opened realm.
+            // Once we handle caching in managed, we should reenable the test.
+            using (var realm = Realm.GetInstance())
+            {
+                var initialSize = new FileInfo(realm.Config.DatabasePath).Length;
+                Assert.IsFalse(Realm.Compact());
```
kristiandupont: I prefer `Assert.That(.., Is.False)` but for whatever reason I use `Assert.Throws` so I should get off my high horse...

PR #969, opened 2016-12-04
Improve the RealmConfiguration inheritance chain

nirinchev:
This is to avoid inheriting from instance class, which messes up equality comparisons.
Somewhat of a prerequisite for #947


Shared/Realm.Sync.Shared/Handles/SharedRealmHandleExtensions.cs
```diff
@@ -59,6 +62,15 @@ public static SharedRealmHandle OpenWithSync(Realms.Native.Configuration configu
             return handle;
         }
 
+        public static string GetRealmPath(string userIdentity, Uri serverUri)
```
fealebenpae: It would be faster to pass in a `UserHandle` and get the identity from that instead of copying the string.

wrappers/src/sync_manager_cs.cpp
```diff
@@ -88,10 +88,21 @@ REALM_EXPORT SharedRealm* shared_realm_open_with_sync(Configuration configuratio
         std::string realm_url(Utf16StringAccessor(sync_configuration.url, sync_configuration.url_len));
         
         config.sync_config = std::make_shared<SyncConfig>(SyncConfig{*sync_configuration.user, realm_url, SyncSessionStopPolicy::AfterChangesUploaded, bind_session, handle_session_error});
-        config.path = SyncManager::shared().path_for_realm((*sync_configuration.user)->identity(), realm_url);
+        config.path = Utf16StringAccessor(configuration.path, configuration.path_len);
         return new SharedRealm(Realm::get_shared_realm(config));
     });
 }
+    
+REALM_EXPORT size_t shared_realm_get_path(uint16_t* identity, size_t identity_len, uint16_t* url, size_t url_len, uint16_t* pathbuffer, size_t pathbuffer_len, NativeException::Marshallable& ex)
```
fealebenpae: `realm_syncmanager_get_path_for_realm`

PR #972, opened 2016-12-05
[CI] Move the android tests away from the Pi

fealebenpae:
/cc @emanuelez


emanuelez:
I'm not sure this is a good idea. You're using a docker slave in a non-dockerized environment.
This mean that this job will start the adb-server and then other jobs would start another one in a dockerized environment, which might cause issues. Also, the keys to access the devices are only really available in a reliable way in Docker. Please revert to using a Dockerfile instead.

PR #973, opened 2016-12-05
Expose User.Current and User.AllLoggedIn

### LINKED ISSUE #970: Add User.Current and User.All
Need to add support for `User.Current` and `User.All` similar to what Cocoa has (https://realm.io/docs/objc/2.1.1/api/Classes/RLMSyncUser.html#/c:objc(cs)RLMSyncUser(cm)allUsers). This will allow us to bypass the login screen.

nirinchev:
OS PR: https://github.com/realm/realm-object-store/pull/279

Resolves #970

Needs to update docs after that: https://github.com/realm/realm.io-middleman/pull/2321

Tests theoretically pass, but due to an OS bug, once they're run, it's game over, because it `reset_for_testing` doesn't delete the metadata realm: https://github.com/realm/realm-object-store/issues/207

### DIFF
--- a/Platform.PCL/Realm.Sync.PCL/UserPCL.cs
+++ b/Platform.PCL/Realm.Sync.PCL/UserPCL.cs
@@ -28,6 +28,46 @@ namespace Realms.Sync
     /// </summary>
     public class User : IEquatable<User>
     {
+        #region static
+
+        /// <summary>
+        /// Gets the currently logged-in user. If none exists, null is returned. If more than one user is currently logged in, an exception is thrown.
+        /// </summary>
+        public static User Current
+        {
+            get
+            {
+                RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
+                return null;
+            }
+        }
+
+        /// <summary>
+        /// Gets all currently logged in users.
+        /// </summary>
+        public static User[] AllLoggedIn
+        {
+            get
+            {
+                RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
+                return null;
+            }
+        }
+
+        /// <summary>
+        /// Logs the user in to the Realm Object Server.
+        /// </summary>
+        /// <param name="credentials">The credentials to use for authentication.</param>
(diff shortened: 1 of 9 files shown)


Platform.XamarinIOS/Realm.Sync.XamarinIOS/Realm.Sync.XamarinIOS.csproj
```diff
@@ -69,6 +69,7 @@
     <Reference Include="Xamarin.iOS" />
     <Reference Include="System.Net.Http" />
     <Reference Include="System.Json" />
+    <Reference Include="System.Core" />
```
nirinchev: I'm not sure why, but without that, the LINQ expression here: https://github.com/realm/realm-dotnet/pull/973/files#diff-6842da5be1f20518257e6c13d8e1e763R143 failed to compile.

Platform.PCL/Realm.Sync.PCL/UserPCL.cs
```diff
@@ -28,6 +28,46 @@ namespace Realms.Sync
     /// </summary>
     public class User : IEquatable<User>
     {
+        #region static
+
+        /// <summary>
+        /// Gets the currently logged-in user. If none exists, null is returned. If more than one user is currently logged in, an exception is thrown.
```
AndyDentFree: What does _logged-in_ mean? This seems to make assumptions about only having a single server connection.

nirinchev: The SyncUser has 3 states - Active, LoggedOut, and Error. Currently logged-in users are those whose state is Active.

AndyDentFree: Definition from the Java docs should be included.

 _ The current user is the last user who logged in and credentials are not expired yet_

It disambiguates the issue of what makes a given user _Current_ if you consider multiple servers being used.

nirinchev: But we don't behave like they do - if more than one user is logged, `User.Current` throws an exception.

PR #977, opened 2016-12-07
Use a SynchronizationContext for WeakRealmNotifier

### LINKED ISSUE #564: Base Realm notification on top of SynchronizationContext
Every Realm has a background watcher that notifies it of changes in the database file - the `ExternalCommitHelper` and `WeakRealmNotifier` tandem.

Because `ExternalCommitHelper` receives change notifications from other Realm instances on a worker thread, `WeakRealmNotifier` has to call back to the original thread, the one that owns the Realm instance that needs to be notified of changes, using platform-specific primitives such as the Android Looper and the iOS RunLoop.

Unfortunately, this doesn't work when a Realm is opened on a thread without an active Looper or RunLoop, or when those aren't given a chance to process their message queue.

To remedy this, I propose that the `WeakRealmNotifier` class be taught to use .NET's `SynchronizationContext` primitive. 
Xamarin already implement `SynchronizationContext` on top of the [Android Looper](https://github.com/xamarin/xamarin-android/blob/master/src/Mono.Android/Android.App/SyncContext.cs) and the [iOS RunLoop](https://github.com/xamarin/xamarin-macios/blob/master/src/UIKit/UIKitSynchronizationContext.cs) so we wouldn't be losing existing functionality or behaviour. On the contrary, we'd gain support for asynchronously observing realms inside [NUnit async tests](https://github.com/nunit/nunitlite/blob/master/src/framework/Internal/AsyncSynchronizationContext.cs) and other environments that do not interact with the Looper and RunLoop we currently rely on.

fealebenpae:
Closes #564


Shared/Realm.Shared/native/SynchronizationContextEventLoopSignal.cs
```diff
@@ -0,0 +1,116 @@
...
+                    {
+                        callback(user_data);
+                    }
+                }, null);
+            }
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            internal void Invalidate()
+            {
+                _isReleased = true;
```
nirinchev: Should we nullify the `SynchronizationContext`?

Shared/Realm.Shared/native/SynchronizationContextEventLoopSignal.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        #if __IOS__
+        [ObjCRuntime.MonoPInvokeCallback(typeof(release_eventloop))]
+        #endif
+        private static void ReleaseSynchronizationContext(IntPtr eventloop)
+        {
+            if (eventloop != IntPtr.Zero)
+            {
+                var gCHandle = GCHandle.FromIntPtr(eventloop);
```
nirinchev: Weird casing - probably `gcHandle` would be better?

Shared/Tests.Shared/AsyncTests.cs
```diff
@@ -27,14 +28,11 @@ namespace IntegrationTests
     [TestFixture, Preserve(AllMembers = true)]
     public class AsyncTests
     {
+        // We capture the current SynchronizationContext when opening a Realm.
+        // However, NUnit replaces the SynchronizationContext after the SetUp method and before the async test method.
+        // That's why we make sure we open the Realm in the test method by accessing it lazily.
         private Realm _realm;
```
nirinchev: Why is this not a `Lazy<Realm>` like in the other tests?

Shared/Tests.Shared/PropertyChangedTests.cs
```diff
@@ -31,24 +31,34 @@ namespace IntegrationTests
...
         public void TearDown()
         {
-            _realm.Dispose();
-            Realm.DeleteRealm(_realm.Config);
+            if (_lazyRealm.IsValueCreated)
+            {
+                _realm.Dispose();
+            }
+
+            Realm.DeleteRealm(RealmConfiguration.DefaultConfiguration.ConfigWithPath(_databasePath));
```
nirinchev: Do we need to delete it if it's value hasn't been created?

Shared/Realm.Shared/native/SynchronizationContextEventLoopSignal.cs
fealebenpae: Would that improve things in any way?

Shared/Realm.Shared/native/SynchronizationContextEventLoopSignal.cs
fealebenpae: Gotcha, will fix.

Shared/Realm.Shared/native/SynchronizationContextEventLoopSignal.cs
nirinchev: Not sure what the lifecycle of SynchronizationContext is and if there'll be a case when a reference to the EventLoop is held that would prevent collection of the context.

fealebenpae: I see. I think this makes sense in the case where a `SynchronizationContext` is stopped and uninstalled with actions posted by the `EventLoop` helper still pending. Will zero out the field just in case.

Shared/Realm.Shared/native/SynchronizationContextEventLoopSignal.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            internal EventLoop(SynchronizationContext context)
+            {
+                Debug.Assert(context != null, "The SynchronizationContext EventLoopSignal implementation always needs a SynchronizationContext");
+                _context = context; 
+            }
+
+            internal void Post(EventLoopPostHandler callback, IntPtr user_data)
+            {
+                _context.Post(_ =>
```
nirinchev: Now that the `_context` can be null, perhaps use null-propagation just to be extra safe? :D

fealebenpae: `Invalidate` is called in the `EventLoopSignal` destructor so there would be no way for further work to be scheduled on the same `EventLoop` instance.

nirinchev: Technically, I could easily write code that calls `invalidate`, then `post`, but I get your point and I agree it's fine to leave like that.

PR #989, opened 2016-12-08
API to control user persistence

### LINKED ISSUE #975: Pass in encryption key on Android in initialize_sync
We want to encrypt the realm that stores user data, but Android doesn't have native keystore API to use for securely storing the encryption key. Retrieve it from .NET and pass it to native.

fealebenpae:
which allows a custom encryption key for the sync metadata realm to be provided on Android since we cannot use the AndroidKeyStore because of its version requirements

Closes #975

### DIFF
--- a/Shared/Realm.Sync.Shared/User.cs
+++ b/Shared/Realm.Sync.Shared/User.cs
@@ -44,6 +44,8 @@ namespace Realms.Sync
         {
             get
             {
+                SharedRealmHandleExtensions.DoInitialFileSystemConfiguration();
+
                 var handle = SyncUserHandle.GetCurrentUser();
                 if (handle == null)
                 {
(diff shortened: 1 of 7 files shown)


Shared/Realm.Sync.Shared/UserPersistenceMode.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// </summary>
+        NotEncrypted = 0,
+
+        /// <summary>
+        /// Persist <see cref="User"/> objects in an encrypted store.
+        /// </summary>
+        Encrypted,
+
+        /// <summary>
+        /// Do not encrypt <see cref="User"/> objects.
```
nirinchev: This seems to be wrong doc - shouldn't it state "do not persist users"?

Shared/Realm.Sync.Shared/User.cs
```diff
@@ -88,11 +93,29 @@ public static User[] AllLoggedIn
             return new User(SyncUserHandle.GetSyncUser(refresh_token["token_data"]["identity"], refresh_token["token"], serverUrl.AbsoluteUri, false));
         }
 
-        static User()
+        /// <summary>
+        /// Configures user persistence.
```
nirinchev: We should mention that this should be called before `User.Current`, `User.All`, or `User.LoginAsync`.

PR #990, opened 2016-12-08
IRealmObjectHelper.CopyToRealm should always set primary keys

### LINKED ISSUE #988: Properly initialize PK values
Modify the weaver behavior to always set PK values and move that instruction to the top of CopyToRealm. As per Simon's explanation:
> primary key fields must always be initialized directly after `InsertEmptyRows` using the `SetUnique` instruction
relying on the default value inserted by `InsertEmptyRows` (which is `0`, `null` or `""` respectively) will not trigger primary key behavior

fealebenpae:
Fixes #988

### DIFF
--- a/Weaver/RealmWeaver.Fody/extensions/PropertyDefinitionExtensions.cs
+++ b/Weaver/RealmWeaver.Fody/extensions/PropertyDefinitionExtensions.cs
@@ -77,7 +77,7 @@ internal static class PropertyDefinitionExtensions
 
     internal static bool IsPrimaryKey(this PropertyDefinition property)
     {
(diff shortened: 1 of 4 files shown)


Weaver/RealmWeaver.Fody/extensions/PropertyDefinitionExtensions.cs
```diff
@@ -74,6 +75,12 @@ internal static FieldReference GetBackingField(this PropertyDefinition property)
             .SingleOrDefault();
     }
 
+    internal static bool IsPrimaryKey(this PropertyDefinition property)
+    {
+        Debug.Assert(property.DeclaringType.BaseType.Name == "RealmObject", "Primary key properties on make sense on RealmObject classes");
```
nirinchev: This comment is somewhat hard to understand

PR #1001, opened 2016-12-10
Ensure required properties are being set

### LINKED ISSUE #1000: CopyToRealm should always set [Required] properties
CopyToRealm currently skips default properties, which won't trigger required validations. Change it similarly to the PK change (#990).

nirinchev:
Closes #1000

### DIFF
--- a/Weaver/WeaverTests/RealmWeaver.Tests/WeaverTests.cs
+++ b/Weaver/WeaverTests/RealmWeaver.Tests/WeaverTests.cs
@@ -606,17 +606,17 @@ namespace RealmWeaver
             List<string> targetList;
             if (required)
             {
-                targetList = ((Type)instance.GetType()).GetProperties(BindingFlags.Public | BindingFlags.Instance)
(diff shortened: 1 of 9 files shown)


Weaver/WeaverTests/RealmWeaver.Tests/WeaverTests.cs
```diff
@@ -593,6 +593,39 @@ public void WovenCopyToRealm_ShouldAlwaysSetPrimaryKeyProperties(string type)
...
+            var objectType = _assembly.GetType($"AssemblyToProcess.{type}");
+            var instance = (dynamic)Activator.CreateInstance(objectType);
+            instance.IsManaged = true;
+
+            CopyToRealm(objectType, instance);
+
+            List<string> targetList;
+            if (required)
+            {
+                targetList = ((Type)instance.GetType()).GetProperties(BindingFlags.Public | BindingFlags.Instance)
```
fealebenpae: Use `objectType` instead of `(Type)instance.GetType()` here.

PR #1003, opened 2016-12-11
Expose IList<T>.Move extension method

### LINKED ISSUE #995: Implement Move operation in list
Our underlying list implementation supports a move operation which we should expose, maybe as an extension method on `IList`.

Collection notifications should also express move operations.

nirinchev:
Resolves #995

### DIFF
--- a/Shared/Realm.Shared/Linq/IRealmCollection.cs
+++ b/Shared/Realm.Shared/Linq/IRealmCollection.cs
@@ -29,26 +29,57 @@ namespace Realms
     public class ChangeSet
     {
         /// <summary>
-        /// The indices in the new version of the <see cref="IRealmCollection{T}" /> which were newly inserted.
+        /// Gets the indices in the new version of the <see cref="IRealmCollection{T}" /> which were newly inserted.
         /// </summary>
-        public readonly int[] InsertedIndices;
+        public int[] InsertedIndices { get; }
 
         /// <summary>
-        /// The indices in the new version of the <see cref="IRealmCollection{T}"/> which were modified. This means that the property of an object at that index was modified
+        /// Gets the indices in the new version of the <see cref="IRealmCollection{T}"/> which were modified. This means that the property of an object at that index was modified
         /// or the property of another object it's related to.
         /// </summary>
-        public readonly int[] ModifiedIndices;
+        public int[] ModifiedIndices { get; }
 
         /// <summary>
-        /// The indices of objects in the previous version of the <see cref="IRealmCollection{T}"/> which have been removed from this one.
+        /// Gets the indices of objects in the previous version of the <see cref="IRealmCollection{T}"/> which have been removed from this one.
         /// </summary>
-        public readonly int[] DeletedIndices;
+        public int[] DeletedIndices { get; }
 
-        internal ChangeSet(int[] insertedIndices, int[] modifiedIndices, int[] deletedIndices)
+        /// <summary>
+        /// Gets the rows in the collection which moved..
+        /// </summary>
+        /// <remarks>
+        /// Every <c>From</c> index will be present in <c>DeletedIndices</c> and every <c>To</c> index will be present in <c>InsertedIndices</c>.
+        /// </remarks>
+        public Move[] Moves { get; }
+
+        internal ChangeSet(int[] insertedIndices, int[] modifiedIndices, int[] deletedIndices, Move[] moves)
         {
             InsertedIndices = insertedIndices;
             ModifiedIndices = modifiedIndices;
             DeletedIndices = deletedIndices;
+            Moves = moves;
+        }
+
+        /// <summary>
+        /// A <see cref="Move" /> contains information about objects that moved within the same <see cref="IRealmCollection{T}"/>.
+        /// </summary>
+        public class Move
+        {
+            /// <summary>
+            /// Gets the index in the old version of the <see cref="IRealmCollection{T}" /> from which the object has moved.
+            /// </summary>
+            public int From { get; }
+
+            /// <summary>
+            /// Gets the index in the new version of the <see cref="IRealmCollection{T}" /> to which the object has moved.
+            /// </summary>
+            public int To { get; }
+
+            internal Move(int from, int to)
+            {
+                From = from;
+                To = to;
+            }
         }
     }
 
--- a/Shared/Realm.Shared/Linq/IRealmCollection.cs
+++ b/Shared/Realm.Shared/Linq/IRealmCollection.cs
@@ -45,7 +45,7 @@ namespace Realms
         public int[] DeletedIndices { get; }
 
         /// <summary>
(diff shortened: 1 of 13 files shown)


Shared/Tests.Shared/NotificationTests.cs
```diff
@@ -389,6 +389,100 @@ public void List_WhenTransactionHasBothAddAndRemove_ShouldReset()
             Assert.That(propertyEventArgs, Is.EquivalentTo(new[] { "Count", "Item[]" }));
         }
 
+        [TestCase(0, 3, 1, 3, NotifyCollectionChangedAction.Reset)] // a b c d e -> b d a c e
+        [TestCase(0, 3, 0, 2, NotifyCollectionChangedAction.Reset)] // a b c d e -> c d b a e
+        [TestCase(0, 3, 4, 0, NotifyCollectionChangedAction.Reset)] // a b c d e -> e b c d a
+        [TestCase(0, 2, 0, 2, NotifyCollectionChangedAction.Move)] // a b c d e -> c a b d e
+        [TestCase(4, 2, 4, 2, NotifyCollectionChangedAction.Move)] // a b c d e -> a b d e c
+        [TestCase(1, 3, 1, 3, NotifyCollectionChangedAction.Move)] // a b c d e -> c d a b e
+        public void List_WhenTransactionHasMultipleMoves(int oldIndex1, int newIndex1, int oldIndex2, int newIndex2, NotifyCollectionChangedAction expectedAction)
```
AndyDentFree: This is a confusing test to read. I had to stop and think through the test steps a bit too much for me to be comfortable with it as a test clearly describing state and expected result.

Shared/Tests.Shared/NotificationTests.cs
```diff
@@ -389,6 +389,100 @@ public void List_WhenTransactionHasBothAddAndRemove_ShouldReset()
...
+            }
+        }
+
+        [TestCase(0, 4)]
+        [TestCase(4, 0)]
+        [TestCase(0, 2)]
+        [TestCase(2, 0)]
+        [TestCase(1, 2)]
+        [TestCase(2, 1)]
+        public void List_WhenTransactionHasSingleMove(int oldIndex, int newIndex)
```
AndyDentFree: What does this test name mean?

Shared/Tests.Shared/NotificationTests.cs
```diff
@@ -389,6 +389,100 @@ public void List_WhenTransactionHasBothAddAndRemove_ShouldReset()
...
+                items.Move(movedObject, newIndex);
+            }, NotifyCollectionChangedAction.Move);
+
+            Assert.That(args.OldStartingIndex, Is.EqualTo(oldIndex));
+            Assert.That(args.NewStartingIndex, Is.EqualTo(newIndex));
+            Assert.That(args.OldItems, Is.EquivalentTo(new[] { movedObject }));
+            Assert.That(args.NewItems, Is.EquivalentTo(new[] { movedObject }));
+        }
+
+        private NotifyCollectionChangedEventArgs TestMoves(Action<IList<OrderedObject>> moveAction, NotifyCollectionChangedAction expectedAction)
```
AndyDentFree: Please add a simple comment saying what you expect to be in the Realm as a result of this test helper.

Shared/Realm.Shared/RealmList.cs
```diff
@@ -171,6 +171,16 @@ private void AddObjectToRealmIfNeeded(T obj)
...
         #endregion
 
+        public void Move(T item, int targetIndex)
+        {
+            if (targetIndex < 0)
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            _listHandle.Move(item.ObjectHandle, (IntPtr)targetIndex);
```
nirinchev: An exception is thrown in native.

nirinchev:
Yes we allow, and as far as I can tell, it is consistent and correct behavior. That's how lists behave and that's how the native move behaves, so don't see a problem there.

Shared/Realm.Shared/RealmCollectionBase.cs
```diff
@@ -176,6 +176,44 @@ private void OnChange(IRealmCollection<T> sender, ChangeSet change, Exception er
             }
             else if (change != null)
             {
+                if (change.Moves.Length > 0 &&
```
nirinchev: I'm not a big fan of assertions - they are useless for our users and we have unit tests to assert correct behavior.

Shared/Realm.Shared/Linq/IRealmCollection.cs
```diff
@@ -29,26 +29,57 @@ namespace Realms
...
             InsertedIndices = insertedIndices;
             ModifiedIndices = modifiedIndices;
             DeletedIndices = deletedIndices;
+            Moves = moves;
+        }
+
+        /// <summary>
+        /// A <see cref="Move" /> contains information about objects that moved within the same <see cref="IRealmCollection{T}"/>.
+        /// </summary>
+        public class Move
```
fealebenpae: Actually it would be better if `Move` were a struct here - `Changeset` contains arrays which are a reference type so making it a struct would have gained us nothing, whereas making `Move` a struct would be beneficial on platforms where two 32bit integers can be stored in a single integers.

Shared/Tests.Shared/CollectionTests.cs
```diff
@@ -0,0 +1,173 @@
...
+
+namespace IntegrationTests
+{
+    [TestFixture, Preserve(AllMembers = true)]
+    public class CollectionTests
+    {
+        private Realm _realm;
+
+        [SetUp]
+        public void SetUp()
```
fealebenpae: This needs a `TearDown` that closes the realm, doesn't it?

nirinchev:
I felt this was a more natural case, especially with data-bindings - you get the object from the binding and move it. In that case, you would have to call into native to ask for the object index, which seemed wasteful. So we have 3 options:
- Keep it as it is - `Move<T>(IList<T>, T, int).
- Change it to accept two indices - `Move<T>(IList<T>, int, int)`
- Expose both methods.

I think all cases work, so would be happy to hear your opinion on the matter.

Shared/Tests.Shared/CollectionTests.cs
nirinchev: It does indeed.

fealebenpae:
Having an overload that takes `T` sounds like a good idea for the cases you describe, but I'd also push for us to have the `int, int` overload and special-case it for `ObservableCollection<T>`. I think that soon we might run into cases where we want the default value of `IList<T>` properties to be `ObservableCollection<T>` instead of `List<T>` when data-binding against unmanaged objects.

fealebenpae:
In any case, we can look at `ObservableCollection<T>` in a separate PR when we need to handle it.

nirinchev:
I tested it on Tasks and seems like move is working fine, which is nice.

PR #1007, opened 2016-12-12
Implement support for backlinks/LinkingObjects

### LINKED ISSUE #219: Implement Backlinks - reverse relationships
Also needs adding to the documentation when implemented.

fealebenpae:
Closes #219

### DIFF
--- a/Shared/Tests.Shared/RelationshipTests.cs
+++ b/Shared/Tests.Shared/RelationshipTests.cs
@@ -479,6 +479,23 @@ namespace IntegrationTests
             Assert.That(person.Dogs.Count, Is.EqualTo(2));
         }
 
+        [Test]
+        public void Backlinks()
+        {
+            var tim = realm.All<Owner>().Single(o => o.Name == "Tim");
+            foreach (var dog in tim.Dogs)
+            {
+                Assert.That(dog.Owners, Is.EquivalentTo(new[] { tim })); 
+            }
+
+            var dani = realm.All<Owner>().Single(o => o.Name == "Dani");
+            var maggie = realm.All<Dog>().Single(d => d.Name == "Maggie Mongrel");
+            Assert.That(maggie.Owners, Is.Empty);
+
+            realm.Write(() => dani.Dogs.Add(maggie));
+            Assert.That(maggie.Owners, Is.EquivalentTo(new[] { dani }));
+        }
+
         #region DeleteRelated
 
         // from http://stackoverflow.com/questions/37819634/best-method-to-remove-managed-child-lists-one-to-many-parent-child-relationsh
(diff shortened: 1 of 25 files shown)


Weaver/RealmWeaver.Fody/ModuleWeaver.cs
```diff
@@ -548,6 +587,68 @@ private void ReplaceListGetter(PropertyDefinition prop, FieldReference backingFi
         Debug.Write("[get list] ");
     }
 
+        // WARNING
```
nirinchev: indendation issue

Weaver/RealmWeaver.Fody/ModuleWeaver.cs
```diff
@@ -548,6 +587,68 @@ private void ReplaceListGetter(PropertyDefinition prop, FieldReference backingFi
...
+            return <backingField>; // supplied by the generated getter OR RealmObject._CopyDataFromBackingFields
+        */
+
+        var start = prop.GetMethod.Body.Instructions.First();  // this is a label for return <backingField>;
+        var il = prop.GetMethod.Body.GetILProcessor();
+
+        il.InsertBefore(start, il.Create(OpCodes.Ldarg_0));  // this for field ref [ -> this]
+        il.InsertBefore(start, il.Create(OpCodes.Ldfld, backingField)); // [ this -> field]
+        il.InsertBefore(start, il.Create(OpCodes.Ldnull)); // [field -> field, null]
+        il.InsertBefore(start, il.Create(OpCodes.Ceq));  // [field, null -> bool result]
```
nirinchev: Isn't null falsy for IL? I thought we could go for `Brfalse` without explicitly checking for null equality.

Shared/Tests.Shared/RelationshipTests.cs
```diff
@@ -479,6 +479,23 @@ public void Insert_WhenIndexIsEqualToCount_ShouldWork()
             Assert.That(person.Dogs.Count, Is.EqualTo(2));
         }
 
+        [Test]
+        public void Backlinks()
```
nirinchev: It seems we're testing a backlink to a List, should we add a test for a `RealmObject`?

fealebenpae: I'm not sure it matters much at runtime. We only do relationship validation in the weaver so there are tests for both kinds of relationships there. In runtime we're basically calling through to core and our code doesn't care whether the other end of a relationship is `Object` or `Array` in the schema.

Weaver/RealmWeaver.Fody/ModuleWeaver.cs
fealebenpae: I'll try it out, thanks.

PR #1031, opened 2016-12-15
Just bumped version number to make it clear what am using when use local

AndyDentFree:
Per discussion last night, bumped to lower of values discussed


AndyDentFree:
retest this please

PR #1039, opened 2016-12-16
Make sure set_unique is always called

### LINKED ISSUE #1002: Objects are not properly merged by PK
It seems #988 wasn't enough to ensure that proper merging occurs after a sync. We still get duplicate PK objects, so further investigation is required.

nirinchev:
Resolves #1002

### DIFF
--- a/Shared/Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
+++ b/Shared/Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
@@ -25,7 +25,7 @@ namespace Realms.Dynamic
     {
         internal static readonly DynamicRealmObjectHelper Instance = new DynamicRealmObjectHelper();
 
-        public void CopyToRealm(RealmObject instance, bool update)
+        public void CopyToRealm(RealmObject instance, bool update, bool setPrimaryKey)
         {
             throw new NotSupportedException("DynamicRealmObjectHelper cannot exist in unmanaged state, so CopyToRealm should not be called ever.");
         }
--- a/Shared/Tests.Shared/AddOrUpdateTests.cs
+++ b/Shared/Tests.Shared/AddOrUpdateTests.cs
@@ -816,6 +816,17 @@ namespace IntegrationTests
         [Test]
         public void AddOrUpdate_WhenObjectHasNonEmptyList_ShouldNotThrow()
         {
+            // This test verifies that updating an object that has PK and non-empty list will not throw an exception.
+            // The reason it *could* throw is a limitation on core in table.cpp: Table::check_lists_are_empty.
+            // The comment states:
+            //     FIXME: Due to a limitation in Sync, it is not legal to change the primary
+            //     key of a row that contains lists (including linklists) after those lists
+            //     have been populated. This limitation may be lifted in the future, but for
+            //     now it is necessary to ensure that all lists are empty before setting a
+            //     primary key (by way of set_int_unique() or set_string_unique() or set_null_unique()).
+            //
+            // So if we set the Primary Key unnecessarily in the .NET binding, we could trigger that case.
+
             var first = new PrimaryKeyWithPKList
             {
                 Id = 42,
--- a/Shared/Tests.Shared/AddOrUpdateTests.cs
+++ b/Shared/Tests.Shared/AddOrUpdateTests.cs
@@ -813,6 +813,39 @@ namespace IntegrationTests
             }, Throws.TypeOf<RealmException>());
         }
 
+        [Test]
+        public void AddOrUpdate_WhenObjectHasNonEmptyList_ShouldNotThrow()
+        {
+            var first = new PrimaryKeyWithPKList
+            {
+                Id = 42,
+                StringValue = "value1"
+            };
+
+            first.ListValue.Add(new PrimaryKeyObject
+            {
+                Id = 1
+            });
+
+            _realm.Write(() => _realm.Add(first));
+
+            var second = new PrimaryKeyWithPKList
+            {
+                Id = 42,
+                StringValue = "value2"
+            };
+
+            second.ListValue.Add(new PrimaryKeyObject
+            {
+                Id = 1
+            });
+
+            Assert.That(() =>
+            {
+                _realm.Write(() => _realm.Add(second, update: true));
+            }, Throws.Nothing);
+        }
+
         private class Parent : RealmObject
         {
             [PrimaryKey]
--- a/Shared/Tests.Sync.Shared/MergeByPKTests.cs
+++ b/Shared/Tests.Sync.Shared/MergeByPKTests.cs
@@ -42,7 +42,7 @@ namespace Tests.Sync
 
         [NUnit.Framework.Explicit]
         [TestCaseSource(nameof(MergeTestCases))]
-        public async void WhenObjectHasPK_ShouldMergeCorrectly(Type objectType, object pkValue, Func<dynamic, bool> pkValueChecker)
+        public async void WhenObjectHasPK_ShouldNotCreateDuplicates(Type objectType, object pkValue, Func<dynamic, bool> pkValueChecker)
         {
             User.ConfigurePersistence(UserPersistenceMode.NotEncrypted);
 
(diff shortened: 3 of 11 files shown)


Shared/Tests.Sync.Shared/MergeByPKTests.cs
```diff
@@ -0,0 +1,122 @@
...
+using NUnit.Framework;
+using Realms;
+using Realms.Sync;
+
+namespace Tests.Sync
+{
+    [TestFixture, Preserve(AllMembers = true)]
+    public class MergeByPKTests
+    {
+        [NUnit.Framework.Explicit]
```
fealebenpae: There is already a using statement for `NUnit.Framework`.

Shared/Tests.Sync.Shared/MergeByPKTests.cs
```diff
@@ -0,0 +1,122 @@
...
+        [TestCaseSource(nameof(MergeTestCases))]
+        public async void WhenObjectHasPK_ShouldMergeCorrectly(Type objectType, object pkValue, Func<dynamic, bool> pkValueChecker)
+        {
+            User.ConfigurePersistence(UserPersistenceMode.NotEncrypted);
```
fealebenpae: Why?

Shared/Tests.Sync.Shared/MergeByPKTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            var pkProperty = objectType.GetProperties(BindingFlags.Instance | BindingFlags.Public)
+                                       .Single(p => p.GetCustomAttribute<PrimaryKeyAttribute>() != null);
+
+            for (var i = 0; i < 5; i++)
+            {
```
fealebenpae: Consider extracting the body of the loop in a separate method to aid readability.

Shared/Tests.Sync.Shared/MergeByPKTests.cs
```diff
@@ -0,0 +1,122 @@
...
+                    }
+                    catch (RealmDuplicatePrimaryKeyValueException)
+                    {
+                        // Sync went through too quickly (that's why we do 5 attempts
+                    }
+
+                    await Task.Delay(500); // A little bit of time for sync
+                }
+
+                User.Current.LogOut();
```
fealebenpae: Why?

Shared/Tests.Sync.Shared/MergeByPKTests.cs
nirinchev: Because it doesn't work on simulator

Shared/Tests.Sync.Shared/MergeByPKTests.cs
nirinchev: it conflicts with the `Realms.ExplicitAttribute` that you know who created ;)

Shared/Tests.Sync.Shared/MergeByPKTests.cs
nirinchev: I wanted to simulate a scenario where a user logs in for the first time and opens a realm. I guess that's too much - I'll simplify it.

nirinchev:
Ping @fealebenpae

Shared/Tests.Sync.Shared/MergeByPKTests.cs
```diff
@@ -30,6 +30,16 @@ namespace Tests.Sync
     [TestFixture, Preserve(AllMembers = true)]
     public class MergeByPKTests
     {
+        [SetUp]
+        public void SetUp()
+        {
+            // We simulate a clean slate first time user login
```
fealebenpae: Maybe this means we should move this code to `ResetForTesting`? It's what realm-js does.

Shared/Tests.Shared/AddOrUpdateTests.cs
```diff
@@ -813,6 +813,39 @@ public void Add_WhenRequiredPropertyIsNotSet_ShouldThrow()
...
+
+            _realm.Write(() => _realm.Add(first));
+
+            var second = new PrimaryKeyWithPKList
+            {
+                Id = 42,
+                StringValue = "value2"
+            };
+
+            second.ListValue.Add(new PrimaryKeyObject
```
AndyDentFree: What is the expected fate of this object, what effect does it have on what is stored in the Realm. I find this test a bit confusing.

Shared/Tests.Sync.Shared/MergeByPKTests.cs
```diff
@@ -0,0 +1,125 @@
...
+            foreach (var u in User.AllLoggedIn)
+            {
+                u.LogOut();
+            }
+        }
+
+        [NUnit.Framework.Explicit]
+        [TestCaseSource(nameof(MergeTestCases))]
+        public async void WhenObjectHasPK_ShouldMergeCorrectly(Type objectType, object pkValue, Func<dynamic, bool> pkValueChecker)
```
AndyDentFree: _Correctly_ is a loaded term - **How** should they merge?

Shared/Tests.Sync.Shared/MergeByPKTests.cs
```diff
@@ -0,0 +1,125 @@
...
+
+                using (var realm = await GetSyncedRealm(objectType))
+                {
+                    try
+                    {
+                        realm.Write(() => realm.Add(instance));
+                    }
+                    catch (RealmDuplicatePrimaryKeyValueException)
+                    {
+                        // Sync went through too quickly (that's why we do 5 attempts
```
AndyDentFree: This comments scares me - is it implying users should handle these exceptions depending on sync performance?

Shared/Tests.Shared/AddOrUpdateTests.cs
nirinchev: It verifies that we're not throwing when updating with an object that has `ListValue == null` (because the getter is never invoked)

AndyDentFree: Please copy your explanation into a comment in the test source.

Shared/Tests.Sync.Shared/MergeByPKTests.cs
nirinchev: If they're adding items with non-unique PK's, yes. Between opening the realm in `GetInstance` and write locking it in `realm.Write`, the sync write transaction may or may not go through, essentially creating the object, which causes `RealmDuplicatePrimaryKeyValueException`. If you want to avoid it, you can use `update: true`, but we're explicitly testing the add case here.

Shared/Tests.Sync.Shared/MergeByPKTests.cs
nirinchev: The problem with `ResetForTesting` is that it forces the bindings to terminate all active sessions and I'm not sure how to do that. Will logging out the user do the trick?

Shared/Tests.Sync.Shared/MergeByPKTests.cs
nirinchev: I feel we should do this for all our tests (at least temporarily) to avoid having 1 failing test each time. I was planning to do it in a separate PR though.

Shared/Realm.Shared/Dynamic/DynamicRealmObjectHelper.cs
```diff
@@ -25,7 +25,7 @@ internal class DynamicRealmObjectHelper : IRealmObjectHelper
     {
         internal static readonly DynamicRealmObjectHelper Instance = new DynamicRealmObjectHelper();
 
-        public void CopyToRealm(RealmObject instance, bool update)
+        public void CopyToRealm(RealmObject instance, bool update, bool setPrimaryKey)
```
nirinchev: Not sure what you mean. DynamicRealmObjectHelper does not exist in PCL.

AndyDentFree: sorry, misread the `NotSupported...` as being because in a PCL.

Shared/Tests.Shared/AddOrUpdateTests.cs
nirinchev: That's the wrong explanation - was thinking about another case. I'll add a comment with the right one.

nirinchev:
Ping @fealebenpae @AndyDentFree for another round of reviews

Shared/Tests.Sync.Shared/MergeByPKTests.cs
fealebenpae: I'll submit a PR with the enhancement so we can use `ResetForTesting`.

PR #1040, opened 2016-12-17
Fix a NRE when adding an object with a non-accessed list

### LINKED ISSUE #1037: Realm.Add(update: true) throws when object contains a list that hasn't been accessed
The following test case doesn't pass:
```csharp
[Test]
public void AddOrUpdate_WhenListIsNull_ShouldNotThrow()
{
    var first = new PrimaryKeyWithPKList
    {
        Id = 42
    };

    first.ListValue.Add(new PrimaryKeyObject { Id = 1 });

    _realm.Write(() => _realm.Add(first));

    var second = new PrimaryKeyWithPKList
    {
        Id = 42
    };

    Assert.That(() =>
    {
        _realm.Write(() => _realm.Add(second, update: true));
    }, Throws.Nothing);
}
```

The reason is that the generated code for lists looks like:

```csharp
if (update || listField != null)
{
    /*...*/
    for (var i = 0; i < listField.Count; i++)
    {
        /*...*/
    }
}
```

which throws a NRE when `update` is `true` but the list property hasn't been accessed, thus `listField` is `null`. We should clear the list and not access properties on a null field.

### LINKED ISSUE #1038: Realm.Add(update: true) merges lists instead of overwriting them
Currently `Realm.Add(.., update: true)` will only add elements to a list, but will never remove the old ones. In that respect, it's not truly overwriting the previous state of the object and rather merging the new with the old one, which doesn't seem correct.

nirinchev:
- clear to-many relationship lists
- only force setting of properties when not-setting PK (e.g. 100% updating)

Motivation for the weaver change (from `update || non-default` to `!setPrimaryKey || non-default`):
`setPrimaryKey` is always true when `update` is false - we always create a new object in that case. However, if we did create a new object, even though `update` was true (i.e. existing object with that PK was not found), `setPrimaryKey` will also be true, allowing us to optimize that case as well.

Closes #1037 
Closes #1038

### DIFF
--- a/Shared/Tests.Shared/AddOrUpdateTests.cs
+++ b/Shared/Tests.Shared/AddOrUpdateTests.cs
@@ -841,6 +841,86 @@ namespace IntegrationTests
             }, Throws.Nothing);
         }
 
+        [Test]
+        public void AddOrUpdate_WhenListIsNull_ShouldNotThrow()
+        {
+            var first = new PrimaryKeyWithPKList
+            {
+                Id = 42
+            };
+
+            Assert.That(() =>
+            {
+                _realm.Write(() => _realm.Add(first, update: true));
+            }, Throws.Nothing);
+        }
+
+        [Test]
+        public void AddOrUpdate_WhenListObjectsHavePK_ShouldOverwriteList()
+        {
+            var first = new PrimaryKeyWithPKList
+            {
+                Id = 1
+            };
+
+            first.ListValue.Add(new PrimaryKeyObject
+            {
+                Id = 1
+            });
+
+            _realm.Write(() => _realm.Add(first));
+
+            var second = new PrimaryKeyWithPKList
+            {
+                Id = 1
+            };
+
+            second.ListValue.Add(new PrimaryKeyObject
+            {
+                Id = 2
+            });
+
+            _realm.Write(() => _realm.Add(second, update: true));
+
+            Assert.That(first.ListValue, Is.EquivalentTo(second.ListValue));
+            Assert.That(first.ListValue.Count, Is.EqualTo(1));
+            Assert.That(first.ListValue[0].Id, Is.EqualTo(2));
+            Assert.That(_realm.All<PrimaryKeyObject>().Count(), Is.EqualTo(2));
+        }
+
+        [Test]
+        public void AddOrUpdate_WhenListObjectsDontHavePK_ShouldOverwriteList()
+        {
+            var first = new PrimaryKeyWithNoPKList
+            {
+                Id = 1
+            };
+
+            first.ListValue.Add(new NonPrimaryKeyObject
+            {
+                StringValue = "1"
+            });
+
+            _realm.Write(() => _realm.Add(first));
+
+            var second = new PrimaryKeyWithNoPKList
+            {
+                Id = 1
+            };
+
+            second.ListValue.Add(new NonPrimaryKeyObject
+            {
+                StringValue = "2"
+            });
+
+            _realm.Write(() => _realm.Add(second, update: true));
+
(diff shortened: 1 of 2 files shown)


Shared/Tests.Shared/AddOrUpdateTests.cs
```diff
@@ -846,6 +846,86 @@ public void AddOrUpdate_WhenObjectHasNonEmptyList_ShouldNotThrow()
...
+            {
+                Id = 2
+            });
+
+            _realm.Write(() => _realm.Add(second, update: true));
+
+            Assert.That(first.ListValue, Is.EquivalentTo(second.ListValue));
+            Assert.That(first.ListValue.Count, Is.EqualTo(1));
+            Assert.That(first.ListValue[0].Id, Is.EqualTo(2));
+            Assert.That(_realm.All<PrimaryKeyObject>().Count(), Is.EqualTo(2));
```
AndyDentFree: This assertion is a bit confusing, could be interpreted as expecting there are two because of 'first' and 'second' above. How about testing adding 3 `ListValue` objects and asserting count is 4 after `first` is added, then can see that count **drops** to 2 when `second` updates `first` and overwrites the list.

Shared/Tests.Shared/AddOrUpdateTests.cs
```diff
@@ -846,6 +846,86 @@ public void AddOrUpdate_WhenObjectHasNonEmptyList_ShouldNotThrow()
...
+            {
+                StringValue = "2"
+            });
+
+            _realm.Write(() => _realm.Add(second, update: true));
+
+            Assert.That(first.ListValue, Is.EquivalentTo(second.ListValue));
+            Assert.That(first.ListValue.Count, Is.EqualTo(1));
+            Assert.That(first.ListValue[0].StringValue, Is.EqualTo("2"));
+            Assert.That(_realm.All<NonPrimaryKeyObject>().Count(), Is.EqualTo(2));
```
AndyDentFree: Same restructuring suggestion as for AddOrUpdate_WhenListObjectsHavePK_ShouldOverwriteList

AndyDentFree: Seems to lack a pair of tests that clear the list when you update with an empty related list (assuming that is the intent).

Shared/Tests.Shared/AddOrUpdateTests.cs
nirinchev: This assertion is verifying that overwriting the list doesn't delete any objects from the database.

AndyDentFree: I get that, I just don't think it proves the point very well because of the confusion of where the **two** objects come from. At the very least, please repeat the assertion so it is before _and_after the `add(...update:true)` - **then** it will be proving the count didn't change.

nirinchev: But the count *did* change, which kind of proves that the test is not very clear. Initially we have 1 PKO (added by the first object). Then we update the first object by Id with a list that has a different PKO, which causes the second object to be added to the realm (bumping the count of PKO's to 2), but overwriting the `PrimaryKeyWithPKList`'s list so that it only contains a single item.

Shared/Tests.Shared/AddOrUpdateTests.cs
nirinchev: The original exception was reproduced by a test like https://github.com/realm/realm-dotnet/pull/1040/files/f9ee927531099a1885cb6329c21fdae5e8081457#diff-601e2ac442a9e2a156fbd52803ca8dc1R850, but I could add a clearing test just in case.

nirinchev:
Ping @fealebenpae @AndyDentFree

PR #1048, opened 2016-12-20
Make PermissionChange immutable

nirinchev:
Previously, it was wrongly implied you could modify a PermissionChange after adding it to the Realm. That is not the case, so I've made all setters private and exposed a constructor similar to what we have in Cocoa.

Additionally, I've exposed a `ManagementObjectStatus` enum to make it friendlier to check the status of the request.

Finally, I've created an explicit unit test to verify that permissions are actually processed and we get notified about it.

### DIFF
--- a/Platform.PCL/Realm.Sync.PCL/Permissions/PermissionChangePCL.cs
+++ b/Platform.PCL/Realm.Sync.PCL/Permissions/PermissionChangePCL.cs
@@ -52,7 +52,7 @@ namespace Realms.Sync
         public string StatusMessage { get; }
 
         /// <inheritdoc />
-        public ManagementObjectStatus ObjectStatus { get; }
+        public ManagementObjectStatus Status { get; }
 
         /// <summary>
         /// Gets the user or users to effect.
--- a/Platform.PCL/Realm.Sync.PCL/Permissions/PermissionChangePCL.cs
+++ b/Platform.PCL/Realm.Sync.PCL/Permissions/PermissionChangePCL.cs
@@ -18,7 +18,7 @@
 
 using System;
 
-namespace Realms.Sync.Permissions
+namespace Realms.Sync
 {
     /// <summary>
     /// Objects of this class allow to change permissions of owned Realms.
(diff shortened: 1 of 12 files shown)


Shared/Realm.Sync.Shared/Permissions/PermissionChange.cs
```diff
@@ -40,59 +39,114 @@ public class PermissionChange : RealmObject, IPermissionObject
...
+            MayManage = mayManage;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PermissionChange"/> class.
+        /// </summary>
+        /// <remarks>
+        /// This constructor is used by Realm to create a new instance from C++. You should use the constructor that accepts parameters instead.
+        /// </remarks>
+        public PermissionChange()
```
fealebenpae: If you make this internal or even private the woven object helper should still be able to access it as it's a nested class.

Platform.PCL/Realm.Sync.PCL/Permissions/PermissionChangePCL.cs
```diff
@@ -18,7 +18,7 @@
 
 using System;
 
-namespace Realms.Sync.Permissions
+namespace Realms.Sync
 {
     /// <summary>
     /// Objects of this class allow to change permissions of owned Realms.
```
AndyDentFree: Bad grammar _allow to change permissions_ maybe rewrite as _are used to change permissions_.

Shared/Realm.Sync.Shared/Permissions/IPermissionObject.cs
```diff
@@ -18,34 +18,34 @@
...
-        DateTimeOffset CreatedAt { get; set; }
+        DateTimeOffset CreatedAt { get; }
 
         /// <summary>
-        /// Gets or sets when the object was updated the last time.
+        /// Gets when the object was updated the last time.
         /// </summary>
         /// <remarks>
         /// This should be filled by the client with the <see cref="CreatedAt"/>
         /// date and is updated by the server with the current object when the object is processed.
```
AndyDentFree: Should  _updated by the server with the current object when_ be _updated by the server with the timestamp when_

Shared/Realm.Sync.Shared/Permissions/IPermissionObject.cs
```diff
@@ -66,15 +66,20 @@ public interface IPermissionObject
         /// </listheader>
         /// </list>
         /// </remarks>
-        int? StatusCode { get; set; }
+        int? StatusCode { get; }
```
AndyDentFree: What are the values of `StatusCode` or where can the user find them?

Shared/Realm.Sync.Shared/Permissions/IPermissionObject.cs
```diff
@@ -66,15 +66,20 @@ public interface IPermissionObject
...
         /// Filled by the server after an object was processed with additional info
         /// explaining the status if necessary.
         /// </remarks>
-        string StatusMessage { get; set; }
+        string StatusMessage { get; }
+
+        /// <summary>
+        /// Gets the <see cref="ManagementObjectStatus"/> as set by the server.
+        /// </summary>
+        ManagementObjectStatus ObjectStatus { get; }
```
AndyDentFree: I don't like the name `ObjectStatus` as you have to go look at the type to infer its meaning. Maybe `ManagementStatus`?

Shared/Realm.Sync.Shared/Permissions/ManagementObjectStatus.cs
```diff
@@ -0,0 +1,41 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync
+{
+    /// <summary>
+    /// The status of the management object as set by the server.
```
AndyDentFree: What is the _management object_? Is it really an object and is the fact that it is an object sufficiently important to keep that in the name? The enum values suggest it is more like a _ServerRequestStatus_

Shared/Realm.Sync.Shared/Permissions/PermissionChange.cs
```diff
@@ -40,59 +39,114 @@ public class PermissionChange : RealmObject, IPermissionObject
...
+        /// Gets a value indicating whether the user(s) have write access to the specified Realm(s).
         /// </summary>
         /// <value><c>true</c> or <c>false</c> to request this new value. <c>null</c> to keep current value.</value>
         [MapTo("mayWrite")]
-        public bool? MayWrite { get; set; }
+        public bool? MayWrite { get; private set; }
 
         /// <summary>
-        /// Gets or sets manage access.
+        /// Gets a value indicating whether the user(s) have manage access to the specified Realm(s).
```
AndyDentFree: What does _manage access_ mean? If we have other documentation that explains this in more detail, please link to it.

Shared/Tests.Sync.Shared/Constants.cs
```diff
@@ -0,0 +1,30 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Tests.Sync
+{
+    public static class Constants
```
AndyDentFree: What is the purpose of these constants? Even though they are only in tests, I think they should have some docs because their values embody **some** assumptions. 

For example, are they a user which is expected to exist or **not** to exist?

Shared/Tests.Sync.Shared/PermissionsTests.cs
```diff
@@ -47,5 +50,35 @@ public void PermissionChange_ShouldNotBeInDefaultSchema()
         {
             Assert.That(RealmSchema.Default.Find(nameof(PermissionChange)), Is.Null);
         }
+
+        [Test, Explicit]
```
AndyDentFree: Does this work (localhost login) for both iOS and Android simulators? What about device testing? If the test has constraints please add them to the `Explicit` attribute

AndyDentFree: Are we going to also have tests about permissible state changes depending on state of server?

Platform.PCL/Realm.Sync.PCL/Permissions/PermissionChangePCL.cs
nirinchev: I haven't changed that line...

Shared/Realm.Sync.Shared/Permissions/IPermissionObject.cs
nirinchev: It's the same name as in cocoa

Shared/Realm.Sync.Shared/Permissions/ManagementObjectStatus.cs
nirinchev: Yes, it's an object and you create it like any other Realm object, that's why I went with that explanation.

Shared/Tests.Sync.Shared/PermissionsTests.cs
nirinchev: Not sure what you mean. All state changes are permissible.

Shared/Realm.Sync.Shared/Permissions/IPermissionObject.cs
nirinchev: And I haven't changed that line.

Shared/Realm.Sync.Shared/Permissions/IPermissionObject.cs
nirinchev: I'm not sure.

Shared/Realm.Sync.Shared/Permissions/PermissionChange.cs
nirinchev: It means you can have the right to manage access for other users.

Shared/Tests.Sync.Shared/PermissionsTests.cs
nirinchev: It works on iOS simulators. Whenever you want to execute the test, you'll have to update the server url to wherever your server is located. This will be updated once we get the testing framework running, so I haven't overdesigned it.

nirinchev:
I don't like the namespacing because it puts the `GetManagementRealm` extension method in a different namespace from the `User` class, so people will not get intellisense that this method exists until they have found about it. If you insist, I can move the other objects back to the `.Permissions` namespace, but I'd rather have the extension methods at the same level as the objects they're extending.

nirinchev:
Ping @realm/dotnet

fealebenpae:
If we put the `GetManagementRealm` extension method outside the namespace so it's accessible, it would still be somewhat useless without the `PermissionChange` class being accessible.

We could come to a compromise where the extension method is moved outside of namespaces so it's global (I'm a proponent of this approach, tbh) and `PermissionChange` is moved to `Realms.Sync`. Auxiliary types such as the interface and any enum types can still be in the namespace.

Should you wish to forego the `.Permissions` namespace completely, make sure that all the types you move or introduce make it clear that they belong to the permissions subsystem. At the very least this means renaming `ManagementObjectStatus` to something like `PermissionObjectStatus`, IMO.

nirinchev:
What is the purpose of that interface anyway? Doesn't seem like we're using it anywhere.

fealebenpae:
It's a mirror of the abstract base class that the object server has internally. It's meant to be a placeholder for when we add support for model inheritance - then the properties shared between the `Permission*` models can be extracted into a `PermissionObject` base class and the interface can be deprecated.

nirinchev:
Spoke with @mrackwitz and he confirmed that there will eventually be other objects (non-permissions) in the management realm, so keeping it as `ManagementObjectStatus` makes sense.

Shared/Realm.Sync.Shared/Permissions/IPermissionObject.cs
nirinchev: My bad, it is just `Status`. I'll rename it to align with their API.

PR #1055, opened 2016-12-21
Tiny API change to indicate exception thrown by User.Current

AndyDentFree:


nirinchev:
Should this be updated for the PCL as well?

PR #1057, opened 2016-12-22
Add PCL comment changes for User.Current and missing GetHashCode

AndyDentFree:
Followup to prematurerly closed PR where I forgot to add PCL version of comment update.

Whilst doing that and difffing PCL version, realised GetHashCode was missing.

### DIFF
--- a/Platform.PCL/Realm.Sync.PCL/UserPCL.cs
+++ b/Platform.PCL/Realm.Sync.PCL/UserPCL.cs
@@ -153,12 +153,5 @@ namespace Realms.Sync
             RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
             return false;
         }
-
-        /// <inheritdoc />
-        public override int GetHashCode()
-        {
-            RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
-            return 0;
-        }
     }
 }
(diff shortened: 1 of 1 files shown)


AndyDentFree:
It seems anything that has an `Equals` override causes warnings without matching `GetHashCode` so I figured having it there didn't hurt, might prevent some warnings and would keep it _more compatible_ with the implementation.

nirinchev:
I guess we can remove the `Equals` as well. I don't expect it's adding any value having those there?

fealebenpae:
Yeah, I would remove any overridden framework members such as `Equals` or `GetHashCode` unless their documentation header is useful to have.

AndyDentFree:
Nope, can't remove `Equals` and still be an `IEquatable`

Error CS0535: `Realms.Sync.User` does not implement interface member `System.IEquatable<Realms.Sync.User>.Equals(Realms.Sync.User)` (CS0535) (Realm.Sync.PCL)

nirinchev:
Sorry, I was mislead because you mentioned "Equals override". Since it's not actually overriding `object.Equals`, but rather implementing `IEquatable<User>.Equals`, you shouldn't get the `GetHashCode` warning.

PR #1061, opened 2016-12-22
Windows

### LINKED ISSUE #784: Add CI for Windows build
Needs a core build automating first.

Currently we have to build Core from a VS project ourselves and then we only use the desktop build of Realm in Windows unit testing.

### LINKED ISSUE #1008: Build Win32 wrappers against prebuilt core
Now that we have a prebuilt package of core for Win32 we should change the wrappers Visual Studio project to download and build against that instead of a local checkout of realm-core.

### LINKED ISSUE #1016: Add the Win32 native binaries and assemblies to the NuGet package

fealebenpae:
Closes #784 
Closes #1008 
Closes #1016


Shared/Realm.Shared/Realm.cs
```diff
@@ -918,7 +925,7 @@ public void RemoveAll()
         /// </summary>
         /// <typeparam name="T">The Type T must be a RealmObject.</typeparam>
         /// <param name="id">Id to be matched exactly, same as an == search.</param>
-        /// <returns>Null or an object matdhing the id.</returns>
+        /// <returns>Null or an object matching the id.</returns>
```
nirinchev: 😓

Shared/Realm.Shared/native/NativeCommon.cs
```diff
@@ -63,8 +67,18 @@ private static unsafe void DebugLogger(IntPtr utf8String, IntPtr stringLen)
 
         public static void Initialize()
         {
+            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
+            {
+                var assemblyLocation = Path.GetDirectoryName(typeof(NativeCommon).Assembly.Location);
+                var architecture = Environment.Is64BitProcess ? "x64" : "x86";
+                var path = Path.Combine(assemblyLocation, "lib", "win32", architecture) + Path.PathSeparator + Environment.GetEnvironmentVariable("PATH");
+                Environment.SetEnvironmentVariable("PATH", path);
```
nirinchev: Why do we have to update the PATH? Also, should this be an #ifdef?

Shared/Tests.Shared/AsyncTests.cs
```diff
@@ -53,10 +53,8 @@ public void TearDown()
             }
         }
 
+#if !WINDOWS
```
nirinchev: Isn't it better to `[Ignore]` the test fixture in an `#if WINDOWS`?

Shared/Tests.Shared/PropertyChangedTests.cs
```diff
@@ -36,6 +39,8 @@ public class PropertyChangedTests
 
         private Realm _realm => _lazyRealm.Value;
 
+        private static readonly Task _completedTask = Task.FromResult<object>(null);
```
nirinchev: What was wrong with `Task.CompletedTask`?

wrappers/wrappers.vcxproj
```diff
@@ -162,93 +137,57 @@
...
+      <AdditionalDependencies>core-windows\lib\Realm$(PlatformArchitecture).lib;%(AdditionalDependencies)</AdditionalDependencies>
     </Link>
   </ItemDefinitionGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
+  <Import Project="$(SolutionDir)packages\MSBuild.Extension.Pack.1.8.0\build\net40\MSBuild.Extension.Pack.targets" Condition="Exists('$(SolutionDir)packages\MSBuild.Extension.Pack.1.8.0\build\net40\MSBuild.Extension.Pack.targets')" />
+  <Target Name="DownloadCore" BeforeTargets="BeforeClCompile">
+    <PropertyGroup>
+      <RealmCoreVersion>2.2.0</RealmCoreVersion>
```
nirinchev: I wonder if we could have a single place to edit those? Not a big deal though, as I imagine it will happen quite rarely.

Shared/Tests.Shared/PropertyChangedTests.cs
fealebenpae: It's not there in version 4.5.2 of the framework.

wrappers/wrappers.vcxproj
fealebenpae: Yeah, we should adopt the `dependencies.list` approach of the other repositories.

NuGet/Realm.Database/Realm.Database.nuspec
```diff
@@ -31,14 +31,18 @@
...
-    <file src="../../wrappers/build/$Configuration$-android/x86/librealm-wrappers.so" target="lib/MonoAndroid44/x86" />
-    <file src="../../wrappers/build/$Configuration$-android/x86_64/librealm-wrappers.so" target="lib/MonoAndroid44/x86_64" />
+    <file src="../../wrappers/build/$Configuration$-android/armeabi-v7a/librealm-wrappers.so" target="native/android/armeabi-v7a" />
+    <file src="../../wrappers/build/$Configuration$-android/arm64-v8a/librealm-wrappers.so" target="native/android/arm64-v8a" />
+    <file src="../../wrappers/build/$Configuration$-android/x86/librealm-wrappers.so" target="native/android/x86" />
+    <file src="../../wrappers/build/$Configuration$-android/x86_64/librealm-wrappers.so" target="native/android/x86_64" />
+    <!-- Win32 files -->
+    <file src="../../Platform.Win32/Realm.Win32/bin/$Configuration$/Realm.*" target="lib/net452" />
+    <file src="../../wrappers/build/$Configuration$-Win32/realm-wrappers.dll" target="native/win32/x86" />
+    <file src="../../wrappers/build/$Configuration$-x64/realm-wrappers.dll" target="native/win32/x64" />
```
fealebenpae: We'll have to have a total overhaul of the nuget package for UWP where we'll use the new `runtimes` special folder and that handles UWP binaries with a special `uap` convention.

Shared/Tests.Shared/PropertyChangedTests.cs
nirinchev: Apf... Should we move that to a `TaskHelper.CompletedTask` to be used in other places as well?

Shared/Realm.Shared/native/NativeCommon.cs
fealebenpae: We update `PATH` so that we can point the CLR to the correct folder where it should look for `realm-wrappers.dll` based on architecture.

I opted not to use a preprocessor guard because that would be deprecated as soon as we switch to .NET Standard.

Shared/Tests.Shared/AsyncTests.cs
fealebenpae: NUnit 3 severely dislikes `async void` tests methods and errors out, even if they're decorated with `[Ignore]`.

I'm going to fix async tests, though, so this is only temporary to get other tests to pass.

Shared/Tests.Shared/PropertyChangedTests.cs
fealebenpae: I could always up the target framework of the tests assembly to get `Task.CompleteTask` back...

PR #1063, opened 2016-12-23
Reuse cached Realm instances

### LINKED ISSUE #933: Investigate potential object store caching issues

nirinchev:
TODO:
- [x] Write tests to verify multiple Dispose calls will not corrupt other instances.
- [x] Throw ObjectDisposed exception when accessing disposed realm methods.

Resolves #933

### DIFF
--- a/Shared/Realm.Shared/Realm.cs
+++ b/Shared/Realm.Shared/Realm.cs
@@ -18,6 +18,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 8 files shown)


Shared/Realm.Shared/Realm.cs
```diff
@@ -985,5 +973,80 @@ public void Close()
...
+        {
+            #region static
+
+            #if __IOS__
+            [MonoPInvokeCallback(typeof(NativeCommon.NotifyRealmCallback))]
+            #endif
+            public static void NotifyRealmChanged(IntPtr stateHandle)
+            {
+                var gch = GCHandle.FromIntPtr(stateHandle);
+                ((RealmState)gch.Target).NotifyChanged(EventArgs.Empty);
```
nirinchev: I don't think so - we keep the `GCHandle` to the state (line 1044) that is only freed after `realm.SharedRealmHandle.CloseRealm` is called (line 1075). I wouldn't expect that notifications would be delivered after calling `CloseRealm`. @fealebenpae you are more versed in OS inner workings, what's your take here?

fealebenpae: This is correct - the handle is only freed once the last realm it caches is closed.

PR #1064, opened 2016-12-23
Refactor tests

fealebenpae:
run all async tests in an explicit AsyncContext and enable more windows tests
make sure tests use unique realm paths where possible and realms are closed after each test

### DIFF
--- a/Shared/Tests.Shared/RealmInstanceTest.cs
+++ b/Shared/Tests.Shared/RealmInstanceTest.cs
@@ -16,26 +16,29 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-using IntegrationTests;
-using NUnit.Framework;
+using System.IO;
+using Realms;
 
-namespace IntegrationTests.Win32
+namespace IntegrationTests
 {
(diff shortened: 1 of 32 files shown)


Shared/Tests.Shared/RealmInstanceTest.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System.IO;
+using Realms;
+
+namespace IntegrationTests
+{
+    [Preserve(AllMembers = true)]
+    public abstract class RealmInstanceTest : RealmTest
+    {
+        protected RealmConfiguration _configuration = new RealmConfiguration(Path.GetTempFileName());
```
nirinchev: Why is this different from `RealmConfiguration.DefaultConfiguration` and in which cases should I use `Realm.GetInstance()` instead of `Realm.GetInstance(_configuration)`?

fealebenpae: Inheritors of this class can customize `_configuration` in their `SetUp` override and thus affect `_realm`. You should very rarely need to `Realm.GetInstance(_configuration)` and should just use `_realm` instead. They only case where it makes sense to not do that is when you need to open a new realm instance with the same configuration as `_realm`, so `Realm.GetInstance(_configuration)` is short for `Realm.GetInstance(_realm.Config)`.

PR #1071, opened 2016-12-26
Expose IPermissionObject.ErrorCode to replace StatusCode

### LINKED ISSUE #1052: Use the ErrorCode enum in IPermissionObject
`IPermissionObject.StatusCode` maps to the `ErrorCode` enum. We should expose that as it has docs and it's easier to use than a simple int.

nirinchev:
Resolves #1052

### DIFF
--- a/Shared/Realm.Sync.Shared/Exceptions/ErrorCodeHelper.cs
+++ b/Shared/Realm.Sync.Shared/Exceptions/ErrorCodeHelper.cs
@@ -30,13 +30,12 @@ namespace Realms.Sync
                 case 0:
                     return null;
                 default:
-                    var errorCode = (ErrorCode)statusCode.Value;
-                    if (!Enum.IsDefined(typeof(ErrorCode), errorCode))
+                    if (!Enum.IsDefined(typeof(ErrorCode), statusCode.Value))
                     {
-                        errorCode = ErrorCode.Unknown;
+                        return ErrorCode.Unknown;
                     }
 
-                    return errorCode;
+                    return (ErrorCode)statusCode.Value;
             }
         }
     }
--- a/Shared/Realm.Sync.Shared/Exceptions/ErrorCodeHelper.cs
+++ b/Shared/Realm.Sync.Shared/Exceptions/ErrorCodeHelper.cs
@@ -1,0 +1,43 @@
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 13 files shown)


nirinchev:
I know, but I got a build error for that one :) I didn't feel like fixing that in this PR, but will investigate and open a separate issue.

Shared/Realm.Sync.Shared/Exceptions/ErrorCodeHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        public static ErrorCode? GetErrorCode(int? statusCode)
+        {
+            switch (statusCode)
+            {
+                case null:
+                case 0:
+                    return null;
+                default:
+                    var errorCode = (ErrorCode)statusCode.Value;
```
AndyDentFree: Checking the value is valid after the cast made me feel a bit uneasy but is (currently) apparently legal.

Shared/Realm.Sync.Shared/Exceptions/ErrorCodeHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+                    if (!Enum.IsDefined(typeof(ErrorCode), errorCode))
+                    {
+                        errorCode = ErrorCode.Unknown;
```
AndyDentFree: Could we include the unknown integer value into the status message in this circumstance, to help with diagnosis if it occurs?

nirinchev:
Ping @fealebenpae I've fixed the weaver in this PR since it was fairly small change. Can you look over the last commit?

Shared/Realm.Sync.Shared/Exceptions/ErrorCodeHelper.cs
nirinchev: Is there another way to check it?

Shared/Realm.Sync.Shared/Exceptions/ErrorCodeHelper.cs
nirinchev: Which status message?

Shared/Realm.Sync.Shared/Exceptions/ErrorCodeHelper.cs
fealebenpae: You can use `Enum.IsDefined` to be extra safe.

nirinchev: I use it on the next line. Or do you suggest I rework it as:
```csharp
if (Enum.IsDefined(typeof(ErrorCode), statusCode.Value))
{
    return (ErrorCode)statusCode.Value;
}
else
{
    return ErrorCode.Unknown;
}
```

fealebenpae: Yeah, use the underlying type instead of casting.

Shared/Realm.Sync.Shared/Exceptions/ErrorCodeHelper.cs
AndyDentFree: The property `StatusMessage` but maybe that's not a great idea as it would require extra mapping.

nirinchev: StatusMessage is populated by the server - it should have the correct error description set.

PR #1072, opened 2016-12-26
Verify that [Required] is applied on correct property types

### LINKED ISSUE #912: Validation on [Required] properties

nirinchev:
Resolves #912


Weaver/RealmWeaver.Fody/extensions/PropertyDefinitionExtensions.cs
```diff
@@ -23,6 +23,8 @@
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 
+using static ModuleWeaver;
```
fealebenpae: 😂

nirinchev: C# 6.0 ftw ;)

PR #1075, opened 2016-12-28
Integrate docfx 

### LINKED ISSUE #854: Api reference docs incorrectly recognizes expression bodied properties as attributes
#### Expected Results

Expression bodied properties are documented as properties in the api reference page.
#### Actual Results

They are marked as attributes.

https://realm.io/docs/xamarin/latest/api/class_realms_1_1_realm_object.html

### LINKED ISSUE #1074: Migrate to docfx

nirinchev:
TODO:
- [ ] ~Add some intro to `api/index.md`.~
- [x] Logo update.
- [ ] Decide if we should use PCL/Xamarin project

Consider namespace changes
- [x] `Realms.ObjectSchema` -> `Realms.Schema.ObjectSchema`
- [ ] ~`Realms.XXXAttribute` -> `Realms.Attributes.XXXAttribute`~
- [x] `Realms.XXXException` -> `Realms.Exceptions.XXXException`

Decide what to filter out
- [x] `IRealmObjectHelper`
- [x] `RealmObject.{Get|Set}XXXValue`
- [x] `WovenAttribute`, `WovenPropertyAttribute`
- [x] `PreserveAttribute`

Functional changes
- [x] Removed `RealmObjectAlreadyManagedByRealmException`.
- [x] Made `RealmSchema.Builder` private.

Resolves #1074 
Resolves #854


kristiandupont:
Is `docfx_project` a standard name? Seems a bit inconsistent with what we otherwise have..

nirinchev:
That's the standard folder you get from running `docfx init -q`, but I guess I could rename it, I don't think any of the functionality is tied to that name.

nirinchev:
Should we exclude obsolete members from the api reference?

AndyDentFree:
>Should we exclude obsolete members from the api reference?

Good question. I think we keep them.

Whilst code using older NuGets will still compile, they need to be documented as being obsolete. We can't erase outdated code samples floating around the net so it is important that anyone doing a search for those methods will see docs showing their obsolete status and the replacements.

kristiandupont:
The view source links point to the PCL version of the code. It should probably point to the shared version instead.

Docs/index.md
```diff
@@ -0,0 +1,75 @@
...
+The API reference is located at [realm.io/docs/xamarin/latest/api](https://realm.io/docs/xamarin/latest/api/).
+
+Source
+------
+Source is available [on Github](https://github.com/realm/realm-dotnet).
+
+Instructions on how to build from source are included in that repository's `README.md`. 
+
+
+Minimal Sample
```
kristiandupont: Perhaps this should be a link to walkthrough docs instead? Seems redundant to maintain another minimal sample with no real added value..

Docs/template/partials/footer.tmpl.partial
```diff
@@ -0,0 +1,14 @@
+{{!Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE file in the project root for full license information.}}
```
kristiandupont: I guess this should say Realm Inc instead.

Docs/linqsupport.md
```diff
@@ -0,0 +1,234 @@
+LINQ support in Realm Xamarin
+=============================
+
+To make a query with Realm, you use the `Realm.All<T>()` method to get a `IQueryable<T>` instance.
```
kristiandupont: All of these should be `cref`'s, but that is probably outside the scope of this PR.

AndyDentFree:
I'd like to remove the _Improve this Doc_ links. We may need to add hooks to notify us of people posting such _improvements_ and I don't think the team is big enough to manage such an open contribution approach.

Platform.PCL/Realm.PCL/Schema/ObjectSchemaPCL.cs
```diff
@@ -23,18 +23,20 @@
...
     /// <summary>
-    /// Public description of a class stored in a Realm, as a collection of managed Property objects.
+    /// Public description of a class stored in a <see cref="Realm"/>, as a collection of managed <see cref="Property"/> objects.
     /// </summary>
     public class ObjectSchema : IReadOnlyCollection<Property>
     {
         /// <summary>
         /// Gets the name of the original class declaration from which the schema was built.
         /// </summary>
+        /// <value>The name of the class.</value>
```
kristiandupont: Are these necessary? Does lacking them make the docs look incomplete or something? They seem really redundant to me.

AndyDentFree:
Frankly, I don't see the appeal. It has better typography (which can be configured) but loses a number of points of functionality:

- poorer navigation
- no easy overview
- more noise
- no search.

Shared/Realm.Shared/Extensions/CollectionNotificationsExtensions.cs
```diff
@@ -139,12 +142,12 @@ public static class CollectionNotificationsExtensions
         }
 
         /// <summary>
-        /// A convenience method that casts <c>IQueryable{T}</c> to <see cref="IRealmCollection{T}"/> which implements INotifyCollectionChanged.
+        /// <b>Deprecated</b> A convenience method that casts <see cref="IQueryable{T}"/> to <see cref="IRealmCollection{T}"/> which implements INotifyCollectionChanged.
```
kristiandupont: Doesn't docfx respect the `[Obsolete]` attribute?

Shared/Realm.Shared/RealmConfigurationBase.cs
```diff
@@ -35,24 +38,32 @@ public abstract class RealmConfigurationBase
...
+        /// <example>
+        /// <code>
+        /// config.ObjectClasses = new Type[] 
+        /// { 
+        ///     typeof(CommonClass), 
+        ///     typeof(RareClass) 
+        /// };
+        /// </code>
+        /// </example>
+        /// <value>Typically left null so by default all <see cref="RealmObject"/>s will be able to be stored in all Realms.</value>
```
kristiandupont: This seems inconsistent with what the `<value>` elements normally contain -- this is more like a remark.

Shared/Realm.Shared/exceptions/RealmInvalidTransactionException.cs
```diff
@@ -19,7 +19,8 @@
 namespace Realms
 {
     /// <summary>
-    /// An exception, thrown when trying to write data to the Realm and you haven't begun a Write transaction or when the realm is opened as read-only.
+    /// An exception, thrown when trying to write data to the <see cref="Realm"/> and you haven't begun a Write 
+    /// <see cref="Transaction"/> or when the <see cref="Realm"/> is opened as read-only.
```
kristiandupont: Is this the exception that is thrown when trying to create a transaction on a readonly realm? Because if not, I think the case is covered by the comment above. Also, it could be reworded to "when trying to write data the realm outside a transaction" or similar.

Shared/Realm.Sync.Shared/Exceptions/HttpException.cs
```diff
@@ -29,16 +29,19 @@ public class HttpException : Exception
         /// <summary>
         /// Gets the <see cref="HttpStatusCode"/> of the response.
         /// </summary>
+        /// <value>A well known <see cref="HttpStatusCode"/>.</value>
         public HttpStatusCode StatusCode { get; }
 
         /// <summary>
         /// Gets the ReasonPhrase of the response.
         /// </summary>
+        /// <value>The ReasonPhrase of the response.</value>
```
kristiandupont: This, as well as the summary, provides very little information about what this actually is. Maybe it could say "HTTP ReasonPhrase" or something?

Shared/Realm.Sync.Shared/Exceptions/SessionErrorException.cs
```diff
@@ -28,11 +28,13 @@ public class SessionErrorException : Exception
         /// <summary>
         /// Gets the kind of session error this exception represents.
         /// </summary>
+        /// <value>An enum value, describing the root cause of the error.</value>
```
kristiandupont: The "root cause" could be that the sysadmin who was supposed to keep the server running was lazy. I think this merely describes the error :-)

Platform.PCL/Realm.PCL/Schema/ObjectSchemaPCL.cs
nirinchev: screen shot 2016-12-29 at 12 46 42 pm

vs

screen shot 2016-12-29 at 12 47 03 pm

Shared/Realm.Shared/Extensions/CollectionNotificationsExtensions.cs
nirinchev: It doesn't display it very prominently.

nirinchev:
In terms of appeal, here are a few points:
- It's the next iteration of Sandcastle and as such, looks quite similar to sandcastle generated docs (e.g. [Newtonsoft.Json](http://www.newtonsoft.com/json/help/html/R_Project_Documentation.htm)).
- Most new .NET projects coming from MS use it to drive docs, e.g. [Orleans](http://dotnet.github.io/orleans/index.html).
- It has search (no idea why that hasn't rendered, investigating).
- It has a link to the source code so people can easily check what exactly is happening.
- It allows plugging in markdown files to supplement the xml docs.
- While it has its own set of issues, it correctly recognizes c# 6.0 syntax.
- It has links to the BCL classes (`INotifyCollectionChanged`, etc.).
- It shows extension methods in the declaring class - `User.GetManagementRealm`, `Realm.GetSession`, etc.
- It looks more "modern".

Docs/template/partials/footer.tmpl.partial
nirinchev: This is the original footer which is licensed by MS, I'm not a lawyer, but I don't think we can copyright it :)

nirinchev: The change is in the footer text - I replace Microsoft with Realm :)

nirinchev:
I deployed the docs on github pages so you can play with it: https://nirinchev.github.io/DocfxTest/index.html

nirinchev:
@fealebenpae I'd say let's keep them private for now (otherwise it's a bit confusing to have them in our public API, but not offer any chance to use them). Once we expose the dynamic opening API, we'll make them public again. How does that sound?

nirinchev:
The reason why "View Source" links to the pcl is that docfx doesn't play nicely with Xamarin projects - it doesn't resolve BCL references. It works fine with the windows project, but we don't have a windows.sync project to generate the sync docs from. So as far as I can see, we have 4 options here:
- Generate docs from PCL and have "View Source" point to the dummy implementation for the time being.
- Generate docs from Xamarin.iOS/Android and have unresolved BCL references (e.g. `INotifyCollectionChanged`, `IList<T>`, etc.). This one is not very appealing, because when we have a BCL reference in the XML docs, it will replace it with an empty string (i.e. not only does it not link to msdn, it also will not show the name of the class, which is quite confusing).
- Create a dummy Win32.Sync project just for docs.
- Have @fealebenpae ship a working Win32.Sync implementation over the weekend. 

Personally, I find Option 4 most appealing, but would be happy to hear your thoughts as well. @kristiandupont @AndyDentFree

nirinchev:
I decided against the attribute namespace change - mainly because I imagine it will be very common scenario to add attributes to your models, which will make it really annoying to have to add both namespaces.

Issue #1053, opened 2016-12-20
Backlinks crash when model is defined in PCL


nirinchev:
As reported in #1030 , you get `System.BadImageFormatException: Method is a pinvoke or open generic` when accessing backlink collection in a PCL. If the model were defined in the native assembly, things work just fine.

Repro project: https://github.com/realm/realm-dotnet/issues/1030#issuecomment-268155628

nirinchev:
After quickly playing around with the repro project, I have found the following:
- If the model is defined in PCL, it doesn't matter if it's used in the native or the PCL project, crashes in both cases.
- Doesn't matter if the object is managed or not - crashes in both cases.
- The generated IL seems identical except for assembly names.

PR #1083, opened 2016-12-30
PCL Testing and pcl backlink crash #1053

AndyDentFree:
The approach I have ended up using to audit our PCL vs standard builds is to have a parallel unit test build that pulls in our `Tests.Shared` and `Tests.Sync.Shared` shared projects to build them inside a PCL.

That has required a little bit of refactoring to supply non-PCL operations.

This PR doesn't yet build that project as the Thread calls have not been solved. If you build the project `TestInPCL.XamarinIOS` you will see the errors.


AndyDentFree:
Even if we adopt this instead of the simplified approach, is waiting for #1105

bmunkholm:
@AndyDentFree We don't use "WIP" i titles as the _state_ of issues is reflected by the "S:" states. In this case the `S:In Progress` which is exactly the same ;-)

AndyDentFree:
Closing and abandoning the pull request due to general disagreement with this approach to unit testing our PCL reference assembly.

PR #1088, opened 2017-01-03
Remove references to CreateObject from our tests

### LINKED ISSUE #1077: Update tests to avoid using obsolete members

nirinchev:
One nice side effect is that it helped catch a bug with `Realm.Add` when supplying an object that is not part of the schema - it threw an obscure `KeyNotFoundException` which is not the case anymore.

Resolves #1077


nirinchev:
That's an interesting question. It seems like the API should still "work", which means it should have tests.

PR #1093, opened 2017-01-03
Fix PCL broken reference assemblies (via debugging Backlink crash)

AndyDentFree:
just fixing a couple of minor PCL declaration issues and addressing the GetBackLinks confusion causing the crash.

Working theory - Having method overloads breaks the way we do lookups in ModuleWeaver.


AndyDentFree:
I am still doing testing to prove that this is actually the fix.

AndyDentFree:
This PR is blocked until #1105

AndyDentFree:
OK, testing passes so this is open for review again.

Platform.PCL/Realm.PCL/Schema/RealmSchemaPCL.cs
```diff
@@ -38,7 +38,14 @@ public class RealmSchema : IReadOnlyCollection<ObjectSchema>
         /// Gets the number of known classes in the schema.
         /// </summary>
         /// <value>The number of known classes.</value>
-        public int Count { get; }
```
fealebenpae: The only constructor of this class is private and throws. Is there a chance that the `Count` property would ever be accessed?

AndyDentFree: I was originally thinking of code calling Count independent of construction? You could have code which was in a PCL which received a `RealmSchema` object from elsewhere then called Count on it. In practice, I guess your simplification is worth it.

AndyDentFree:
> I would revert the changes that rename the GetBacklinks overload to GetBacklinksForHandle.

I am uneasy relying on overloads because that also seemed to be the cause of one error with Fody. As this is an internal method I would rather leave it renamed, to be on the safe side.

Platform.PCL/Realm.PCL/Schema/RealmSchemaPCL.cs
fealebenpae: For code to be able to get ahold of a `RealmSchema` instance to call `Count` on then the realm assembly must be loaded instead of the reference one.

fealebenpae:
The changes in #1117 take care of the overload resolution problem.

AndyDentFree:
@fealebenpae unless you're really set on me changing it back, I'd prefer to leave `GetBacklinksForHandle`. I hadn't managed to quantify my unease before but this discussion helped - I **really** don't like having overloads which are purely internal implementation, that match an API method. Overloads of the public API of a class should only exist to be part of the API at the same level.

Note that _public_ in this sense is referring to `GetBackLinks` being a protected method thus used by classes inheriting from `RealmObject`.

PR #1117, opened 2017-01-08
Use hard-coded MethodReferences in the weaver

### LINKED ISSUE #1053: Backlinks crash when model is defined in PCL

### LINKED ISSUE #1105: Hard-code type and method references in the weaver instead of resolving
The weaver relies on resolving assemblies and importing type and method references. This is error-prone for two reasons: the weaver doesn't always resolve to the correct assembly reference for the target framework of the woven assembly, and it sometimes selects the wrong overload.

The solution would be to craft `TypeReference` and `MemberReference` objects manually, describing in detail method parameters so as to guarantee selecting the correct overload. What's more, the weaver should respect the `TargetFramework` attribute of the woven assembly and construct `TypeReference` objects based on existing framework API surfaces instead of trying to best guess. This is vitally important for types such as `List<T>` or `Queryable<T>` which exist in different assemblies in different frameworks.

### LINKED ISSUE #1116: Failed To Resolve Reference When Adding Backlink
#### Goal

I am trying to add inverted relationship to `ItemModel : RealmObject`

#### Expected Results

Successful compilation.

#### Actual Results

> /iOS/MTOUCH: Error MT2002: Failed to resolve "System.Linq.IQueryable System.Linq.Queryable::AsQueryable(System.Collections.IEnumerable)" reference from "System.Linq.Queryable, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" (MT2002) (*.iOS)

I tried:
- removing/adding Realm
- updating packages
- cleaning/building

Also tried to change `IQueryable` to `IList`, but get when accessing object:

> ../realm/table.hpp:1634: [realm-core-2.2.0] Assertion failed: typeid(Col) == typeid(col)
> 0   *.iOS                          0x000000010bc5f4f8 _ZN5realm4utilL18terminate_internalERNSt3__118basic_stringstreamIcNS1_11char_traitsIcEENS1_9allocatorIcEEEE + 40
> 1   *.iOS                          0x000000010bc5f7fa _ZN5realm4util9terminateEPKcS2_lOSt16initializer_listINS0_9PrintableEE + 506
> 2   *.iOS                          0x000000010bdfd25d _ZN5realm5Table12get_linklistEmm + 173
> 3   *.iOS                          0x000000010bc1ddb1 object_get_list + 156
> 4   ???                                 0x000000011e43fbe1 0x0 + 4802739169
> IMPORTANT: if you see this error, please send this log to help@realm.io.

#### Code Sample

```csharp
public class ItemModel : RealmObject
{
	[Backlink(nameof(UserModel.Items))]
	public IQueryable<UserModel> Owners 
	{
		get;
	}
}
```

Realm version(s): 0.81.0

Xamarin/Visual Studio version: Xamarin Studio 6.1.3

Which operating system version and device: iOS 10.2

fealebenpae:
Closes #1105 
Fixes #1116 
Fixes #1053


Weaver/RealmWeaver.Fody/ImportedReferences.cs
```diff
@@ -0,0 +1,353 @@
...
+
+            ICollectionOfT = new TypeReference("System.Collections.Generic", "ICollection`1", Module, Types.CoreLibrary);
+            ICollectionOfT.GenericParameters.Add(new GenericParameter(ICollectionOfT));
+
+            IListOfT = new TypeReference("System.Collections.Generic", "IList`1", Module, Types.CoreLibrary);
+            IListOfT.GenericParameters.Add(new GenericParameter(IListOfT));
+
+            System_Type = new TypeReference("System", "Type", Module, Types.CoreLibrary);
+
+            System_DateTimeOffset = new TypeReference("System", "DateTimeOffset", Module, Types.CoreLibrary, true);
```
nirinchev: Named args for the boolean :)

Weaver/RealmWeaver.Fody/ImportedReferences.cs
```diff
@@ -0,0 +1,353 @@
...
+
+            // If the assembly has a reference to PropertyChanged.Fody, let's look up the DoNotNotifyAttribute for use later.
+            var PropertyChanged_Fody = Module.AssemblyReferences.SingleOrDefault(a => a.Name == "PropertyChanged");
+            if (PropertyChanged_Fody != null)
+            {
+                InitializePropertyChanged_Fody(PropertyChanged_Fody);
```
nirinchev: This method is fairly small, we could inline it I guess. Alternatively, we could move the check for PropertyChanged_Fody inside the method as well.

Weaver/RealmWeaver.Fody/ImportedReferences.cs
```diff
@@ -0,0 +1,353 @@
...
+                    PublicKeyToken = corlib.PublicKeyToken
+                };
+
+                Module.AssemblyReferences.Add(assembly);
+            }
+
+            return assembly;
+        }
+
+        private sealed class NETFramework : ImportedReferences
```
nirinchev: I would redesign the inheritance structure here. Instead of making the properties abstract and forcing people to initialize them in inheriting class's ctor, I would expose an abstract `GetXXAssembly` method and initialize them in the base class. This way, we would be able to inline `InitializeFrameworkMethods` and `InitializeRealm` in the base class ctor and make all properties `readonly` :D This is definitely not a blocker though, just throwing it out there :)

Weaver/RealmWeaver.Fody/ModuleWeaver.cs
```diff
@@ -175,6 +142,8 @@ private IEnumerable<TypeDefinition> GetMatchingTypes()
         }
     }
 
+    private RealmWeaver.ImportedReferences _references;
```
nirinchev: This should be above the `GetMatchingTypes()` method.

Weaver/RealmWeaver.Fody/ImportedReferences.cs
fealebenpae: IMO making it explicit with abstract properties makes it easier to reason about which types we're resolving according to framework from reading the source.

PR #1140, opened 2017-01-10
Logout all active user prior to calling reset_for_testing

### LINKED ISSUE #1045: ResetForTesting should close all active sessions
From OS's comment on `sync_manager.reset_for_testing`:
> // Reset the singleton state for testing purposes. DO NOT CALL OUTSIDE OF TESTING CODE.
// Precondition: any synced Realms or `SyncSession`s must be closed or rendered inactive prior to
// calling this method.

nirinchev:
Closes #1045

### DIFF
--- a/Shared/Realm.Sync.Shared/Handles/SharedRealmHandleExtensions.cs
+++ b/Shared/Realm.Sync.Shared/Handles/SharedRealmHandleExtensions.cs
@@ -117,7 +117,14 @@ namespace Realms.Sync
 
         public static void ResetForTesting(UserPersistenceMode? userPersistenceMode = null)
         {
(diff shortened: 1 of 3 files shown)


nirinchev:
I tried to stick with the way [Cocoa does it](https://github.com/realm/realm-cocoa/blob/master/Realm/ObjectServerTests/RLMSyncManager+ObjectServerTests.m) and it seemed like they do it in Obj-C, although there's less of a distinction on their side. Do you think I should move it to our wrappers?

fealebenpae:
Cocoa does clever things on the Objective-C side, whereas we are more straight-forward. I think even just skipping all these extra allocations between each test case would be a good reason for doing it in wrappers.

PR #1141, opened 2017-01-10
Add support for case-insensitive contains

### LINKED ISSUE #917: Case insensitive string.Contains
The other string comparison operators can all be used with case insensitive search. `string.Contains` can't at the moment and the reason is partly that it doesn't take a culture parameter. But we should work around that.

nirinchev:
Resolves #917

### DIFF
--- a/Shared/Realm.Shared/Linq/RealmResultsVisitor.cs
+++ b/Shared/Realm.Shared/Linq/RealmResultsVisitor.cs
@@ -55,6 +55,8 @@ namespace Realms
             {
                 internal static readonly LazyMethod Contains = Capture<string>(s => s.Contains(string.Empty));
 
+                internal static readonly LazyMethod ContainsStringComparison = Capture<string>(s => s.Contains(string.Empty, StringComparison.Ordinal));
+
                 internal static readonly LazyMethod StartsWith = Capture<string>(s => s.StartsWith(string.Empty));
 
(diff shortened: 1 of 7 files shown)


Shared/Realm.Shared/Linq/RealmResultsVisitor.cs
```diff
@@ -331,14 +333,23 @@ internal override Expression VisitMethodCall(MethodCallExpression m)
                 }
             }
 
-            if (m.Method.DeclaringType == typeof(string))
+            if (m.Method.DeclaringType == typeof(string) ||
+                m.Method.DeclaringType == typeof(StringExtensions))
```
nirinchev: Initially I had it in an `else if`, but felt a lot of code was duplicated that way so moved it inside with the other string methods.

I'm not sure what's the argument about using it less frequently?

nirinchev:
Yes, that's unfortunate, but it's MS's fault for breaking the pattern :P TBH, I like the `a.ToUpper().Contains("xx".ToUpper())` even less. It just doesn't seem intuitive (not to me at least), so I'll definitely need to look up some sort of documentation on how to do case-insensitive contains with Realm.

If we're to redesign this approach, I would prefer to support `IndexOf("aa") >= 0`, but that would imply we also support `IndexOf("bb") >= 1`, so once again, dead end.

Long story short, I think the current approach is the "least bad" one and opens the door for introducing support for `like` (should be quite cheap now).

PR #1145, opened 2017-01-11
Redirect the temp folder when running Windows tests

fealebenpae:
This allows the temp files created during testing to be easily deleted with the rest of the Jenkins workspace.


fealebenpae:
Yes, it deletes the current directory.

PR #1166, opened 2017-01-20
Produce a weaver error when [Backlink] is applied incorrectly

### LINKED ISSUE #1163: Backlinks to the current class fail
Trying to declare circular relationships with a Backlink result in an `InvalidOperationException` _Sequence contains no elements_ being thrown during `RealmSchema.BuildDefaultSchema`.

eg: with a class such as:

```
    public class Employee : RealmObject
    {

        public IList<Employee> Reports { get; }

        [Backlink(nameof(Employee.Reports))]
        public Employee Boss { get; set; }
    }

```

nirinchev:
Additionally, fix some weaver message inconsistencies.

Resolves #1163


nirinchev:
The test failure seems unrelated. I'll go ahead and merge it.

PR #1176, opened 2017-01-23
Bump version number to 0.83.0 so we can build local nugets with sync

AndyDentFree:
with “higher” version than release


AndyDentFree:
retest this please

PR #1178, opened 2017-01-24
Minor doc corrections, making PCL match main interface comments

AndyDentFree:
needed to prevent errors when building PCL in VS2017RC


Platform.Win32/Tests.Win32/Tests.Win32.csproj
```diff
@@ -82,6 +82,7 @@
       <Link>ForMigrationsToCopyAndMigrate.realm</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Include="ClassDiagram1.cd" />
```
nirinchev: Hm... seems like this was unintentional?

AndyDentFree: Well-spotted, I deleted the temp file but VS got it in there.

Issue #1084, opened 2016-12-30
Nullable DateTimeOffset and quering by LINQ


hrach:
I would like to filter entities by nullable DateTimeOffset property. I have exactly the same problem as desribed here: http://stackoverflow.com/questions/38763893/how-to-check-for-null-on-nullable-types-using-linq-expression-in-realm

The answers refers to nullable support for primitives, not for DateTimeOffset. Also, the referenced unit  tests doesn't test this behavior.

Documentation also doesn't describe any usage.

```csharp
public class EventEntity : RealmObject
{
	[PrimaryKey]
	public int Id { get; set; }
	public DateTimeOffset? StartAt { get; set; }
}

var realm = Realm.GetInstance();
DateTimeOffset maxStartAt = DateTimeOffset.Now.AddDays(-2);
return realm.All<EventEntity>()
	.Where(e => e.StartAt == null || e.StartAt > maxStartAt);
```
> `System.NotSupportedException: The rhs of the binary operator 'GreaterThan' should be a constant or closure variable expression. 
Unable to process `Convert(value(TickinCzHost.ViewModel.EventsViewModel+<>c__DisplayClass7_0).maxStartAt)` occurred`

#### Version of Realm and tooling
Realm version(s): 0.81
Xamarin/Visual Studio version: VS 2017 RC
Which operating system version and device: A6.0 & OnePlus 3

hrach:
Hm, the same happens for long?

```csharp
public class EventEntity : RealmObject
{
	[PrimaryKey]
	public int Id { get; set; }
	public long? StartAt { get; set; }
}

var realm = Realm.GetInstance();
long maxStartAt = DateTimeOffset.Now.AddDays(-2).ToEpochTime();
return realm.All<EventEntity>()
	.Where(e => e.StartAt == null || e.StartAt > maxStartAt);
```

fealebenpae:
Hey @hrach,

This seems like a bug. Could you try making the `maxStartAt` nullable as well? That should simplify the expression tree a bit and work around the problem area in our expression visitor.

hrach:
Thank you! The workaround nullable works! So this works:

```csharp
DateTimeOffset? maxStartAt = DateTimeOffset.Now.AddDays(-2);
return realm.All<EventEntity>()
	.Where(e => e.StartAt == null || e.StartAt > maxStartAt)
	.OrderByDescending(e => e.StartAt);
```

PR #1182, opened 2017-01-25
Added more testing and idiom to show how to get the single Boss

AndyDentFree:
Want to merge these tests in, with `RecursiveBacklinksQueryFailures` disabled as Explicit, so that can be just enabled for the fix later.

### DIFF
--- a/Shared/Tests.Shared/RelationshipTests.cs
+++ b/Shared/Tests.Shared/RelationshipTests.cs
@@ -438,7 +438,7 @@ namespace IntegrationTests
         }
 
         [Test]
-        public void RecursiveBacklinksBossTestOnly()
+        public void RecursiveBacklinkWorksViaOneToManyProperty()
         {
             realm.Write(() =>
             {
@@ -460,9 +460,9 @@ namespace IntegrationTests
             Assert.That(bossesBoss, Is.Null);
         }
 
-        // Demonstrate the bug reported in issue 1177
+        // Test is named assuming it succeeds. Currently fails and demonstrates the bug reported in issue 1177
         [Test, NUnit.Framework.Explicit("Demonstrates issue 1177 so disable from auto run")]
-        public void RecursiveBacklinksQueryFailures()
+        public void RecursiveBacklinksQueriesOnlySeeRelatedItems()
         {
             realm.Write(() =>
             {
--- a/Shared/Tests.Shared/RelationshipTests.cs
+++ b/Shared/Tests.Shared/RelationshipTests.cs
@@ -460,8 +460,7 @@ namespace IntegrationTests
             Assert.That(bossesBoss, Is.Null);
         }
 
-        // Test is named assuming it succeeds. Currently fails and demonstrates the bug reported in issue 1177
-        [Test, NUnit.Framework.Explicit("Demonstrates issue 1177 so disable from auto run")]
+        [Test]
         public void RecursiveBacklinksQueriesOnlySeeRelatedItems()
         {
             realm.Write(() =>
--- a/Shared/Tests.Shared/RelationshipTests.cs
+++ b/Shared/Tests.Shared/RelationshipTests.cs
@@ -421,12 +421,12 @@ namespace IntegrationTests
         }
 
         [Test]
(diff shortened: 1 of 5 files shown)


Shared/Tests.Shared/TestObjects.cs
```diff
@@ -209,6 +210,25 @@ public class Owner : RealmObject
...
+        public IList<Employee> Reports { get; }
+
+        [Backlink(nameof(Employee.Reports))]
+        public IQueryable<Employee> Bosses { get; }
+
+        public Employee Boss
+        {
+            get
+            {
+                return Bosses.ToList().SingleOrDefault();
```
fealebenpae: Shouldn't this be just `Bosses.SingleOrDefault()` to trigger the error condition?

AndyDentFree: I wrote it that way to show a workaround for the period up until it is fixed. Would you like a comment on there to make that clear (sorry I should have had one)?

Shared/Tests.Shared/RelationshipTests.cs
```diff
@@ -437,6 +437,56 @@ public void Backlinks()
             Assert.That(maggie.Owners, Is.EquivalentTo(new[] { dani }));
         }
 
+        [Test]
+        public void RecursiveBacklinksBossTestOnly()
```
kristiandupont: I'd prefer if we could try and stick with a more spec-style naming of our tests. The fact that this involves a boss is not relevant. This test could be named `BacklinksShouldWorkRecursively`

Shared/Tests.Shared/RelationshipTests.cs
```diff
@@ -437,6 +437,56 @@ public void Backlinks()
...
+
+            // simple query of immediate Boss 
+            var jake = realm.All<Employee>().Single(p => p.Name == "Jake");
+            Assert.That(jake.Boss?.Name, Is.EqualTo("Sally"));
+            var bossesBoss = jake.Boss?.Boss;
+            Assert.That(bossesBoss, Is.Null);
+        }
+
+        // Demonstrate the bug reported in issue 1177
+        [Test, NUnit.Framework.Explicit("Demonstrates issue 1177 so disable from auto run")]
```
kristiandupont: Do we need to specify the namespace here?

Shared/Tests.Shared/TestObjects.cs
nirinchev: Why do our tests need to show a workaround?

Shared/Tests.Shared/RelationshipTests.cs
fealebenpae: Yes, because we have the internal `Realms.ExplicitAttribute` and they conflict.

AndyDentFree: Yes, it wasn't building until I did that.

Shared/Tests.Shared/TestObjects.cs
AndyDentFree: It was unclear if we are going to ship a fix before the next release - as there's a simple but not necessarily obvious workaround I wanted to include it.

nirinchev: I am quite opposed to using tests as documentation. Putting myself in users' shoes, if I get weird behavior, I wouldn't start browsing unit tests, I'll check the docs and if there's nothing in there, I'd open an issue.

hrach: @nirinchev I'm just novice user to c# and unit test were the first I'd go for when I didn't google any info. #1084 I have no strong opinion on this, I want just add that users may take a look at tests when they are desperate. Generally, 👍  for putting info about workarounds to docs!

nirinchev:
This is superseded by #1206 for backlink issues. AFAICT the other tests were used for debugging purposes only and don't demonstrate the problematic behavior in a testable manner. I suggest extracting the gist of these in a new issue to improve our query provider performance and closing this PR.

AndyDentFree:
Relevant test has been grabbed, this is no longer needed.

Issue #1171, opened 2017-01-22
Suggestion: Realm.Error event static


nirinchev:
As a preparation for https://github.com/realm/realm-dotnet/issues/1158, we could consider making the error events static. I feel it will be easier to work with them when you don't have to subscribe every time you open a Realm.

AndyDentFree:
This feels like it could go horribly wrong but maybe I am misunderstanding things. There are a few issues mentioned on [this SO post](http://stackoverflow.com/questions/7045595/how-do-static-events-compare-to-non-static-events-in-c)

My feeling is that the main problem is that it's annoying to have to subscribe every time you open a new instance **on the same Realm**.

However, you don't necessarily want the same subscriptions to **functionally different Realms**.

When people write richer apps using different Realms in different roles, having static events would mean they would have to filter on sender, which is pushing unnecessary complexity into their app logic.

So, maybe we need a way to propagate event registration, **as an option** onto new instances on the same Realm?

sushihangover:
Two cents: I currently have apps that have more than one `Realm` instance for **different**  `RealmConfiguration`s in order to minimize memory consumption, isolating static data or very dynamic data (artificial on-disk size inflation due to multiple thread access until a Compact can be done), etc... So I would prefer a way to isolate events with different handlers *without* filtering, ... (unless I misunderstood : `don't have to subscribe every time you open a Realm` 😕 )

kristiandupont:
@sushihangover I am not sure I follow -- would you prefer handlers on the instances rather than subscribe once-static handlers? Off the top of my head I can't think of a situation where I'd keep specific error handlers for different realms. It should obviously be possible to determine which realm triggered the error, but you could do that by inspecting the `sender` argument, if you really do need special behavior

AndyDentFree:
@kristiandupont you just added ammunition for my point as to why static is a bad idea ;-)

Setting a different handler for a given Realm is a once-off that vectors to the correct handling code and keeps that code small.

Having to filter on the sender means:

1. added runtime cost per notification
2. Mixing code concerns - you now have a central handler in the application code, which makes life **really** hard or someone writing a component that uses Realm.

kristiandupont:
Runtime cost in an error notification is completely irrelevant as I see it, but #2 is a good point. If you are writing a library that relies on Realm, you could perceivably want to handle errors without affecting other users. Do the events belong on the configuration classes perhaps?

nirinchev:
[...]

PR #1198, opened 2017-01-30
Make Session.Error event static

nirinchev:
Resolves half of #1171


nirinchev:
Seems like CI fails with unrelated test failure. Looks like a timing/race issue so will merge it anyway.

PR #1200, opened 2017-01-30
Expose client reset callback

### LINKED ISSUE #1158: Client reset error event

nirinchev:
TODO:
- [x] Docs: https://github.com/realm/realm.io/pull/2736

Resolves #1158


Shared/Tests.Sync.Shared/SessionTests.cs
```diff
@@ -117,6 +116,49 @@ public void Session_Error_ShouldPassCorrectSession()
...
         }
 
+        [Test]
+        public void Session_DivergingHistories_ShouldRaiseClientResetException()
+        {
+            AsyncContext.Run(async () =>
+            {
+                var realm = await SyncTestHelpers.GetFakeRealm(isUserAdmin: true);
+                var tcs = new TaskCompletionSource<Tuple<Session, ClientResetException>>();
+                Session.Error += (sender, e) =>
```
AndyDentFree: What happens if you repeat the unit test, with the error handler being added twice?

Shared/Realm.Sync.Shared/Handles/SharedRealmHandleExtensions.cs
```diff
@@ -153,11 +171,32 @@ private static unsafe void RefreshAccessTokenCallback(IntPtr userHandlePtr, IntP
...
+                    break;
+            }
+
+            Session.RaiseError(session, exception);
+        }
+
+        private static Dictionary<string, string> MarshalErrorUserInfo(IntPtr userInfoPairs, int userInfoPairsLength)
+        {
+            return Enumerable.Range(0, userInfoPairsLength)
+                             .Select(i => Marshal.PtrToStructure<KeyValuePair<string, string>>(IntPtr.Add(userInfoPairs, i * kvpStringStringSize)))
```
AndyDentFree: Will this marshal correctly convert from C++ char* ANSI null-terminated strings?

Shared/Realm.Sync.Shared/Handles/SharedRealmHandleExtensions.cs
```diff
@@ -153,11 +171,32 @@ private static unsafe void RefreshAccessTokenCallback(IntPtr userHandlePtr, IntP
...
         {
             var session = Session.Create(sessionHandlePtr);
             var message = new string(messageBuffer, 0, (int)messageLength, System.Text.Encoding.UTF8);
-            Session.RaiseError(session, new SessionErrorException(message, errorCode));
+
+            SessionException exception;
+
+            switch (errorCode)
+            {
+                case ErrorCode.DivergingHistories:
```
AndyDentFree: Should this handle the full list of four cases as per the [server docs](https://realm.io/docs/realm-object-server/#client-recovery-from-a-backup)?

nirinchev: Good catch. Was thinking about adding them after testing that it works but eventually forgot 😓

Platform.PCL/Realm.Sync.PCL/Exceptions/ClientResetException.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Realms.Sync.Exceptions
+{
+    /// <summary>
+    /// An exception describing a condition where a reset of the local Realm is required.
+    /// </summary>
+    public class ClientResetException : SessionException
+    {
+        /// <summary>
+        /// Gets the path where the backup copy of the realm will be placed once the client reset process is complete.
```
nirinchev: Local - the client should not care about server-side paths.

Shared/Realm.Sync.Shared/Handles/SharedRealmHandleExtensions.cs
nirinchev: Yes - that's how we do it with the schema as well.

nirinchev:
Ping @AndyDentFree I decided to introduce a new extension method to simplify checking for client reset. Let me know if you think it makes sense.

PR #1206, opened 2017-02-01
Fix query creation on backlink properties

### LINKED ISSUE #1177: Queries on backlinks fail
Given a declaration:

```
    public class Employee : RealmObject
    {

        public IList<Employee> Reports { get; }

        [Backlink(nameof(Employee.Reports))]
        public IQueryable<Employee> Bosses { get; }

        public Employee Boss
        {
            get
            {
                return Bosses.FirstOrDefault();
            }
        }
    }
```

The following unit test fails because the value of `bossesBoss` is the Sally object, instead of null.

```
        [Test]
        public void RecursiveBacklinks()
        {
            realm.Write(() =>
            {
                realm.Add(new Employee
                {
                    Name = "Sally",
                    Reports =
                    {
                        new Employee { Name = "Sree" },
                        new Employee { Name = "Jake" }
                    }
                });
            });
            var jake = realm.All<Employee>().Single(p => p.Name == "Jake");
            Assert.That(jake.Boss?.Name, Is.EqualTo("Sally"));
            var bossesBoss = jake.Boss?.Boss;
            Assert.That(bossesBoss, Is.Null);
        }
```

nirinchev:
Fixes #1177


Shared/Realm.Shared/Linq/IRealmCollection.cs
```diff
@@ -110,12 +110,6 @@ internal Move(int from, int to)
     public interface IRealmCollection<out T> : IReadOnlyList<T>, INotifyCollectionChanged, INotifyPropertyChanged
     {
         /// <summary>
-        /// Gets the <see cref="ObjectSchema"/> of the contained objects.
-        /// </summary>
-        /// <value>The ObjectSchema of the contained objects.</value>
-        ObjectSchema ObjectSchema { get; }
```
fealebenpae: Isn't removing this a breaking change? Also, why remove it?

nirinchev: I'd be surprised if anyone was using it, but yes, it's a breaking change. I'm removing it because I'd be surprised if anyone was using it. Do you suggest deprecating it instead?

wrappers/src/table_cs.cpp
```diff
@@ -55,10 +55,10 @@ REALM_EXPORT Object* table_add_empty_object(Table* table_ptr, SharedRealm* realm
     });
 }
 
-REALM_EXPORT Query* table_where(Table* table_ptr, NativeException::Marshallable& ex)
+REALM_EXPORT Query* table_where(Table* table_ptr, Results* results_ptr, NativeException::Marshallable& ex)
 {
     return handle_errors(ex, [&]() {
-        return new Query(table_ptr->where());
+        return new Query(table_ptr->where(new TableView(results_ptr->get_tableview())));
```
nirinchev: This may leak.

fealebenpae: Use `new Query(*table_ptr, std::make_unique<TableView>(results_ptr->get_tableview()))` or something along those lines to prevent the `TableView` from leaking.

AndyDentFree:
I'd really like to see this tested against the recursive tests I introduced in #1182 as well.

Shared/Realm.Shared/Linq/IRealmCollection.cs
fealebenpae: I suggest leaving it as it is. It'll be extremely convenient in the dynamic API.

Shared/Realm.Shared/RealmCollectionBase.cs
```diff
@@ -85,8 +89,6 @@ public int Count
             }
         }
 
-        public Schema.ObjectSchema ObjectSchema => TargetMetadata.Schema;
```
fealebenpae: `RealmCollectionBase<T>` is internal so removing public properties on it won't affect our users, but the same property has also been removed from the public interface, as discussed in https://github.com/realm/realm-dotnet/pull/1206#pullrequestreview-19596696.

nirinchev:
@AndyDentFree my thinking is our unit tests should cover the simplest case that confirms something works. There is nothing special about recursive relationships to justify adding tests for those.

AndyDentFree:
> our unit tests should cover the simplest case that confirms something works

I think this gives us inadequate coverage.

>There is nothing special about recursive relationships to justify adding tests for those.

Well, there's already been one bug found with regards to them, so that kind of proves my point.

nirinchev:
> Well, there's already been one bug found with regards to them, so that kind of proves my point.

Not really. I can write a super complicated test that has 32 levels of indirection and nesting and so on that would expose the problem. Would it make sense to include it in our unit tests? Both your tests and the ones I've written in this PR validate the same codepaths. If there's a case not covered by our current tests, but by your tests, I'd be happy to include them, but as it stands, those are just redundant.

nirinchev:
Ping @fealebenpae @AndyDentFree

Shared/Realm.Shared/Linq/IRealmCollection.cs
```diff
@@ -113,6 +113,7 @@ public interface IRealmCollection<out T> : IReadOnlyList<T>, INotifyCollectionCh
         /// Gets the <see cref="ObjectSchema"/> of the contained objects.
         /// </summary>
         /// <value>The ObjectSchema of the contained objects.</value>
+        [Obsolete("This property is no longer necessary and will be removed in a subsequent release.")]
```
fealebenpae: Just because we don't use it internally anymore doesn't mean it lacks intrinsic value in and of itself. Please leave as it was.

Shared/Tests.Shared/RelationshipTests.cs
```diff
@@ -437,6 +437,113 @@ public void Backlinks()
...
+        public void Backlinks_WhenTargetsDifferentClass_ShouldReturnOnlyRelatedData()
+        {
+            var john = new Owner
+            {
+                Name = "John"
+            };
+
+            var doggy = new Dog
+            {
+                Name = "Doggy"
```
fealebenpae: You missed a great opportunity for a Game of Thrones reference here :/

Shared/Realm.Shared/Linq/IRealmCollection.cs
nirinchev: What is the value of exposing the `ObjectSchema`? What would people use it for?

fealebenpae: It's useful in meta-programming - either when working with the dynamic API, or when you're handed a downcasted type, such as `IRealmCollection<RealmObject>` (or some other type once we support inheritance).

nirinchev: Meh, I'd be surprised if anyone ever used it. I'll revert it though as it's not directly related to this PR (I'll still obsolete it in another one 👻)

nirinchev:
Test failures are unrelated and should be addressed by #1215

PR #1216, opened 2017-02-06
Extract a new interface exposing ObjectSchema

### LINKED ISSUE #1214: Obsolete IRealmCollection.ObjectSchema
It is not relevant for IRealmCollection. I suggest we obsolete it and introduce a new interface that will expose the same property (which we can then apply on RealmObject as well).

nirinchev:
Resolves #1214

### DIFF
--- a/Shared/Realm.Shared/ISchemaObject.cs
+++ b/Shared/Realm.Shared/ISchemaObject.cs
@@ -1,0 +1,37 @@
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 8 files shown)


Shared/Realm.Shared/ISchemaObject.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System;
+using Realms.Schema;
+
+namespace Realms
+{
+    /// <summary>
+    /// An object describing its properties in terms of a <see cref="ObjectSchema"/>.
+    /// </summary>
+    public interface ISchemaObject
```
AndyDentFree: I would have preferred if we are renaming things, to get away from the confusing `ObjectSchema` vs `SchemaObject` naming - maybe call this `ISchemaSource` instead?

nirinchev: Yes, TBH, the hardest part of this PR was coming up with a proper name for that :/ I hate `ISchemaObject`, but the alternatives I could think of were `ISchemableObject`, `IObjectWithSchema` and variations of those which were even worse.

nirinchev:
The test failures are sporadic and should be fixed in #1215

PR #1231, opened 2017-02-14
Work toward resolving Stylecop crashes

nirinchev:


Jenkinsfile
```diff
@@ -385,12 +387,18 @@ def nodeWithCleanup(String label, Closure steps) {
   }
 }
 
-def xbuildSafe(String command) {
+def xbuild(String arguments) {
   try {
-    sh "${command}"
+    sh "${xbuildCmd} ${arguments} 1> xbuildOutput"
```
emanuelez: You could do something like:
```groovy
def stdOut = sh(returnStdout: true, script: "${xbuildCmd} ${arguments}")
```

Jenkinsfile
```diff
@@ -385,12 +387,18 @@ def nodeWithCleanup(String label, Closure steps) {
...
+    def stdOut = readFile 'xbuildOutput'
+
+    echo stdOut
   } catch (err) {
-    if (err.getMessage().contains("Assertion at gc.c:910, condition `ret != WAIT_TIMEOUT' not met")) {
-      echo "StyleCop crashed. No big deal."
+    def stdOut = readFile 'xbuildOutput'
+    echo stdOut
+
+    if (stdOut.contains("Assertion at gc.c:910, condition `ret != WAIT_TIMEOUT' not met")) {
```
emanuelez: Use single quotes when you don't need to expand a variable in the string

Jenkinsfile
```diff
@@ -385,12 +387,18 @@ def nodeWithCleanup(String label, Closure steps) {
...
+      echo "StyleCop crashed, no big deal."
```
emanuelez: Single quotes

Jenkinsfile
nirinchev: I may be wrong but what will happen if sh returns non-0 result? I assume it will throw and stdOut will not be assigned? I need the stdOut in the catch clause as well to check against the GC assertion.

Jenkinsfile
nirinchev: There is a single quote in the string - I thought using double quotes looks better than escaping it, but I can switch to single quotes if that's the preferred way.

Jenkinsfile
emanuelez: I see. You could also do something like:
```groovy
def errCode = sh returnStatus: true, script: "${xbuildCmd} ${arguments} > stdout"
def out = readFile('stdout')
```
and fail if `errCode != 0 && !out.contains(blabla)`

PR #1240, opened 2017-02-20
Win32 wrappers path should not be dynamic

### LINKED ISSUE #1239: Win32 wrappers not available when downloading NuGet
#### Goal

Run realm as part of a WPF app, targeting .Net 4.5.2

#### Expected Results

To build

#### Actual Results

It doesn't build claiming the Win32 wrapper dlls are unavailable: `C:\Dev\AdGroupSearch\packages\Realm.1.0.3\native\win32\x86\realm-wrappers.dll" because it was not found.` and `C:\Dev\AdGroupSearch\packages\Realm.1.0.3\native\win32\x64\realm-wrappers.dll" because it was not found.`

Checking in the `C:\Dev\AdGroupSearch\packages\Realm.1.0.3\native` folder, there's only an `android` and an `iOS` folder present.

The following has been tried to resolve:

* Remove Nuget packages and re-add them
* Remove package folders in `C:\Dev\AdGroupSearch\packages` and restore packages
* Download Nugets using both VS 2015 and VS2017RC
* Download Nuget in Linqpad and test there, but same error

#### Code Sample

```csharp

// N/A

```

Realm version(s): 1.0.3

Xamarin/Visual Studio version: VS2017RC mainly, but also tried VS2015

Which operating system version and device: Windows 10 Enterprise 1607 targeting .Net 4.5.2

fealebenpae:
The path should always point to the location of the `Realm.Database` package and should not be affected by the presence of the `Realm` package like the iOS and Android wrappers because `Realm` doesn't provide Win32 wrappers.

Fixes #1239

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+x.y.z (TBD)
+------------------
+
+### Bug fixes
+
+- The `Realm` NuGet package no longer clobbers the path to Win32 native binaries in `Realm.Database`. (#1239)
+
 1.0.3 (2017-02-14)
 ------------------
 # Out of Beta!
--- a/NuGet/Realm.Database/Realm.Database.targets
+++ b/NuGet/Realm.Database/Realm.Database.targets
@@ -50,11 +50,11 @@
     </AndroidNativeLibrary>
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
-    <None Include="$(_RealmNugetNativePath)win32\x86\realm-wrappers.dll">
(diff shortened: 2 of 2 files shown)


nirinchev:
In that case, should we provide some other warning/error to let people know they shouldn't be using sync classes? I'm not sure, but I guess with that change, you'll get the PCL assembly if you try to use them, which might be just as confusing for people?

fealebenpae:
Unfortunately if it were not for the PCL assembly users would get a NuGet error about `net452` not being supported by the package. I'm not sure what other approach we can take.

This doesn't change the error that pops up when using Sync APIs on Windows, it merely makes the Database APIs available through the `Realm` package on Windows. If we come up with a proper Windows Sync warning I believe it should be in its own PR.

PR #1248, opened 2017-02-22
Migrate to netstandard

### LINKED ISSUE #1080: .NET Standard package
In order to support UWP we need to turn our packages into .NET Standard packages. This will require us to update the uses of a number of apis that have been removed, mostly reflection. Parts of this work has been done already in #865.

nirinchev:
## Description

Notable changes:

- System.Json is not part of netstandard, so I've taken dependency on Newtonsoft.Json
- Schema is woven at compile time in `ModuleWeaver.WeaveSchema`. Had to expose public property `RealmSchema.DefaultTypes` and set it in the application entry point. Theoretically that would allow people to modify it before the first call to `Realm.GetInstance()`.
- Realm.BuildTasks defaults to `mscorlib` when resolving dependencies. Had to do that as the default assembly resolver fails to resolve dependencies such as System.Runtime, etc. that are just reference assemblies.

Fixes #1080 

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)
* [x] Address `IReflectableType` (`RealmObject.cs`)
* [x] Address Win32 Path (`NativeCommon.cs`)
* [x] Fix the schema generation (`RealmSchema.cs`)
* [x] Fix realm assembly weaving (task fails). Currently falls back to `mscorlib` if an assembly can't be resolved. Not sure how correct that behavior is.
* [x] Add Realm.Sync
* [x] Properly create realms in SpecialFolder.Personal (`RealmConfigurationBase`, `SharedRealmHandleExtensions`)
* [x] **Compiler crash** in `SharedRealmHandleExtensions.ResetForTesting`. Filed a [bug report](https://bugzilla.xamarin.com/show_bug.cgi?id=52898). Works fine in VS. Fixed in mono master.
* [x] **Compiler crash** in `SyncConfiguration.CreateRealm`. Filed a [bug report](https://bugzilla.xamarin.com/show_bug.cgi?id=52897). Works fine in VS. Fixed in mono master.
* [x] Jenkinsfile (needs update for DataBinding projects)
* [x] Nuspecs (needs update for DataBinding projects)
* [x] Validate docfx generates proper documentation
* [x] Weave schema properly, even if assembly does not have EntryPoint (needed for tests)
* [x] Failing Session test: `Session_Error_ShouldPassCorrectSession`

## High-level overview of what's changed
- Added `Realm/Realm` and `Realm/Realm.Sync` .NET standard 1.4 projects. Those will be packaged in the NuGet and will be the only .dll's we ship.
- Renamed `Tests` to `Manual Tests` as those are run manually rather than by CI.
- Moved all test projects (platform specific + shared) to `Tests`.
- Removed platform specific Realm projects.
- Got rid of the `try-finally` pattern for assigning handles.
- **Hacky:** BuildTasks falls back to `mscorlib` if it fails to resolve an assembly (as `System.Runtime`, etc. are not real ones). As which types we use is very deterministic, it's not an issue, but it's something to watch out for.
- **[Weave the default schema](https://github.com/realm/realm-dotnet/pull/1248/files#diff-ecf2dc2a90c6d654b5f280b1b6b2a411R29) on compile** - as we don't have access to `AppDomain.Current` in .NET Standard, we can't inspect the default types at runtime. The current solution will weave logic to call [`AddDefaultTypes`](https://github.com/realm/realm-dotnet/pull/1248/files#diff-5d74b376a3f324fdc778483ddb221612R49) either in the `Assembly.EntryPoint` or if it doesn't exist, in the module initializer (that's needed for unit tests on windows). It will walk the graph of all referenced assemblies and pull in their types. **It's good enough for merging, but should be tested with complex graphs**.
- System.Json is gone, so add Newtonsoft.Json as a dependency for Realm.Sync.
- **[Use Reflection](https://github.com/realm/realm-dotnet/pull/1248/files#diff-ce3f8a31e7c0c676029f61bfbe343cbcR26)** to find `Environment.SpecialFolder.Personal` (default location of realm files). **This will not work on UWP** and will need to be extended.
- **Add new NuGet - Realm.DataBinding** that Realm.Database depends on. It is platform specific and has PCL (bait and switch), iOS, and Android versions (a shared project) and contains all the `IReflectableType`-related implementations. Needed as `TypeDelegator` will not be available until netstandard 2.0. The [weaver will weave](https://github.com/realm/realm-dotnet/pull/1248/files#diff-8b5fb211f1829c087b0c315a95462731R25) `IReflectableType` implementation to user objects by calling [`TypeInfoHelper.GetInfo`](https://github.com/realm/realm-dotnet/pull/1248/files#diff-8b45c294f3ec2e146e297958a03492edR35).
- Updated Jenkinsfile, nuspecs, docfx.json

## After merging
- Test that NuGet dependencies are correct, especially on Windows (i.e. I can create a new app, add Realm and have everything working as expected).
- Test default schema with non-trivial project graphs.
- Verify that `IReflectableType` is properly woven on PCL models as well.


nirinchev:
I'm happy to announce that non-sync tests pass (except `IReflectableType`) 🎉

Realm/Realm/Native/NativeCommon.cs
```diff
@@ -71,12 +71,20 @@ private static unsafe void DebugLogger(IntPtr utf8String, IntPtr stringLen)
...
-                var assemblyLocation = Path.GetDirectoryName(typeof(NativeCommon).Assembly.Location);
-                var architecture = Environment.Is64BitProcess ? "x64" : "x86";
-                var path = Path.Combine(assemblyLocation, "lib", "win32", architecture) + Path.PathSeparator + Environment.GetEnvironmentVariable("PATH");
-                Environment.SetEnvironmentVariable("PATH", path);
+                var osVersion = osVersionPI.GetValue(null);
+                var platformPI = osVersion.GetType().GetProperty("Platform", BindingFlags.Public | BindingFlags.Instance);
+                if (platformPI.GetValue(osVersion).ToString() == "Win32NT")
+                {
+                    // We know we're on Win32 so Assembly.Location is available
+                    var assemblyLocationPI = typeof(Assembly).GetProperty("Location", BindingFlags.Public | BindingFlags.Instance);
```
nirinchev: UWP might also report Win32NT, so this should be guarded

DataBinding/DataBindingAssemblyInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+using System.Reflection;
+
+[assembly: AssemblyDescription("Realm is a mobile database: a replacement for SQLite")]
+[assembly: AssemblyCopyright("Copyright © 2017 Realm")]
```
kristiandupont: I read recently that the year you put in a copyright statement is supposed to represent the earliest year from which the copyright applies. Thus, it should not be updated. I am not sure if that info was valid and I don't think it really matters as *at least* half of the web then gets this wrong, but since we're being overly pedantic on the .net team, I thought I'd bring it up.. :-)

DataBinding/Realm.DataBinding.TypeDelegator/TypeInfoHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System;
+using System.Collections.Concurrent;
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
+
+namespace Realms
+{
+    [SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1600:ElementsMustBeDocumented")]
+    public static class TypeInfoHelper
```
kristiandupont: Why does this need to be public?

NuGet/Realm.DataBinding/Realm.DataBinding.nuspec
```diff
@@ -0,0 +1,29 @@
...
+    <version>$version$</version>
+    <title>Realm.DataBinding</title>
+    <authors>Realm</authors>
+    <owners>Realm</owners>
+    <developmentDependency>false</developmentDependency>
+    <licenseUrl>https://opensource.org/licenses/Apache-2.0</licenseUrl>
+    <projectUrl>http://github.com/realm/realm-dotnet</projectUrl>
+    <iconUrl>http://realm.io/img/favicon-32x32.png</iconUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>Realm is a mobile database: a replacement for SQLite and ORMs</description>
```
kristiandupont: Maybe we'd want a specific description for this package?

Platform.PCL/Realm.PCL/RealmObjectPCL.cs
```diff
@@ -503,11 +503,5 @@ protected void RaisePropertyChanged([CallerMemberName] string propertyName = nul
         protected virtual void OnPropertyChanged(string propertyName)
         {
         }
-
-        TypeInfo IReflectableType.GetTypeInfo()
```
kristiandupont: Are automatic transactions not supported any more?

Platform.PCL/Realm.PCL/RealmPCL.cs
```diff
@@ -293,13 +293,15 @@ public void Write(Action action)
         }
 
         /// <summary>
-        /// Execute an action inside a temporary <see cref="Transaction"/> on a worker thread. If no exception is thrown,
+        /// Execute an action inside a temporary <see cref="Transaction"/> on a worker thread, <b>if</b> called from UI thread. If no exception is thrown,
```
kristiandupont: Maybe "Execute an action in a temporary transaction. If called from the UI thread, this will take place on a worker thread."

Realm/Realm.Sync/Handles/SharedRealmHandleExtensions.cs
```diff
@@ -45,7 +45,7 @@ private static class NativeMethods
 
             public unsafe delegate void RefreshAccessTokenCallbackDelegate(IntPtr session_handle_ptr);
 
-            public unsafe delegate void SessionErrorCallback(IntPtr session_handle_ptr, ErrorCode error_code, sbyte* message_buf, IntPtr message_len, IntPtr user_info_pairs, int user_info_pairs_len);
+            public unsafe delegate void SessionErrorCallback(IntPtr session_handle_ptr, ErrorCode error_code, byte* message_buf, IntPtr message_len, IntPtr user_info_pairs, int user_info_pairs_len);
```
kristiandupont: Was the use of `sbyte` just a typo?

DataBinding/Realm.DataBinding.TypeDelegator/TypeInfoHelper.cs
nirinchev: It's called by the woven `IReflectableType.GetType` implementation.

NuGet/Realm.DataBinding/Realm.DataBinding.nuspec
nirinchev: Yes, I didn't feel like coming up with something, so I just copy-pasted the old one.

Platform.PCL/Realm.PCL/RealmObjectPCL.cs
nirinchev: They are - the new package Realm.Databinding contains the logic for automatic transactions and the `IReflectableType` implementation is woven to all user types.

Platform.PCL/Realm.PCL/RealmPCL.cs
nirinchev: That's from @AndyDentFree's PR: #1253, haven't changed it.

Realm/Realm.Sync/Handles/SharedRealmHandleExtensions.cs
nirinchev: I can't say I'm 100% sure :) The thing is, there's no longer a string ctor that accepts `sbyte*` in .NET standard, but converting to `byte*` and using `Encoding.GetString` seems to do the trick (passes the tests at least).

kristiandupont: I see. Well those calls look nicer to me anyway.

Platform.PCL/Realm.PCL/RealmPCL.cs
kristiandupont: No, but you did add to it :-)

nirinchev: I don't think I did - it was merged in this branch which is why you're seeing it compared to master :P

DataBinding/DataBindingAssemblyInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+[assembly: AssemblyCompany("Realm Inc.")]
+[assembly: AssemblyProduct("Realm C#")]
+
+[assembly: AssemblyVersion("1.0.0.0")]
```
nirinchev: My thinking was that it should not be tied to the Realm version numbers as that is a package we'll rarely update (I hope). But it does add a bit of complexity to the release procedure if we decide to update it as that would require a bump in the dependency versions (which I'll document once this PR is finalized and merged).

Jenkinsfile
```diff
@@ -2,15 +2,19 @@ wrapperConfigurations = [
   Debug: 'dbg',
   Release: ''
 ]
-configuration = 'Release'
+configuration = 'Debug'
 
 xbuildCmd = '/usr/local/bin/xbuild'
-def nuget = '/usr/local/bin/nuget'
+nugetCmd = '/usr/local/bin/nuget'
+def windowsNugetCmd = 'C:\\ProgramData\\chocolatey\\bin\\NuGet.exe'
```
nirinchev: Yes - it requires a chocolatey install of nuget on our CI slaves. I'm not sure if we have documentation about CI setup anywhere. Ping @fealebenpae.

NuGet/Realm.Database/Realm.Database.nuspec
```diff
@@ -17,8 +17,18 @@
     <tags>Realm database db persistence sqlite</tags>
     <dependencies>
       <dependency id="Fody" version="1.29.4"/>
+      <dependency id="Realm.DataBinding" version="1.0.0"/>
       <dependency id="DotNetCross.Memory.Unsafe" version="0.2.2"/>
```
nirinchev: Let's leave it for later as this one's already huge and would prefer not to add more points where things might break :)

DataBinding/Realm.DataBinding.TypeDelegator/TypeInfoHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        // Holds Type -> RealmObjectTypeInfo map to avoid creating a new TypeDelegator for each IReflectableType.GetTypeInfo invocation.
```
nirinchev: Yes - it's still valid, as we're weaving call to `TypeInfoHelper.GetInfo` in the user's assembly, so caching types is useful.

Jenkinsfile
fealebenpae: Nope, we don't. And that only matters to Jenkins. People building manually (us, included) will just restore packages like they always would - through Visual Studio.

PR #1253, opened 2017-02-28
Ad/doc/write async comments

AndyDentFree:
Major updates to match coming in the main realm.io docs

### DIFF
--- a/Platform.PCL/Realm.PCL/RealmPCL.cs
+++ b/Platform.PCL/Realm.PCL/RealmPCL.cs
@@ -292,14 +292,17 @@ namespace Realms
             RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
         }
 
+
         /// <summary>
-        /// Execute an action inside a temporary <see cref="Transaction"/> on a worker thread. If no exception is thrown,
+        /// Execute an action inside a temporary <see cref="Transaction"/> on a worker thread, <b>if</b> called from UI thread. If no exception is thrown,
         /// the <see cref="Transaction"/> will be committed.
         /// </summary>
         /// <remarks>
         /// Opens a new instance of this Realm on a worker thread and executes <c>action</c> inside a write <see cref="Transaction"/>.
         /// <see cref="Realm"/>s and <see cref="RealmObject"/>s are thread-affine, so capturing any such objects in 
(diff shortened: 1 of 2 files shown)


nirinchev:
I've changed the base branch so it's easier to see what's been changed

Platform.PCL/Realm.PCL/RealmPCL.cs
```diff
@@ -292,14 +292,17 @@ public void Write(Action action)
...
+        /// Execute an action inside a temporary <see cref="Transaction"/> on a worker thread, <b>if</b> called from UI thread. If no exception is thrown,
         /// the <see cref="Transaction"/> will be committed.
         /// </summary>
         /// <remarks>
         /// Opens a new instance of this Realm on a worker thread and executes <c>action</c> inside a write <see cref="Transaction"/>.
         /// <see cref="Realm"/>s and <see cref="RealmObject"/>s are thread-affine, so capturing any such objects in 
-        /// the <c>action</c> delegate will lead to errors if they're used on the worker thread.
+        /// the <c>action</c> delegate will lead to errors if they're used on the worker thread. Note that it checks the
+        /// <see cref="SynchronizationContext"/> to determine if <c>Current</c> is null, as a test to see if you are on the UI thread
+        /// and will otherwise just call Write without starting a new thread. So if you know you are invoking from a worker thread, just call Write instead.
```
nirinchev: This can be `...just call <see cref="Write"> instead.`.

Platform.PCL/Realm.PCL/RealmPCL.cs
```diff
@@ -292,14 +292,17 @@ public void Write(Action action)
             RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
         }
 
+
         /// <summary>
-        /// Execute an action inside a temporary <see cref="Transaction"/> on a worker thread. If no exception is thrown,
+        /// Execute an action inside a temporary <see cref="Transaction"/> on a worker thread, <b>if</b> called from UI thread. If no exception is thrown,
```
nirinchev: While that would be the majority of the cases, it's realistic to have `SynchronizationContext` on non-UI threads, so the updated text might be a little misleading.

PR #1260, opened 2017-03-03
More fixes related to BuildTasks

nirinchev:
Make skipping logic more robust

## Description

- Uses Cecil from Nuget - means we're compatible with XF's version of Cecil, which may get loaded into the appdomain before ours, so that caused a very obscure build error.
- Ensures we're not falsely skipping weaving - if the user project references but doesn't use realm, we would incorrectly ignore it.


fealebenpae:
Shouldn't we add a dependency on the Mono.Cecil nuget in the Realm.Database nuspec?

nirinchev:
We can't add Cecil as a dependency, as it doesn't provide version for Xamarin.X and installing it on the client's project will fail.

PR #1275, opened 2017-03-13
Move the Android tests to another Jenkins node

fealebenpae:


emanuelez:
This should not be needed anymore

PR #1292, opened 2017-03-21
Call OnCompleted in the SyncProgressObservable

### LINKED ISSUE #1241: SyncProgressObservable doesn't call OnCompleted
When created with `ProgressMode.ForCurrentlyOutstandingWork`, we should call `OnCompleted` once `transferredBytes == transferableBytes`.

nirinchev:
## Description

Fixes #1241

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Realm/Realm.Sync/ProgressNotifications/SyncProgressObservable.cs
+++ b/Realm/Realm.Sync/ProgressNotifications/SyncProgressObservable.cs
@@ -76,7 +76,8 @@ namespace Realms.Sync
                 {
                     _observer.OnNext(new SyncProgress(transferredBytes, transferableBytes));
 
-                    if (_mode == ProgressMode.ForCurrentlyOutstandingWork && transferableBytes == transferredBytes)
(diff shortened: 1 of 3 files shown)


Realm/Realm.Sync/ProgressNotifications/SyncProgressObservable.cs
```diff
@@ -73,6 +75,11 @@ public void Notify(ulong transferredBytes, ulong transferableBytes)
                 Task.Run(() =>
                 {
                     _observer.OnNext(new SyncProgress(transferredBytes, transferableBytes));
+
+                    if (_mode == ProgressMode.ForCurrentlyOutstandingWork && transferableBytes == transferredBytes)
```
AndyDentFree: I think this should be `transferableBytes <= transferredBytes` just in case there's some overhead and we go past the target. Or, if you prefer, swap the order  `transferredBytes >= transferableBytes`.

PR #1293, opened 2017-03-21
Fix default DateTimeOffset value

### LINKED ISSUE #1225: DateTimeOffset property with MinDate value is set to Unix Epoch start after insert
#### Goal

I have a Task object with DateTimeOffset properties.

````csharp
public class Task : RealmObject 
{
  public DateTimeOffset Reminder { get; set; }
  ...
}
````

Task instance has Remidner property set to MinValue (I'm using DateTimeOffset.MinValue as "date not set" identifier). I want to save Task instance to Realm with

````csharp
using (var transaction = realm.BeginWrite()) {
  ...
  realm.Add (task);
  transaction.Commit();
}
````

#### Expected Results

Task instance is stored and  `task.Reminder` property should be `DateTimeOffset.MinValue` (i.e. "1/1/0001 12:00:00 AM +00:00").

#### Actual Results

Task instance is stored but `task.Reminder`property  is Unix epoch start (i.e. "1/1/1970 12:00:00 AM +00:00") and  `task.Reminder == DateTimeOffset.MinValue` returns false.

Realm version(s): 0.82.1
Xamarin/Visual Studio version: Xamarin Studio Community 6.1.5
Which operating system version and device: iOS simulator 10.2

### PR DESCRIPTION
**Pull request by nirinchev**
## Description

Fixes #1225

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Weaver/RealmWeaver.Fody/ModuleWeaver.cs
+++ b/Weaver/RealmWeaver.Fody/ModuleWeaver.cs
@@ -187,8 +187,6 @@ public partial class ModuleWeaver
             }
         }
 
-        LogDebug("trololo");
-
         WeaveSchema(matchingTypes);
 
         submitAnalytics.Wait();
--- a/Weaver/RealmWeaver.Fody/ModuleWeaver.cs
+++ b/Weaver/RealmWeaver.Fody/ModuleWeaver.cs
@@ -187,6 +187,8 @@ public partial class ModuleWeaver
             }
         }
 
+        LogDebug("trololo");
+
         WeaveSchema(matchingTypes);
 
         submitAnalytics.Wait();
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: Weaver/RealmWeaver.Fody/ModuleWeaver.cs [hunk 1]
```diff
@@ -187,6 +187,8 @@ public void Execute()
             }
         }
 
+        LogDebug("trololo");
```
fealebenpae: Probably you want to remove that.

nirinchev: Haha, yes - I was getting surprising results so wanted to confirm that the correct weaver was processing the assembly.

PR #1296, opened 2017-03-22
Update to latest dependencies

nirinchev:
## Description

Cherry pick from #1284 to get the new progress notification fixes


nirinchev:
There's a change in the .lock file layout, so maybe we could mention it as a breaking change.

PR #1311, opened 2017-03-27
Add support for like queries on strings

nirinchev:
##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


fealebenpae:
I don't like the super-generic extension method on `System.String` :/

Can we use a static method, like Entity Framework? http://stackoverflow.com/a/6814315/438778

nirinchev:
We could, but we set a precedent with `Contains`, so I figured we should reuse that. The benefit of using an extension method is the supposedly better discoverability, but if we go for a container class, we could package other methods that we add there, making it easier to see what's supported and what is not. I don't have a strong opinion regarding either option. @kristiandupont you?

fealebenpae:
To be honest I didn't much like our adding `Contains`, either. I agree it's better for discoverability, but it can also lead people to unintentionally use Realm APIs thinking they are system-provided.

kristiandupont:
>lead people to unintentionally use Realm APIs thinking they are system-provided.

I don't think that is a huge risk? 
I agree that it's less than perfect, but I don't think it warrants changing `Contains` by now, and aligning the API's seems more important then?

nirinchev:
> lead people to unintentionally use Realm APIs thinking they are system-provided

Even better for us then - making it harder for people to switch to a competitor 😆 In any case, we provide a "standalone" implementation for both, so using them on regular objects should work just fine.

PR #1316, opened 2017-03-29
POC of the new object notifications

nirinchev:
## Description

This is a POC of using the new object notifications API from OS. There are two breaking changes compared to our old approach:

- You can't subscribe for notifications during a write transaction. That means that if you open a page with an ongoing transaction (like we do with the quickjournal), you won't be able to subscribe for notifications until the transaction is committed.
- You don't get notifications until a write transaction is committed. Previously, we would send notifications to same-thread subscribers as soon as the change is made. This means that if you subscribe on the main thread and then start manipulating realm objects there, the UI will be updated only after the transaction has completed.

These may or may not be a deal breaker. The huge benefit we get from that approach is the radically simplified notifications code as well as (hopefully) fixing all those annoying deleting subscriber while sending notifications race conditions.

##  TODO

* [x] Changelog entry

### DIFF
--- a/wrappers/src/notifications_cs.hpp
+++ b/wrappers/src/notifications_cs.hpp
@@ -62,49 +62,51 @@ namespace realm {
         return -1;
     }
     
-    inline ManagedNotificationTokenContext* subscribe_for_notifications(void* managed_object, ManagedNotificationCallback callback, std::function<NotificationToken(CollectionChangeCallback)> subscriber, ObjectSchema* schema = nullptr)
+    static void handle_changes(ManagedNotificationTokenContext* context, CollectionChangeSet changes, std::exception_ptr e) {
+        if (e) {
+            try {
+                std::rethrow_exception(e);
+            } catch (...) {
+                auto exception = convert_exception();
+                auto marshallable_exception = exception.for_marshalling();
+                context->callback(context->managed_object, nullptr, &marshallable_exception);
+            }
+        } else if (changes.empty()) {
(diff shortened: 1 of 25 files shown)


wrappers/src/notifications_cs.hpp
```diff
@@ -0,0 +1,116 @@
...
+                { changes.moves.data(), changes.moves.size() },
+                { properties.data(), properties.size() }
+            };
+            
+            context->callback(context->managed_object, &marshallable_changes, nullptr);
+        }
+    }
+
+    template<typename Subscriber>
+    inline ManagedNotificationTokenContext* subscribe_for_notifications(void* managed_object, ManagedNotificationCallback callback, Subscriber subscriber, ObjectSchema* schema = nullptr)
```
nirinchev: Ping @fealebenpae was this what you had in mind?

fealebenpae: Yep, looks great.

nirinchev:
I tested this with RealmTasks and seems to work just fine. Any outstanding issues may be addressed as bug fixes.

EmmanVazz:
Currently experiencing crashes that this PR seems to fix. Any timeline to when master will be pushed to nuget?

nirinchev:
I'm sorry to hear you're having problems. Here's a prerelease package based on the latest master: https://drive.google.com/open?id=0B45yfBWnWXwAYXNqRDg2ekVyeGs.

EmmanVazz:
Thank I will give it a try.

PR #1361, opened 2017-04-20
Propose convenience API for working with permissions

nirinchev:
## TODO
- [x] Update PCL
- [x] Some more docs
- [ ] Use `GetInstanceAsync` from #1390 
- [x] Tests

## Description

A reference implementation for exposing a thin convenience wrapper around the current permission API. This also implements setting permissions by email/username. The PR is reviewable, even though there are a few polishing tasks.

## Product issues:
- https://github.com/realm/product/issues/231 - Permissions
- https://github.com/realm/product/issues/303 - Email

## Proposed API:

All permission objects are immutable and their constructors are internal, meaning a user can't perform any changes outside the convenience API exposed. The underlying implementation uses `RealmObject`s, so no infrastructure changes are necessary. All API are instance methods on the `User` class.

### GrantPermissions/RevokePermissions

A thin wrapper around PermissionChange that avoids the "null" case. Usage:

```csharp
// Signature
// Permissions is a flag enum with values Read, Write, Manage
Task GrantPermissions(string userId, string realmUrl, Permissions permissions)

// Usage

// Grants read and write permissions
// Equivalent to PermissionChange - mayRead: true, mayWrite: true, mayManage: null
await user.GrantPermissions("other-user-id", "realm://localhost/~/myrealm", Permissions.Read | Permissions.Write);

// Revokes just write permissions
// Equivalent to PermissionChange - mayRead: null, mayWrite: false, mayManage: null
await user.RevokePermissions("other-user-id", "realm://localhost/~/myrealm", Permissions.Write);
```

In the future we may introduce overloads accepting a complex type instead of an Id to allow more convoluted scenarios, e.g. granting permissions to all users matching a query or combining it with the user search functionality.

#### OfferPermissions

```csharp
// Signature
// Returns the offer token
Task<string> OfferPermissions(string realmUrl, Permissions permissions, DateTimeOffset? expiresAt = null)

// Usage
var token = await user.OfferPermissions("realm://localhost/~/myrealm", Permissions.Read | Permissions.Write);
```

#### AcceptPermissionOffer

```csharp
// Signature
// Returns the shared realm's url
Task<string> AcceptPermissionOffer(string offerToken)

// Usage
var realmUrl = await user.AcceptPermissionOffer(token);
```

#### InvalidateOffer

```csharp
// Signature
Task InvalidateOffer(PermissionOffer offer)

// Usage
var offer = ...;
await user.InvalidateOffer(offer);
```

The invalidate mechanics uses an offer object obtained from the `GetPermissionObjects` API. 

### Getting pending management objects

In case of app restart, users might want to reattach subscribers for any unprocessed management objects. Alternatively, they may want to review offers that are not yet expired and either get their tokens or invalidate them.

```csharp
// Signature
// The IQueryable<T> is RealmResults<T>
// ManagementObjectStatus is flag enum with values NotProcessed, Success, Error
IQueryable<T> GetPermissionObjects<T>(ManagementObjectStatus status) where T : RealmObject, IPermissionObject

// Usage
var offers = user.GetPermissionObjects<PermissionOffer>(ManagementObjectStatus.NotProcessed);
var changes = user.GetPermissionObjects<PermissionChange>(ManagementObjectStatus.NotProcessed);
var response = user.GetPermissionObjects<PermissionOfferResponse>(ManagementObjectStatus.NotProcessed);
```

The resulting type is `RealmResults<T>` so any queries, notifications mechanics, etc. will work out of the box.

Here's how to "reattach a subscriber" to an unprocessed object:

```csharp
var offer = user.GetPermissionObjects<PermissionOffer>(ManagementObjectStatus.NotProcessed)
                .FirstOrDefault(o => o.RealmUrl == "some-url");

if (offer != null)
{
    await offer.WaitForProcessing();
    var token = offer.Token;
}
```

### Getting granted permissions

Permissions are still backed by the `__permission` realm. Even if we decide to swap that for something else, it's likely that locally those will be stored in realm anyway. The fact it's an async API means we'll be able to create those realms on demand on the server if need be.

```csharp
// Signature
// Receiver options are CurrentUser, OtherUser, Any
// The IQueryable<T> is RealmResults<T>
Task<IQueryable<Permission>> GetGrantedPermissions(Receiver receiver = Receiver.Any, string realmUrl = null)

// Usage
var someUrlPermissions = await GetGrantedPermissions(Receiver.CurrentUser, realmUrl: "some-url");
var allPermissions = await GetGrantedPermissions(Receiver.CurrentUser);
```

The resuling collection is `RealmResults<Permission>`, so again, queries and notifications work as expected.

## Alternative API

The proposed design suffers from the disjoint between the reading and writing API. That could be addressed by simplifying the 
write APIs to accept the backing objects as parameters at the cost of clarity, e.g.

```csharp
Task ChangePermissions(PermissionChange change);

Task<string> OfferPermissions(PermissionOffer offer);

Task<string> AcceptOffer(PermissionOfferResponse response);
```


Realm/Realm.Sync/Permissions/PermissionCondition.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// </summary>
+    [SuppressMessage("StyleCop.CSharp.MaintainabilityRules", "SA1402:FileMayOnlyContainASingleClass")]
+    public abstract class PermissionCondition
+    {
+        /// <summary>
+        /// Apply permissions based on the user's Id.
+        /// </summary>
+        /// <returns>A <see cref="PermissionCondition"/> containing information about the user's Id.</returns>
+        /// <param name="userId">The Id of the user or <c>*</c> to change the permissions for all users.</param>
+        public PermissionCondition UserId(string userId)
```
kristiandupont: How come these aren't `static`?

Realm/Realm.Sync/User.cs
```diff
@@ -238,5 +242,187 @@ public override int GetHashCode()
...
+        #region Permissions
+
+        private readonly Lazy<Realm> _permissionRealm;
+        private readonly Lazy<Realm> _managementRealm;
+
+        internal Realm PermissionRealm => _permissionRealm.Value;
+
+        internal Realm ManagementRealm => _managementRealm.Value;
+
+		/// <summary>
```
kristiandupont: Some tabs have snuck in here..

Realm/Realm.Sync/Permissions/Receiver.cs
```diff
@@ -0,0 +1,41 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync
+{
+    /// <summary>
+    /// The user affected by the permission change.
+    /// </summary>
+    public enum Receiver
```
kristiandupont: Recipient?

Realm/Realm.Sync/User.cs
```diff
@@ -238,5 +242,187 @@ public override int GetHashCode()
...
+            else
+            {
+                throw new NotSupportedException("Invalid PermissionCondition.");
+            }
+
+            return WriteToManagementRealm(change);
+        }
+
+        /// <summary>
+        /// TODO lorem ipsum
```
kristiandupont: ..

Realm/Realm.Sync/Permissions/AccessLevel.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// Users who have read-only access to a Realm should open the Realm using 
+        /// <see cref="Realm.GetInstanceAsync(RealmConfigurationBase)"/> Attempting to directly open the Realm is 
+        /// an error; in this case the Realm must be deleted and re-opened.
+        /// </remarks>
+        Read = 1,
+
+        /// <summary>
+        /// User can read and write the contents of the Realm.
+        /// </summary>
+        Write,
```
nirinchev: tl;dr - I'll remove the values :)

Long version: I initially had only read, write, admin, so specified Read to be non-default. Then I added None, which was supposed to be default, and forgot to remove the values.

Realm/Realm.Sync/Permissions/ManagementObjectStatus.cs
```diff
@@ -16,26 +16,29 @@
...
 ////////////////////////////////////////////////////////////////////////////
 
+using System;
+
 namespace Realms.Sync
 {
     /// <summary>
     /// The status of the management object as set by the server.
     /// </summary>
+    [Flags]
```
nirinchev: These are flags only to enable: `user.GetPermissionOffers(ManagementObjectStatus.Success | ManagementObjectStatus.Error)` to get all offers in the success or error states. Perhaps I can make it a `params` argument and avoid the flags?

kristiandupont: Ah, that does make sense. Interesting situation actually -- proves that there are times when you want to be able to combine mutually exclusive flags. I am ok with keeping it like it is for this.

Realm/Realm.Sync/Realm.Sync.csproj
```diff
@@ -8,7 +8,7 @@
     <OutputType>Library</OutputType>
     <RootNamespace>Realms.Sync</RootNamespace>
     <AssemblyName>Realm.Sync</AssemblyName>
-    <StyleCopTreatErrorsAsWarnings>false</StyleCopTreatErrorsAsWarnings>
```
nirinchev: Needed due to the c# 7 syntax

PR #1372, opened 2017-04-27
Convert databinding project to netstandard

nirinchev:
## Description

Should fix .NET Core issues when adding the nuget

**Note**: this will not fix runtime crashes. Those will be addressed in follow-up PR.

### DIFF
--- a/DataBinding/Realm.DataBinding.PCL/TypeInfoHelper.cs
+++ b/DataBinding/Realm.DataBinding.PCL/TypeInfoHelper.cs
@@ -27,8 +27,7 @@ namespace Realms
         [SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1600:ElementsMustBeDocumented")]
         public static TypeInfo GetInfo<T>() where T : RealmObject
         {
-            RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
-            return null;
+            return typeof(T).GetTypeInfo();
         }
     }
 }
--- a/Realm/Realm/InteropConfig.cs
+++ b/Realm/Realm/InteropConfig.cs
@@ -17,6 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 using System;
+using System.IO;
 using System.Reflection;
 
 namespace Realms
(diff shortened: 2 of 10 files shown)


DataBinding/Realm.DataBinding.PCL/TypeInfoHelper.cs
```diff
@@ -27,8 +27,7 @@ public static class TypeInfoHelper
         [SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1600:ElementsMustBeDocumented")]
         public static TypeInfo GetInfo<T>() where T : RealmObject
         {
-            RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
-            return null;
+            return typeof(T).GetTypeInfo();
```
kristiandupont: Why this change?

Realm/Realm/InteropConfig.cs
```diff
@@ -38,18 +39,33 @@ internal static class InteropConfig
...
+            {
+            }
+
+            var assemblyLocationPI = typeof(Assembly).GetProperty("Location", BindingFlags.Public | BindingFlags.Instance);
+            if (assemblyLocationPI != null)
+            {
+                var assemblyLocation = Path.GetDirectoryName((string)assemblyLocationPI.GetValue(typeof(InteropConfig).GetTypeInfo().Assembly));
+                var folder = Path.Combine(assemblyLocation, "Documents");
+                Directory.CreateDirectory(folder);
+                return folder;
```
kristiandupont: So the folder will be `Documents` beneath the assembly location?

DataBinding/Realm.DataBinding.PCL/TypeInfoHelper.cs
nirinchev: I felt it's semantically incorrect to throw pcl exception and tell people to reference the platform specific version when they can't really do that (since we only have platform specific versions for iOS and Android). They shouldn't be using the library anyway, but if they somehow did, we can at least not throw :)

nirinchev: To elaborate on that - our weaver will weave `IReflectableType` on pcl/net standard projects, because we can't know whether those will be used in an iOS application or a console app. While it's unlikely someone will be casting to `IReflectableType` in a console application, a 3rd party library may be doing it.

Realm/Realm/InteropConfig.cs
nirinchev: Yes - it's far from ideal, which is why we should instruct users writing .NET Core apps to always specify the full path to the database.

PR #1383, opened 2017-05-05
Expose RealmObject.OnManaged virtual method

### LINKED ISSUE #1382: Custom initialization of RealmObject
## Goals
I work with a lot of hierarchical RealmObjects, and there are times when one or more parent objects need to be notified when a child changes (including child properties).
As I work in Xamarin.Forms, I also need it to be compatible with data binding, so changes are automatically reflected on the UI without any/much manual intervention.

## Expected Results
So far I've been trying to do this in the parent object's constructor, subscribing for notifications in the children IList, then manually raising property change notifications from that, but this doesn't work with Realm, since the object is (most of the time) not managed when the constructor is called. It does seem to work sometimes, and when it does, it works very well, but it's far from reliable and really shouldn't even be working.

@nirinchev mentioned in the Xamarin Slack that a virtual initializer for RealmObjects would apply to this situation, if it would be called (internally) after the object is managed.

From what I can tell, this should solve my issue 100%.

nirinchev:
## Description

Fixes #1382

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------------------
 
 ### Enhancements
+- Expose `RealmObject.OnManaged` virtual method that can be used for init purposes, since the constructor is run before the object has knowledge of its Realm. (#1383)
 
 ### Bug fixes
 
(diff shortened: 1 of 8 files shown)


nirinchev:
Here's a link to the [prerelease nuget](https://drive.google.com/file/d/0B45yfBWnWXwAV19kWXhvZmVrQ0E/view?usp=sharing).

PR #1399, opened 2017-05-19
Build wrappers for macOS and Linux

fealebenpae:


Jenkinsfile
```diff
@@ -177,6 +215,38 @@ stage('Build without sync') {
...
     },
+    'macOS': {
+      nodeWithCleanup('osx') {
+        getArchive()
+
+        dir('wrappers') {
+          cmake 'build-osx', "${pwd()}/build", configuration
+        }
+
+        stash includes: "wrappers/build/Darwin/${configuration}/**/*", name: 'macos-wrappers-nosync'
```
nirinchev: I wonder if it makes sense to ship nosync wrappers for OSX and Linux. I guess we might just as well do it, but I don't expect people to be upset if we didn't :)

Jenkinsfile
```diff
@@ -177,6 +215,38 @@ stage('Build without sync') {
         stash includes: 'wrappers/build/**/*.dll', name: 'uwp-wrappers-nosync'
       }
     },
+    'macOS': {
+      nodeWithCleanup('osx') {
+        getArchive()
+
+        dir('wrappers') {
```
nirinchev: It bugs me a little that everywhere else we unstash core or core & sync and here we don't 🤔

NuGet/Realm/Realm.nuspec
```diff
@@ -17,10 +17,10 @@
     <tags>Realm database db persistence sqlite</tags>
     <dependencies>
       <group targetFramework="portable-net45+sl5+wp8+wpa81+win8+monoandroid+Xamarin.iOS10+monotouch+Xamarin.Mac">
-        <dependency id="Realm.Database" version="$version$"/>
+        <dependency id="Realm.Database" version="$version$" exclude="native" />
```
nirinchev: Should we start putting the iOS and Android artifacts in the runtimes folder? Seems like that would allow us to get rid of the workaround in the targets file.

Jenkinsfile
fealebenpae: The macOS without Sync build uses the prebuilt core binaries from our CDN

Jenkinsfile
fealebenpae: Weeeell, I would agree with that, except it would mean Realm.Database wouldn't be usable on those platforms.

NuGet/Realm/Realm.nuspec
fealebenpae: I don't think NuGet handles iOS and Android in the runtimes folder. And technically the iOS wrappers are a build-time dependency because they're a static library.

fealebenpae: Maybe we can use `contentFiles` instead, though.

Jenkinsfile
nirinchev: Meh 😛 If it saves us CI time and doesn't hurt anybody, I don't see a problem with that.

fealebenpae: I thought about that, but the macOS and Linux branches complete before the Android branch does, and they're executed on the generic macOS and Docker nodes so removing them won't buy us anything in CI time.

PR #1423, opened 2017-05-31
Expose SyncConfiguration.TrustedCAPath

### LINKED ISSUE #1422: Add support for specifying trusted root CA
See the Java PR: https://github.com/realm/realm-java/pull/4591

nirinchev:
## Description

Fixes #1422 

Also fixes a case when `client_validate_ssl` wasn't respected when opening a realm with `GetInstanceAsync`.

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm.Sync/SyncConfiguration.cs
```diff
@@ -87,11 +109,18 @@ internal override Realm CreateRealm(RealmSchema schema)
 
         private Native.SyncConfiguration ToNative()
         {
+            if (!string.IsNullOrEmpty(TrustedCAPath) &&
+                !File.Exists(TrustedCAPath))
+            {
+                throw new RealmFileNotFoundException($"File not found: {TrustedCAPath}");
```
kristiandupont: Is that the proper exception here? (In fact, I am not even sure why we have an exception by that name?)

nirinchev: Well... the file's not there, the exception contains "FileNotFound", so I went with it :P Perhaps we can throw the system's `FileNotFoundException`?

kristiandupont: Right. Well, I can't think of any reason not to use the system one, unless we for some reason want all exceptions thrown by our SDK to be `Realm*`-exceptions. But that just seems silly to me. As does prefixing them with `Realm` in general, unless they are related to something in a realm, not just in the SDK.

nirinchev: I agree - we throw a bunch of system exceptions anyway. I'll update it here, but will leave the rest of the codebase for another time ;)

PR #1426, opened 2017-06-01
Remove unnecessary delete lock

### LINKED ISSUE #1356: DeleteRealm lock is meaningless
The lock `lockOnWhileDeleting` in `Realm.DeleteRealm` doesn't seem to serve any purpose. Should we remove it?

nirinchev:
## Description

Fixes #1356 

Also enables two ignored tests.

### DIFF
--- a/Realm/Realm/Realm.cs
+++ b/Realm/Realm/Realm.cs
@@ -161,19 +161,21 @@ namespace Realms
         /// <param name="configuration">A <see cref="RealmConfigurationBase"/> which supplies the realm path.</param>
         public static void DeleteRealm(RealmConfigurationBase configuration)
         {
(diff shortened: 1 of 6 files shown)


Realm/Realm/Realm.cs
```diff
@@ -429,8 +431,7 @@ public dynamic CreateObject(string className)
         {
             ThrowIfDisposed();
 
-            RealmObject.Metadata ignored;
-            return CreateObject(className, out ignored);
+            return CreateObject(className, out var _);
```
kristiandupont: Nice! Didn't know you could do this.

nirinchev: C# 7 😎

PR #1433, opened 2017-06-06
Raise PropertyChanged with correct property name, regardless of order.

### LINKED ISSUE #1432: Incorrect property name given on property changed event when Backlinks are listed before the property in the class definition
## Goals
When subscribed to property changed events on a Realm object, the property name that is given is incorrect if the property is listed after any properties marked with the Backlink attribute in the class definition.

## Expected Results
The correct property name to be passed in to the OnPropertyChanged method.

## Actual Results /  Steps to Reproduce
For the following class definition:
```
public class Captain : RealmObject
{
    [Backlink(nameof(Ship.Captain))]
    public IQueryable<Ship> Ships { get; }
    public string Age { get; set; }
}
```
When the "Age" property is changed, "Ships" is being passed into OnPropertyChanged."Name" works as expected. 

## Code Sample
I have [uploaded a Xamarin Forms project](https://github.com/ChrisWay/RealmTest) which shows the issue. The first Entry is bound to Name which works as expected but Age outputs the wrong property name.

- Realm: 1.4.0

nirinchev:
## Description

Fixes #1432

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Tests/Tests.Shared/PropertyChangedTests.cs
+++ b/Tests/Tests.Shared/PropertyChangedTests.cs
@@ -750,8 +750,46 @@ namespace Tests.Database
 
             _realm.Refresh();
             Assert.That(notifiedPropertyNames, Is.EqualTo(new[] { nameof(Person.FirstName) }));
-            
+
             person.PropertyChanged -= handler;
+        }
+
+        [Test]
+        public void ManagedObject_WhenPropertyIsAfterBacklinks()
+        {
+            var notifiedPropertyNames = new List<string>();
+            var obj = new BacklinkObject
+            {
+                AfterBacklinks = "a",
+                BeforeBacklinks = "a"
+            };
+            _realm.Write(() =>
+            {
+                _realm.Add(obj);
+            });
+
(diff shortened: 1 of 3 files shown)


Tests/Tests.Shared/PropertyChangedTests.cs
```diff
@@ -750,10 +750,48 @@ public void ManagedObject_WhenSubscribedDuringDeletion_AfterRollback_ShouldRecei
...
+            });
+
+            _realm.Refresh();
+
+            var expected = new[]
+            {
+                nameof(BacklinkObject.BeforeBacklinks),
+                nameof(BacklinkObject.AfterBacklinks)
+            };
+            Assert.That(notifiedPropertyNames, Is.EquivalentTo(expected));
```
kristiandupont: So this shouldn't assert anything about the order of the events?

nirinchev: No, that only tests that the correct property names are passed.

PR #1434, opened 2017-06-06
Don’t throw exceptions in the module initializer of PCL assemblies

### LINKED ISSUE #1431: UnitTests failure with Realm Xamarin
We have a separate solution that houses our POCO's for use in our Xamarin.Forms mobile app. We have added Realm to one of the POCO's. The solution also has a Unit Test project. We have one unit test, that doesn't have anything to do with the POCO that we added Realm to, but we are getting a failure. This solution does not contain our droid and ios projects, so I'm not sure how to resolve.

> Test method MyApp.Common.UnitTests.Utilities.UtilitiesTest.GetStateAbbreviationTest threw exception: 
> 
> System.TypeInitializationException: The type initializer for '<Module>' threw an exception. ---> System.PlatformNotSupportedException: The PCL build of Realm is being linked which probably means you need to use NuGet or otherwise link a platform-specific Realm.dll to your main application.
>     at Realms.Schema.RealmSchema.AddDefaultTypes(IEnumerable`1 types)
>    at RealmModuleInitializer.Initialize()
>    at .cctor()
>  --- End of inner exception stack trace ---
>     at MyApp.Common.UnitTests.Utilities.UtilitiesTest.GetStateAbbreviationTest()

We've tried adding Realm to the UnitTests project as well, but it didn't change and still failed the test. The test is below:

```
[TestClass]
	public class UtilitiesTest
	{
		[TestMethod]
		public void GetStateAbbreviationTest()
		{
			var abbreviation = new Common.Utilities.Utilities().GetStateAbbreviation("SOUTH CAROLINA");

			Assert.AreEqual(abbreviation, "SC");
		}
	}
```

The method the test is executing looks like this (removing most states for brevity):

```
public string GetStateAbbreviation(string stateName)
	{
		Dictionary<string, string> stateList = new Dictionary<string, string>();
		stateList.Add("ALABAMA", "AL");
		stateList.Add("SOUTH CAROLINA", "SC");
		stateList.Add("WYOMING", "WY");
		if (!string.IsNullOrEmpty(stateName) && stateList.ContainsKey(stateName.ToUpper()))
			return stateList[stateName.ToUpper()];

		return null;
	}
```

Why does this test fail with a Realm error if we aren't trying to test something for Realm? And is there any solution?

- Realm Version: v1.4.0

nirinchev:
## Description

Fixes #1431

##  TODO

* [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Fix a crash when querying over properties that have `[MapTo]` applied. ([#1405](https://github.com/realm/realm-dotnet/pull/1405))
 - Fix an issue where synchronized Realms did not connect to the remote server in certain situations, such as when an application was offline when the Realms were opened but later regained network connectivity. ([#1407](https://github.com/realm/realm-dotnet/pull/1407))
 - Fix an issue where incorrect property name will be passed to `RealmObject.PropertyChanged` subscribers when the actual changed property is below a `Backlink` property. ([#1433](https://github.com/realm/realm-dotnet/pull/1433))
+- Fix an exception being thrown when referencing Realm in a PCL test assembly without actually using it. ([#1434](https://github.com/realm/realm-dotnet/pull/1434))
 
 ### Breaking Changes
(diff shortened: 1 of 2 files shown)


nirinchev:
The failed unit test is sporadic, should be fine to merge.

PR #1435, opened 2017-06-06
Cleanup some trailing spaces and apply some c# 7 syntax here and there

nirinchev:
## Description
No functional changes apart from SuppressFinalize in some dispose calls.

### DIFF
--- a/Realm/Realm/Linq/TypeSystem.cs
+++ b/Realm/Realm/Linq/TypeSystem.cs
@@ -50,7 +50,7 @@ namespace Realms
 
             if (seqType.GetTypeInfo().IsGenericType)
             {
-                foreach (Type arg in seqType.GetGenericArguments())
+                foreach (var arg in seqType.GetGenericArguments())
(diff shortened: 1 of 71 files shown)


Realm/Realm/Linq/TypeSystem.cs
```diff
@@ -63,7 +63,7 @@ private static Type FindIEnumerable(Type seqType)
             var ifaces = seqType.GetInterfaces().ToArray();
             if (ifaces != null && ifaces.Length > 0)
             {
-                foreach (Type iface in ifaces)
+                foreach (var iface in ifaces)
```
kristiandupont: Unrelated, but I think our unwritten convention would be to use names `@interface` and `interfaces`?

nirinchev: Yes, that would be correct - I was just cleaning some warnings/suggestions and haven't looked too deeply into the code

PR #1440, opened 2017-06-07
Ensure tests always use temp filenames

nirinchev:
## Description
On non-sandboxed environments (e.g. Windows, macOS, etc.), our tests operate in the same documents folder, so parallel builds often report false negatives.

### DIFF
--- a/Tests/Tests.Shared/ConfigurationTests.cs
+++ b/Tests/Tests.Shared/ConfigurationTests.cs
@@ -27,8 +27,25 @@ using ExplicitAttribute = NUnit.Framework.ExplicitAttribute;
 namespace Tests.Database
 {
     [TestFixture, Preserve(AllMembers = true)]
-    public class ConfigurationTests
+    public class ConfigurationTests : RealmTest
     {
+        private RealmConfiguration _configuration;
+
+        protected override bool OverrideDefaultConfig => false;
+
+        protected override void CustomSetUp()
+        {
+            base.CustomSetUp();
+            _configuration = new RealmConfiguration(Path.GetTempFileName());
+        }
+
+        protected override void CustomTearDown()
+        {
+            base.CustomTearDown();
+
+            Realm.DeleteRealm(_configuration);
+        }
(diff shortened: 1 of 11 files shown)


Tests/Tests.Shared/ConfigurationTests.cs
```diff
@@ -66,7 +83,7 @@ public void PathIsCanonicalised()
             // Assert
             Assert.That(Path.IsPathRooted(config.DatabasePath));
             Assert.That(config.DatabasePath, Does.EndWith(Path.Combine("Documents", "fred.realm")));
-            Assert.IsFalse(config.DatabasePath.Contains(".."));  // our use of relative up and down again was normalised out
+            Assert.That(config.DatabasePath.Contains(".."), Is.False);  // our use of relative up and down again was normalised out
```
kristiandupont: Could probably be `Assert.That(config.DatabasePath, Does.Not.Contain(".."))`. I think.

Tests/Tests.Shared/ConfigurationTests.cs
```diff
@@ -126,23 +143,15 @@ public void UnableToOpenWithNoKey()
             ReliesOnEncryption();
 
             // Arrange
-            var config = new RealmConfiguration("UnableToOpenWithNoKey.realm");
-            Realm.DeleteRealm(config);  // ensure guarded from prev tests
-            var emptyKey = new byte[64];
-            config.EncryptionKey = emptyKey;
-            using (Realm.GetInstance(config))
+            _configuration.EncryptionKey = new byte[64];
```
kristiandupont: I remember we wanted to make configurations immutable. This will work now but maybe we will have to change it then?

nirinchev: I guess there are quite a few lines of code we'll have to change to adopt for immutable configurations. While I don't object to the idea, I feel that is best addressed when we make the change (primarily because we don't have a clone method yet).

PR #1443, opened 2017-06-07
Replace xbuild with msbuild

nirinchev:


fealebenpae:
Can you check if `msbuild` needs the same workaround for the StyleCop crash as `xbuild` does?

nirinchev:
It's a sporadic issue so will be hard to verify. Will disable it and see how it goes.

PR #1458, opened 2017-06-22
Allow building for Xamarin.Mac

nirinchev:
This is very hacky, but can be used for a quick smoke test.

### DIFF
--- a/NuGet/Realm.Database/Realm.Database.targets
+++ b/NuGet/Realm.Database/Realm.Database.targets
@@ -41,6 +41,14 @@
       <LinkerFlags>-lz -framework Security</LinkerFlags>
     </NativeReference>
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == 'Xamarin.Mac'">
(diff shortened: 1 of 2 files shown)


NuGet/Realm.Database/Realm.Database.targets
```diff
@@ -41,6 +41,14 @@
       <LinkerFlags>-lz -framework Security</LinkerFlags>
     </NativeReference>
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == 'Xamarin.Mac'">
+    <NativeReference Include="$(_RealmNugetNativePath)..\runtimes\osx.10.10-x64\native\librealm-wrappers.dylib">
+      <Kind>Dynamic</Kind>
```
fealebenpae: You only need the `Kind` metadata here.

nirinchev: What about `IsCxx`?

fealebenpae: That comes into play when we statically link a library that contains C++code that needs the C++ runtime shared library. Our wrappers already link it directly because they're a dynamic library.

PR #1466, opened 2017-06-28
Add support for RealmInteger

nirinchev:
## Description

Related to https://github.com/realm/product/issues/202

##  TODO

* [ ] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Handles/ObjectHandle.cs
+++ b/Realm/Realm/Handles/ObjectHandle.cs
@@ -96,7 +96,10 @@ namespace Realms
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_set_int64_unique", CallingConvention = CallingConvention.Cdecl)]
             public static extern void set_int64_unique(ObjectHandle handle, IntPtr propertyIndex, Int64 value, out NativeException ex);
 
-            [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_get_int64", CallingConvention = CallingConvention.Cdecl)]
+			[DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_add_int64", CallingConvention = CallingConvention.Cdecl)]
+			public static extern void add_int64(ObjectHandle handle, IntPtr propertyIndex, Int64 value, out NativeException ex);
+
+			[DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_get_int64", CallingConvention = CallingConvention.Cdecl)]
             public static extern Int64 get_int64(ObjectHandle handle, IntPtr propertyIndex, out NativeException ex);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_get_nullable_int64", CallingConvention = CallingConvention.Cdecl)]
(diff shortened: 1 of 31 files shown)


fealebenpae:
We might need extra work for queries on `RealmInteger<T>` properties. At least we should add tests for those.

nirinchev:
Good call, I'll look into it.

Weaver/WeaverTests/Realm.FakeForWeaverTests/RealmInteger.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        public RealmInteger<T> Decrement()
+        {
+            return Increment(default(T));
+        }
+
+        public RealmInteger<T> Increment(T value)
+        {
+            return new RealmInteger<T>(default(T));
```
kristiandupont: Shouldn't that be `return new RealmInteger<T>(_value + value)`?

Realm/Realm/Handles/ObjectHandle.cs
```diff
@@ -96,6 +96,9 @@ private static class NativeMethods
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_set_int64_unique", CallingConvention = CallingConvention.Cdecl)]
             public static extern void set_int64_unique(ObjectHandle handle, IntPtr propertyIndex, Int64 value, out NativeException ex);
 
+            [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_add_int64", CallingConvention = CallingConvention.Cdecl)]
+            public static extern void add_int64(ObjectHandle handle, IntPtr propertyIndex, Int64 value, out NativeException ex);
```
kristiandupont: Shouldn't this be called `increment_int64`?

Realm/Realm/RealmObject.cs
```diff
@@ -601,6 +434,42 @@ protected void SetByteArrayValue(string propertyName, byte[] value)
             _objectHandle.SetByteArray(_metadata.PropertyIndices[propertyName], value);
         }
 
+        [SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1600:ElementsMustBeDocumented")]
+        protected void SetRealmIntegerValue<T>(string propertyName, RealmInteger<T> value)
```
nirinchev: Yes, but it should be negligibly cheap.

Tests/Tests.Shared/APITests.cs
```diff
@@ -33,5 +33,56 @@ public void RealmCollectionContravariance(Type type)
...
+            Assert.That(five + 5, Is.EqualTo(10));
+            Assert.That(five > 5, Is.False);
+            Assert.That(five < 5, Is.False);
+            Assert.That(five > 3);
+            Assert.That(five < 7);
+            Assert.That(five / 5, Is.EqualTo(1));
+            Assert.That(five * 5, Is.EqualTo(25));
+            Assert.That(five - 3, Is.EqualTo(2));
+
+            var six = five.Increment();
```
nirinchev: It should be a method - it’s what you would call on an object’s property to increment it with one. It returns the new value because the struct is immutable.

Weaver/WeaverTests/Realm.FakeForWeaverTests/RealmInteger.cs
nirinchev: This is the fake assembly so we don’t care about correctness of methods.

Realm/Realm/Handles/ObjectHandle.cs
nirinchev: The native method is called add, so I wanted to keep the original naming - we can change it though - I don’t have strong opinion on the matter.

PR #1477, opened 2017-07-06
Add .NET core test runner

nirinchev:
Supersedes #1475


Jenkinsfile
```diff
@@ -469,6 +481,19 @@ def AndroidTest(stashName) {
...
 }
 
+def NetCoreTest(String node, String stashName) {
+  return {
+    nodeWithCleanup(node) {
+      getArchive()
+      unstash stashName
+
+      dir ("Tests/Tests.NetCore/bin/${configuration}/netcoreapp1.1") {
+        sh 'dotnet Tests.NetCore.dll'
```
fealebenpae: You might also want to report the result to Jenkins.

nirinchev: Yes, I'm taking baby steps - first make sure it runs, next - transform results and report them :)

Tests/Tests.NetCore/Tests.NetCore.csproj
```diff
@@ -0,0 +1,49 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\Realm\Realm\Realm.csproj" />
+    <ProjectReference Include="..\..\Realm\Realm.Sync\Realm.Sync.csproj" Condition="'$(RealmNoSync)' != 'true'" />
+  </ItemGroup>
+
+  <Import Project="..\Tests.Shared\Tests.Shared.projitems" Label="Shared" />
+  <Import Project="..\Tests.Sync.Shared\Tests.Sync.Shared.projitems" Label="Shared" Condition="'$(RealmNoSync)' != 'true'" />
+  
+  <Import Project="..\..\packages\Fody.2.0.6\build\netstandard1.4\Fody.targets" />
```
nirinchev: Look into why that's needed

Tests/Tests.Shared/NotificationTests.cs
```diff
@@ -32,8 +32,6 @@ namespace Tests.Database
     [TestFixture, Preserve(AllMembers = true)]
     public class NotificationTests
     {
-        private const int MillisecondsToWaitForCollectionNotification = 50;
```
nirinchev: We should investigate if there's a real bug with autorefreshing realms - on .NET Framework apps, this seems to work.

Jenkinsfile
```diff
@@ -406,10 +405,33 @@ stage('Build with sync') {
...
 }
 
+stage ('Build .NET Core with sync') {
+  nodeWithCleanup('windows') {
+    getArchive()
+    unstash 'macos-wrappers-sync'
+    unstash 'linux-wrappers-sync'
+    unstash 'tools-weaver'
+
+    bat 'dotnet restore Tests/Tests.NetCore/Tests.NetCore.csproj'
```
fealebenpae: This shouldn't be necessary if you're calling the Restore target in msbuild

Jenkinsfile
```diff
@@ -748,7 +770,7 @@ def transform(String original, String transform) {
     """
   } else {
     bat """
-      powershell \"\$xml = Resolve-Path ${original};\$output = Join-Path (\$pwd) temp.xml;\$xslt = New-Object System.Xml.Xsl.XslCompiledTransform;\$xslt.Load(\"${transform}\");\$xslt.Transform(\$xml, \$output);\"
+      powershell \"\$xml = Resolve-Path ${original};\$output = Join-Path (\$pwd) temp.xml;\$xslt = New-Object System.Xml.Xsl.XslCompiledTransform;\$xslt.Load(\\\"${transform}\\\");\$xslt.Transform(\$xml, \$output);\"
```
fealebenpae: Can't you just do that in Groovy? 😛

nirinchev: Hm... I wasn't able to find how, but I'm not a groovy expert, so maybe I wasn't looking at the right place :)

@emanuelez is there an easy way to apply an xslt transform in groovy?

Jenkinsfile
nirinchev: Not sure why, but it is. `dotnet restore` creates a `.targets` file that imports the Fody targets, whereas I wasn't able to see such created by `msbuild /t:Restore`. Consequently, if we don't run `dotnet restore`, the weaver is not run.

Jenkinsfile
emanuelez: There should really be as little logic as possible in Groovy so I think this approach is fine. What are you transforming by the way? If it's unit test reports I'm sure the xunit plugin could help.

nirinchev: Yes - it's a test result. I'm not familiar with xunit, but so far we've been transforming all our test results, so perhaps there's a reason which @fealebenpae knows :)

emanuelez: ok :) here's the wiki page so you can check whether your test framework is supported. https://wiki.jenkins.io/display/JENKINS/xUnit+Plugin

nirinchev: Ah... I [guess it's not](https://issues.jenkins-ci.org/browse/JENKINS-33450), which is why we're not using it.

fealebenpae: The xunit plugin also prematurely sets the build result to successful, if I remember correctly.

Tests/Tests.Shared/NotificationTests.cs
```diff
@@ -645,7 +645,13 @@ public void TestCollectionChangedAdapter(int[] initial, NotifyCollectionChangedA
...
         public void WhenSynchronizationContextExists_ShouldAutoRefresh()
         {
-            AsyncContext.Run(async () =>
+#if NETCOREAPP1_1
+            if (TestHelpers.IsMacOS)
+            {
+                Assert.Ignore("Seems like there's a bug with autorefreshing on .NET Core on macOS");
+            }
+#endif
+			AsyncContext.Run(async () =>
```
fealebenpae: indentation

PR #1479, opened 2017-07-07
Add a few more dependencies to try and get the linux build to pass

nirinchev:


nirinchev:
I'm not familiar with any of those tools so I just copied the dependencies from https://github.com/realm/realm-sync/blob/master/packaging/ubuntu-1604/build-image/Dockerfile. I'll get rid of those.

nirinchev:
I ran 7 builds with the new dockerfile and the yaml issue hasn't manifested, so probably it's safe to merge?

PR #1485, opened 2017-07-18
Add Xamarin.Mac test runner

nirinchev:


fealebenpae:
Have you looked at https://blog.xamarin.com/preview-bringing-macos-to-xamarin-forms? Is it worth it to use it for our Xamarin.Forms-based runner?

nirinchev:
Yes, I looked into it briefly, but decided against using it. My reasoning was:
1. The majority of our test runners are command-line based, so this will be familiar experience.
2. I imagine it will be at least slightly more effort to use the forms based one.
3. It's a preview and recently I've been less than happy with the quality of Xamarin/MS alpha/beta software 😞 
4. Our current build infrastructure makes it a little hard to build wrappers locally, so I don't expect us to run the Xamarin.Mac test runner during regular development.

I'll admit I didn't ponder too much on it, so if you think it's valuable to have another GUI test runner, I'm open for debate 😸

PR #1486, opened 2017-07-19
Add user lookup API

### LINKED ISSUE #1457: Expose User Lookup API
### Goal

Our users need a way to look up the random generated user IDs to connect those with their own user models of existing backends and can interact with user-specific Realms. We've an API which covers a minimum useful solution to achieve a user lookup. As this is subject to change, it would be desirable to wrap this API endpoint, so that our users can rely on a stable API and we can change the implementation details.

nirinchev:
## Description

Fixes #1457 

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Platform.PCL/Realm.Sync.PCL/UserPCL.cs
+++ b/Platform.PCL/Realm.Sync.PCL/UserPCL.cs
@@ -214,10 +214,10 @@ namespace Realms.Sync
         /// The exact names of built-in providers can be found in <see cref="Credentials.Provider"/>.
         /// </remarks>
         /// <returns>
-        /// A <see cref="UserLookupResponse"/>, containing information about the User's Identity in Realm's authentication system,
+        /// A <see cref="UserInfo"/>, containing information about the User's Identity in Realm's authentication system,
         /// or <c>null</c> if a user has not been found.
         /// </returns>
-        public Task<UserLookupResponse> LookupUserAsync(string provider, string providerId)
+        public Task<UserInfo> RetrieveInfoForUserAsync(string provider, string providerId)
         {
             RealmPCLHelpers.ThrowProxyShouldNeverBeUsed();
             return null;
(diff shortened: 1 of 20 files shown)


Platform.PCL/Realm.Sync.PCL/UserPCL.cs
```diff
@@ -202,6 +202,27 @@ public Task ChangePasswordAsync(string userId, string newPassword)
             return null;
         }
 
+        /// <summary>
+        /// Looks up user's information by provider id. This is useful when you know the id of a user in a provider's system,
+        /// e.g. on Facebook and want to find the associated Realm user's Id.
+        /// </summary>
+        /// <param name="provider">The provider that the user has signed up with.</param>
+        /// <param name="providerId">The id of the user in the provider's system.</param>
```
kristiandupont: I hate this name. Is it publicly exposed in the other SDK's?

nirinchev: In js it's `provider, username`, in Java, it's `provider, providerId`.

nirinchev: I'm struggling to find a better name, but I'm definitely open for suggestions!

bdash:
Did y'all take a look at what the Java and JavaScript versions of these APIs are like?

nirinchev:
I did, why?

Realm/Realm.Sync/Dtos/UserInfo.cs
```diff
@@ -21,7 +21,7 @@ namespace Realms.Sync
     /// <summary>
     /// The response of a user lookup operation.
```
fealebenpae: The comment could use updating now that the class is renamed.

nirinchev:
CI failures will be resolved by #1489.

PR #1489, opened 2017-07-21
Run ResetForTesting between tests

nirinchev:
## Description

Fixes #

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)
* [ ] Update PCL (if applicable)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -836,6 +836,9 @@ def insideDocker(String imageTag, String dockerfile = null, Closure steps) {
 }
 
 def reportTests(String file) {
-  archive file
-  junit file
+  stash includes: file, name: file
+  nodeWithCleanup('xamarin-mac') {
+    unstash file
+    junit file
+  }
 }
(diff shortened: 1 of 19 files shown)


Jenkinsfile
```diff
@@ -832,4 +833,12 @@ def insideDocker(String imageTag, String dockerfile = null, Closure steps) {
   image.inside() {
     steps()
   }
+}
+
+def reportTests(String file) {
+  stash includes: file, name: file
+  nodeWithCleanup('xamarin-mac') {
```
fealebenpae: I suggest using the `docker` label as we have more nodes with it

Jenkinsfile
```diff
@@ -837,7 +837,7 @@ def insideDocker(String imageTag, String dockerfile = null, Closure steps) {
 
 def reportTests(String file) {
   stash includes: file, name: file
-  nodeWithCleanup('xamarin-mac') {
+  nodeWithCleanup(!'osx') {
```
fealebenpae: I think the expression should be entirely inside the string.

PR #1496, opened 2017-07-26
Build for macOS and Linux with prebuilt sync binaries

fealebenpae:


wrappers/Dockerfile.xenial
```diff
@@ -0,0 +1,25 @@
+FROM ubuntu:xenial
```
nirinchev: I may be missing something, but looks like this is unused?

wrappers/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
nirinchev: Would be nice to update the readme on how to build.

nirinchev:
If you rebase on master, you should get the fixes for the sporadic CI issues

fealebenpae:
> do you think it makes sense to run tests on other linux distros just in case?

We don't do that in JS, but I think it might. I wanted Debian 8 because that's what the Microsoft dotnet Docker images are based on, but we ought to add others as soon as we decide on a subset of distros we'll officially support.

wrappers/Dockerfile.xenial
fealebenpae: It is unused, but I wanted to keep it in case we need it again - it was a pain to set up.

PR #1513, opened 2017-08-09
Minor cleanup

nirinchev:


Tests/Tests.Sync.Shared/SyncTestHelpers.cs
```diff
@@ -57,7 +57,10 @@ public static Task<User> GetUserAsync()
             return new SyncConfiguration(user, serverUri);
         }
 
-        public static Task<User> GetFakeUserAsync(string token = "foo:bar", string scheme = "http") => User.LoginAsync(Credentials.AdminToken(token), new Uri($"{scheme}://some.fake.server:9080"));
+        public static Task<User> GetFakeUserAsync(string token = "foo:bar", string scheme = "http")
```
fealebenpae: We should still pass an identity. Let's add another optional parameter `string identity = null` and in the body `identity = identity ? Guid.NewGuid().ToString();`.

nirinchev: `token` is used as identity

nirinchev: I can probably rename it though - I guess it's not obvious

PR #1574, opened 2017-09-23
Revoke refresh token on Logout

### LINKED ISSUE #1561: Revoke refresh token upon Logout
On Logout, we should revoke the user's refresh token: https://github.com/realm/ros/pull/290

nirinchev:
## Description

Fixes #1561

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Platform.PCL/Realm.Sync.PCL/UserPCL.cs
```diff
@@ -170,6 +170,16 @@ public void LogOut()
         }
 
         /// <summary>
+        /// Logs out the user from the Realm Object Server. Once the Object Server has confirmed the logout the user credentials will be deleted from this device.
+        /// </summary>
+        [Obsolete("Use LogOutAsync instead")]
```
fealebenpae: I think this attribute should be on the `LogOut` method.

PR #1602, opened 2017-10-24
Expose [Explicit] to the world

### LINKED ISSUE #1467: Expose mechanism to prevent adding classes to the default schema
Library authors may wish to opt out of the module initializer weaving to avoid their objects leaking into users' default schema. Additionally, if a library is loaded lazily, it will try to add its classes after the default schema has materialized, throwing an exception.

nirinchev:
## Description

Fixes #1467

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Weaver/RealmWeaver.Fody/ModuleWeaver.Schema.cs
+++ b/Weaver/RealmWeaver.Fody/ModuleWeaver.Schema.cs
@@ -32,7 +32,7 @@ public partial class ModuleWeaver
             return;
         }
 
-        var referencedTypes = GetReferencedTypes().Where(t => t?.CustomAttributes.Any(a => a.AttributeType.Name == "WovenAttribute") == true);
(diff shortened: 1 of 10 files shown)


Weaver/RealmWeaver.Fody/ModuleWeaver.Schema.cs
```diff
@@ -151,4 +159,9 @@ private MethodDefinition FindOrCreateCctor(TypeDefinition moduleClass)
 
         return cctor;
     }
+
+    private static bool ShouldInclude(ICustomAttributeProvider provider)
+    {
+        return provider.CustomAttributes.All(a => a.AttributeType.Name != "ExplicitAttribute");
```
fealebenpae: I think we should add an imported reference for `ExplicitAttribute` and compare against that. Otherwise we risk mistaking this for any other type with the same name but in a different namespace.

Comparing against the type name was part of the original implementation, but we ought to be safer now that this feature is public.

nirinchev: Good catch, and especially with Explicit being somewhat generic. Will update it.

PR #1610, opened 2017-10-27
Expose internal members to server assembly

nirinchev:


Realm/Realm/RealmConfigurationBase.cs
```diff
@@ -118,9 +118,12 @@ public byte[] EncryptionKey
             }
         }
 
-        internal RealmConfigurationBase(string optionalPath)
+        internal RealmConfigurationBase(string optionalPath = null, bool calculatePath = true)
```
fealebenpae: Maybe we should just add a default empty constructor instead?

nirinchev: Much better, yes. Wonder why it didn't occur to me 😔

PR #1614, opened 2017-10-31
Expose mechanisms to control cache by the configuration

### LINKED ISSUE #1531: Expose RealmHandle + Schema constructor
For GN purposes, we'll need an additional Realm constructor that avoids all caches (including the managed state one).

nirinchev:
## Description

Fixes #1531

### DIFF
--- a/Realm/Realm/Native/Configuration.cs
+++ b/Realm/Realm/Native/Configuration.cs
@@ -62,5 +62,7 @@ namespace Realms.Native
 
         internal ShouldCompactCallback should_compact_callback;
         internal IntPtr managed_should_compact_delegate;
+
+        internal bool enable_cache;
     }
 }
--- a/Realm/Realm/Native/Configuration.cs
+++ b/Realm/Realm/Native/Configuration.cs
@@ -63,6 +63,7 @@ namespace Realms.Native
         internal ShouldCompactCallback should_compact_callback;
         internal IntPtr managed_should_compact_delegate;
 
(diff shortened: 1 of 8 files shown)


Realm/Realm/Native/Configuration.cs
```diff
@@ -62,5 +62,7 @@ internal string Path
 
         internal ShouldCompactCallback should_compact_callback;
         internal IntPtr managed_should_compact_delegate;
+
+        internal bool enable_cache;
```
fealebenpae: This needs to be explicitly marshaled as `UnmanagedType.I1`.

Realm/Realm/RealmConfigurationBase.cs
```diff
@@ -45,6 +45,8 @@ public abstract class RealmConfigurationBase
 
         internal bool Dynamic;
 
+        internal virtual bool EnableCache => true;
```
fealebenpae: I think we're better off with making this a simple field, like `Dynamic, not a virtual property.

nirinchev: Then we'll have to set it which is annoying :P

fealebenpae: Like virtual dispatch is much better :P

nirinchev: I don't see it => it doesn't exist ;) Okay, okay... 😩

PR #1616, opened 2017-11-01
Fix the usage of Maps in the Jenkinsfile

emanuelez:
There's a known issue regarding the serialization of Map iterators so this PR works around that.


nirinchev:
I replayed the pipeline with windows tests disabled. If it passes, let's merge and we'll look into why they hang later on. It's not a problem caused by this PR.

PR #1617, opened 2017-11-01
Refactor Realm’s static constructor

fealebenpae:
Extract general native init in NativeCommon and realm-specific init in the SharedRealmHandle static constructor

### DIFF
--- a/Realm/Realm/Native/NativeCommon.cs
+++ b/Realm/Realm/Native/NativeCommon.cs
@@ -56,9 +56,7 @@ namespace Realms
         internal static unsafe void Initialize()
         {
             if (Interlocked.CompareExchange(ref _isInitialized, 1, 0) == 0)
-            {
-                SynchronizationContextEventLoopSignal.Install();
-                
+            {   
                 try
                 {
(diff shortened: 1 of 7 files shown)


Realm/Realm/Native/NativeCommon.cs
```diff
@@ -51,17 +46,16 @@ private static unsafe void DebugLogger(byte* utf8String, IntPtr stringLen)
...
-        public static extern void register_callbacks(NotifyRealmCallback notifyRealmCallback, GetNativeSchemaCallback nativeSchemaCallback);
-
         [DllImport(InteropConfig.DLL_NAME, EntryPoint = "delete_pointer", CallingConvention = CallingConvention.Cdecl)]
         public static extern unsafe void delete_pointer(void* pointer);
 
         [DllImport(InteropConfig.DLL_NAME, EntryPoint = "realm_reset_for_testing", CallingConvention = CallingConvention.Cdecl)]
         public static extern void reset_for_testing();
 
-        public static unsafe void Initialize()
+        private static unsafe void InitializeOnce()
```
nirinchev: Can't we move this to `Initialize`? Don't see much value in keeping them separate.

PR #1623, opened 2017-11-04
Bump version and update dependencies

nirinchev:


nirinchev:
The unit test failure is one that we've had for a while so no cause for concern.

PR #1637, opened 2017-11-15
Expose config.Dynamic

### LINKED ISSUE #778: Make dynamic Realm opening public
On top of the current Schema and other dynamic classes, this would allow a user app to open a completely unknown Realm and browse its contents (unless encrypted).

nirinchev:
Fixes #778 

TODO:
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ x.y.z (TBD)
 ------------------
 
 ### Enhancements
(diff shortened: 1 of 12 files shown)


nirinchev:
There are two cases that I have come across already:
- Writing a .NET browser/studio app
- Compacting a large Realm

I agree that ultimately, exposing a mechanism to provide the schema is what we should do, but until that's done, it's a cheap enhancement to unblock some people.

HideDev:
@nirinchev hi, when it wlll be release?

Zhuinden:
@listepo-alterpost 2.2.0 (2017-03-22)

PR #1679, opened 2018-01-30
Improved partial sync API

nirinchev:
## Description

This is a WIP of the partial sync API. It deviated slightly from the original proposal due to the introduction of a `Subscription` class to match Cocoa's API.

Remaining questions
- Should we expose `Unsubscribe(query)` method? The idea behind it relied on results generating deterministic names. Not sure if that is still the case. We could also expose `Unsubscribe(subscription)` that could be useful for shortly lived queries (e.g. searching).
- Should `Subscribe` be async in OS? I believe it makes sense to make it symmetrical to `Unsubscribe` and to make the API a bit more predictable (as it stands today, a well-timed crash would leave you unsubscribed despite the `Subscribe` call returning).

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


ironage:
FYI, this PR was taking over a day on CI Windows executors so I cancelled the job to free up some CPU.

Realm/Realm.Sync/Subscription.cs
```diff
@@ -0,0 +1,175 @@
...
+                        tcs.TrySetResult(null);
+                        break;
+                    case SubscriptionState.Error:
+                        tcs.TrySetException(Error);
+                        break;
+                }
+            });
+
+            PropertyChanged += handler;
+            await tcs.Task;
```
fealebenpae: If the task is faulted the following two lines will never execute. Maybe `try/finally`?

PR #1717, opened 2018-04-09
Expose string-based queries

nirinchev:
## Description

This exposes the query parser as-is. We can improve with argument support later on. The tests are non-exhaustive by far, but cover the basic operations and assume that between core, js, and OS, the parser has been thoroughly tested.

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)
  * [x] Named backlinks
  * [x] Unnamed backlinks
  * [x] Following lists
  * [x] Following links
  * [x] Notifications
  * [x] LINQ before
  * [x] LINQ after
  * [x] Sort
  * [x] Distinct


wrappers/src/results_cs.cpp
```diff
@@ -118,14 +134,35 @@ REALM_EXPORT ManagedNotificationTokenContext* results_add_notification_callback(
...
 }
     
+REALM_EXPORT Results* results_get_filtered_results(const Results& results, uint16_t* query_buf, size_t query_len, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        Utf16StringAccessor query_string(query_buf, query_len);
+        auto query = results.get_query();
+        auto const &realm = results.get_realm();
+        
+        parser::ParserResult result = parser::parse(query_string);
```
fealebenpae: Shouldn't we check whether the query was successfully parsed?

wrappers/src/results_cs.cpp
```diff
@@ -43,6 +44,21 @@ inline T get(Results* results, size_t ndx, NativeException::Marshallable& ex)
     });
 }
 
+inline void alias_backlinks(parser::KeyPathMapping &mapping, const realm::SharedRealm &realm)
```
ironage: Please use the updated version so that this works across classes, see the fix here: https://github.com/realm/realm-js/pull/1734/files

wrappers/src/results_cs.cpp
nirinchev: I copied this verbatim from the js implementation and my assumption was that it would throw on error which would then get proxied by handle_errors.

ironage: I can confirm it will throw runtime errors if something goes wrong.

PR #1718, opened 2018-04-12
Add missing GCHandle guard for callback method

### LINKED ISSUE #1716: Race conditions in partial sync subscriptions
It looks like there is some race condition in the partial sync subscription logic that manifests itself in peculiar ways.

This subscription doesn't appear to work:

```csharp
var subscription = realm.All<Foo>().Subscribe();
await subscription.WaitForSynchronizationAsync(); // Never resolves
```

Similarly, subscribing twice results in similar behavior:

```csharp
var subscription = realm.All<Foo>().Where(f => f.Bar > 0).Subscribe();
await subscription.WaitForSynchronizationAsync(); // <-- works

var subscription2 = realm.All<Foo>().Where(f => f.Bar > 0).Subscribe();
await subscription2.WaitForSynchronizationAsync(); // <-- never resolves

```

fealebenpae:
Fixes #1716

### DIFF
--- a/Realm/Realm.Sync/Subscription.cs
+++ b/Realm/Realm.Sync/Subscription.cs
@@ -40,7 +40,14 @@ namespace Realms.Sync
     [SuppressMessage("StyleCop.CSharp.MaintainabilityRules", "SA1402:FileMayOnlyContainASingleClass")]
     public static class Subscription
     {
-        internal static readonly SubscriptionHandle.SubscriptionCallbackDelegate SubscriptionCallback = SubscriptionCallbackImpl;
+        internal static readonly SubscriptionHandle.SubscriptionCallbackDelegate SubscriptionCallback;
+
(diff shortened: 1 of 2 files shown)


blimkemann:
I am using 3.2.1 and still having this issue. When I do a bunch (meaning 50+) writes in a tight loop the subscription will hang eventually. It often happens at different spots, but it WILL hang without question. Even NOT calling WaitForSubsriptionAsync will cause the application to hang - it gets a bit slow and then hangs. It's almost as though something internally is getting filled, or confused or something and then the entire process crashes.

nirinchev:
Do you think you can isolate a repro for that?

blimkemann:
I have one. I will open a support ticket and link to the repro. For me, the system hangs on the very first WaitForSubscriptionAsync...

PR #1729, opened 2018-05-04
Make WriteAsync wait for changes asynchronously

### LINKED ISSUE #1728: WriteAsync blocks on Refresh

fealebenpae:
## Description

Fixes #1728

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Realm.cs
+++ b/Realm/Realm/Realm.cs
@@ -695,22 +695,15 @@ namespace Realms
             if (SynchronizationContext.Current != null)
             {
                 var scheduler = TaskScheduler.FromCurrentSynchronizationContext();
-                ulong otherVersion = 0;
                 return Task.Run(() =>
                 {
                     using (var realm = GetInstance(Config))
                     {
                         realm.Write(() => action(realm));
-                        otherVersion = realm.SharedRealmHandle.GetTransactionVersion();
                     }
                 }).ContinueWith(_ =>
                 {
-                    if (otherVersion == SharedRealmHandle.GetTransactionVersion())
-                    {
-                        return Task.CompletedTask;
-                    }
-
-                    return WaitForChangesAsync();
+                    return RefreshAsync();
                 }, scheduler).Unwrap();
             }
             else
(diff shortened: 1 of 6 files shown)


Realm/Realm/Realm.cs
```diff
@@ -695,13 +695,23 @@ public Task WriteAsync(Action<Realm> action)
...
                 {
                     using (var realm = GetInstance(Config))
                     {
                         realm.Write(() => action(realm));
+                        otherVersion = realm.SharedRealmHandle.GetTransactionVersion();
                     }
-                }).ContinueWith(_ => Refresh(), scheduler);
+                }).ContinueWith(_ =>
+                {
+                    if (otherVersion == SharedRealmHandle.GetTransactionVersion())
```
nirinchev: Should this be strict equality? I imagine if the current version is higher (for some reason) than the other version, we should still return?

Realm/Realm/Realm.cs
```diff
@@ -724,6 +734,30 @@ public bool Refresh()
...
+            {
+                Refresh();
+                return Task.CompletedTask;
+            }
+
+            var tcs = new TaskCompletionSource<object>();
+            RealmChangedEventHandler handler = null;
+            handler = (sender, e) =>
+            {
+                ((Realm)sender).RealmChanged -= handler;
```
nirinchev: I'm a bit hazy on the subtleties of doing it like this (via the `sender`) rather than unsubscribing `this.RealmChanged`. Is it to avoid capturing `this` in the closure and what would the harm be if we did capture it?

fealebenpae: Exactly - I didn’t want to capture `this` and create a reference cycle. If we capture it and the wvent never fires then the realm will leak.

Realm/Realm/Realm.cs
```diff
@@ -724,6 +734,30 @@ public bool Refresh()
             return SharedRealmHandle.Refresh();
         }
 
+        /// <summary>
+        /// Waits for the realm and outstanding objects to be updated to a newer version.
+        /// </summary>
+        /// <returns>An awaitable <see cref="Task"/>.</returns>
+        private Task WaitForChangesAsync()
```
nielsenko: I hope you don't mind a comment from an outsider. Could you consider making this public? And why the change in name compared to `Refresh`? I would expect it to be called `RefreshAsync`

nielsenko: Hmm.. I think I get it. You need to compare the current transaction version number with something, to avoid waiting in vane for a change that may never come. Is it not possible to query the realm for the highest transaction version seen?

nirinchev:
@tgoyne I added you as a reviewer to give your opinion on the high-level approach and confirm whether it matches what we discussed, thanks.

nirinchev:
Here are some packages:
https://drive.google.com/file/d/1bQ7U5jBzg4xSvcbkaxNObn1s8JIGvVrV/view?usp=sharing
https://drive.google.com/file/d/19b0mie_oHmQbZb2imje2XJhxUZqmyjEC/view?usp=sharing

Issue #1743, opened 2018-05-21
Unable to use Fody v3 with Realm


Hawxy:
## Goals

Use the latest version of Fody alongside Realm

## Expected Results

Build successfully with Fody v3.

## Actual Results

- Inability to use Fody v3.0.3 due to:
`Fody: The weaver assembly 'RealmWeaver.Fody, Version=3.0.0.0, Culture=neutral, PublicKeyToken=null' references an out of date version of Mono.Cecil.dll. Expected strong name token of '1C-A0-91-87-7D-12-CA-03' but got '50-CE-BF-1C-CE-B9-D0-5E'. The weaver needs to update to at least version 3.0 of FodyHelpers.	
`

From a quick search, this seems to be a repeat of the last time Fody had a major update, and thus it needs to be updated within Realm.

As a side note, there's a _lot_ of warnings produced when you add realm to XF, between version conflicts and osx libraries being referenced that don't exist. Maybe something to investigate.

## Steps to Reproduce
(I'm currently working on a XF v3.0 project)
1. Open Visual Studio 2017 15.7.1
2. Create a new Xamarin.Forms project using .NET Standard with Droid/iOS/UWP targets
3. Add Realm via nuget 
4. Build solution, get various warnings, but build successfully.
5. Add Fody v3.0.3 via nuget
6. Build, get above error.

## Version of Realm and Tooling

- Client SDK Version: 3.0.0
- Client OS & Version: Windows 10 1803

anhhtz:
Downgrade Foddy to 2.5.0 and it works fine.

andrekiba:
There is a way to fix this? Every time there is an update in Fody something breaks. In my project I also have PrepertyChanged.Fody and AsyncErrorHandler.Fody and it is almost impossible to find the correct formula to make them together! What is the correct way? Thank you!

Hawxy:
Looks like this'll be fixed whenever #1744 gets merged

Jyosua:
@andrekiba The latest Realm, 2.2.10 for PropertyChanged.Fody, and 2.5.0 for Fody. Just go to the desired nuget's page on nuget.org and click on each release until you see the dependency list 2.5.0 for Fody instead of 3+.

Issue #1693, opened 2018-02-27
WPF App Build Error


justinrogowski:
Apologies if I have missed a solution somewhere. I have searched through the current issue library and help docs. 

I am experiencing a problem when building a WPF App project in VS2017. I have tried this on Windows 7 and Windows 10. I install VS2017 community edition, create a new 4.6.2 WPF App project and then add Realm from NuGet.

After this the project will not build. I get the following error:

Severity	Code	Description	Project	File	Line	Suppression State
Error		Unknown build error, 'Binary format of the specified custom attribute was invalid.'

I am new to realm and I appreciate that this might be specific to my configuration so any pointers or things to check for would be beneficial.

I have also tried updating all packages that can be updated via NuGet and still experience the error.

Does anyone have any ideas?

## Goals

Build a simple WPF App in VS2017 using Realm from NuGet

## Expected Results

Build Succeeded
## Actual Results

Severity	Code	Description	Project	File	Line	Suppression State
Error		Unknown build error, 'Binary format of the specified custom attribute was invalid.'

C:\Windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.WinFx.targets(268,9): error MC1000: Unknown build error, 'Binary format of the specified custom attribute was invalid.' 

## Steps to Reproduce

1. Open VS2017
2. Create 'WPF App' Project (Target 4.6.2 - Does the same for 4.7 & 4.7.1 etc.)
3. Add Realm from NuGet - wait for install to complete
4. Build project - This is where I get the error

## Version of Realm and Tooling

- Realm Object Server Version: N/A
- Flavor:
  - [x] Developer
  - [ ] Professional
  - [ ] Enterprise
- Server OS & Version: N/A
- Client SDK Version: Realm 2.1
- Client OS & Version: Windows 10 Pro (1709 - 16299.248)

justinrogowski:
Could anyone confirm at all if they are experiencing a similar issue?

If I don't get any feedback to this issue due to time constraints (which I understand), could someone at least let me know when a new release of the dotnet version might be released so I could try with that?

My current workaround:

I am able to test Realm by adding it to a Class Library. I then put all Realm code in that library and reference it in my WPF App. I make sure I do NOT reference Realm.Sync in the WPF App. This allows me to compile and run (I need to make sure I swap out the realm-wrappers.dll with the one from the Class Library or else I can't use the Sync features of course). 

This isn't ideal of course and need a solution that solves the main build error for WPF Apps.

[...]

PR #1744, opened 2018-05-21
Upgrade to new Fody

fealebenpae:
## Description

Update the Fody dependency to 3.0.

Fixes https://github.com/realm/realm-dotnet/issues/1693
Fixes https://github.com/realm/realm-dotnet/issues/1743

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


nirinchev:
Looks like CI is upset, but other than that, it looks solid. We should check how this behaves with the Server assembly - I remember it was somewhat flimsy when it came to Fody versions.

fealebenpae:
I'm updating it precisely because of the Server assembly - I started getting CustomAttributeFormatExceptions there and this fixes them.

PR #1774, opened 2018-09-27
Fix subscription state size

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet-private/issues/128

##  TODO

* [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+3.2.1 (TBD)
+------------------
+
+### Bug fixes
+- Fixed a bug with marshalling subscription state from native objects to .NET which could result in the .NET type reading a broader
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+3.2.1 (TBD)
+------------------
+
+### Bug fixes
+- Fixed a bug with marshalling subscription state from native objects to .NET which could result in the .NET type reading a broader
```
bmunkholm: We can leave out the implementation details.

bmunkholm: Perhaps just:
- When subscribing to queries with (?how) you could get an exception as `An unknown error has occurred. State: *some-number-larger than 127*`. ([dotnet-private#128](https://github.com/realm/realm-dotnet-private/issues/128), since v?)

PR #1788, opened 2018-11-06
Fix memory leak.

aTakayukiMorita:
The instance of Migration class is not freed after dispose the instance of Realm class when MigrationCallback is not called.
It should call GCHandle.Free() method after checking migration even if MigrationCallback is not called.


nirinchev:
Hey, sorry for getting back at you after such a long time - it's embarrassing and we don't have an excuse for it. I'd like to thank you for bringing the issue to our attention - it's absolutely a problem and this PR would have been a great fix at the time. Unfortunately some of the mechanics around lifecycle management of migrations have changed and I implemented an alternative solution in https://github.com/realm/realm-dotnet/pull/2331, which will be released with the next version of the SDK.

I'll close your PR, but I'd like to once again thank you for you contribution and we promise to be a lot more responsive in the future!

PR #1839, opened 2019-03-07
Add session.stop/start methods

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet-private/issues/138

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+3.5.0 (TBD)
+------------------
+
+### Enhancements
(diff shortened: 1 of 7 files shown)


ianpward:
is it possible to get a tarball of the nuget package? doesn't have to be a formal release

fealebenpae:
You can download nuget packages of our nightly builds from https://www.myget.org/gallery/realm-nightly. Just click on each of the three packages and then, all the way down, you can find the download link.

pfedotovsky:
When this will be released officially?

Issue #1758, opened 2018-07-14
ArgumentNullException when using both PropertyChanged.Fody and Realm


tuyen-vuduc:
## Goals
I am trying to use latest PropertyChanged.Foday v2.5.8 and latest Realm v3.1.0 installed.

I am working on Xamarin project with MvvmCross and here is my weavers.xml
```
<?xml version="1.0" encoding="utf-8"?>
<Weavers>
  <PropertyChanged />
  <RealmWeaver/>
</Weavers>
```
## Expected Results
Everything just works.

## Actual Results
```
Value cannot be null.
Parameter name: instruction
System.ArgumentNullException: Value cannot be null.
Parameter name: instruction
   at Mono.Cecil.Cil.InstructionOffset..ctor(Instruction instruction)
   at Mono.Cecil.Cil.CodeReader.ReadScope(ScopeDebugInformation scope)
   at Mono.Cecil.Cil.CodeReader.ReadDebugInfo()
   at Mono.Cecil.Cil.CodeReader.ReadMethodBody()
   at Mono.Cecil.Cil.CodeReader.ReadMethodBody(MethodDefinition method)
   at Mono.Cecil.MethodDefinition.<>c.<get_Body>b__41_0(MethodDefinition method, MetadataReader reader)
   at Mono.Cecil.ModuleDefinition.Read[TItem,TRet](TRet& variable, TItem item, Func`3 read)
   at Mono.Cecil.MethodDefinition.get_Body()
   at Mono.Cecil.MethodDefinition.get_DebugInformation()
   at Mono.Debugging.Soft.SoftDebuggerSession.LoadPdbType(TypeDefinition type, Dictionary`2 fileToSourceFileInfos) in E:\A\_work\91\s\external\debugger-libs\Mono.Debugging.Soft\SoftDebuggerSession.cs:line 2645
   at Mono.Debugging.Soft.SoftDebuggerSession.LoadPdbFile(String assemblyFileName, String pdbFileName) in E:\A\_work\91\s\external\debugger-libs\Mono.Debugging.Soft\SoftDebuggerSession.cs:line 2637
   at Mono.Debugging.Soft.SoftDebuggerSession.LoadDebugFile(String assemblyFileName, String debugFileName, Func`3 loadDebugFile) in E:\A\_work\91\s\external\debugger-libs\Mono.Debugging.Soft\SoftDebuggerSession.cs:line 2733
[...]

PR #1849, opened 2019-04-13
Refactor packaging

fealebenpae:
Fixes https://github.com/realm/realm-dotnet/issues/1860
Fixes https://github.com/realm/realm-dotnet/issues/1758


listepo:
@fealebenpae maybe it makes sense to replace Jenkins with Azure Pipelines?

nirinchev:
Thank you for the suggestion, but the entire Realm organization relies on Jenkins for CI and we do have all the build agents provisioned and shared between different projects. It would be an unreasonable added cost to build just the .NET SDK on Azure rather than reuse the infrastructure we already have.

Do you see some particularly appealing aspect of Azure Pipelines that I'm missing and would justify the additional time and cost to migrate to?

listepo:
@nirinchev ready infrastructure is a distinct advantage. But now I see two advantages:
1. Free for open source projects.
2. The ability of the community to see what problems there are in build and fix them.

About the cost of time is difficult to say. Since I just started learning Azure Pipelines.

PR #1851, opened 2019-04-16
Update test runners

nirinchev:
Updates test runners to NUnit 3.11.0

- [ ] iOS build fails due to linking
- [ ] Android build succeeds but tests fail (looks like it picks up the app.config that's intended for iOS only)
- [ ] UWP build fails


Jenkinsfile
```diff
@@ -233,6 +233,8 @@ stage('Test') {
             sh '''
               adb uninstall io.realm.xamarintests
               adb install io.realm.xamarintests-Signed.apk
+              adb shell pm grant io.realm.xamarintests android.permission.READ_EXTERNAL_STORAGE
```
nirinchev: We're targeting a newer android version which comes with runtime permissions. We can just grant them via adb to avoid a UI prompt.

Issue #1834, opened 2019-02-12
Selected backlink closures for subscriptions


kneth:
See Epic: [realm/engineering#13](https://github.com/realm/engineering/issues/13)
Helps unblock a adapter performance issue.

Guiorgy:
@kneth, link is dead BTW

nirinchev:
It's a link to a private repo, but essentially, the proposal is to enable specifying which backlinks should be included when subscribing to results in query based Realms (currently, backlinks are ignored and only forward links are processed).

PR #1858, opened 2019-04-24
Add support for subscription inclusions

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/1834

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Sync/Subscription.cs
+++ b/Realm/Realm/Sync/Subscription.cs
@@ -20,6 +20,7 @@ using System;
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Realms.Exceptions;
@@ -85,6 +86,11 @@ namespace Realms.Sync
         /// <param name="options">
         /// Options that configure some metadata of the subscription, such as its name or time to live.
         /// </param>
+        /// <param name="includedBacklinks">
+        /// An array of property expressions which specifies which linkingObjects relationships should be included in
+        /// the subscription. Subscriptions already include link and list properties (in the forward direction)
+        /// automatically by default.
+        /// </param>
         /// <returns>
         /// A <see cref="Subscription{T}"/> instance that contains information and methods for monitoring
         /// the state of the subscription.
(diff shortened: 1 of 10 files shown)


Realm/Realm/Sync/Subscription.cs
```diff
@@ -85,6 +86,11 @@ public static Subscription<T> Subscribe<T>(this IQueryable<T> query, string name
         /// <param name="options">
         /// Options that configure some metadata of the subscription, such as its name or time to live.
         /// </param>
+        /// <param name="includedBacklinks">
```
ironage: I think "backlinks" is a term used internally, are .NET users familiar with it in any other context than here? Instead of introducing new terminology, I recommend using something similar to the js binding "includeLinkingObjects" because linkingObject should be something that users know about.

Realm/Realm/Sync/Subscription.cs
```diff
@@ -85,6 +86,11 @@ public static Subscription<T> Subscribe<T>(this IQueryable<T> query, string name
         /// <param name="options">
```
ironage: Why not add the new option as part of the existing options parameter?

wrappers/src/schema_cs.hpp
```diff
@@ -83,6 +88,21 @@ REALM_FORCEINLINE SchemaObject SchemaObject::for_marshalling(const realm::Object
     return ret;
 }
 
-realm::util::Optional<realm::Schema> create_schema(SchemaObject* objects, int objects_length, SchemaProperty* properties);
+util::Optional<Schema> create_schema(SchemaObject* objects, int objects_length, SchemaProperty* properties);
 
+REALM_FORCEINLINE void alias_backlinks(parser::KeyPathMapping &mapping, const SharedRealm &realm)
```
ironage: Hmm maybe I should move this to the object store as a utility function since both js and .NET use it.

Tests/Realm.Tests/Sync/QueryBasedSyncTests.cs
```diff
@@ -455,12 +454,55 @@ public void WaitForSynchronization_OnBackgroundThread_Throws(bool openAsync)
...
+        [TestCase(false)]
+        public void Subscribe_WithInclusions(bool openAsync)
+        {
+            SyncTestHelpers.RunRosTestAsync(async () =>
+            {
+                using (var realm = await GetQueryBasedRealm(openAsync))
+                {
+                    var subscription = realm.All<ObjectB>()
+                                            .Where(b => b.BoolValue)
+                                            .Subscribe(includedBacklinks: b => b.As);
```
ironage: I'm not familiar with this query syntax, but it looks good to me. How does it look if you wanted to included more than one relationship eg: b.As and b.Cs (if that relationship were named)?

Realm/Realm/Sync/Subscription.cs
ironage: Oh I see that "Backlink" is indeed already part of the syntax in this binding.

wrappers/src/schema_cs.hpp
nirinchev: Yeah, I believe it was initially copied from js and it would make sense to move to OS to avoid the duplication. That piece of code is also copied verbatim from the js PR, so probably it would make sense to move to OS as well: https://github.com/realm/realm-dotnet/pull/1858/files#diff-f34349b48d6a286b0ef21c9c05bd390bR40

Tests/Realm.Tests/Sync/QueryBasedSyncTests.cs
nirinchev: It would look like `.Subscribe(options, b => b.As, b => b.Cs)`. If you don't specify options, it's a bit more cumbersome: `.Subscribe(includedBacklinks: new[] { b => b.As, b => b.Cs })`.

Issue #1710, opened 2018-03-31
Realms.Exceptions.RealmFileAccessErrorException: Unable to open a realm at path


dajbych:
## Goals

Reliable realm database creation.

## Expected Results

Realm instance.

## Actual Results

_Realms.Exceptions.RealmFileAccessErrorException: 'Unable to open a realm at path 'C:\Users\Václav\AppData\Local\Packages\7539ef42-e9e8-4e55-8c45-547c7a6ee489_v1yjhd5ekz562\LocalCache\realmtest.realm.management': make_dir() failed: No such file or directory.'_

## Steps to Reproduce

Simple project for bug reproduction: https://1drv.ms/u/s!AlFEHk1e5ZDCoL1NodRClQwLeQdO6A

## Code Sample

```
class RealmEntity : RealmObject {
    [PrimaryKey]
    public long Id { get; set; }
}

readonly static RealmConfiguration config = new RealmConfiguration(Path.Combine(ApplicationData.Current.LocalCacheFolder.Path, "realmtest.realm")) {
    SchemaVersion = 1,
    ObjectClasses = new Type[] { typeof(RealmEntity) },
    ShouldDeleteIfMigrationNeeded = true
};

async void Page_Loaded(object sender, RoutedEventArgs e) {
    using (var realm = await Realm.GetInstanceAsync(config)) {
    }
}
```

## Version of Realm and Tooling

- Client NuGet package version: 2.2.0
- Client OS & Version: Windows 10 version 1709 (build 16299.248)

## Hint

I think (but I cannot be sure) that instance creation fails because of lack of UTF-8 support. The directory path contains non-ascii characters. I can reproduce this issue only on user accounts with non-ascii characters in the profile directory path.

dajbych:
Still reproducible in version 3.0.0 of the client NuGet package.

nirinchev:
You are correct - non-ascii characters in the path are not supported yet on Windows, sorry about that.

dajbych:
Will it be fixed in the next version? What is expected release date of the next version?

nirinchev:
We don't have a timeframe for when this will be addressed.

juhasuni:
[...]

PR #1890, opened 2019-07-29
Update to latest sync/core/OS

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/1710

##  TODO

* [x] Changelog entry
* [ ] Tests (if applicable)


lindahassini:
which versions supports UTF8 characters while createInstance and accessing path with special characters please??

nirinchev:
This is a PR that was merged two years ago. If you do have general inquiries or are facing problems trying to open a Realm instance where the path contains non-ASCII characters, please create a new issue and we'll try to help you there.

Issue #1892, opened 2019-08-07
Bug: Infinite recursion


nielsenko:
`RealmResult<T>.IndexOf(object o)` recurses infinitely.

Test case:

```
public class Dummy : RealmObject
{
    public IList<Dummy> Friends { get; }
}

[Test]
public void IndexOfTest()
{
    using (var realm = OpenRealm())
    {
        var f = new Dummy();
        var d = new Dummy { Friends = { f } };

         var realmList = d.Friends;
         Assert.That(realmList.IndexOf(f), Is.Zero);

         var realmResult = realm.All<Dummy>() as RealmCollectionBase<Dummy>;
         // Assert.That(realmResult.IndexOf(f), Is.Not.Negative); // this causes infinite recursion!
    }
}
```

This happens because `RealmResult<T>` implements `IndexOf(object o)` as:
```
public override int IndexOf(object value) => IndexOf((T)value);
```
but doesn't implement a `IndexOf(T item)`, unlike `RealmList<T>` which also derive from `RealmCollectionBase<T>`.

nielsenko:
I can confirm this is fixed with 4.2.0

PR #1893, opened 2019-08-07
Fix IndexOf for Results

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/1892

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Linq/IRealmCollection.cs
+++ b/Realm/Realm/Linq/IRealmCollection.cs
@@ -117,7 +117,22 @@ namespace Realms
     /// <typeparam name="T">Type of the <see cref="RealmObject"/> which is being returned.</typeparam>
     public interface IRealmCollection<out T> : IReadOnlyList<T>, INotifyCollectionChanged, INotifyPropertyChanged
     {
+        /// <summary>
+        /// Searches for the specified object and returns the zero-based index of the first        /// occurrence within the entire <see cref="IRealmCollection{T}"/>
+        /// </summary>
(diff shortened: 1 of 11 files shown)


Realm/Realm/Linq/IRealmCollection.cs
```diff
@@ -117,6 +117,23 @@ internal Move(int from, int to)
     /// <typeparam name="T">Type of the <see cref="RealmObject"/> which is being returned.</typeparam>
     public interface IRealmCollection<out T> : IReadOnlyList<T>, INotifyCollectionChanged, INotifyPropertyChanged
     {
+        /// <summary>
+        /// Searches for the specified object and returns the zero-based index of the first        /// occurrence within the entire <see cref="IRealmCollection{T}"/>
```
fealebenpae: Line ending whitespace here looks fishy.

Realm/Realm/Linq/RealmResults.cs
```diff
@@ -70,9 +72,23 @@ internal override CollectionHandleBase CreateHandle()
...
+            Argument.NotNull(value, nameof(value));
+
+            if (_argumentType != (PropertyType.Object | PropertyType.Nullable))
+            {
+                throw new NotSupportedException("IndexOf on non-object results is not supported.");
+            }
 
-        public override int IndexOf(object value) => IndexOf((T)value);
+            var obj = Operator.Convert<T, RealmObject>(value);
+            if (!obj.IsManaged)
```
fealebenpae: Shouldn't we rather check instead that the object belongs to the same realm, not that it belongs to a realm?

PR #1912, opened 2019-10-15
Allow required attribute on string arrays

### LINKED ISSUE #1911: Realm/Fody does reports error on primitive array properties marked [Required]
In a class exported to C# using Realm Studio, I have a field which is a primitive array of strings, such as follows:
```
        [Required]
        [MapTo("settings")]
        public IList<string> Settings { get; }
```
When I try to build the code, Realm/Fody errors out as follows:
error : Fody/Realm: User.Settings is marked as [Required] which is only allowed on strings or nullable scalar types, not on System.Collections.Generic.IList`1<System.String>.

Removing [Required] from the generated C# code will eliminate the build error, but the code is unusable as it no longer matches the schema.

Realm/Fody needs to be modified to allow [Required] on primitive arrays.

braudabaugh:
## Description

Fixes #1911 for string arrays only.  

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/Realm/Realm.Fody/Extensions/PropertyDefinitionExtensions.cs
+++ b/Realm/Realm.Fody/Extensions/PropertyDefinitionExtensions.cs
@@ -59,6 +59,11 @@ internal static class PropertyDefinitionExtensions
         return IsIList(property) && ((GenericInstanceType)property.PropertyType).GenericArguments.Single().IsSameAs(elementType);
     }
 
+    internal static bool IsIList(this PropertyDefinition property, System.Type elementType)
+    {
+        return IsIList(property) && ((GenericInstanceType)property.PropertyType).GenericArguments.Single().FullName == elementType.FullName;
+    }
+
     internal static bool IsIQueryable(this PropertyDefinition property)
     {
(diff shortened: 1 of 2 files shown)


nirinchev:
Thanks so much for your contribution - do you mind signing the [CLA](https://docs.google.com/forms/d/1bVp-Wp5nmNFz9Nx-ngTmYBVWVdwTyKj4T0WtfVm0Ozs/viewform?fbzx=4154977190905366979) so that we can merge it? Also, it would be great if you added a simple case that verifies that
1) the project builds and
2) adding `null` to a `[Required] List<string>` will result in an exception

braudabaugh:
Hi @nirinchev, I've signed the CLA.  As far as the test cases, I'm not sure I understand what or how to do what you are asking.

nirinchev:
I created https://github.com/realm/realm-dotnet/pull/1924 based on your change - it adds a few tests to ensure that we don't accidentally break it in the future. Thanks again for your contribution - it's very much appreciated!

nirinchev:
Closing this as #1924 was merged.

PR #1927, opened 2019-11-26
Fix the FodyHelpers dependency

### LINKED ISSUE #1910: Cannot hit breakpoints with Realm 4.2.0 in Xamarin.Android project
## Goals
Use Nuget package Realm 4.2.0 in a Xamarin.Android project

## Expected Results
I can debug the project and hit breakpoints
image

## Actual Results
The app is not debuggable / is not hitting breakpoints
image

## Steps to Reproduce
The problem was introduced with Realm 4.2.0

## Code Sample
Example project: [RealmDebuggerProblem.zip](https://github.com/realm/realm-dotnet/files/3716972/RealmDebuggerProblem.zip)

Likely important:
The "debugging information" dropdown in the project properties jumps back to "Embedded" even though "Portable" or "Full" were selected previously (and that value is still correct in the .csproj file):
image
image

Android Target version: Android 10 / Q
Xamarin.Android: 10.0.3
Vistual Studio 16.3.4

fealebenpae:
## Description
Make the [FodyHelpers](https://www.nuget.org/packages/fodyhelpers) package a private dependency of the Realm.Fody package. This will stop it being included as a transient dependency in user projects, which in turn breaks debugging.

Thanks to @Dids and @SimonCropp for diagnosing the issue.

Fixes #1910

##  TODO

* [x] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Realm.csproj
+++ b/Realm/Realm/Realm.csproj
@@ -10,7 +10,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
-    <PackageReference Include="Fody" Version="6.0.0" PrivateAssets="None" />
+    <PackageReference Include="Fody" Version="6.*" PrivateAssets="None" />
     <PackageReference Include="Nito.AsyncEx.Context" Version="5.0.0" />
     <PackageReference Include="System.Dynamic.Runtime" Version="4.3.0" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" />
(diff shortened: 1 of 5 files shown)


Realm/Realm/Realm.csproj
```diff
@@ -10,7 +10,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
-    <PackageReference Include="Fody" Version="6.0.0" PrivateAssets="None" />
+    <PackageReference Include="Fody" Version="6.*" PrivateAssets="None" />
```
SimonCropp: shouldnt this be `PrivateAssets="All"` as well?

fealebenpae: We've always had Fody as an explicit dependency on the Realm SDK itself as it's impossible to use the SDK without Fody to run the Realm weaver.

SimonCropp: ahhh. makes sense. sorry for the noise

fealebenpae: Not at all - thank you for the input.

PR #1997, opened 2020-08-19
Upgrade to Core 6

nirinchev:
## Description

Upgrades to Core6, but doesn't add frozen object support or support for syncing with MongoDB Realm. These will come later.

##  TODO

* [x] Reintroduce caching in Object Store: https://github.com/realm/realm-object-store/issues/1069
* [x] `Subscribe_WithInclusions_ExtraHop` fails with "No such table exists".
* [x] Notifier tests fail, likely due to caching (the test is written in such a way that the Notifier Realm gets disposed before the objects are actually used - this would be fine with caching, but not without).
* [x] Changelog entry


Realm/Realm/RealmObject.cs
```diff
@@ -518,21 +518,26 @@ public override bool Equals(object obj)
...
+            if (ObjectSchema.Name != robj.ObjectSchema.Name)
             {
                 return false;
             }
 
             // Return true if the fields match.
             // Note that the base class is not invoked because it is
             // System.Object, which defines Equals as reference equality.
-            return ObjectHandle.GetKey() == ((RealmObject)obj).ObjectHandle.GetKey();
+            return ObjectHandle.Equals(((RealmObject)obj).ObjectHandle);
```
nirinchev: @fealebenpae I reverted that from your change - any reason to use `GetKey()` instead of `Equals`?

nirinchev:
The failing test is due to a race condition in the code for partial sync - we're aggressively closing the Realm and terminating sessions, so by the time the subscription callback fires, the Realm is closed and the results are invalidated. It's not a blocker for merging.

Realm/Realm.Fody/Analytics.cs
```diff
@@ -222,8 +218,8 @@ private static void ComputeHostOSNameAndVersion(out string name, out string vers
                     name = "windows";
                     break;
                 default:
-                    name = "osx";  // proved "windows" detected so default to "osx" for now
-                                   //                    name = Environment.OSVersion.Platform.ToString();
+                    name = "osx";  //// proved "windows" detected so default to "osx" for now
```
peppy, line 221: double commenting here looks like an oversight

nirinchev: This is actually on purpose to silence a StyleCop warning - it doesn't it like when single line comments span multiple lines. By doubly commenting it, you tell it that you know what you're doing, so it stops complaining 😄

peppy: wow, that's a weird default rule! might be worth hard-disabling it.

nirinchev: Yeah, I wasn't too worried about it because this code is going to be heavily reworked as part of the MongoDB Realm effort, so did the minimum amount of effort required to suppress the warnings there 😅

PR #1998, opened 2020-08-19
RNET-184: Add support for frozen objects

### LINKED ISSUE #1945: Frozen Objects

nirinchev:
## Description

This introduces support for "frozen objects". These are objects, collections, or Realms that have been pinned at a particular version. This allows moving them between threads, but also means that they are not immutable and will not be updated as changes to the database happen. Trying to open a write transaction on a frozen Realm or subscribe for changes on an object/collection will throw an exception.

Below is a high level overview of the newly added API

```csharp
var liveRealm = Realm.GetInstance();

// Get a frozen Realm from a live one
var frozenRealm = liveRealm.Freeze();

// Check if a Realm is frozen
frozenRealm.IsFrozen; // true

// Frozen Realms can be queried as normal
var frozenFoos = frozenRealm.All<Foo>().Where(f => f.Bar > 5);

// Queries in frozen Realms are frozen themselves
frozenFoos.IsFrozen(); // true

// Writing to a frozen Realm or subscribing for changes is not allowed
// frozenRealm.Write(...) <-- throws
// frozenRealm.RealmChanged += (...) <-- throws

// Get a frozen query from a live one
var frozenFoos = liveRealm.All<Foo>().Where(f => f.Bar > 5).Freeze(); // extension method on IQueryable<T>
frozenFoos.IsFrozen(); // extension method on IQueryable<T> -> true

// Objects in a frozen query are frozen themselves
frozenFoos().First().IsFrozen; // true

// Get a frozen list from a live one
var liveFoo = liveRealm.Find<Foo>("123");
var frozenBars = liveFoo.Bars.Freeze(); // extension method on IList<T>
frozenBars.IsFrozen(); // extension method on IList<T> -> true

// Get a frozen object from a live one
var frozenFoo = liveFoo.Freeze();
frozenFoo.IsFrozen; // true

// RealmObjects can be frozen in place too if you no longer need the live version
liveFoo.FreezeInPlace();
liveFoo.IsFrozen; // true
```

Having multiple frozen Realms at different versions will cause file growth. To control the maximum number of pinned versions, a new property - `MaxNumberOfActiveVersions` is added to all `RealmConfiguration` classes. When set, Realm will keep track of the number of active versions and throw an exception if the limit is exceeded.

Fixes #1945, RNET-184

##  TODO

* [x] Wire up MaxNumberOfActiveVersions
* [x] Changelog entry
* [x] Tests for collections
* [x] Tests for MaxNumberOfActiveVersions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@ vNext(TBD)
 ------------------
 
 ### Enhancements
-* None
+* Added the notion of "frozen objects" - these are objects, queries, lists, or Realms that have been "frozen" at a specific version. This allows you to access the data from any thread, but it will never change. All frozen objects can be accessed and queried as normal, but attempting to mutate them or add change listeners will throw an exception. (Issue [#1945](https://github.com/realm/realm-dotnet/issues/1945))
+  * Added `Realm.Freeze()`, `RealmObject.Freeze()`, `RealmObject.FreezeInPlace()`, `IQueryable<RealmObject>.Freeze()`, `IList<T>.Freeze()`, and `IRealmCollection<T>.Freeze()`. These methods will produce the frozen version of the instance on which they are called.
+  * Added `Realm.IsFrozen`, `RealmObject.IsFrozen`, and `IRealmCollection<T>.IsFrozen`, which returns whether or not the data is frozen.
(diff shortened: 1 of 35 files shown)


nirinchev:
@fealebenpae this is mostly ready for review - I have a few more tests to write, but the core logic should be there.

fealebenpae (approved):

PR #2008, opened 2020-08-31
Update to latest Object Store

fealebenpae:


Realm/Realm/Realm.csproj
```diff
@@ -41,61 +41,61 @@
     <None Include="..\..\wrappers\build\Darwin\$(Configuration)\librealm-wrappers.dylib">
       <Pack>true</Pack>
       <PackagePath>runtimes\osx-x64\native</PackagePath>
-      <Link>wrappers\Darwin\%(Filename).%(Extension)</Link>
+      <Link>wrappers\Darwin\librealm-wrappers..dylib</Link>
```
nirinchev, line 44: These were likely unintentional?

nirinchev (approved):
Neat!

Realm/Realm/Native/SynchronizationContextScheduler.cs
```diff
@@ -0,0 +1,150 @@
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
```
nirinchev, line 3: Ah... I miss the good old days too 🤤

```suggestion
// Copyright 2020 Realm Inc.
```

wrappers/src/scheduler_cs.cpp
```diff
@@ -0,0 +1,102 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
```
nirinchev, line 3:
```suggestion
// Copyright 2020 Realm Inc.
```

Realm/Realm/Native/SynchronizationContextScheduler.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            private volatile bool _isReleased;
+
+            internal Scheduler(SynchronizationContext context)
+            {
+                Debug.Assert(context != null, "The SynchronizationContext Scheduler implementation always needs a SynchronizationContext");
+                Context = context;
+            }
+
+            internal void Post(ContextPostHandler callback, IntPtr user_data)
```
nirinchev, line 61: Isn't this a bit roundabout? The way I'm reading the code, the native schedule invokes this function with a function to be called (the static `handle` function) and an argument another function (`user_data`), that `handle` will invoke? Since `handle` is static, can't we expose it as a C function and always call that instead of passing a pointer to `handle` on every `Post` invocation?

nirinchev (commented):
Some follow-up comments while waiting for the build to complete

PR #2010, opened 2020-09-02
Use junit for test reporting

### LINKED ISSUE #2009: Failing tests don't mark the build unstable
Example: https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-dotnet/detail/PR-2008/9/pipeline

(adding screenshot in case I don't have time to look into it before the build expires)

Screen Shot 2020-09-02 at 00 09 18

nirinchev:
## Description

Fixes #2009 by replacing xunit reporting with junit.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -202,7 +202,7 @@ stage('Test') {
           }
         }
 
-        reportTests 'temp/TestResults.macOS.xml'
+        junit 'temp/TestResults.macOS.xml'
       }
     },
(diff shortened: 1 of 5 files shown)


nirinchev:
Decided to use junit, because xunit uses some global unstable threshold and marks only the first step that crosses it as unstable:

Screen Shot 2020-09-03 at 00 34 12

In fact, all variants have failing tests, but because .NET Core for windows completed first, it's the only one marked unstable.

PR #2014, opened 2020-09-04
Add decimal support

### LINKED ISSUE #1938: Decimal128 data type
See RPM-88 for details.

nirinchev:
## Description

This adds support for persisting and querying decimal properties. The database stores these values as a 128-bit decimal value, but we're also allowing defining the properties as .NET's built-in `decimal` type, even though it's only 96-bit. We'll automatically convert between the 96 and the 128-bit version where appropriate, but keep in mind that you may lose precision or range when using the built-in type.

Example code:

```csharp
public class MyClassWithDecimals : RealmObject
{
    // Declare decimal properties as usual - with an automatic getter and setter
    public decimal VeryPreciseNumber { get; set; }

    // To use the full range of the native decimal, use Decimal128 from the MongoDB.Bson package.
    // The .NET SDK already pulls that as a dependency, so you don't have to NuGet install it.
    public Decimal128 EvenMorePreciseNumber { get; set; }

    // Nullable decimal or Decimal128 are supported as usual
    public decimal? MaybeDecimal { get; set; }
    public Decimal128? MaybeDecimal128 { get; set; }
}

// To store decimal values:
realm.Write(() =>
{
    myObj.VeryPreciseNumber = 1.234567890123456789M;
    myObj.EvenMorePreciseNumber = Decimal128.Parse("987654321.123456789");

    // Decimal128 implicitly converts from decimal, int, long, etc.
    myOtherObj.EvenMorePreciseNumber = 123.456M;
    myThirdObj.EvenMorePreciseNumber = 6;

    // Decimal128 has explicit constructors that take a float or a double
    myFourthObj.EvenMorePreciseNumber = new Decimal128(9.99999);
});

// To query decimal values:

var largerThanFive = realm.All<MyClassWithDecimals>()
                          .Where(o => o.VeryPreciseNumber > 5)
                          .OrderBy(o => o.EvenMorePreciseNumber);

var smallerThanZero = realm.All<MyClassWithDecimals>()
                           .Where(o => o.EvenMorePreciseNumber < 0)
                           .OrderByDescending(o => o.VeryPreciseNumber);
```

Fixes #1938, RNET-177

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/wrappers/src/object_cs.cpp
+++ b/wrappers/src/object_cs.cpp
@@ -27,7 +27,6 @@
 #include <object_accessor.hpp>
 #include <thread_safe_reference.hpp>
 
-
 using namespace realm;
 using namespace realm::binding;
 
@@ -149,6 +148,19 @@ extern "C" {
                 value.value.int_value = result.is_null() ? 0 : to_ticks(result);
                 break;
             }
+            case realm::PropertyType::Decimal: {
+                auto result = object.obj().get<Decimal128>(column_key).raw();
+                value.value.low_bytes = result->w[0];
+                value.value2.high_bytes = result->w[1];
+                break;
+            }
+            case realm::PropertyType::Decimal | realm::PropertyType::Nullable: {
+                auto result = object.obj().get<Decimal128>(column_key);
+                value.has_value = !result.is_null();
+                value.value.low_bytes = result.is_null() ? 0 : result.raw()->w[0];
+                value.value2.high_bytes = result.is_null() ? 0 : result.raw()->w[1];
+                break;
+            }
             default:
                 REALM_UNREACHABLE();
             }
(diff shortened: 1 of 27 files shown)


wrappers/src/object_cs.cpp
```diff
@@ -149,6 +148,22 @@ extern "C" {
                 value.value.int_value = result.is_null() ? 0 : to_ticks(result);
                 break;
             }
+            case realm::PropertyType::Decimal: {
+                auto result = *object.obj().get<Decimal128>(column_key).raw();
+                value.value.low_bytes = result.w[0];
+                value.value2.high_bytes = result.w[1];
```
nirinchev, line 154:
```suggestion
                auto result = object.obj().get<Decimal128>(column_key);
                value.value.low_bytes = result.raw()->w[0];
                value.value2.high_bytes = result.raw()->w[1];
```

wrappers/src/object_cs.cpp
```diff
@@ -149,6 +148,22 @@ extern "C" {
...
+                break;
+            }
+            case realm::PropertyType::Decimal | realm::PropertyType::Nullable: {
+                auto result = object.obj().get<Decimal128>(column_key);
+                value.has_value = !result.is_null();
+                if (value.has_value) {
+                    auto dec = *result.raw();
+                    value.value.low_bytes = dec.w[0];
+                    value.value2.high_bytes = dec.w[1];
```
nirinchev, line 163:
```suggestion
                    value.value.low_bytes = dec.raw()->w[0];
                    value.value2.high_bytes = dec.raw()->w[1];
```

Issue #1937, opened 2020-01-07
Embedded Objects


RealmBot:
See RPM-72 for details.

PR #2017, opened 2020-09-08
Add support for embedded objects

nirinchev:
## Description

Adds support for embedded objects. Embedded objects are objects which are owned by a single parent object, and are deleted when that parent object is deleted or their parent no longer references them. Embedded objects are declared by subclassing `EmbeddedObject` instead of `RealmObject`. Reassigning an embedded object is not allowed and neither is linking to it from multiple parents. Querying for embedded objects directly is also disallowed as they should be viewed as complex structures belonging to their parents as opposed to standalone objects. A trivial example is:

```csharp
public class Address : EmbeddedObject
{
    public string Street { get; set; }

    public string City { get; set; }
}

public class Person : RealmObject
{

    // Address is an embedded object - you reference it as usual
    public Address Address { get; set; }
}

public class Company : RealmObject
{
    public string PhoneNumber { get; set; }

    // Embedded objects can be contained in lists too
    public IList<Address> OfficeAddresses { get; }
}
```

Added new dynamic methods for instantiating embedded objects:
* `Realm.DynamicApi.CreateEmbeddedObjectForProperty` should be used to create an embedded object and assign it to a parent's property. For example:

  ```csharp
  // static API
  var person = new Person();
  person.Address = new Address
  {
      City = "New York"
  };

  // dynamic API
  var dynamicPerson = realm.DynamicApi.CreateObject("Person");
  var address = realm.DynamicApi.CreateEmbeddedObjectForProperty(dynamicPerson, "Address")
  address.City = "New York";
  ```

* `Realm.DynamicApi.AddEmbeddedObjectToList` should be used to create an embedded object and add it to a parent's list property.
* `Realm.DynamicApi.InsertEmbeddedObjectInList` should be used to create an embedded object and insert it in a parent's list property at a specified index.
* `Realm.DynamicApi.SetEmbeddedObjectInList` should be used to create an embedded object and set it at an index in a parent's list property.

  ```csharp
  // static API
  var company = new Company();
  company.OfficeAddresses.Add(new Address
  {
      City = "New York"
  });

  company.OfficeAddresses.Insert(0, new Address
  {
      City = "Palo Alto"
  });

  company.OfficeAddresses[1] = new Address
  {
      City = "New Jersey"
  };

  // dynamic API
  var dynamicCompany = realm.DynamicApi.CreateObject("Company");
  var officeToAdd = realm.DynamicApi.AddEmbeddedObjectToList(dynamicCompany.OfficeAddresses)  ;
  officeToAdd.City = "New York";

  var officeToInsert = realm.DynamicApi.InsertEmbeddedObjectInList(dynamicCompany.  OfficeAddresses, 0);
  officeToInsert.City = "Palo Alto";

  var officeToSet = realm.DynamicApi.SetEmbeddedObjectInList(dynamicCompany.  OfficeAddresses, 1);
  officeToSet.City = "New Jersey";
  ```

Fixes https://github.com/realm/realm-dotnet/issues/1937

## Breaking changes

This moves all the string-based/dynamic API from `Realm` to `Realm.DynamicAPI`. This is necessary because of the increased number of dynamic API which would unnecessarily pollute the interface for strongly typed access which is the primary use case.

##  TODO

* [x] Changelog entry
* [x] Test recursive objects
* [x] Implement lists of embedded objects
* [x] Test lists
* [x] Test dynamic API
* [ ] Wire up EmbeddedObject.Parent (blocked on https://github.com/realm/realm-core/issues/3898)
* [x] Test backlinks
* [x] Read embeddedness from object store schema


Realm/Realm/Attributes/BacklinkAttribute.cs
```diff
@@ -25,7 +25,7 @@ namespace Realms
     /// </summary>
     /// <example>
     /// <code>
-    /// class Dog : RealmObject
+    /// class Dog : RealmObjectB
```
nirinchev, line 28: ??

wrappers/src/list_cs.cpp
```diff
@@ -143,6 +143,13 @@ REALM_EXPORT void list_add_binary(List& list, char* value, size_t value_len, boo
         add(list, BinaryData(), ex);
     }
 }
+
+REALM_EXPORT Object* list_add_embedded(List& list, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        return new Object(list.get_realm(), list.add_embedded());
```
nirinchev, line 150:
```suggestion
        return new Object(list.get_realm(), list.get_object_schema(), list.add_embedded());
```

wrappers/src/list_cs.cpp
```diff
@@ -292,6 +311,18 @@ REALM_EXPORT void list_insert_binary(List& list, size_t list_ndx, char* value, s
...
 
+REALM_EXPORT Object* list_insert_embedded(List& list, size_t list_ndx, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        const size_t count = list.size();
+        if (list_ndx > count) {
+            throw IndexOutOfRangeException("Insert into RealmList", list_ndx, count);
+        }
+
+        return new Object(list.get_realm(), list.insert_embedded(list_ndx));
```
nirinchev, line 322: same

dev-thinks:
@nirinchev The description over here is clearly explaining the usage of IList<T>. I think this need to be transferred to the documentation as well. Thank you.

PR #2082, opened 2020-10-23
Add Write overloads with return value

### LINKED ISSUE #2081: Expose Write and WriteAsync overloads that return a value
Currently `Realm.Write` and `Realm.WriteAsync` accept an Action and don't return anything. It would be nice to expose overloads that take a `Func<T>` and return the value returned by the func:

```csharp
var foo = realm.Write(() =>
{
    return realm.Add(new Foo());
});

var foosCount = realm.Write(() =>
{
    realm.Add(new Foo());
    return realm.All<Foo>().Count();
});
```

Similarly for `WriteAsync`:

```csharp
var foo = await realm.WriteAsync(r =>
{
    return r.Add(new Foo());
});

var foosCount = await realm.WriteAsync(r =>
{
    realm.Add(new Foo());
    return realm.All<Foo>().Count();
});
```

papafe:
## Description

This is a work in progress PR that add new overloads for `Write` and `WriteAsync` methods, so that they can return a value. 

Fixes #2081 

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm/Realm.cs
```diff
@@ -675,6 +675,44 @@ public void Write(Action action)
             transaction.Commit();
```
nirinchev, line 675: Can we implement this `Write` in terms of the new one? I know it's a fairly trivial method, but the less code we have, the less bugs we'll ship 😄

nirinchev (commented):
Looks good - just a few documentation nits from me + a request to deduplicate the code in the existing Write method.

Realm/Realm/Realm.cs
```diff
@@ -675,6 +675,44 @@ public void Write(Action action)
...
+        ///     return realm.Add(new Dog
+        ///     {
+        ///         Name = "Eddie",
+        ///         Age = 5
+        ///     });
+        /// });
+        /// </code>
+        /// </example>
+        /// <param name="function">
+        /// Delegate with one return value to perform inside a <see cref="Transaction"/>, creating, updating or removing objects.
```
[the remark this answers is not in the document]
papafe: > `perform` -> `execute`?

Regarding this and the following comment, I have copied almost verbatim the documentation for the plain `Write`, so I'll correct both of them there too. 
I'm not a fan of the Oxford comma actually, but let's agree on doing it ;)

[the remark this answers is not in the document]
nirinchev: Yeah, I'm guessing the person who wrote the docs for the original `Write` method felt that "perform an action" makes sense and it kinda does, although I still think "execute" is better. But "perform a delegate" is awkward. So yeah, let's change in both places.

papafe:
I have corrected the documentation and modified the old `Write` method to use the new one, and it seems that the tests are still fine.

I think that the test cases specifically for the new function need to be moved to a new test class, as it seems that there is none of the current ones is related to this. At the moment the tests are in the `AddOrUpdateTests` just for convenience.

nirinchev:
The CI test failures are legitimate - the `Compact_ShouldReduceSize` are flaky and you can ignore them, but `Realms.Tests.Database.DynamicRelationshipTests` are legitimate failures and need to be fixed. My hunch is that the binder is mistakenly using the new `Write(Func<T>)` overload instead of the old `Write(Action)` one, so we might need to be more explicit about which overload to use.

papafe:
Yes, it was actually using the new `Write(Func<T>` instead of the old one. This can be easily fixed, but made me realise that this is actually true for all cases in which the input value of `Write` is an expression lambda (that returns something else than void) instead of a statement lambda.

nirinchev (approved):
Nice job!

Realm/Realm/Realm.cs
```diff
@@ -742,6 +782,263 @@ async Task DoWorkAsync()
...
+                            return realm.Write(() => function(realm));
+                        });
+                        var didRefresh = await RefreshAsync();
+                        return result;
+                    }
+
+                    return DoWorkAsync();
+                }
+            }
+            else
```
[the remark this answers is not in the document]
nirinchev: Perhaps also invert the if clause to do an early return and reduce the nesting of the majority of the method:

```csharp
if (! AsyncHelper.HasValidContext)
{
    return Task.FromREsult(Write(() => function(this)));
}

async Task<T> DoWorkAsync()
{
    // ...
}

return DoWork();
```

Realm/Realm/Realm.cs
```diff
@@ -742,6 +782,263 @@ async Task DoWorkAsync()
...
+        /// <remarks>
+        /// Opens a new instance of this Realm on a worker thread and executes <paramref name="function"/> inside a write <see cref="Transaction"/>.
+        /// <see cref="Realm"/>s and <see cref="RealmObject"/>s/<see cref="EmbeddedObject"/>s are thread-affine, so capturing any such objects in
+        /// the <c>action</c> delegate will lead to errors if they're used on the worker thread. Note that it checks the
+        /// <see cref="SynchronizationContext"/> to determine if <c>Current</c> is null, as a test to see if you are on the UI thread
+        /// and will otherwise just call Write without starting a new thread. So if you know you are invoking from a worker thread, just call Write instead.
+        /// </remarks>
+        /// <example>
+        /// <code>
+        /// await realm.WriteAsync(tempRealm =&gt;
```
nirinchev, line 972: Perhaps add `var dogs = await realm.WriteAsync...`

Realm/Realm/Realm.cs
```diff
@@ -742,6 +782,263 @@ async Task DoWorkAsync()
...
+        /// });
+        /// </code>
+        /// <b>Note</b> that inside the action, we use <c>tempRealm</c>.
+        /// </example>
+        /// <param name="function">
+        /// Delegate with return type <see cref="IList{T}"/> to execute inside a <see cref="Transaction"/>, creating, updating, or removing objects.
+        /// </param>
+        /// <typeparam name="T">The type of data in the <see cref="IList"/>.</typeparam>
+        /// <returns>An awaitable <see cref="Task"/> with return type <see cref="IList{T}"/>.</returns>
+        public Task<IList<T>> WriteAsync<T>(Func<Realm, IList<T>> function)
```
nirinchev, line 996: The comments for the `IQueryable` overload apply to this one too.

Tests/Realm.Tests/Database/WriteOverloads.cs
```diff
@@ -0,0 +1,272 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+using System;
+using System.Linq;
+using NUnit.Framework;
+
+namespace Realms.Tests.Database
+{
+    [TestFixture, Preserve(AllMembers = true)]
+    public class WriteOverloads : RealmInstanceTest
```
nirinchev, line 26: Perhaps it's worth adding a test that ensures that an exception is thrown if the user returns a managed object/query/list indirectly, e.g.:

```csharp
var temp = await _realm.WriteAsync(bgRealm =>
{
    var obj = bgRealm.Add(new IntPrimaryKeyWithValueObject());
    return new
    {
        ManagedObject = obj,
        SomeValue = 123
    };
});

Assert.Throws<*not-sure-what-exception*>(() => temp.ManagedObject.Id);
```

nirinchev (changes requested):
Looking solid, but I think there are some areas where we can simplify the code a little.

Tests/Realm.Tests/Realm.Tests.csproj
```diff
@@ -87,4 +87,9 @@
   <ItemGroup>
     <AdditionalFiles Include="..\..\stylecop.json" />
   </ItemGroup>
+  <ItemGroup>
+    <Compile Update="Database\WriteOverloads.cs">
```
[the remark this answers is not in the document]
papafe: I'll remove it

papafe:
- Added new test case for managed objects returned indirectly
- Changed a little bit documentation to make it clearer that methods return a value
- Refactored code to remove unnecessary `else`
- Simplified code of `WriteAsync`

nirinchev (approved):
Nice job!

CHANGELOG.md
```diff
@@ -8,7 +8,8 @@
 * None
 
 ### Enhancements
-* Added an extra compile-time check to detect erroneous List<T> declarations and suggest IList<T> for collection properties in Realm objects (Issue [#2083](https://github.com/realm/realm-dotnet/pull/2083))
+* Added an extra compile-time check to detect erroneous List<T> declarations and suggest IList<T> for collection properties in Realm objects. (Issue [#2083](https://github.com/realm/realm-dotnet/pull/2083))
+* Added overloads for `Realm.Write` and `Realm.WriteAsync` that can return a value. (Issue [#2081 (https://github.com/realm/realm-dotnet/issues/2081)])
```
```suggestion
* Added overloads for `Realm.Write` and `Realm.WriteAsync` that can return a value. (Issue [#2081](https://github.com/realm/realm-dotnet/issues/2081))
```

nirinchev:
The failing tests are flaky, so feel free to ignore them.

Issue #2077, opened 2020-10-16
Weaver should detect List<T> properties and warn that they should be IList<T>


nirinchev:
Collection properties should be declared as `IList<T> MyProp { get; }`, but it's easy to mistake that for `List<T>`. The weaver should detect that and emit a helpful warning.

PR #2083, opened 2020-10-27
Detect List in weaver and suggest IList instead

LaPeste:
## Description

Weaver should detect List<T> properties and warn that they should be IList<T>. [Jira issue](https://jira.mongodb.org/browse/RNET-279)

One test was modified to check if the change did what expected. All tests were Realm.Fondy.Test were executed and successful.

In the one before the last commit I added code to start a discussion. The concept is to chain all the errors in one sweep. If you, Nikola, have time to look at it and think that we could think about it or want to hear what I mean in more detail, just let me know.

Closes https://github.com/realm/realm-dotnet/issues/2077

### DIFF
--- a/Realm/Realm.Fody/ModuleWeaver.cs
+++ b/Realm/Realm.Fody/ModuleWeaver.cs
@@ -497,10 +497,6 @@ Analytics payload
                               new GenericInstanceMethod(_references.RealmObject_GetListValue) { GenericArguments = { elementType } },
                               concreteListConstructor);
         }
-        else if (prop.PropertyType.GetElementType().FullName == "System.Collections.Generic.List`1")
-        {
-            return WeaveResult.Error($"{type.Name}.{prop.Name} is a 'System.Collections.Generic.List`1' but only its interface is supported by Realm. Did you mean IList?");
-        }
         else if (prop.ContainsRealmObject(_references) || prop.ContainsEmbeddedObject(_references))
         {
             // with casting in the _realmObject methods, should just work
@@ -537,6 +533,11 @@ Analytics payload
             }
 
             ReplaceBacklinksGetter(prop, backingField, columnName, elementType);
+        }
+        else if (prop.PropertyType.GetElementType().FullName == "System.Collections.Generic.List`1")
+        {
+            var genericType = ((GenericInstanceType)prop.PropertyType).GenericArguments.Single().Name;
+            return WeaveResult.Error($"{type.Name}.{prop.Name} is declared as List<{genericType}> which is not the correct way to declare to-many relationships in Realm. If you want to persist the collection, use the interface IList<{genericType}>, otherwise annotate the property with the [Ignored] attribute.");
         }
         else if (prop.SetMethod == null)
         {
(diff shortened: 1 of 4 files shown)


Realm/Realm.Fody/ModuleWeaver.cs
```diff
@@ -497,6 +497,10 @@ private WeaveResult WeaveProperty(PropertyDefinition prop, TypeDefinition type,
                               new GenericInstanceMethod(_references.RealmObject_GetListValue) { GenericArguments = { elementType } },
                               concreteListConstructor);
         }
+        else if (prop.PropertyType.GetElementType().FullName == "System.Collections.Generic.List`1")
```
nirinchev, line 500: I would move this check toward the end - perhaps before the `SetMethod == null` check. The motivation is primarily to keep "legit" cases toward the top and erroneous ones toward the bottom.

Realm/Realm.Fody/ModuleWeaver.cs
```diff
@@ -497,6 +497,10 @@ private WeaveResult WeaveProperty(PropertyDefinition prop, TypeDefinition type,
...
+        {
+            return WeaveResult.Error($"{type.Name}.{prop.Name} is a 'System.Collections.Generic.List`1' but only its interface is supported by Realm. Did you mean IList?");
```
nirinchev, line 502: We can reword this error message to be a bit more precise and actionable. I'm thinking of something like:

```
$"{type.Name}.{prop.Name} is declared as List<{*prop.genericargument*}> which is not the correct way to declare to-many relationships in Realm. If you want to persist the collection, use the interface IList<{*prop.genericargument*}, otherwise annotate the property with the [Ignored] attribute."
```

It's a bit more verbose, but it achieves a few things:
1. It makes `System.Collections.Generic.List`1` more readable.
2. Makes the `only its interface is supported by Realm` a bit more precise - `List<T>` implements multiple interfaces and while the most obvious one is `IList<T>`, it's imprecise to call it "its interface".
3. It tells people what to do in case they do want to have a `List<T>` property that is not persisted in Realm.

nirinchev (commented):
I have a few minor suggestions.

Regarding including the extra check for `SetMethod == null` - I'm on the fence about that one. It's obviously a good idea to point out as many mistakes as possible in one go, however it does make the error fairly verbose. I would probably keep it out and let people uncover it the next time they build, but it's your call 😄

Realm/Realm.Fody/ModuleWeaver.cs
LaPeste: What you're saying makes totally sense. Good points.
I'll apply this at the end of the skunworks week.

LaPeste:
> I have a few minor suggestions.
> 
> Regarding including the extra check for `SetMethod == null` - I'm on the fence about that one. It's obviously a good idea to point out as many mistakes as possible in one go, however it does make the error fairly verbose. I would probably keep it out and let people uncover it the next time they build, but it's your call 😄

The only thing holding me back is: either all do the same or no should. Just to keep the code consistent for the readers (us). So that's why I wanted to hear your opinion in person. We'll postpone this to next week.

nirinchev (approved):

nirinchev:
This was manually merged as https://github.com/realm/realm-dotnet/commit/ed321e4a62707d75494015d4fa7b7901fa2d7365.

PR #2098, opened 2020-11-04
Simplified WriteAsync code

### LINKED ISSUE #2088: Check if we streamline WriteAsync
WriteAsync declares a local function to avoid using async/await due to a mono bug that existed years ago. Hopefully, that's fixed, so we can streamline the method.

https://github.com/realm/realm-dotnet/blob/6228b00777793780a24e0348c0d6ad6cc85cc43e/Realm/Realm/Realm.cs#L712-L743

papafe:
## Description
Simplifies code for `WriteAsync` methods, using `async/await`. This was not done in the past due to a mono bug.

https://github.com/realm/realm-dotnet/blob/6228b00777793780a24e0348c0d6ad6cc85cc43e/Realm/Realm/Realm.cs#L712-L743

Fixes #2088

### DIFF
--- a/Realm/Realm/Realm.cs
+++ b/Realm/Realm/Realm.cs
@@ -789,7 +789,6 @@ namespace Realms
         /// <returns>An awaitable <see cref="Task"/> with return type <typeparamref name="T"/>.</returns>
         public async Task<T> WriteAsync<T>(Func<Realm, T> function)
         {
-            // Can't use async/await due to mono inliner bugs
             ThrowIfDisposed();
 
             Argument.NotNull(function, nameof(function));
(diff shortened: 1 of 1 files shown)


Realm/Realm/Realm.cs
```diff
@@ -748,11 +748,11 @@ public T Write<T>(Func<T> function)
         /// Action to execute inside a <see cref="Transaction"/>, creating, updating, or removing objects.
         /// </param>
         /// <returns>An awaitable <see cref="Task"/>.</returns>
-        public Task WriteAsync(Action<Realm> action)
+        public async Task WriteAsync(Action<Realm> action)
         {
             Argument.NotNull(action, nameof(action));
 
-            return WriteAsync(tempRealm =>
+            await WriteAsync(tempRealm =>
```
nirinchev, line 755: I don't think we need to change this. Not that it's that big of a deal but there's no need to allocate a state machine for this method.

nirinchev (approved):
Nice! Glad to see it has been fixed. As you probably guessed, for PRs like this, there's no need for a changelog entry as it doesn't do anything user-facing.

papafe:
Corrected according to comment, the CI results are fine (apart from the usual flaky tests)

nirinchev:
Go for it

PR #2109, opened 2020-11-05
Removed passing pointer

### LINKED ISSUE #2062: Experiment if we can avoid pointer passing primitive values
.NET Core had a bug on macOS/Linux with marshaling our `PrimitiveValue` struct, which is why we're passing it by reference:

https://github.com/realm/realm-dotnet/blob/52607b9b4d4a0aaaf6557e54c5bd8faeeee7278e/Realm/Realm/Handles/ListHandle.cs#L36-L39

Check if that has been addressed and we can start using the struct directly.

papafe:
## Description
Avoid passing `PrimitiveValue` as pointer. It was done like that due to a bug in .NET Core for macOS/Linux.

Fixes #2062

### DIFF
--- a/Realm/Realm/Handles/ListHandle.cs
+++ b/Realm/Realm/Handles/ListHandle.cs
@@ -33,10 +33,8 @@ namespace Realms
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "list_add_object", CallingConvention = CallingConvention.Cdecl)]
             public static extern void add_object(ListHandle listHandle, ObjectHandle objectHandle, out NativeException ex);
 
-            // value is IntPtr rather than PrimitiveValue due to a bug in .NET Core on Linux and Mac
-            // that causes incorrect marshalling of the struct.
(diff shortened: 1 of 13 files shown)


papafe:
I have changed all the places in which there was a comment. It seemed to work fine with the testing locally, and also with the CI. 
I've run the CI again because there was an error that I wasn't sure about if it was flaky (`RefreshAsync_Tests ` that failed with a timeout). Now the CI is failing when building the wrappers though.

nirinchev:
I think you're missing https://github.com/realm/realm-dotnet/blob/ef9e1431a2dc1fa108993b0151f39c70e52e2d9f/Realm/Realm/Handles/SyncUserHandle.cs#L115-L136

papafe:
> I think you're missing
> 
> https://github.com/realm/realm-dotnet/blob/ef9e1431a2dc1fa108993b0151f39c70e52e2d9f/Realm/Realm/Handles/SyncUserHandle.cs#L115-L136

Definitely, just corrected that too. I was looking for other places to correct using part of the comment string, but it seems I used the one that was different.

nirinchev (approved):

PR #2114, opened 2020-11-12
Update README.md with adjusted license info

bmunkholm:
Also add license file.


nirinchev:
There's already a license in `license.txt` - we should either remove it or update that one instead.

bmunkholm:
Haha - yeah I overlooked that one as I was looking for LICENSE :-) Since that's the name in all other repos, I would suggest we use that here as well.

nirinchev (approved):

PR #2115, opened 2020-11-12
Add some primitive benchmarks

nirinchev:
## Description

This adds some very primitive benchmarks and attempts to run them on Jenkins.

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)

### Query Tests

|    Method | ObjectCount |      Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|---------- |------------ |----------:|----------:|----------:|-------:|-------:|------:|----------:|
|     **Count** |          **10** | **254.65 μs** | **166.04 μs** |  **9.101 μs** | **0.7324** | **0.2441** |     **-** |     **11 KB** |
|     Count |         100 | 202.05 μs | 512.95 μs | 28.117 μs | 0.7324 | 0.2441 |     - |     11 KB |
|     Count |        1000 | 292.22 μs | 318.27 μs | 17.445 μs | 0.7324 | 0.2441 |     - |     11 KB |
| **Enumerate** |          **10** |  **35.98 μs** |  **79.75 μs** |  **4.371 μs** | **0.3662** | **0.1831** |     **-** |      **5 KB** |
| Enumerate |         100 |  89.38 μs | 121.22 μs |  6.644 μs | 0.9766 | 0.4883 |     - |     13 KB |
| Enumerate |        1000 | 643.31 μs | 727.21 μs | 39.861 μs | 6.8359 | 2.9297 |     - |     94 KB |

### String Tests

|           Method | StringSize |        Mean |       Error |    StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|----------------- |----------- |------------:|------------:|----------:|-------:|-------:|------:|----------:|
| **GetPropertyValue** |         **20** |    **444.0 ns** |   **544.55 ns** |  **29.85 ns** | **0.0119** |      **-** |     **-** |     **160 B** |
| GetPropertyValue |        100 |  1,480.0 ns |    20.25 ns |   1.11 ns | 0.0229 |      - |     - |     320 B |
| GetPropertyValue |       1000 | 10,975.5 ns |   466.76 ns |  25.58 ns | 0.1526 |      - |     - |   2,120 B |
|       **LookupByPK** |         **20** |  **4,798.8 ns** |   **975.86 ns** |  **53.49 ns** | **0.0381** | **0.0153** |     **-** |     **520 B** |
|       LookupByPK |        100 |  4,954.5 ns |   615.81 ns |  33.75 ns | 0.0381 | 0.0153 |     - |     520 B |
|       LookupByPK |       1000 | 21,369.1 ns | 3,111.51 ns | 170.55 ns | 0.0305 |      - |     - |     520 B |
| **SetPropertyValue** |         **20** |    **592.1 ns** |   **976.76 ns** |  **53.54 ns** |      **-** |      **-** |     **-** |         **-** |
| SetPropertyValue |        100 |  1,334.0 ns |   861.83 ns |  47.24 ns |      - |      - |     - |         - |
| SetPropertyValue |       1000 |  9,922.9 ns |   808.99 ns |  44.34 ns |      - |      - |     - |         - |

### Write Tests

|     Method | objectCount | binarySize | stringLength |      Mean |      Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------- |------------ |----------- |------------- |----------:|-----------:|----------:|------:|------:|------:|----------:|
|     **Binary** |          **10** |        **128** |            **?** |  **1.229 ms** |  **1.5442 ms** | **0.0846 ms** |     **-** |     **-** |     **-** |     **530 B** |
|     Binary |          10 |       1024 |            ? |  1.884 ms |  0.3507 ms | 0.0192 ms |     - |     - |     - |     530 B |
|     Binary |          10 |    1048576 |            ? |  2.040 ms |  1.1299 ms | 0.0619 ms |     - |     - |     - |     532 B |
|     Binary |         100 |        128 |            ? |  1.844 ms |  1.9821 ms | 0.1086 ms |     - |     - |     - |     530 B |
|     Binary |         100 |       1024 |            ? |  1.784 ms |  2.5881 ms | 0.1419 ms |     - |     - |     - |     530 B |
|     Binary |         100 |    1048576 |            ? | 13.214 ms |  9.7873 ms | 0.5365 ms |     - |     - |     - |   1,680 B |
| **EmptyWrite** |           **?** |          **?** |            **?** |  **1.106 ms** |  **0.5453 ms** | **0.0299 ms** |     **-** |     **-** |     **-** |     **251 B** |
|    **Strings** |          **10** |          **?** |           **20** |  **1.803 ms** | **14.5738 ms** | **0.7988 ms** |     **-** |     **-** |     **-** |     **530 B** |
|    Strings |          10 |          ? |          200 |  1.801 ms |  2.9158 ms | 0.1598 ms |     - |     - |     - |     530 B |
|    Strings |         100 |          ? |           20 |  1.799 ms |  1.0604 ms | 0.0581 ms |     - |     - |     - |     530 B |
|    Strings |         100 |          ? |          200 |  1.708 ms |  0.3419 ms | 0.0187 ms |     - |     - |     - |     530 B |


nirinchev:
## Output from CI

### Query Tests

|    Method | ObjectCount |      Mean |       Error |     StdDev |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|---------- |------------ |----------:|------------:|-----------:|--------:|-------:|------:|----------:|
|     **Count** |          **10** | **440.06 μs** | **1,967.65 μs** | **107.853 μs** |  **3.4180** | **1.4648** |     **-** |  **21.93 KB** |
|     Count |         100 | 400.74 μs |   530.28 μs |  29.066 μs |  3.4180 | 1.4648 |     - |  21.93 KB |
|     Count |        1000 | 383.48 μs |   335.86 μs |  18.410 μs |  3.4180 | 1.4648 |     - |  21.93 KB |
| **Enumerate** |          **10** |  **51.22 μs** |    **49.03 μs** |   **2.688 μs** |  **0.9155** | **0.4272** |     **-** |   **6.06 KB** |
| Enumerate |         100 | 172.40 μs |   604.85 μs |  33.154 μs |  2.1973 | 1.0986 |     - |  14.74 KB |
| Enumerate |        1000 | 870.82 μs |   439.02 μs |  24.064 μs | 14.6484 | 6.8359 |     - |  99.17 KB |

### String Tests

|           Method | StringSize |           Mean |         Error |       StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|----------------- |----------- |---------------:|--------------:|-------------:|-------:|-------:|------:|----------:|
| **GetPropertyValue** |         **20** |       **434.5 ns** |      **50.20 ns** |      **2.75 ns** | **0.0257** |      **-** |     **-** |     **168 B** |
| GetPropertyValue |        100 |     1,023.5 ns |      52.64 ns |      2.89 ns | 0.0496 |      - |     - |     328 B |
| GetPropertyValue |       1000 |     7,449.1 ns |     576.61 ns |     31.61 ns | 0.3281 |      - |     - |    2128 B |
|       **LookupByPK** |         **20** |     **6,190.3 ns** |      **90.10 ns** |      **4.94 ns** | **0.0763** | **0.0381** |     **-** |     **552 B** |
|       LookupByPK |        100 |     7,478.5 ns |     144.34 ns |      7.91 ns | 0.0763 | 0.0381 |     - |     552 B |
|       LookupByPK |       1000 |    16,794.7 ns |   2,888.57 ns |    158.33 ns | 0.0305 |      - |     - |     552 B |
| **SetPropertyValue** |         **20** | **1,263,965.4 ns** | **123,006.57 ns** |  **6,742.41 ns** |      **-** |      **-** |     **-** |     **416 B** |
| SetPropertyValue |        100 | 1,514,912.0 ns | 363,904.33 ns | 19,946.82 ns |      - |      - |     - |     416 B |
| SetPropertyValue |       1000 | 1,726,601.6 ns | 739,698.49 ns | 40,545.37 ns |      - |      - |     - |     416 B |

Tests/PerformanceTests/StringTests.cs
```diff
@@ -0,0 +1,294 @@
...
+            for (var i = 0; i < _primaryKeys.Length; i++)
+            {
+                _primaryKeys[i] = _realm.All<StringClass>().ElementAt(_faker.Random.Int(0, ObjectCount - 1)).Value;
+            }
+
+            _newPropertyValues = new string[32];
+            for (var i = 0; i < _newPropertyValues.Length; i++)
+            {
+                var prop = typeof(StringClass).GetProperty($"Value{i}", BindingFlags.Public | BindingFlags.Instance);
+                _newPropertyValues[i] = (string)prop.GetValue(_robject);
```
papafe, line 99: Why do we need to get the property values of the random object? Can't we just create it randomly as before?

papafe:
Overall I think it's a nice idea. In my opinion It'll also be good to move all the performance tests we already have in this new project, as they are not testing much at the moment.

Tests/PerformanceTests/StringTests.cs
nirinchev: Here's how I understand the BenchmarkDotNet's pipeline (note that I only did some cursory reading last night, so completely possible I'm missing something):

1. It calls GlobalSetup
2. It starts calling the benchmark method until it feels it has a fairly confident measurement

Ideally, a benchmark should have no side effects, but that's kinda impossible in our case (particularly in the SetProperty case). This means, that, if we generated random values, the first time the benchmarked method is run, it would set the property from X -> Y, then all the other invocations it would be Y -> Y. Now, BenchmarkDotNet is pretty smart, so it'll do some warmup invocations and also remove outliers, so it's very unlikely that this has any effect on the measurements, but I felt it's a good idea to ensure all invocations do the same thing.

Tests/PerformanceTests/StringTests.cs
```diff
@@ -0,0 +1,294 @@
...
+                    });
+                }
+            });
+
+            _robject = _realm.All<StringClass>().ElementAt(_faker.Random.Int(0, ObjectCount - 1));
+
+            _primaryKeys = new string[32];
+            for (var i = 0; i < _primaryKeys.Length; i++)
+            {
+                _primaryKeys[i] = _realm.All<StringClass>().ElementAt(_faker.Random.Int(0, ObjectCount - 1)).Value;
```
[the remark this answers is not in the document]
nirinchev: Agree - that code can probably be simplified.

Tests/PerformanceTests/StringTests.cs
papafe: Maybe I am missing something, but I don't think there is a difference between what you wrote and just doing
```
            for (var i = 0; i < _newPropertyValues.Length; i++)
            {
                _newPropertyValues[i] =_faker.Random.Utf16String(StringSize, StringSize);
            }
```
As you said the `SeedData` method should be called once before all the iterations start, so the elements of `_newPropertyValues` will be already defined in both cases.

nirinchev: `_newPropertyValues` is just an array of in-memory values and at each benchmark invocation, we're going to call `rObject.ValueX = _newPropertyValues[x]`. The issue is that the first invocation will behave differently than subsequent invocations. Let me try to illustrate:

1. SeedData is called and let's say `rObject.Value1 = "abc"`.
2. `newPropertyValues[1] = _faker.Random...` is called, so let's say `newPropertyValues[1] = "xyz"`.
3. Benchmark.NET starts invoking the `SetPropertyValue` method.
4. First invocation, we have `rObject.Value1 = "xyz"`. This actually changes the value of `rObject.Value1` because it was previously set to `"abc"`.
5. Second invocation, we again have `rObject.Value1 = "xyz"`. This time, however, the value in the database won't be changed because it was already set to `xyz`.
6. Repeat 5. until we get a high confidence measurement.

Now, I don't think that one iteration out of many will skew the results significantly. Additionally, Benchmark.NET is smart about it and will do several warmup iterations, so it will ignore the first invocation where we mutate the underlying data. That being said, I felt it's a good idea to design the test not to rely on this knowledge and instead ensure that all invocations do the right thing - i.e. set `"abc" = "abc"`. Happy to jump on a call and chat about it if my explanation is not understandable enough 😄

LaPeste (approved):

Tests/PerformanceTests/StringTests.cs
papafe: Now I think I got the point. I misunderstood one thing :)

PR #2116, opened 2020-11-12
Ac/add targets dotnet5

### LINKED ISSUE #2113: Add .NET 5 target for tests
While .NET 5 is supposed to be fully backward compatible, we should ensure this is the case.

- [x] Add .NET 5 target in Realm.Tests
- [x] Install .NET 5 on CI nodes
  - [x] Linux
  - [x] macOS
  - [x] Windows
- [x] Run .NET 5 tests on Linux/macOS/Windows

LaPeste:
## Description
PR opened to see if the changes in Jenkinsfile return the expected results: trigger and pass tests for all platforms targeting .Net5

Fixes #2113 

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


nirinchev:
Seems like the unit test failures are due to nunit not supporting .NET 5- https://github.com/nunit/nunit/pull/3666.

Realm/Realm/Dynamic/MetaRealmObject.cs
```diff
@@ -127,15 +100,13 @@ public override DynamicMetaObject BindGetMember(GetMemberBinder binder)
...
-                            getter = GetGetMethod(DummyHandle.GetList<DateTimeOffset>);
-                        }
-
+                        getter = property.Type.IsNullable() ? GetGetMethod(DummyHandle.GetList<DateTimeOffset?>) : GetGetMethod(DummyHandle.GetList<DateTimeOffset>);
+                        break;
+                    case PropertyType.ObjectId:
+                        getter = property.Type.IsNullable() ? GetGetMethod(DummyHandle.GetList<ObjectId?>) : GetGetMethod(DummyHandle.GetList<ObjectId>);
+                        break;
+                    case PropertyType.Decimal:
+                        getter = property.Type.IsNullable() ? GetGetMethod(DummyHandle.GetList<Decimal128?>) : GetGetMethod(DummyHandle.GetList<Decimal128>);
```
nirinchev, line 109: These were forgotten (and apparently untested 🙈 @papafe you should add a case for UUID once this goes in.

Realm/Realm/Dynamic/MetaRealmObject.cs
```diff
@@ -345,11 +279,26 @@ public override DynamicMetaObject BindSetMember(SetMemberBinder binder, DynamicM
...
+            var argumentShouldBeTheSameType = BindingRestrictions.GetExpressionRestriction(
+                Expression.Equal(
+                    Expression.Constant(_metadata.Schema.Name),
+                    Expression.Property(
+                        Expression.Field(
+                            Expression.Field(
+                                self,
+                                RealmObjectMetadataField),
+                            ObjectMetadataSchemaField),
+                        SchemaGetNameProperty)));
```
nirinchev, line 299: This is the real fix for the dynamic tests - it creates an expression restriction that will prevent the runtime from caching the set/get member if the schema name is different (i.e. if the dynamic object represents a different object type).

nirinchev (approved):
Need a changelog entry for the dynamic fix. I'm thinking something like:

```
Fixed a bug that could cause incorrect property values to be read during a migration for apps running on .NET Core 3.0 or newer. The issue manifests itself when different classes have persisted properties with the same name and could result in the wrong property being accessed - e.g. `foo.Name` could return `foo.Bar`. This could only happen when using the dynamic API during a migration and does not affect apps that use the strongly typed API or run on platforms other than .NET Core 3.x/.NET 5.
```

PR #2124, opened 2020-11-20
[Don't merge] Test windows notification deadlock

nirinchev:
## Description

Fixes #

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/wrappers/src/mongo_collection_cs.cpp
+++ b/wrappers/src/mongo_collection_cs.cpp
@@ -24,9 +24,9 @@
 #include "sync/sync_user.hpp"
 #include "sync/sync_session.hpp"
 #include "sync/app.hpp"
-#include "sync/remote_mongo_client.hpp"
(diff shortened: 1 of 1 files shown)


nirinchev:
Fix was tested and verified it works.

peppy:
Still following progress on this one so we can continue to prototype our migration to realm. Just wondering if a release/pre-release containing the fix will be available in the near future. Thanks again for helping make a fix happen.

nirinchev:
Hi @peppy, we don't have a release containing the fix yet, but hope to do a .beta-3 one fairly soon. A fair warning though - there's still an outstanding issue with notifications when processes exit. We're actively working on it and will ping you whenever we have a PR that resolves it. Even so, it will be super helpful to get feedback of the current state of affairs and I'll let you know when we get beta-3 out.

PR #2129, opened 2020-11-24
Moved benchmarking project go Github Actions

### LINKED ISSUE #2119: Move benchmarks to Github actions
There's a nice benchmark runner action - https://github.com/rhysd/github-action-benchmark - that can generate graphs, report performance regressions, etc. .NET is not supported, but there's a PR and a fork that adds support here: https://github.com/rhysd/github-action-benchmark/issues/43.

papafe:
## Description
Moved benchmarking project to Github Actions. The workflow takes care of building the wrappers, and cache the results to speed up subsequent runs. In case the wrappers have been modified, those are built again.
For now the benchmarks run only on Windows, and a series of graphs is shown on Github Pages. 

Fixes #2119


papafe:
As discussed with @nirinchev, it would also be possible to make it run on PR, but there is no way of constraining it on the PR labels. We can still do this check inside the workflow, but the workflow will run nevertheless, and we can make it fail, or in the best case cancel.

.github/workflows/benchmark-test.yml
```diff
@@ -0,0 +1,82 @@
...
+          name: benchmark-results
+          # Could be put in an environmental variable
+          path: .\BenchmarkDotNet.Artifacts\results
+      # Download previous benchmark result from cache (if exists)    
+      - name: Store benchmark results
+        uses: Happypig375/github-action-benchmark@v1.8.2
+        with:
+          tool: 'benchmarkdotnet'
+          # Where the output from the benchmark tool is stored
+          output-file-path: .\BenchmarkDotNet.Artifacts\results\PerformanceTests.QueryTests-report-full-compressed.json
```
nirinchev, line 77: Should this include the other tests as well?

Tests/PerformanceTests/Program.cs
```diff
@@ -49,33 +41,9 @@ public static void Main(string[] args)
                 .WithArtifactsPath(defaultConfig.ArtifactsPath)
                 .AddDiagnoser(MemoryDiagnoser.Default)
                 .AddJob(Job.ShortRun.WithToolchain(InProcessEmitToolchain.Instance))
```
nirinchev, line 43: TODO: check if these parameters make sense - I used them to get quick results locally, but maybe on CI we'd want a default/long job. The toolchain is also something I used as a bit of a workaround and we may want to adjust later. The problem that I hit was that it looked like the weaver is not running for the generated assembly. I'm guessing there are ways around that, but I didn't feel investigating is justified. We can keep it for now and revisit in another PR.

.github/workflows/benchmark-test.yml
papafe: Yes it should, I was testing on a subset and I forgot :)

papafe:
@nirinchev Regarding making the graphs for the different test sets, it's not possible at the moment with that action. Only one file can be specified as the benchmark output, and if I try to run that step multiple times with different `output-file-path` it does not work. 
So probably our best bet would be forking the action ourselves and do what we need to do. This will also allow us to just specify the output folder, and not all the single output files.

nirinchev:
Wouldn't using `--join` work for that? It'll produce a single file for the results, but I'm not sure if the action will like it.

papafe:
> Wouldn't using `--join` work for that? It'll produce a single file for the results, but I'm not sure if the action will like it.

You're correct, I didn't find it in the documentation for `BenchmarkDotNet`, so I thought it wasn't possible to merge the output files.

nirinchev (approved):
Looks good; we should remove the jenkins job at some point, but let's keep it for now just to be able to easily compare results.

papafe:
@nirinchev regarding the Jenkins job, I removed the exporter for that from the benchmark. Do you think I should put it back to keep both?

nirinchev:
Hm, good point - I guess that it will now fail, so let's remove it I guess.

wrappers/build.ps1
```diff
@@ -22,6 +22,8 @@ param(
     [Parameter(Position=0)]
     [string]$Target = 'Windows',
 
+    [string]$Toolchain = 'c:\\src\\vcpkg\\scripts\\buildsystems\\vcpkg.cmake'
```
nirinchev, line 25:
```suggestion
    [string]$Toolchain = 'c:\\src\\vcpkg\\scripts\\buildsystems\\vcpkg.cmake',
```

PR #2139, opened 2020-12-02
Trigger weaving on editor plugin load

nirinchev:
## Description

* Fixes first part of #2136
* Fixes the casing of info.plist
* Adds [Preserve] to UnityUtils to prevent stripping

### DIFF
--- a/Tools/SetupUnityPackage/SetupUnityPackage/Program.cs
+++ b/Tools/SetupUnityPackage/SetupUnityPackage/Program.cs
@@ -50,7 +50,7 @@ namespace SetupUnityPackage
                 { "lib/netstandard2.0/Realm.dll", "Realm.dll" },
                 { "native/ios/Realm.dll.config", "Realm.dll.config" },
                 { "native/ios/universal/realm-wrappers.framework/realm-wrappers", "iOS/realm-wrappers.framework/realm-wrappers" },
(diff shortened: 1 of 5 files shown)


nirinchev:
Need to downgrade System.Buffers to 4.4.0 because that's the version where lib and ref assemblies have the same version. Ref: https://github.com/dotnet/runtime/issues/1830.

Tools/SetupUnityPackage/SetupUnityPackage/Program.cs
```diff
@@ -50,7 +50,7 @@ public static class Program
                 { "lib/netstandard2.0/Realm.dll", "Realm.dll" },
                 { "native/ios/Realm.dll.config", "Realm.dll.config" },
                 { "native/ios/universal/realm-wrappers.framework/realm-wrappers", "iOS/realm-wrappers.framework/realm-wrappers" },
-                { "native/ios/universal/realm-wrappers.framework/Info.plist", "iOS/realm-wrappers.framework/info.plist" },
+                { "native/ios/universal/realm-wrappers.framework/Info.plist", "iOS/realm-wrappers.framework/Info.plist" },
```
LaPeste, line 53: Oh wow.... I assume that iOS is case sensitive when looking for this. Has iOS always disregarded this? Can the OS make one on the spot if it's missing?

nirinchev: It can't - I did get errors about a missing `Info.plist` from the framework.

PR #2149, opened 2020-12-05
Replace the expressions based operator with T4

### LINKED ISSUE #2148: Replace Operator with T4 generated code
IL2CPP doesn't compile lambdas, so we need to replace Operator with something else. Since Convert.ChangeType is significantly slower and direct casting doesn't work, the best I can think of is to generate explicit conversions for all supported types.

nirinchev:
## Description

Fixes #2148

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


papafe (approved):
Looks good to me for what I can understand! 
I also learned what is a T4 Text Template 👍

LaPeste (approved):

PR #2150, opened 2020-12-07
Added RuntimeInitializeOnLoad attribute

### LINKED ISSUE #2144: [Unity] Replace module initializer with [RuntimeInitializeOnLoad]
Unity doesn't appear to respect the module initializer, so we should instead annotate the RealmInitializer with https://docs.unity3d.com/ScriptReference/RuntimeInitializeOnLoadMethodAttribute.html.

LaPeste:
## Description
Added `RuntimeInitializeOnLoadMethod` on `RealmModuleInitializer` when is in UnityEngine with backend il2cpp.

Fixes #2144

##  TODO

* [ ] Changelog entry      -->   (should I? I'm not sure there's anything to really add)

### DIFF
--- a/Realm/Realm.Unity/Editor/Common/RealmWeaver.Schema.cs
+++ b/Realm/Realm.Unity/Editor/Common/RealmWeaver.Schema.cs
@@ -121,10 +121,8 @@ namespace RealmWeaver
 
             if (!(_references.RuntimeInitializeOnLoadAttribute_Unity is null))
             {
-                _logger.Warning("*** unityEngine wasn't null");
                 initialize.CustomAttributes.Add(new CustomAttribute(_references.RuntimeInitializeOnLoadAttribute_Unity_Constructor));
             }
-            else _logger.Warning("*** unityEngine WAS NULL!!!");
 
             initialize.Body.Instructions.Add(Instruction.Create(OpCodes.Ret));
             initializerType.Methods.Add(initialize);
--- a/Realm/Realm.Unity/Editor/Common/RealmWeaver.Schema.cs
+++ b/Realm/Realm.Unity/Editor/Common/RealmWeaver.Schema.cs
@@ -119,6 +119,14 @@ namespace RealmWeaver
                 HasThis = false
             };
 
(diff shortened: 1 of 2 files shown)


Realm/Realm.Unity/Editor/Common/RealmWeaver.Schema.cs
```diff
@@ -119,6 +119,14 @@ private MethodDefinition GetModuleInitializer()
                 HasThis = false
             };
 
+#if UNITY_IOS || ENABLE_IL2CPP
+            var isUnityEngineAvailable = _moduleDefinition.AssemblyReferences.Any(a => a.Name.Contains("UnityEngine"));
+            if (isUnityEngineAvailable)
+            {
+                var attributeConstructor = _moduleDefinition.ImportReference(typeof(UnityEngine.RuntimeInitializeOnLoadMethodAttribute).GetConstructor(Type.EmptyTypes));
```
nirinchev, line 126: We don't want to assume that the build pipeline has a dependency on UnityEngine. Instead, we should construct the type and the constructor reference by hand, similar to what we're doing for Fody.PropertyChanged:

https://github.com/realm/realm-dotnet/blob/4f900bdda9bc1ebbdb5f8ae9088895532e1c30fc/Realm/Realm.Unity/Editor/Common/ImportedReferences.cs#L372-L373

Realm/Realm.Unity/Editor/Common/RealmWeaver.Schema.cs
```diff
@@ -119,6 +119,14 @@ private MethodDefinition GetModuleInitializer()
                 HasThis = false
             };
 
+#if UNITY_IOS || ENABLE_IL2CPP
```
nirinchev, line 122: Instead of doing preprocessor directives, we should always check for a reference to UnityEngine, but that check should move to ImportedReferences.

nirinchev (commented):
A few code style comments, otherwise looks good

nirinchev (approved):
No need for changelog entry as Unity support is not something we officially have.

PR #2151, opened 2020-12-07
Added attr to avoid dll stripping

### LINKED ISSUE #2145: [Unity] Ensure that UnityUtils gets included in the built binary
Since the methods in UnityUtils only get invoked via reflection, it appears Unity doesn't copy that to the final binary. We should probably weave a dummy call in there to force it to include and then rely on [Preserve] to protect the methods.

LaPeste:
## Description
The addition of the attribute you, @nirinchev, suggested seems to work.
I say "seems" because the only way I had to test if the change was making a difference was to check the temp/staging area during the building process. It was not possible to use breakpoints because the debugging session was being stopped by unity regularly at the important point.
Anyway, what I could observe was that with the attribute added, the `UnityUtils.dll` was kept in the staging area until the very last step (the one that cleans away all the dll but leaving some pdb but not all). Instead, Without such attribute the dll was initially copied in the staging area but delete almost immediately.
This is not the most scientific way of testing this, but my expectation is that when we've merged my changes for adding the attribute to the module initializer and your changes for fixing the issue on iOS and UWP we should have a clearer indication of how much this change helps.

If you can think of better ways to test this, please advice.

Fixes #2145

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
+++ b/Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
@@ -19,9 +19,6 @@
 
 using System;
 using UnityEngine;
-using UnityEngine.Scripting;
-
-[assembly: AlwaysLinkAssembly]
 
 namespace UnityUtils
 {
--- a/Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
+++ b/Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
@@ -19,6 +19,9 @@
 
 using System;
 using UnityEngine;
+using UnityEngine.Scripting;
+
+[assembly: AlwaysLinkAssembly]
 
 namespace UnityUtils
 {
--- a/Tools/SetupUnityPackage/UnityUtils/Properties/AssemblyInfo.cs
+++ b/Tools/SetupUnityPackage/UnityUtils/Properties/AssemblyInfo.cs
@@ -1,32 +1,25 @@
-﻿using System.Reflection;
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2020 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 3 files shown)


Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
```diff
@@ -19,6 +19,9 @@
 
 using System;
 using UnityEngine;
+using UnityEngine.Scripting;
+
+[assembly: AlwaysLinkAssembly]
```
nirinchev, line 24: Let's put this in `AssemblyInfo.cs` - while I agree that it's a bit silly to add yet another file for only one attribute, we should follow best practices as much as possible.

nirinchev (approved):
Can confirm that this did the trick for UWP. Let's just move the attribute to the right file, then it's ready to merge.

Tools/SetupUnityPackage/UnityUtils/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+// Change them to the values specific to your project.
+
+[assembly: AssemblyTitle("UnityUtils")]
+[assembly: AssemblyDescription("Utils to support the Realm SDK in Unity projects")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("Realm")]
+[assembly: AssemblyProduct("")]
+[assembly: AssemblyCopyright("")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
```
nirinchev, line 15: These will fail the build because assembly information is automatically generated on build. It's a good idea to add them, but they should be specified in the csproj instead.

Tools/SetupUnityPackage/UnityUtils/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Reflection;
```
nirinchev, line 1: Missing license header.

nirinchev (changes requested):
Let's clean up the AssemblyInfo.cs file to remove everything except the unity attribute - the other attributes should be generated automatically on build.

nirinchev (approved):

PR #2152, opened 2020-12-07
Unity fixes for iOS

### LINKED ISSUE #2136: [Unity] Weaver doesn't trigger in all cases
It appears that the way we hook the weaver in the editor plugin leaves cases where the weaver isn't run, even though the user code is compiled. For now the ones we've discovered are:

* [x] When the user loads the project for the first time. My guess is that this is related to the editor plugin getting compiled at the same time/after the user code is compiled, which means that by the time it subscribes to the `assemblyCompilationFinished` event, the user code has already been compiled. Perhaps we can trigger user code recompilation on initialization.
* [x] When building for iOS. When generating the XCode project, it appears that IL2CPP is invoked on non-woven code.

nirinchev:
## Description

* Add a build callback to weave assemblies
* Replace realm-wrappers with __Internal for dll imports on iOS

Fixes #2136

### DIFF
--- a/Realm/Realm.Unity/Editor/UnityWeaver.cs
+++ b/Realm/Realm.Unity/Editor/UnityWeaver.cs
@@ -21,14 +21,10 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
-using System.Text;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
-using Mono.Cecil;
 using Mono.Cecil.Cil;
-using Mono.Cecil.Pdb;
 using UnityEditor;
 using UnityEditor.Build;
 using UnityEditor.Build.Reporting;
@@ -138,7 +134,18 @@ namespace RealmWeaver
 
         public void OnPostBuildPlayerScriptDLLs(BuildReport report)
         {
-            var referencePaths = report.files.Select(f => f.path).ToArray();
+            // This is a bit hacky - we need actual references, not directories, containing references, so we pass folder/dummy.dll
+            // knowing that dummy.dll will be stripped.
+            var systemAssemblies = CompilationPipeline.GetSystemAssemblyDirectories(ApiCompatibilityLevel.NET_Standard_2_0).Select(d => Path.Combine(d, "dummy.dll"));
+            var referencePaths = systemAssemblies
+                .Concat(report.files.Select(f => f.path))
+                .ToArray();
+
+            var assembliesToWeave = report.files.Where(f => f.role == "ManagedLibrary");
+            foreach (var file in assembliesToWeave)
+            {
+                WeaveAssemblyCore(file.path, referencePaths);
+            }
 
             if (report.summary.platform == BuildTarget.iOS)
             {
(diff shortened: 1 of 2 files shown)


Realm/Realm.Unity/Editor/UnityWeaver.cs
```diff
@@ -103,35 +105,64 @@ private static async Task WeaveAssembly(string assemblyPath)
...
+            var assembliesToWeave = report.files.Where(f => f.role == "ManagedLibrary");
+            foreach (var file in assembliesToWeave)
+            {
+                WeaveAssemblyCore(file.path, referencePaths);
+            }
+
+            if (report.summary.platform == BuildTarget.iOS)
+            {
+                var realmAssemblyPath = report.files
+                    .SingleOrDefault(r => "Realm.dll".Equals(Path.GetFileName(r.path), StringComparison.OrdinalIgnoreCase))
```
LaPeste, line 153: Out of curiosity, did you do this for performance optimization or did you actually thought about issues in paths from locale with non-english characters (even danish vowels for example)?

LaPeste (approved):
So in the end of the day the bin was not found because iOS links plugins statically into the executable and the way to refer to those is by referring to it as `__Internal`.
Good to know.

Realm/Realm.Unity/Editor/UnityWeaver.cs
```diff
@@ -103,35 +105,64 @@ private static async Task WeaveAssembly(string assemblyPath)
...
+            }
+        }
+
+        public void OnPostBuildPlayerScriptDLLs(BuildReport report)
+        {
+            // This is a bit hacky - we need actual references, not directories, containing references, so we pass folder/dummy.dll
+            // knowing that dummy.dll will be stripped.
+            var systemAssemblies = CompilationPipeline.GetSystemAssemblyDirectories(ApiCompatibilityLevel.NET_Standard_2_0).Select(d => Path.Combine(d, "dummy.dll"));
+            var referencePaths = systemAssemblies
+                .Concat(report.files.Select(f => f.path))
```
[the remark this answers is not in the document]
nirinchev: Yeah - that was a bit tricky to express in a readable way. Essentially, what happens is that we have two collections of paths:
1. System assembly directories - it contains all the **folders** where we have system assemblies
2. report.files.path - it contains the paths of all **files** that were compiled

These two are not at the same level - one is a folder, containing files, the other is a file. There were two approaches I could take to bring them to represent the same thing:
1. Get the directory containing the files in report.files.path - essentially doing something like `report.files.Select(f => Path.GetDirectoryName(f.path))`. This would give us the parent folder for each file (duplicates are fine).
2. Add a dummy file to the system directories, to make them look like filepaths.

I chose option 2 because it worked better with the weaver invocation by the CompilationPipeline flow and reduced the amount of code needed.

Realm/Realm.Unity/Editor/UnityWeaver.cs
nirinchev: It was mostly because I was overly cautious about potential casing differences. I would be extremely surprised if it ever happens, but wanted to cover a case where the build renames Realm.dll to realm.dll. Again, I don't think it will ever happen and we're only looking at the filename, which shouldn't include non-ASCII characters, so it's probably not needed. That being said, it satisfies my OCD, so I'd keep it 😄

PR #2163, opened 2020-12-10
Prepare for 10.0.0-beta.3

nirinchev:
## Description

Release 10.0.0-beta.3 with sync fixes + new google credentials method.

##  TODO

* [ ] Revert the changes to disable Guid/ISet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## vNext (TBD)
+## 10.0.0-beta.3 (2020-12-10)
 ------------------
 
 ### Breaking Changes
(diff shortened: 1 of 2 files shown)


nirinchev:
@peppy 10.0.0-beta.3 has been uploaded to NuGet. It includes a fix for the multiprocess notifications issue on Windows, but it doesn't  fix the outstanding problem where notifications may stop getting delivered when a process accessing the Realm and listening for notifications gets terminated.

If you have a situation where multiple processes are going to access the same Realm and they are not terminated at the same time (e.g. process X outlives process Y), the clunky workaround is to handle process termination (if possible) and dispose all open Realm instances. We do have a POC of a proper fix for that issue and hope to get it polished and released later this year or early next one.

peppy:
Thanks a lot! Just to clarify, that remaining issue only applies to separate processes, not accessing the same realm from multiple threads in the same process?

nirinchev:
That is correct - as long as the thread listening for notifications exits cleanly, I don't believe this will manifest itself in a multithreaded scenario in the same process. That being said, if you encounter issues, do let us know and we'll do our best to track it down.

PR #2167, opened 2020-12-15
Ac/final round realm in unity

### LINKED ISSUE #2105: [Unity] Smoke test remaining platforms
We should smoke test that the following platforms work and that the native dlls are copied in the correct locations:

* [x] UWP
* [ ] Linux
* [x] Android
* [x] iOS (also ensure that Realm.dll.config is respected)

LaPeste:
## Description
General round of tests on each platform in both IL2CPP and Mono when both backends are available.

@nirinchev :
Currently `[RuntimeInitializeOnLoadMethod]` is needed on the method `GetInternalStorage`.
If that's not in place when the android application runs an exception with the following stack trace is thrown:
```
E/Unity: Exception: * it's gone *
      at Realms.InteropConfig+<>c.<TryGetUnityFolder>b__8_0 () [0x00000] in <00000000000000000000000000000000>:0 
      at Realms.InteropConfig.TryGetDatabaseFolder (System.Func`1[TResult] getter, System.String& folder) [0x00000] in <00000000000000000000000000000000>:0 
      at Realms.InteropConfig.TryGetUnityFolder (System.String& folder) [0x00000] in <00000000000000000000000000000000>:0 
      at Realms.InteropConfig+<>c.<.cctor>b__12_0 () [0x00000] in <00000000000000000000000000000000>:0 
      at System.Lazy`1[T].CreateValue () [0x00000] in <00000000000000000000000000000000>:0 
      at System.Lazy`1[T].LazyInitValue () [0x00000] in <00000000000000000000000000000000>:0 
      at System.Lazy`1[T].get_Value () [0x00000] in <00000000000000000000000000000000>:0 
      at Realms.InteropConfig.get_DefaultStorageFolder () [0x00000] in <00000000000000000000000000000000>:0 
      at Realms.RealmConfigurationBase.GetPathToRealm (System.String optionalPath) [0x00000] in <00000000000000000000000000000000>:0 
```

The problem is in line 80 of of `InteropConfig.cs` where `getInternalStorage` is null because the method isn't found if `[RuntimeInitializeOnLoadMethod]` isn't in place.

IMPORTANT: In order for the exception of line 80 to be thrown the method `TryGetDatabaseFolder` needs to be modified not to skip exceptions ( = comment out the empty try-catch).

Fixes #2105

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


Realm/Realm.Unity/Editor/Common/ImportedReferences.cs
```diff
@@ -143,7 +143,7 @@ internal abstract class ImportedReferences
 
         public TypeReference RealmSchema_PropertyType { get; private set; }
 
-        public MethodReference UnityEngine_RuntimeInitializeOnLoadAttribute_Constructor { get; private set; }
+        public MethodReference UnityEngine_PreserveAttribute_Constructor { get; private set; }
```
nirinchev, line 146: You don't need the unity preserve attribute. You can use [PreserveAttribute_Constructor](https://github.com/realm/realm-dotnet/blob/aff5bb2bd02e029cc847c0d16f9bfef47f713802/Realm/Realm.Unity/Editor/Common/ImportedReferences.cs#L120).

Realm/Realm/Schema/RealmSchema.cs
```diff
@@ -150,6 +152,35 @@ internal static RealmSchema CreateSchemaForClasses(IEnumerable<Type> classes, Re
...
+                    foreach (var a in allAssemblies)
+                    {
+                        var initializer = a.GetType("RealmModuleInitializer");
+                        if (initializer != null)
+                        {
+                            moduleInitializers.Add(initializer);
+                        }
+                    }
+
+                    moduleInitializers.ForEach(m => m.GetMethod("Initialize").Invoke(null, null));
```
nirinchev, line 174: This can be mostly rewritten with LINQ.

Realm/Realm/Schema/RealmSchema.cs
```diff
@@ -150,6 +152,35 @@ internal static RealmSchema CreateSchemaForClasses(IEnumerable<Type> classes, Re
             return builder.Build();
         }
 
+        internal static RealmSchema GetSchema(HashSet<Type> types)
```
nirinchev, line 155: No need to pass `types` here - this function can operate on `_defaultTypes` directly.

LaPeste: I agree with you, I just thought to follow the way it was done before. I'll update.

Realm/Realm.Unity/Editor/Common/ImportedReferences.cs
LaPeste: Oh, I missed that. I'll update.

Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
```diff
@@ -33,16 +35,35 @@ public static string GetInternalStorage()
...
+                    using (var currentActivity = unityPlayer.GetStatic<AndroidJavaObject>("currentActivity"))
+                    {
+                        var getFilesDir = AndroidJNIHelper.GetMethodID(AndroidJNI.FindClass("android/content/ContextWrapper"), "getFilesDir", "()Ljava/io/File;");
+                        var internalFilesDir = AndroidJNI.CallObjectMethod(currentActivity.GetRawObject(), getFilesDir, Array.Empty<jvalue>());
+                        var getAbsolutePath = AndroidJNIHelper.GetMethodID(AndroidJNI.FindClass("java/io/File"), "getAbsolutePath", "()Ljava/lang/String;");
+
+                        path = AndroidJNI.CallStringMethod(internalFilesDir, getAbsolutePath, Array.Empty<jvalue>());
+
+                        // if not on the unity thread
+                        if (Thread.CurrentThread.ManagedThreadId != 1)
```
LaPeste, line 54: I know this is a weak way of checking if the code is running on the Unity thread. The value 1 was hand checked and consistently worked. May it change in future release? Unlikely, nevertheless it could happen.
Naming the Unity thread could be a good idea, but that thread can be named only within Unity's Monobehaviours in the scene. This is tricky for us to enforce.
If you have a suggestion, I'll look into it.

Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
```diff
@@ -33,16 +35,35 @@ public static string GetInternalStorage()
...
-                var internalFilesDir = AndroidJNI.CallObjectMethod(currentActivity.GetRawObject(), getFilesDir, Array.Empty<jvalue>());
-                var getAbsolutePath = AndroidJNIHelper.GetMethodID(AndroidJNI.FindClass("java/io/File"), "getAbsolutePath", "()Ljava/lang/String;");
+                if (AndroidJNI.AttachCurrentThread() == 0)
+                {
+                    using (var unityPlayer = new AndroidJavaClass("com.unity3d.player.UnityPlayer"))
+                    using (var currentActivity = unityPlayer.GetStatic<AndroidJavaObject>("currentActivity"))
+                    {
+                        var getFilesDir = AndroidJNIHelper.GetMethodID(AndroidJNI.FindClass("android/content/ContextWrapper"), "getFilesDir", "()Ljava/io/File;");
+                        var internalFilesDir = AndroidJNI.CallObjectMethod(currentActivity.GetRawObject(), getFilesDir, Array.Empty<jvalue>());
+                        var getAbsolutePath = AndroidJNIHelper.GetMethodID(AndroidJNI.FindClass("java/io/File"), "getAbsolutePath", "()Ljava/lang/String;");
```
LaPeste, line 49: All these AndroidJNI methods return pointers. We could check for their "nullity" (IntPtr.Zero), but everything is a try and catch that does nothing if exceptions are raised. We can't do much if some of those methods return "null"/break, so I think it's safe to leave it as is.

nirinchev (commented):
I have a few nits regarding the schema lookup. I'll need to read the Android threading code more carefully and try to understand what's going on there though.

Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
nirinchev: After reading up on it a bit, I don't think we need to use JNI here at all. These methods are all defined on the Java classes, so can't we do something like 

```
var filesDir = currentActivity.Call<AndroidJavaObject>("getFilesDir");
return filesDir.Call<string>("getAbsolutePath");
```

Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
```diff
@@ -33,16 +33,22 @@ public static string GetInternalStorage()
...
+                using (var unityPlayer = new AndroidJavaClass("com.unity3d.player.UnityPlayer"))
+                using (var currentActivity = unityPlayer.GetStatic<AndroidJavaObject>("currentActivity"))
+                {
+                    var filesDir = currentActivity.Call<AndroidJavaObject>("getFilesDir");
+                    path = filesDir.Call<string>("getAbsolutePath");
+                }
             }
+            catch
+            {
+            }
```
LaPeste, line 49: This try/catch may not be that useful but
- `getAbsoluteFile` could throw `SecurityException`, which I guess it's pretty rare)
- `GetStatic` may return null if it can't find what's asked, which I can't think of an easy example for currentActivity out of the top of my head

In summary, I think that for precaution is good to be there.

Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
```diff
@@ -33,16 +33,22 @@ public static string GetInternalStorage()
...
 
-                var path = AndroidJNI.CallStringMethod(internalFilesDir, getAbsolutePath, Array.Empty<jvalue>());
-                return path ?? $"/data/data/{Application.identifier}/files";
+            try
+            {
+                using (var unityPlayer = new AndroidJavaClass("com.unity3d.player.UnityPlayer"))
+                using (var currentActivity = unityPlayer.GetStatic<AndroidJavaObject>("currentActivity"))
+                {
+                    var filesDir = currentActivity.Call<AndroidJavaObject>("getFilesDir");
+                    path = filesDir.Call<string>("getAbsolutePath");
```
nirinchev, line 44: You can return here - no need to declare `path` at all.

Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
```diff
@@ -33,16 +33,22 @@ public static string GetInternalStorage()
...
-            using (var currentActivity = unityPlayer.GetStatic<AndroidJavaObject>("currentActivity"))
-            {
-                var getFilesDir = AndroidJNIHelper.GetMethodID(AndroidJNI.FindClass("android/content/ContextWrapper"), "getFilesDir", "()Ljava/io/File;");
-                var internalFilesDir = AndroidJNI.CallObjectMethod(currentActivity.GetRawObject(), getFilesDir, Array.Empty<jvalue>());
-                var getAbsolutePath = AndroidJNIHelper.GetMethodID(AndroidJNI.FindClass("java/io/File"), "getAbsolutePath", "()Ljava/lang/String;");
+            var path = string.Empty;
 
-                var path = AndroidJNI.CallStringMethod(internalFilesDir, getAbsolutePath, Array.Empty<jvalue>());
-                return path ?? $"/data/data/{Application.identifier}/files";
+            try
```
nirinchev, line 38: We don't need a try-catch here - the consumer will handle the exceptions ([it just suppresses them](https://github.com/realm/realm-dotnet/blob/64bfe35ff4226a3e0c18b5db361ab28a8e36a775/Realm/Realm/InteropConfig.cs#L100-L117), but we may eventually want to log them somehow).

Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
```diff
@@ -17,13 +17,13 @@
...
 
 
-using System;
 using UnityEngine;
 
 namespace UnityUtils
 {
     public static class FileHelper
     {
+        [RuntimeInitializeOnLoadMethod]
```
nirinchev, line 26: Now that we're not using JNI, do we still need this?

Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
LaPeste: This was just set that way just because of the try/catch. It'll change since, as you noticed, the consumer handles exceptions.

Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
LaPeste: I tested it and it appears that we still need that.

Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
nirinchev: For future reference, even in a try-catch, you can do a return:

```csharp
try
{
    return DoSomethingThatMayThrow();
}
catch
{
}

return null;
```

LaPeste: That makes sense since the return type is a reference type.

Tools/SetupUnityPackage/UnityUtils/FileHelper.cs
LaPeste: Should I look into this already? If so, what did you have in mind? A logfile or something else?

nirinchev: No, that's outside of the scope of this change. It was more of a recommendation to always let the final caller handle exceptions.

nirinchev (approved):

Issue #2169, opened 2020-12-22
Realm .NET 10.0.0-beta.3 crashes on UWP/iOS release build


kyurkchyan:
## Goals

We’ve decided to update to 10.* version in order to be able to use decimal support and the rest of the new features. We were ready to make the first release with the new SDK, and soon after we made our release build we discovered that the application crashes with the following exception when built in release mode:

`Incompatible MarshalAs detected in parameter named 'value'. Please refer to MCG's warning message for more information.`

This only happens in release mode though, both on iOS and UWP. This is a major blocker for as and we would have to revert the effort of 2 weeks of refactoring if we decide to revert the changes. Anything we could do to work around this issue?

## Expected Results

Realm should work on UWP and iOS Release mode

## Actual Results

Realm crashes with above mentioned exception

## Steps to Reproduce

1. Download the sample project [RealmPlayground.zip](https://github.com/realm/realm-dotnet/files/5731104/RealmPlayground.zip)
2. Set UWP app as startup project
3. Set the build to Release x64
4. Launch the app
5. Observe the exception

## Version of Realm and Tooling

- Client SDK Version: 10.0.0-beta.3
- Client OS & Version:
    - Build Machine: Windows 10, 18363.1256
    - App Target OS - UWP, iOS 14

nirinchev:
When building for iOS, do you get the crash on device or simulator? Also, can you paste the MCG warning message?

kyurkchyan:
@nirinchev  here is one of the warning messages from Release x64 UWP build. I get 46 of them in that simple sample project! Here is the file  with all of them
[MCG_warnings.txt](https://github.com/realm/realm-dotnet/files/5736688/MCG_warnings.txt)

```
Severity	Code	Description	Project	File	Line	Suppression State
Warning		MCG : warning MCG0001: [MarshalAs(UnmanagedType=(Default), ElementType=(Default))] with type 'Realms.Native.PrimitiveValue' is not supported.This signature is located on 'System.Void Realms.ResultsHandle.NativeMethods.get_value(Realms.ResultsHandle, System.IntPtr, Realms.Native.PrimitiveValue, Realms.NativeException)'.	RealmPlayground.UWP		
```	

As for iOS, I can't repro the issue right now, will get back after Christmas and put more details! Thank you @nirinchev  for the support.

nirinchev:
[...]

PR #2170, opened 2021-01-04
Fix .NET Native compilation

nirinchev:
## Description

Hopefully fixes https://github.com/realm/realm-dotnet/issues/2169.

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Native/PrimitiveValue.cs
+++ b/Realm/Realm/Native/PrimitiveValue.cs
@@ -29,10 +29,6 @@ namespace Realms.Native
     internal unsafe struct PrimitiveValue
     {
         [FieldOffset(0)]
-        [MarshalAs(UnmanagedType.U1)]
-        private bool bool_value;
-
-        [FieldOffset(0)]
         private long int_value;
 
         [FieldOffset(0)]
@@ -76,7 +72,7 @@ namespace Realms.Native
         public static PrimitiveValue Bool(bool value) => new PrimitiveValue
         {
             Type = RealmValueType.Bool,
-            bool_value = value
+            int_value = value ? 1 : 0,
         };
 
         public static PrimitiveValue NullableBool(bool? value) => value.HasValue ? Bool(value.Value) : Null();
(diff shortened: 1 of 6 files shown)


Realm/Realm/Native/PrimitiveValue.cs
```diff
@@ -28,10 +28,6 @@ namespace Realms.Native
     [DebuggerDisplay("PrimitiveValue({Type})")]
     internal unsafe struct PrimitiveValue
     {
-        [FieldOffset(0)]
-        [MarshalAs(UnmanagedType.U1)]
-        private bool bool_value;
```
nirinchev, line 33: Removing the `[MarshalAs]` attribute from `bool_value` appears to fix the issue.

Tests/Tests.UWP/Tests.UWP.csproj
```diff
@@ -11,15 +11,14 @@
     <AssemblyName>Tests.UWP</AssemblyName>
```
nirinchev, line 11: Reviewers can ignore changes to this file - since we don't run UWP tests automatically, the project file had rotted a little and needed to bring it up to date to be able to debug things.

wrappers/CMakeLists.txt
```diff
@@ -46,6 +46,9 @@ if(APPLE)
   endif()
 elseif(CMAKE_SYSTEM_NAME MATCHES "^Windows")
   add_definitions(/DWIN32_LEAN_AND_MEAN)
+  add_compile_options(
+      /MP # Enable multi-processor compilation
```
nirinchev, line 50: @fealebenpae added this since it seemed like we didn't compile the wrapper sources in parallel.

fealebenpae (approved):
I suppose this might be a legitimate bug in the MCG - after all in Win32 `BOOL` is usually 4 bytes, so maybe the marshaler hasn't been exercised with the more portable 1-byte `bool`? Whatever the reason, piggy-backing on `int_value` is a good workaround.

PR #2173, opened 2021-01-06
Ac/validate pk in realm.find

### LINKED ISSUE #2111: Validate PK type in Realm.Find
Seems like `Realm.Find` will pass the PK argument as is to native without validating that the PK type matches the argument.

LaPeste:
## Description
There was an error in PropertyTypeMismatchException's .ctor. That made an assert in the `to_string::format` trigger because of the lack of parameters. This made the exception never come up, hence the bug report.

Fixes #2111

##  TODO

* [ ] Changelog entry
* [x] Tests (if applicable)


nirinchev:
This diff is quite weird - can you rebase on master so that we don't see a bunch of unrelated changes?

LaPeste:
I have reopened another PR. That is clean from that weird diff.

PR #2175, opened 2021-01-07
Fixed bug in PropertyTypeMismatchException's .ctor and added unitTest…

### LINKED ISSUE #2111: Validate PK type in Realm.Find

LaPeste:
## Description
There was an error in PropertyTypeMismatchException's .ctor. That made an assert in the `to_string::format` trigger because of the lack of parameters. This made the exception never come up, hence the bug report.

Fixes #2111

##  TODO

* [ ] Changelog entry
* [x] Tests (if applicable)


LaPeste (commented):
Please select "Hide whitespace changes" in the settings of the diff because error_handling.hpp was in a bad formatting shape and Visual studio is reformatting to follow all the other files.
I tried to make it stop formatting completely but it doesn't seem to follow my will. With the setting I suggested turned on, you should see only the real change.
In the end of the day, I guess having a well formatted file is just a good thing. If you strongly disagree, I'll try to see what I can do to revert it to the incorrect formatting it was in, before.

Tests/Realm.Tests/Database/PrimaryKeyTests.cs
```diff
@@ -192,6 +192,46 @@ private RealmObjectBase FindByPKGeneric(Type type, object primaryKeyValue, PKTyp
             return (RealmObjectBase)genericMethod.MakeGenericMethod(type).Invoke(_realm, new[] { primaryKeyValue });
         }
 
+        [Test]
+        public void ExceptionIfMismatchingPKType()
```
nirinchev, line 196: We try to avoid writing overly broad tests that cover a bunch of parameters. A better approach would be to split these into smaller tests and extract a core test that does the verifications:

```csharp
private readonly IEnumerable<dynamic> _primaryKeyValues = new[] { "42", 123, ObjectId.GenerateNewId(), Guid.NewGuid() };

[Test]
public void RealmFind_WhenPKIsIntAndArgumentIsNot_Throws() => RealmFind_IncorrectPKArgument_Throws<IntPrimaryKeyWithValueObject>();

// Rest of PK types.

private void RealmFind_IncorrectPKArgument_Throws<T>()
{
    foreach (var pk in _primaryKeyValues)
    {
        // Skip the PK if its type is equal to the type of the object
        Assert.That(() => _realm.Find<T>(pk), Throws...);

        // Also add a test for the dynamic API: _realm.DynamicApi.Find("")
    }
}
```

Tests/Realm.Tests/Database/PrimaryKeyTests.cs
```diff
@@ -192,6 +195,70 @@ private RealmObjectBase FindByPKGeneric(Type type, object primaryKeyValue, PKTyp
...
+        public void RealmFind_WhenPKIsGuidAndArgumentIsNot_Throws() => RealmFind_IncorrectPKArgument_Throws<PrimaryKeyGuidObject>();
+
+        [Test]
+        public void RealmFind_WhenPKIsObjectIdAndArgumentIsNot_Throws() => RealmFind_IncorrectPKArgument_Throws<PrimaryKeyObjectIdObject>();
+
+        // In Find<T> anything that is a primitive type (according to Realm's view) is converted to long?.
+        // FindCore<T> converts its input in RealmValue. Which in turn treats long as PritiveValue of type Int.
+        // Without this pass it'd be impossible to match any PK that's of primitive type, since a PK defined as e.g. short would always be
+        // compared against an int
+        private static Type ConvertToRealmValue(Type toConvert)
```
LaPeste, line 229: I don't think of this as a good way to achieve what's needed, but after various attempts I folded back to a simple solution.
I'm confident you can advice for better. At least this allowed to write a working version of the tests following your feedback.

nirinchev: This should better be renamed to `GetDatabaseType` as initially I was confused how is that related to RealmValue (which is our implementation of Mixed).

It can be simplified a little with a few suggestions:

1. `nativeTypes` is the incorrect variable name here because those types are not really native - instead it should be something like `intBackedTypes` or `intConvertibleTypes`.
2. You don't need it to be a `ReadOnlyCollection` - a simple array works just fine.
3. You can use LINQ's `Contains` to avoid doing the manual loop.
4. For int-backed types, you should return `typeof(long)` as that is more semantically correct. This will require to change `123` in `_primaryKeyValues` to `123L`.
5. If the type is not stored as an Int64 internally, you can return the argument rather than returning `null`.

Tests/Realm.Tests/Database/PrimaryKeyTests.cs
```diff
@@ -192,6 +195,70 @@ private RealmObjectBase FindByPKGeneric(Type type, object primaryKeyValue, PKTyp
...
+
+        private void RealmFind_IncorrectPKArgument_Throws<T>()
+            where T : RealmObject
+        {
+            var pkInClass = typeof(T).GetProperties().Single(prop => Attribute.IsDefined(prop, typeof(PrimaryKeyAttribute)));
+            var pkType = ConvertToRealmValue(pkInClass.PropertyType) ?? pkInClass.PropertyType;
+
+            foreach (var pk in _primaryKeyValues)
+            {
+                if (pk.GetType() == pkType)
```
nirinchev, line 252: This check can be folded into a `.Where` clause - I have a modification of Atwood's Law, which goes something like "if something can be rewritten as LINQ, it must be rewritten as LINQ" 😄

wrappers/src/error_handling.hpp
```diff
@@ -92,16 +92,16 @@ class NotNullableException : public std::runtime_error {
 class PropertyTypeMismatchException : public std::runtime_error {
 public:
     PropertyTypeMismatchException(std::string object_type, std::string property, std::string property_type, std::string actual_type)
-        : std::runtime_error(util::format("Property type mismatch: %1.%2 is of type %3, but the value is of type %4")) {}
+        : std::runtime_error(util::format("Property type mismatch: %1. %2 is of type %3, but the supplied value is of type %4", object_type, property, property_type, actual_type)) {}
```
nirinchev, line 95: There shouldn't be a space between `%1` and `%2`. The idea here is to have a message like `Foo.Bar is of type string, but the supplied value is of type int`. With the space, it looks like end of a sentence after `Foo`.

nirinchev (changes requested):

wrappers/src/error_handling.hpp
LaPeste: In my mind that was a real full stop. Like `Mistmatch: Foo. "Property" Bar (clearly of Foo) is of type....`
But now that you say that, it just makes a lot more sense. 😅

Tests/Realm.Tests/Database/PrimaryKeyTests.cs
```diff
@@ -192,6 +195,57 @@ private RealmObjectBase FindByPKGeneric(Type type, object primaryKeyValue, PKTyp
...
+        [Test]
+        public void RealmFind_WhenPKIsGuidAndArgumentIsNot_Throws() => RealmFind_IncorrectPKArgument_Throws<PrimaryKeyGuidObject>();
+
+        [Test]
+        public void RealmFind_WhenPKIsObjectIdAndArgumentIsNot_Throws() => RealmFind_IncorrectPKArgument_Throws<PrimaryKeyObjectIdObject>();
+
+        // In Find<T> anything that is a primitive type (according to Realm's view) is converted to long?.
+        // FindCore<T> converts its input in RealmValue. Which in turn treats long as PritiveValue of type Int.
+        // Without this pass it'd be impossible to match any PK that's of primitive type, since a PK defined as e.g. short would always be
+        // compared against an int
```
nirinchev, line 228: I understand what you're trying to say here, but I think we should reword it to make it easier to understand for people who don't have context. I'm thinking of something like:

```
// In RealmFind_IncorrectPKArgument_Throws we're using reflection to check if the primary
// key type and the property type match. Since Realm stores all integral properties as Int64, we
// want to treat the narrower types as "long" to match the value type in _primaryKeyValues.
```

Tests/Realm.Tests/Database/PrimaryKeyTests.cs
```diff
@@ -192,6 +195,57 @@ private RealmObjectBase FindByPKGeneric(Type type, object primaryKeyValue, PKTyp
...
+            return intBackedTypes.Contains(toConvert) ? typeof(long) : toConvert;
+        }
+
+        private void RealmFind_IncorrectPKArgument_Throws<T>()
+            where T : RealmObject
+        {
+            var pkInClass = typeof(T).GetProperties().Single(prop => Attribute.IsDefined(prop, typeof(PrimaryKeyAttribute)));
+            var pkType = GetDatabaseType(pkInClass.PropertyType) ?? pkInClass.PropertyType;
+
+            var keys = _primaryKeyValues.Where(pk => pk.GetType() != pkType);
```
nirinchev, line 241: Perhaps `keys` -> `keysWithIncorrectType`?

nirinchev (commented):
Few final nits and I think it's ready to go.

nirinchev (approved):

Issue #2180, opened 2021-01-08
RealmObject.FreezeInPlace doesn't work for collections


nirinchev:
If a user has accessed a list property and then tries to freeze a RealmObject in place, the list will not be frozen. This is because we replace the RealmObject handle, but it has already cached a reference to the live list and will not be replaced after the freezing process.

PR #2181, opened 2021-01-08
Remove RealmObject.FreezeInPlace

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2180

##  TODO

* [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## vNext (TBD)
 ------------------
 ### Breaking Changes
-* None
+* Removed `RealmObject.FreezeInPlace`. To freeze a realm object use the `Freeze` extension method. (Issue [#2180](https://github.com/realm/realm-dotnet/issues/2180))
 
 ### Fixed
 * Worked around an issue with the .NET Native compiler (used in UWP projects) that would result in the following exception being thrown in Release: `Incompatible MarshalAs detected in parameter named 'value'. Please refer to MCG's warning message for more information.`. (Issue [#2169](https://github.com/realm/realm-dotnet/issues/2169))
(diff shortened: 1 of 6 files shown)


LaPeste (approved):

LaPeste (approved):
As far as I understood the goal was to remove a non-working method and this is what I can see done with this PR. No comment about that.

About the detail "This is because we replace the RealmObject handle, but it has already cached a reference to the live list and will not be replaced after the freezing process." I don't have enough knowledge of this so maybe you could point at where this caching happens or spend a few words on it.

If I misunderstood something about this PR, then it's good I wrote things down, so I can be explained :)

nirinchev:
That part wasn't worded particularly well, sorry. What's happening is, for collections, the first time a property is accessed, we create a RealmList/Set/Dictionary and store a reference in a class level field (we're reusing the backing field for the property). This is safe, because the collection will dynamically look up all its members, so we're not caching any actual data, just avoiding recreating the collection on every property access.

When freezing an object in place, what we did was replace it's ObjectHandle (the pointer to the native memory) with the frozen handle that we get from Core. That means that from then on every property access would use the frozen handle and read the frozen data.

Now, this is all good for primitive properties where we read them on every access, but for lists presented a problem because the local field was still holding on to the live list. There were two ways to resolve that:
1. Iterate all list properties and replace their handles with a frozen handle.
2.Iterate all list fields and set them to `null` (which would then result in the properties recreating the collection next time it's accessed).

Ultimately, I chose to remove the API because:
1. Both options would be fairly involved to write without using reflection and using reflection would be fairly inefficient.
2. RealmObject was the only type that had a FreezeInPlace method - all other types have a `Freeze` which returns a new frozen instance.
3. We're still in beta, so I would rather make a breaking change now and maybe reinstate the API if enough users request it than spend a whole lot of time implementing something we don't know how many people will use.

LaPeste:
Thank you for the detailed explanation. It all makes sense.

PR #2120, opened 2020-11-13
Add UUID support

papafe:
## Description
Add support for `UUID` data type. 
In order to use this, the object property must be declared of type `Guid`:

```csharp
public MyObject : RealmObject
{
    public Guid Id {get; set;}

    .... 
}
```
The new data type:
- Can be used as a primary key
- Can be used in queries: equal, not equal
- Is indexable

Fixes #2065 

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

Issue #123, opened 2015-09-03
Remove Implicit Write Transactions feature


AndyDentFree:
As per Slack discussion with Brian  
_We’ve discussed the option of implicit write transactions and concluded that there are a number of issues with that. Including that many will avoid using the explicit, and then do a bunch of mutating operations as single transactions. Each will do a commit to disk and each time queries will be redone and accessors updated and the state changed.
The longer term solution is merge based write transactions.

PR #2185, opened 2021-01-14
Migrate to bison parser

LaPeste:
##  TODO

* [x] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Internal
 * Using Core 10.3.0.
+* Migrated to Bison parser
 
 ## 10.0.0-beta.3 (2020-12-10)
 ------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Enhancements
 * Add support for the `GUID` data type. It can be used as primary key and is indexable. (PR [#2120](https://github.com/realm/realm-dotnet/pull/2120))
+* Replaced the implementation of the string query parser (the one used for realm.All().Filter("some-string-query")). This results in ~5% reduction of the size of the native binary while keeping the query execution times on par with the old parser.
 
 ### Compatibility
 * Realm Studio: 10.0.0 or later.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Enhancements
 * Add support for the `GUID` data type. It can be used as primary key and is indexable. (PR [#2120](https://github.com/realm/realm-dotnet/pull/2120))
-* Replaced the implementation of the string query parser (the one used for realm.All().Filter("some-string-query")). This results in ~5% reduction of the size of the native binary while keeping the query execution times on par with the old parser.
+* Replaced the implementation of the string query parser (the one used for [`realm.All().Filter("some-string-query")`](https://docs.mongodb.com/realm-sdks/dotnet/10.0.0-beta.3/reference/Realms.CollectionExtensions.html#Realms_CollectionExtensions_Filter__1_System_Linq_IQueryable___0__System_String_)). This results in ~5% reduction of the size of the native binary while keeping the query execution times on par with the old parser. (PR [#2185](https://github.com/realm/realm-dotnet/pull/2185), Core upgrade)
 
 ### Compatibility
 * Realm Studio: 10.0.0 or later.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Internal
 * Using Core 10.3.0.
+* Migrate to bison parser
 
 ## 10.0.0-beta.3 (2020-12-10)
 ------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@
 
 ### Internal
 * Using Core 10.3.0.
-* Migrated to Bison parser
 
 ## 10.0.0-beta.3 (2020-12-10)
 ------------------
--- a/Tests/Realm.Tests/Database/CollectionTests.cs
+++ b/Tests/Realm.Tests/Database/CollectionTests.cs
@@ -611,7 +611,7 @@ namespace Realms.Tests.Database
         {
             Assert.That(
                 () => _realm.All<A>().Filter("Foo == 5"),
-                Throws.TypeOf<RealmException>().And.Message.Contains("No property 'Foo' on object of type 'A'"));
+                Throws.TypeOf<RealmException>().And.Message.Contains("'class_A' has no property: 'Foo'"));
         }
 
         [Test]
--- a/Tests/Realm.Tests/Database/CollectionTests.cs
+++ b/Tests/Realm.Tests/Database/CollectionTests.cs
@@ -611,7 +611,7 @@ namespace Realms.Tests.Database
         {
             Assert.That(
                 () => _realm.All<A>().Filter("Foo == 5"),
-                Throws.TypeOf<RealmException>().And.Message.Contains("'class_A' has no property: 'Foo'"));
+                Throws.TypeOf<RealmException>().And.Message.Contains("'A' has no property: 'Foo'"));
         }
 
         [Test]
--- a/wrappers/src/results_cs.cpp
+++ b/wrappers/src/results_cs.cpp
@@ -128,9 +128,8 @@ REALM_EXPORT Results* results_get_filtered_results(const Results& results, uint1
         query_parser::KeyPathMapping mapping;
         realm::populate_keypath_mapping(mapping, *realm);
 
-        auto table = results.get_table();
         query_parser::NoArguments no_args;
-        auto query = table->query(query_string, no_args, mapping);
+        auto query = results.get_table()->query(query_string, no_args, mapping);
         auto ordering = query.get_ordering();
         return new Results(realm, results.get_query().and_query(query), *ordering);
     });
--- a/wrappers/src/results_cs.cpp
+++ b/wrappers/src/results_cs.cpp
@@ -129,10 +129,13 @@ REALM_EXPORT Results* results_get_filtered_results(const Results& results, uint1
         realm::populate_keypath_mapping(mapping, *realm);
 
         query_parser::NoArguments no_args;
-        auto query = results.get_table()->query(query_string, no_args, mapping);
-        auto ordering = query.get_ordering() ? *query.get_ordering() : results.get_descriptor_ordering();
+        Query parsed_query = results.get_table()->query(query_string, no_args, mapping);
+        DescriptorOrdering new_order = results.get_descriptor_ordering();
+        if (auto parsed_ordering = parsed_query.get_ordering()) {
+            new_order.append(*parsed_ordering);
+        }
 
-        return new Results(realm, results.get_query().and_query(query), ordering);
+        return new Results(realm, results.get_query().and_query(std::move(parsed_query)), new_order);
     });
 }
 
--- a/wrappers/src/results_cs.cpp
+++ b/wrappers/src/results_cs.cpp
@@ -123,7 +123,7 @@ REALM_EXPORT Results* results_get_filtered_results(const Results& results, uint1
 {
     return handle_errors(ex, [&]() {
         Utf16StringAccessor query_string(query_buf, query_len);
-        auto const &realm = results.get_realm();
+        auto const& realm = results.get_realm();
 
         query_parser::KeyPathMapping mapping;
         realm::populate_keypath_mapping(mapping, *realm);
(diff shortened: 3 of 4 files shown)


wrappers/src/results_cs.cpp
```diff
@@ -124,20 +123,18 @@ REALM_EXPORT Results* results_get_filtered_results(const Results& results, uint1
...
-        query_builder::apply_predicate(query, result.predicate, no_args, mapping);
+        auto table = realm->read_group().get_table(object_schema.table_key);
 
-        DescriptorOrdering ordering;
-        query_builder::apply_ordering(ordering, query.get_table(), result.ordering);
-        return new Results(realm, std::move(query), std::move(ordering));
+        query_parser::NoArguments no_args;
+        auto query = table->query(query_string, no_args, mapping);
+        auto ordering = query.get_ordering();
+        return new Results(realm, query, *ordering);
```
LaPeste, line 137: I had to brush up on move semantics. I don't think the `std::move` is needed in either of the case. One special rule says that if the move constructor and operator are defined, returning automatic l-values by copy from a method or function moves such values and not copy.
If my understanding is incorrect, then I'll put the `std::move` back where it was.

nirinchev (changes requested):

wrappers/src/results_cs.cpp
```diff
@@ -124,20 +123,16 @@ REALM_EXPORT Results* results_get_filtered_results(const Results& results, uint1
...
+        query_parser::KeyPathMapping mapping;
         realm::populate_keypath_mapping(mapping, *realm);
 
-        query_builder::NoArguments no_args;
-        query_builder::apply_predicate(query, result.predicate, no_args, mapping);
-
-        DescriptorOrdering ordering;
-        query_builder::apply_ordering(ordering, query.get_table(), result.ordering);
-        return new Results(realm, std::move(query), std::move(ordering));
+        auto table = results.get_table();
```
nirinchev, line 131: [nit] Since `table` is used only once, we can remove the extra variable here and use `results.get_table().query(...)` on line 133.

nirinchev (approved):

nirinchev:
Needs a changelog entry and I asked @jedelbo to double check the C++ code in case we're missing something about the new API.

wrappers/src/results_cs.cpp
```diff
@@ -124,20 +123,16 @@ REALM_EXPORT Results* results_get_filtered_results(const Results& results, uint1
...
+        query_parser::NoArguments no_args;
+        auto query = table->query(query_string, no_args, mapping);
+        auto ordering = query.get_ordering();
+        return new Results(realm, results.get_query().and_query(query), *ordering);
```
jedelbo, line 135: `ordering` can be NULL!

jedelbo (changes requested):

Tests/Realm.Tests/Database/CollectionTests.cs
```diff
@@ -611,7 +611,7 @@ public void Results_GetFiltered_WhenPredicateIsInvalid_Throws()
         {
             Assert.That(
                 () => _realm.All<A>().Filter("Foo == 5"),
-                Throws.TypeOf<RealmException>().And.Message.Contains("No property 'Foo' on object of type 'A'"));
+                Throws.TypeOf<RealmException>().And.Message.Contains("'class_A' has no property: 'Foo'"));
```
nirinchev, line 614: This is a regression - the change in wording is fine, but we shouldn't be surfacing the internal naming of tables. Might need to ping @ironage or @jedelbo about the best way to convert these error messages to use the expected type names.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 ### Internal
 * Using Core 10.3.0.
+* Migrated to Bison parser
```
nirinchev: This should also have a user-facing note. Perhaps sync with @jedelbo on what would be an appropriate user-facing note about it. As far as I can see, the package is now 6% smaller, so perhaps something like:

> Replaced the implementation of the string query parser (the one used for [realm.All<Foo>().Filter("some-string-query")](https://docs.mongodb.com/realm-sdks/dotnet/10.0.0-beta.3/reference/Realms.CollectionExtensions.html#Realms_CollectionExtensions_Filter__1_System_Linq_IQueryable___0__System_String_)). This results in ~5% reduction of the size of the native binary while keeping the query execution times on par with the old parser. (**if Jørgen has other comments here, be sure to include them**).

wrappers/src/results_cs.cpp
```diff
@@ -124,20 +123,21 @@ REALM_EXPORT Results* results_get_filtered_results(const Results& results, uint1
...
-        query_builder::NoArguments no_args;
-        query_builder::apply_predicate(query, result.predicate, no_args, mapping);
+        query_parser::NoArguments no_args;
+        auto query = results.get_table()->query(query_string, no_args, mapping);
+        auto ordering = query.get_ordering();
 
-        DescriptorOrdering ordering;
-        query_builder::apply_ordering(ordering, query.get_table(), result.ordering);
-        return new Results(realm, std::move(query), std::move(ordering));
+        if (ordering) {
```
nirinchev, line 135: If ordering is `null`, you should probably fall back to the existing ordering defined on the results: `DescriptorOrdering ordering = query.get_ordering() ? *query.get_ordering() :results. get_descriptor_ordering();` (not sure if C++ will like this syntax, but I hope the idea is clear).

CHANGELOG.md
jedelbo: This is just fine

jedelbo (approved):

Tests/Realm.Tests/Database/CollectionTests.cs
ironage: we missed a few places, it should be fixed by https://github.com/realm/realm-core/pull/4290

wrappers/src/results_cs.cpp
```diff
@@ -124,20 +123,16 @@ REALM_EXPORT Results* results_get_filtered_results(const Results& results, uint1
...
-
-        parser::KeyPathMapping mapping;
+        query_parser::KeyPathMapping mapping;
         realm::populate_keypath_mapping(mapping, *realm);
 
-        query_builder::NoArguments no_args;
-        query_builder::apply_predicate(query, result.predicate, no_args, mapping);
+        query_parser::NoArguments no_args;
+        auto query = results.get_table()->query(query_string, no_args, mapping);
+        auto ordering = query.get_ordering() ? *query.get_ordering() : results.get_descriptor_ordering();
```
ironage, line 133: query.get_ordering() returns the ordering, but it also removes it from the query, so you can't call it twice (this is a bit of an unusual pattern). Try this: 
```
    query_parser::KeyPathMapping mapping;
    realm::populate_keypath_mapping(mapping, *this->get_realm());
    Query parsed_query = get_query().get_table()->query(query_string, no_args, mapping);
    DescriptorOrdering new_order = results.get_descriptor_ordering();
    if (auto parsed_ordering = parsed_query.get_ordering()) {
        new_order.append(*parsed_ordering);
    }
    return Results(m_realm, results.get_query().and_query(std::move(parsed_query)), new_order);
```
I'm adding the append in https://github.com/realm/realm-core/pull/4291

Tests/Realm.Tests/Database/CollectionTests.cs
LaPeste: Thank you. Then I'll update core in our SDK to point to the right commit when the PR 4290 is merged. And of course change the string in the unit test.

wrappers/src/results_cs.cpp
LaPeste: Thank you very much. I've modified as you suggested and I'll wait for the PR 4291 to be merged, to try test the code.

nirinchev: You can temporarily point this PR to James's branch to get things building.

LaPeste: As expected it works. Thank you.

wrappers/src/results_cs.cpp
```diff
@@ -124,20 +123,19 @@ REALM_EXPORT Results* results_get_filtered_results(const Results& results, uint1
...
+        Query parsed_query = results.get_table()->query(query_string, no_args, mapping);
+        DescriptorOrdering new_order = results.get_descriptor_ordering();
+        if (auto parsed_ordering = parsed_query.get_ordering()) {
+            new_order.append(*parsed_ordering);
+        }
 
-        DescriptorOrdering ordering;
-        query_builder::apply_ordering(ordering, query.get_table(), result.ordering);
-        return new Results(realm, std::move(query), std::move(ordering));
+        return new Results(realm, results.get_query().and_query(std::move(parsed_query)), new_order);
```
nirinchev, line 138: `std::move(new_order)`?

nirinchev (approved):

nirinchev (changes requested):
Oops, hit approve too fast - still needs a user-facing changelog entry.

wrappers/src/results_cs.cpp
LaPeste: I don't know, to be honest. I noticed that as well but @irongage didn't use it in the example. I thought it was a stylistic error, but then I went to check and only `Query` has a custom move constructor, while `DescriptorOrdering` has only the default one. After seeing that I thought that it was not a mistake but a wanted choice.
What is your opinion?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Enhancements
 * Add support for the `GUID` data type. It can be used as primary key and is indexable. (PR [#2120](https://github.com/realm/realm-dotnet/pull/2120))
+* Replaced the implementation of the string query parser (the one used for realm.All().Filter("some-string-query")). This results in ~5% reduction of the size of the native binary while keeping the query execution times on par with the old parser.
```
nirinchev, line 11:
```suggestion
* Replaced the implementation of the string query parser (the one used for [`realm.All().Filter("some-string-query")`](https://docs.mongodb.com/realm-sdks/dotnet/10.0.0-beta.3/reference/Realms.CollectionExtensions.html#Realms_CollectionExtensions_Filter__1_System_Linq_IQueryable___0__System_String_)). This results in ~5% reduction of the size of the native binary while keeping the query execution times on par with the old parser. (PR [#2185](https://github.com/realm/realm-dotnet/pull/2185), Core upgrade)
```

At the end of the change entry we try to add some additional information, depending on the change type:
* If there's an issue filed for the change (typically a bug or a feature request), we add (Issue [#123](https://github.com/realm/realm-dotnet/issue/123)). This allows the person who filed the issue and everyone interested to see that the new release includes what they were asking for.
* If there's no issue we add a link to the PR that introduced the change - e.g. (PR [#2120](https://github.com/realm/realm-dotnet/pull/2120)) like above. This allows us to quickly find the context of a change by looking at the changelog, which is (typically) more readable than the git history.
* Finally, if all a PR does is to bump the core version and we get some fixes for free, we add (Core upgrade). 

In this case, I proposed a combination of 2 and 3 because the filesize reduction is thanks to the Core upgrade, but the PR includes some changes of its own, so worth linking to it.

As for the link to the API docs - we typically don't bother linking to the API from the changelog, but in this case, it concerns a feature that may not be super heavily used (as I'm fairly certain most devs will prefer to use LINQ), so adding a link makes it more prominent and will allow people who are curious about it to find more info quickly.

nirinchev (approved):
Final nit from me. Otherwise, should be ready to merge once https://github.com/realm/realm-core/pull/4290 goes in.

wrappers/src/results_cs.cpp
nirinchev: I don't know - I'm not a C++ dev 😄 As mentioned, I always add `std::move` when I want to move something, just to be on the safe side. Maybe @ironage can educate us on the pros/cons of `std::move`-ing the ordering?

LaPeste: That feedback would be very welcome. 🙂

ironage: Yes you can move the ordering, I just forgot. Either way is functionally correct, but moving it is an efficient transfer of ownership, while not moving it invokes the [copy constructor ](https://github.com/realm/realm-core/blob/master/src/realm/sort_descriptor.cpp#L545-L550) which inefficiently makes copies of all the contents into the new object only to then discard the previous object entirely.

LaPeste: My only doubt was coming from the fact that while brushing on `std::move` I read that returning automatic local variables generally (not sure if always) uses the move semantics instead of the copy ones, as long as a move constructor and assignment exist. So I was not sure if stylistically people still added a `std::move` on the returned values. You clearly answered that people generally prefer to be explicit, thank you for the feedback.

PR #2187, opened 2021-01-18
Fix memory leak due to handle

### LINKED ISSUE #2184: Realm Memory Leak When Get Data.
## Goals
We receiver feed back from some user that host server is slow and disconnect many time. 
We see Ram Overload and we check detail, it maybe memory leak. 
We stop all function, test with just very simple get data from Ream for check bigger issue. 

## Expected Results

Memory is ok.

## Actual Results

Memory none stop Increasing. 

## Steps to Reproduce

We check memory leak, and we thing it is leaking. 
Test function by function to finding where it leak => We found it in very simple function.
Test many way but can't solve. 
Finding google and issue in Realm but can't find anything like that. 

## Code Sample

Calling Function: 
 
        [HttpPost ("{id}/Measures", Name = "PostSchedules")]
        public string PostMeasure(string id, int child,  [FromBody] object value, string token)
        {
            Database.GWTokenDB.GetTokenGW(id); 
            return "{\"Result\":\"Successful\"}"; ;
        }

Action function: ( In test case it will return null - So Clone() function don't run in this test)

        public static GatewayTokenModel GetTokenGW(string gatewayId)
        {
            using (var realm = Realm.GetInstance(Configuration))
            {
                try
                {
                    var data = realm.Find<GatewayTokenModel>(gatewayId);
                    if (data != null) return data.Clone();
                    else return null;
                }
                catch (Exception ex)
                {

                }
            }
            return null;
        }

- Realm Object Server Version: ? 5.1.2
- Flavor:
  - [ ] Realm Cloud
  - [x] Self-Hosted
- Server OS & Version: ? CentOS 6. 
- Client SDK Version: http request.
- Client OS & Version: Window 10, request http to server.

Adding Image Test Memory Leak: 
image

Capture Information at end of test => Stop send data more and press GC Collect many times until memory don't decreasing more.
image

papafe:
Added freeing of the handle in `State` when disposing of realm.

Fixes #2184

### DIFF
--- a/Realm/Realm/Handles/SharedRealmHandle.cs
+++ b/Realm/Realm/Handles/SharedRealmHandle.cs
@@ -47,7 +47,7 @@ namespace Realms
             public unsafe delegate void OpenRealmCallback(IntPtr task_completion_source, IntPtr shared_realm, int error_code, byte* message_buf, IntPtr message_len);
 
             [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
-            public delegate void FreeGCHandleCallback(IntPtr handle);
+            public delegate void OnBindingContextDestructedCallback(IntPtr handle);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_open", CallingConvention = CallingConvention.Cdecl)]
             public static extern IntPtr open(Configuration configuration,
@@ -144,7 +144,7 @@ namespace Realms
             public static extern void get_schema(SharedRealmHandle sharedRealm, IntPtr callback, out NativeException ex);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_install_callbacks", CallingConvention = CallingConvention.Cdecl)]
-            public static extern void install_callbacks(NotifyRealmCallback notifyRealmCallback, GetNativeSchemaCallback nativeSchemaCallback, OpenRealmCallback openCallback, FreeGCHandleCallback freeGcHandleCallback);
+            public static extern void install_callbacks(NotifyRealmCallback notifyRealmCallback, GetNativeSchemaCallback nativeSchemaCallback, OpenRealmCallback openCallback, OnBindingContextDestructedCallback contextDestructedCallback);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_has_changed", CallingConvention = CallingConvention.Cdecl)]
             [return: MarshalAs(UnmanagedType.U1)]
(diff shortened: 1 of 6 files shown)


Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -46,6 +46,9 @@ private static class NativeMethods
             [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
             public unsafe delegate void OpenRealmCallback(IntPtr task_completion_source, IntPtr shared_realm, int error_code, byte* message_buf, IntPtr message_len);
 
+            [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+            public delegate void FreeGCHandleCallback(IntPtr handle);
```
nirinchev, line 50: I know this was the name used previously, but let's rename to something that leaks less implementation details. I'm thinking either `DisposeStateCallback` or `OnBindingContextDestructedCallback`.

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -420,6 +425,15 @@ private static unsafe void HandleOpenRealmCallback(IntPtr taskCompletionSource,
             }
         }
 
+        [MonoPInvokeCallbackAttribute(typeof(NativeMethods.FreeGCHandleCallback))]
+        public static void FreeGCHandle(IntPtr handle)
+        {
+            if (handle != IntPtr.Zero)
+            {
+                GCHandle.FromIntPtr(handle).Free();
```
nirinchev, line 433: Tied to my renaming suggestion - I'm not a big fan of this approach because it leaks the internal design of the managed state. Instead, what we could do is have `State` implement `IDisposable` and in the callback, we can cast the target to `State` and call the `Dispose` method.

nirinchev (commented):
Is there a unit test we can write that would exhibit the memory leak? Maybe open a Realm and take a weak reference to the Realm and to its state, then run GC.Collect until the Realm gets collected and verify that the state is also collected?

papafe:
> Is there a unit test we can write that would exhibit the memory leak? Maybe open a Realm and take a weak reference to the Realm and to its state, then run GC.Collect until the Realm gets collected and verify that the state is also collected?

At the moment the state is private, so to do that test I will need to change its accessibility at least to internal, and I don't feel it's worth it. I'll think about if we can do in a different way

nirinchev:
You can cast to `IDisposable` and call the `Dispose` method.

LaPeste (approved):
An educational PR for me, since I haven't seen before the way C++ could be passed a callback from C#.
The unit test seems more than reasonable.
In summary I have no comments.

Realm/Realm/Realm.cs
```diff
@@ -198,6 +198,7 @@ private static bool IsRealmOpen(string path)
 
         #endregion static
 
+        [SuppressMessage("Usage", "CA2213:Disposable fields should be disposed", Justification = "A realm is responsible for disposing of state as it could be shared")]
```
nirinchev, line 201: This comment is a bit unclear - it says that "a realm is responsible for disposing", but it attempts to explain why the Realm will not dispose the state. I would suggest rewording it to something like "A State can be shared between multiple Realm instances. It is disposed when the native instance and its BindingContext is destroyed and OnBindingContextDestructedCallback is invoked."

wrappers/src/shared_realm_cs.cpp
```diff
@@ -43,13 +43,19 @@ namespace binding {
     void (*s_open_realm_callback)(void* task_completion_source, ThreadSafeReference* ref, int32_t error_code, const char* message, size_t message_len);
     void (*s_realm_changed)(void* managed_state_handle);
     void (*s_get_native_schema)(SchemaForMarshaling schema, void* managed_callback);
+    void (*s_free_gc_handle)(void* managed_handle);
```
nirinchev, line 46: This variable should be renamed to loosely match the C# name.

nirinchev (approved):
Two nits from me, looks good otherwise.

Realm/Realm/Realm.cs
papafe: Agreed, it should have been "A realm is not responsible"

papafe:
Corrected according to @nirinchev suggestions. I made the `Justification` for the `SuppressMessage` shorter, otherwise it feels more like a full fledged comment in my opinion

nirinchev (approved):

PR #2200, opened 2021-01-25
Fix date marshaling

nirinchev:
## Description

Fixes an issue reported on Forums where date properties are not read correctly from pre-beta.5 versions of Realm.

##  TODO

* [x] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Native/PrimitiveValue.cs
+++ b/Realm/Realm/Native/PrimitiveValue.cs
@@ -58,6 +58,9 @@ namespace Realms.Native
         private BinaryValue data_value;
 
         [FieldOffset(0)]
+        private TimestampValue timestamp_value;
+
+        [FieldOffset(0)]
         private LinkValue link_value;
 
         [FieldOffset(16)]
@@ -104,7 +107,7 @@ namespace Realms.Native
         public static PrimitiveValue Date(DateTimeOffset value) => new PrimitiveValue
         {
             Type = RealmValueType.Date,
-            int_value = value.ToUniversalTime().Ticks
+            timestamp_value = new TimestampValue(value.ToUniversalTime().Ticks)
         };
 
         public static PrimitiveValue NullableDate(DateTimeOffset? value) => value.HasValue ? Date(value.Value) : Null();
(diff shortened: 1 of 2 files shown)


Realm/Realm/Native/PrimitiveValue.cs
```diff
@@ -271,5 +274,25 @@ private struct LinkValue
...
         }
+
+        [StructLayout(LayoutKind.Sequential)]
+        private struct TimestampValue
+        {
+            private const long UnixEpochTicks = 621355968000000000;
+            private const long TicksPerSecond = 10000000;
+            private const long NanosecondsPerTick = 100;
+
+            private long seconds;
```
papafe, line 285: Shouldn't these two be public?

nirinchev: This struct is only used for marshaling - private members are correctly marshaled by the runtime and there's a constructor that accepts ticks that sets `seconds` and `nanoseconds` to the correct value based on the conversion formula.

So from a consumer's point of view, they should only care about the `ToTicks` and `FromTicks` (a.k.a. constructor) API.

LaPeste (approved):
No comment.
Just a question: how did this regression happen? Was the struct `TimestampValue` deleted or what happened?

Realm/Realm/Native/PrimitiveValue.cs
papafe: Makes sense. Maybe it'll make sense to still make those public to make it consistent with `StringValue`, `BinaryValue` and `LinkValue`

papafe (approved):
It makes sense, and also the constants seems to be correct. I also wonder how come it was not picked up by the unit tests.

nirinchev:
It was not picked up by tests because we don't have a test that verifies that values written with a previous versions are read correctly with the new one. Since the bug was symmetric (i.e. it affected both reading and writing values), results were consistent for Realm files created with the regressed version.

Realm/Realm/Native/PrimitiveValue.cs
nirinchev: I would rather make the fields in `StringValue` and friends private and encapsulate any relevant logic in those structs, but will do that in a separate PR.

PR #2201, opened 2021-01-25
Add support for value substitution in string based queries

### LINKED ISSUE #1822: Add support for value substitution in string based queries
Right now people have to manually serialize objects in the string-based query. We should expose a mechanism for value substitution similar to the one in the js sdk: `Filter("Birthday > $0 && Owner == $1", someDate, someOwner)`.

LaPeste:
## Description
Adding support for value substitution meant updating the wrapper to supply an array of mixed value to core. And from the C# side passing a marshable array of PrimitiveValue to the wrappers.
Documentation for the syntax of value substitution can be found [here](https://github.com/realm/realm-js/blob/master/docs/tutorials/query-language.md).

Fixes #1822

##  TODO
Nothing left

* [x] Changelog entry
* [x] Tests (if applicable)

=============================
UPDATE:
1. The test **QueryFilter_WithArgumentsUnmanagedObjects_ShouldThrow** is currently failing because the SDK does not throw when an in-memory object is passed; instead the SDK sends a `RealmValueType.Null` to core when an in-memory object is supplied. This will need to be corrected, but not by this PR.
2. The test **QueryFilter_WithArgumentsObject_ShouldMatch** is currently failing because of an issue in core.
3. The test **QueryFilter_WithAnyEmbeddedObjectArguments_ShouldMatch** fails for the same issue of test 2.


CHANGELOG.md
```diff
@@ -29,12 +29,17 @@
 * Fix a race condition which would lead to "uncaught exception in notifier thread: N5realm15InvalidTableRefE: transaction_ended" and a crash when the source Realm was closed or invalidated at a very specific time during the first run of a collection notifier (Core upgrade)
 
 ### Enhancements
+* Add support for the `GUID` data type. It can be used as primary key and is indexable. (PR [#2120](https://github.com/realm/realm-dotnet/pull/2120))
+* Replaced the implementation of the string query parser (the one used for [`realm.All().Filter("some-string-query")`](https://docs.mongodb.com/realm-sdks/dotnet/10.0.0-beta.3/reference/Realms.CollectionExtensions.html#Realms_CollectionExtensions_Filter__1_System_Linq_IQueryable___0__System_String_)). This results in ~5% reduction of the size of the native binary while keeping the query execution times on par with the old parser. (PR [#2185](https://github.com/realm/realm-dotnet/pull/2185), Core upgrade)
+* Added support for value substitution in string based queries. So [`realm.All<T>().Filter("field1 = $0 && field2 = $1", value1, value2)`](https://github.com/realm/realm-js/blob/master/docs/tutorials/query-language.md) is now possible
```
LaPeste, line 34: I don't know why what's above my line looks like added by me, but I clearly only added this third point in this pr.

papafe: Mmh, I think it happened to me too at some point. Maybe you did not branche from master when you created this branch?

CHANGELOG.md
```diff
@@ -29,12 +29,17 @@
...
 * Replaced the implementation of the string query parser (the one used for [`realm.All().Filter("some-string-query")`](https://docs.mongodb.com/realm-sdks/dotnet/10.0.0-beta.3/reference/Realms.CollectionExtensions.html#Realms_CollectionExtensions_Filter__1_System_Linq_IQueryable___0__System_String_)). This results in ~5% reduction of the size of the native binary while keeping the query execution times on par with the old parser. (PR [#2185](https://github.com/realm/realm-dotnet/pull/2185), Core upgrade)
 
 ### Compatibility
 * Realm Studio: 10.0.0 or later.
 
 ### Internal
+* Using Core 10.3.0.
```
papafe, line 41: Also here there is something strange

LaPeste: You're right, that should be 10.3.3, according to master. I'm not sure why that is. I'd really like to learn where the problem comes from and get it done right once and for all.

CHANGELOG.md
LaPeste: No I did not branch from master, but from the bison branch, since I needed that.
What's strange is that if you look at the commits, they count also the once coming from where I branched. I am not sure why that is. It leaves me a bit confused.

papafe:
Regarding the "unexpected" stuff in the merge, I think it's due to the fact that Github uses[ three-dots comparison](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-comparing-branches-in-pull-requests) instead of two-dots. [Here](https://matthew-brett.github.io/pydagogue/git_diff_dots.html) it's an explanation of the difference. 
So I suspect you created this branch before you merged the bson one, and that's why you're getting this result. Probably the best thing to do would be to merge first master into your branch, and this should fix the problem.
It's an annoying kind of comparison though :)

CHANGELOG.md
nirinchev: I would suggest doing interactive rebase on top of master as opposed to doing a merge. The issue is that the branch you branched off of got squashed into `master`, which means that it now sees two divergent sequences of changes that are in fact identical.

LaPeste: Oh, you're totally right. Merge was the mistake, rebase is what I should have done. Note taken for the next time.

.editorconfig
```diff
@@ -12,7 +12,7 @@ indent_style = space
 tab_width = 4
 
 # New line preferences
-end_of_line = crlf
+end_of_line = lf
```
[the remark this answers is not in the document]
LaPeste: Yes, correct.

.gitignore
```diff
@@ -198,3 +198,4 @@ Realm/Realm.Unity/Runtime/Dependencies/*.meta
 Realm/Realm.Unity/Runtime/Dependencies.meta
 Tests/PerformanceTests/BenchmarkDotNet.Artifacts/
 Realm/Realm.Unity/**/*.tgz
+wrappers/src/object-store/
```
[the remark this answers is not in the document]
LaPeste: Git keeps asking me to add that folder. Since we don't want it pushed, I added it to the .gitignore file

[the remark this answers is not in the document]
nirinchev: You should just delete it - that folder shouldn't exist.

Tests/Realm.Tests/Database/CollectionTests.cs
```diff
@@ -713,7 +828,7 @@ public void Query_Freeze_ReturnsAFrozenCopy()
                 _realm.Add(new Dog { Name = "Betazaurus" });
             });
 
-            var query = _realm.All<Dog>().Where(d => d.Name.StartsWith("B"));
+            var query = _realm.All<Dog>().Where(d => d.Name.StartsWith("B", StringComparison.Ordinal));
```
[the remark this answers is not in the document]
nirinchev: We shouldn't change this - it's not super critical here, but we don't know if users will follow best practices and always specify a string comparison.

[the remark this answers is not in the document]
LaPeste: Shall we then try to suppress the warning concerning the lack of a `StringComparison`?

nirinchev (changes requested):
Good start. I have suggestions on how to simplify the tests with using a little bit of reflection. In addition to the proposed tests, it will be a good idea to add:
1. Tests for any failure scenarios/ambiguous scenarios you can think of - such as passing `null` in the arguments or passing incorrect value types - e.g. a property is a type `string` and you try to compare it to a bool argument.
2. Test where we pass multiple arguments, ideally of different types.
3. Tests where we compare numeric types of different sizes - e.g. `int` property to a `int64` argument or `double` with `int`, etc. Some of these should convert automatically, others should error out.

Tests/Realm.Tests/Database/CollectionTests.cs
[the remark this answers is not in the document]
nirinchev: I thought it's suppressed - at least in .editorconfig the severity is set to `none`:

https://github.com/realm/realm-dotnet/blob/725dee61c439defbf6ce8980c944e59333e77232/.editorconfig#L226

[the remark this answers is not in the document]
LaPeste: You're correct in all you said but I still get that CA1307

image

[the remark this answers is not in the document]
LaPeste: It's CA1310. I'm gonna add it to `.editorconfig`.

wrappers/src/results_cs.cpp
```diff
@@ -128,8 +128,17 @@ REALM_EXPORT Results* results_get_filtered_results(const Results& results, uint1
...
 
-        query_parser::NoArguments no_args;
-        Query parsed_query = results.get_table()->query(query_string, no_args, mapping);
+        std::vector<Mixed> mixed_args;
+        for (int i = 0; i < args_count; ++i ) {
+            if (arguments[i].type != realm_value_type::RLM_TYPE_LINK) {
+                mixed_args.push_back(from_capi(arguments[i]));
+            }
+            else {
+                mixed_args.push_back(from_capi(arguments[i].link.object, false));
```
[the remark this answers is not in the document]
LaPeste: This took me a little to figure out. I initially started with true, as you recommended. But in that way it was not working. Digging deeper it turned out that a comparison between objects can only be done between [Link](https://github.com/realm/realm-core/blob/6311eb2028fa34ff714c00c47dbbea34c5dbdbf7/src/realm/data_type.hpp#L142) and not [TypedLink](https://github.com/realm/realm-core/blob/6311eb2028fa34ff714c00c47dbbea34c5dbdbf7/src/realm/data_type.hpp#L145)

You can see that [here](https://github.com/realm/realm-core/blob/6311eb2028fa34ff714c00c47dbbea34c5dbdbf7/src/realm/parser/driver.cpp#L945). There's no case for `type_Typed_Link`.

If you think this is a mistake, I'll contact who you recommended to understand better the situation.

Realm/Realm/Extensions/CollectionExtensions.cs
```diff
@@ -179,10 +180,10 @@ public static void Move<T>(this IList<T> list, int from, int to)
         /// Examples of the NSPredicate syntax
         /// </seealso>
         /// <seealso href="https://academy.realm.io/posts/nspredicate-cheatsheet/">NSPredicate Cheatsheet</seealso>
-        public static IQueryable<T> Filter<T>(this IQueryable<T> query, string predicate)
+        public static IQueryable<T> Filter<T>(this IQueryable<T> query, string predicate, params RealmValue[] arguments)
         {
             var realmResults = Argument.EnsureType<RealmResults<T>>(query, $"{nameof(query)} must be a query obtained by calling Realm.All.", nameof(query));
-            return realmResults.GetFilteredResults(predicate);
+            return realmResults.GetFilteredResults(predicate, arguments);
```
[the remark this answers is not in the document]
LaPeste: I was thinking to throw an exception (`Argument.NotNull(obj, paramName)`). Would that be fine or would you just return an empty result and let the dev figure out the why?

Tests/Realm.Tests/Database/CollectionTests.cs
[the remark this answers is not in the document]
nirinchev: Might be better to add it to global suppressions instead.

Realm/Realm/Extensions/CollectionExtensions.cs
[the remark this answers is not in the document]
nirinchev: Yes - Argument.NotNull sounds correct.

Realm/Realm/Handles/ResultsHandle.cs
```diff
@@ -62,7 +63,7 @@ private static class NativeMethods
             public static extern IntPtr snapshot(ResultsHandle results, out NativeException ex);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "results_get_filtered_results", CallingConvention = CallingConvention.Cdecl)]
-            public static extern IntPtr get_filtered_results(ResultsHandle results, [MarshalAs(UnmanagedType.LPWStr)] string query_buf, IntPtr query_len, out NativeException ex);
+            public static extern IntPtr get_filtered_results(ResultsHandle results, [MarshalAs(UnmanagedType.LPWStr)] string query_buf, IntPtr query_len, PrimitiveValue[] arguments, int args_count, out NativeException ex);
```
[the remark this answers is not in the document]
LaPeste: I did some research on these _Directional Attributes_ and _Marshalling types_.
Maybe we should be more specific about the "In / Out" parameter for all other parameters, just for consistency. We'll discuss in person (call) about this. But the bottom line is, `In` seems correct to me and same for UnmanagedType.LPArray.

Tests/Realm.Tests/Database/CollectionTests.cs
```diff
@@ -16,10 +16,13 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+#nullable enable
```
nirinchev, line 19: Let's not use nullable reference annotations yet - while I agree we should get it done at some point, I'd prefer if we did it for all our code at the same time rather than bit by bit.

Tests/Realm.Tests/Database/CollectionTests.cs
```diff
@@ -411,6 +414,258 @@ public void Results_GetFiltered_SanityTest()
...
+                // boxed value types can't be casted to Decimal128
+                if (PropertyName == "Decimal128Property")
+                {
+                    return (TryToConvert(MatchingValue, (value) => value.AsDecimal128()), TryToConvert(NonMatchingValue, (value) => value.AsDecimal128()));
+                }
+                else
+                {
+                    return (TryToConvert(MatchingValue.AsAny(), targetType), TryToConvert(NonMatchingValue.AsAny(), targetType));
+                }
+            }
```
nirinchev, line 448: Instead of this, let's try a simpler approach:

```csharp
public static object BoxValue(RealmValue val, Type targetType)
{
    var boxed = val.AsAny();
    if (boxed != null && boxed.GetType() != targetType)
    {
        boxed = Convert.ChangeType(boxed, targetType);
    }

    return boxed;
}

// And we use it as:

var boxedMatch = data.BoxValue(data.MatchingValue, prop.PropertyType);
var boxedNonMatch = data.BoxValue(data.NonMatchingValue, prop.PropertyType);
```

Tests/Realm.Tests/Database/CollectionTests.cs
```diff
@@ -411,6 +414,258 @@ public void Results_GetFiltered_SanityTest()
...
+            yield return new StringQueryTestData(nameof(AllTypesObject.DecimalProperty), 0.9999999999999999999999999999, 0.3333333333333333333333333333);
+            yield return new StringQueryTestData(nameof(AllTypesObject.Decimal128Property), new Decimal128(564.42343424323), new Decimal128(666.42300000003));
+            yield return new StringQueryTestData(nameof(AllTypesObject.ObjectIdProperty), new ObjectId("5f64cd9f1691c361b2451d96"), new ObjectId("5ffffffffffff22222222222"));
+            yield return new StringQueryTestData(nameof(AllTypesObject.GuidProperty), new Guid("0f8fad5b-d9cb-469f-a165-70867728950e"), new Guid("0ffffffb-dddd-4444-aaaa-70000000000e"));
+            yield return new StringQueryTestData(nameof(AllTypesObject.StringProperty), "hello world", "no salute");
+            yield return new StringQueryTestData(nameof(AllTypesObject.ByteArrayProperty), new byte[] { 0x5, 0x4, 0x3, 0x2, 0x1 }, new byte[] { 0x1, 0x1, 0x1 });
+        }
+
+        public static IEnumerable<StringQueryTestData> StringQueryTestCases_NumbericValues_ImplicitConvertion_Matches()
```
nirinchev, line 494: Typo: `Numberic` -> `Numeric`

nirinchev: Also, instead of reusing `StringQueryTestData`, perhaps create a different type for these - it will allow you to be more explicit about what the test is doing:
1. Instead of overloading the meaning of `NonMatchingValue`, you can call it something more meaningful for the test - e.g. `ValueToQueryFor`.
2. Instead of having 3 different tests for matches, non-matches due to value difference and non-matches due to type difference, you can add `ExpectedMatches` field to the test object.

Tests/Realm.Tests/Database/CollectionTests.cs
```diff
@@ -411,6 +414,258 @@ public void Results_GetFiltered_SanityTest()
...
+                _realm.Add(new ObjectPropertyObject { Object = new PrimaryKeyInt32Object { Int32Property = 42 } });
+            });
+
+            Assert.Throws<RealmException>(() => _realm.All<ObjectPropertyObject>().Filter("Object = $0", new PrimaryKeyInt32Object { Int32Property = 42 }), "**put your message here!**");
+        }
+
+        [Test]
+        public void QueryFilter_WithNullArguments_ShouldThrow()
+        {
+            Assert.Throws<ArgumentNullException>(() => _realm.All<Dog>().Filter("Name = $0", null));
```
nirinchev, line 650: This test passes due to a bug/peculiarity in the way implicit RealmValue conversions happen. Right now `null` is not implicitly converted to `RealmValue` so the compiler will interpret this as the `RealmValue[]` argument being null. When/If we manage to fix the implicit conversion issue, the test will start to fail. To ensure it's testing what you want it to test, it may make sense to assign it to a variable first - `RealmValue[] argumentsArray = null` and then pass that to `Filter`.

Tests/Realm.Tests/Database/CollectionTests.cs
```diff
@@ -411,6 +414,258 @@ public void Results_GetFiltered_SanityTest()
...
+                _realm.Add(new ObjectPropertyObject { Object = nonMatchingObj });
+            });
+
+            var matches = _realm.All<ObjectPropertyObject>().Filter("Object = $0", matchingObj);
+            var foundObj = matches.Single();
+            Assert.AreEqual(matchingObj, foundObj.Object);
+        }
+
+        [Test]
+        public void QueryFilter_WithArgumentsInMemoryObjects_ShouldThrow()
```
nirinchev, line 637: `InMemory` -> `Unmanaged`

nirinchev (commented):
Some final comments mostly related to tests/simplifying the conversion logic as that was quite hard to grasp/follow. In addition to that, some extra tests you might want to add:

1. Tests for object matches that use embedded objects. You can use `ObjectWithEmbeddedProperties` and `EmbeddedAllTypesObject`. Then, the tests should verify that this works: `realm.All<ObjectWithEmbeddedProperties>().Filter("AllTypesObject = $0", theEmbedded)`
2. Test that `null` is passed correctly as a value. e.g. `realm.All<Foo>().Filter("Prop = null")` should be equivalent by `realm.All<Foo>().Filter("Prop = $0", (int?)null)` (assuming `Prop` is `int?`). Note that you need the explicit cast as implicit null casts don't work yet and may not work due to reasons @papafe will explain on Wednesday.

Tests/Realm.Tests/Database/CollectionTests.cs
LaPeste: So you'd like to have only one unit test, for the sake of argument called `QueryFilter_WithNumericArguments`, which does not specify in the name what is the expected outcome (shouldThrow, shouldMatch, shouldNotMatch), instead of 

1. `QueryFilter_WithImplicitlyConvertibleNumericArguments_ShouldMatch`
2. `QueryFilter_WithImplicitlyConvertibleNumericArguments_ShouldNotMatch`
3. `QueryFilter_WithNonImplicitlyConvertibleArguments_ShouldNotMatch`

Did I understand correctly?

nirinchev: yep

Tests/Realm.Tests/Database/CollectionTests.cs
```diff
@@ -411,6 +412,317 @@ public void Results_GetFiltered_SanityTest()
...
+            var boxed = val.AsAny();
+            if (boxed != null && boxed.GetType() != targetType)
+            {
+                try
+                {
+                    boxed = Convert.ChangeType(boxed, targetType);
+                }
+                catch (System.InvalidCastException e)
+                {
+                    if (e.Message.Contains("RealmInteger"))
```
LaPeste, line 457: This is added so that `RealmInteger` can be processed by the unit tests already in place.
Even though at line 577 in test `QueryFilter_WithAnyArguments_ShouldMatch` we just pass a RealmValue I still consider the test correct because the implicit conversion operator of RealmValue is already put at work during the creation of the test data (line 504-507).

Realm/Realm/Extensions/CollectionExtensions.cs
```diff
@@ -210,16 +211,18 @@ public static IDisposable SubscribeForNotifications<T>(this IDictionary<string,
         /// var results1 = realm.All&lt;Foo&gt;("Bar.IntValue > 0");
         /// var results2 = realm.All&lt;Foo&gt;("Bar.IntValue > 0 SORT(Bar.IntValue ASC Bar.StringValue DESC)");
         /// var results3 = realm.All&lt;Foo&gt;("Bar.IntValue > 0 SORT(Bar.IntValue ASC Bar.StringValue DESC) DISTINCT(Bar.IntValue)");
+        /// var results4 = realm.All&lt;Foo&gt;("Bar.IntValue > $0 || (Bar.String == $1 &amp;&amp; Bar.Bool == $2)" 5, "small", true");
```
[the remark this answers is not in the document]
LaPeste: How is this posted only now? I do remember not seeing the suggestion after the talk in the call. Anyway, it's already fixed.

PR #2227, opened 2021-02-04
Fix analytics version

nirinchev:
In v10 we're sending the SDK version to analytics - this fixes it and removes a bunch of Guid/Set tests so that we don't get test failures on release PRs.


Realm/Realm.Unity/Editor/Common/Analytics.cs
```diff
@@ -117,7 +117,7 @@ private string JsonPayload
                     .Replace("%SYNC_ENABLED%", _isSyncEnabled.ToString())
 
                     // Version of weaver is expected to match that of the library.
-                    .Replace("%REALM_VERSION%", "System.Reflection.Assembly.GetExecutingAssembly().GetName().Version.ToString()")
```
LaPeste, line 120: Can I assume that in the past `"System.Reflection.Assembly.GetExecutingAssembly().GetName().Version.ToString()"` was evaluated at a later time? Otherwise how could it work if passed as a string? Just curiosity

nirinchev: This is what the PR fixes - it shouldn't have been inside quotes 😅

LaPeste (approved):

PR #2229, opened 2021-02-04
Log warning in Weaver when using UUID/ISet/IDictionary property

### LINKED ISSUE #2191: Log warning in Weaver when using UUID/ISet/IDictionary property
We don't want to fully disable the ability to use the new datatypes but should log an explicit warning. Should be something like `Foo.Bar is of type Guid which is not officially supported yet. Some functionality may not exist yet or there may be bugs/known issues that can result in undefined behavior, including data loss. Official support for the datatype will come in a future Realm release.`

LaPeste:
## Description
Adds warning at the weaver level when it's encountered a new datatype that isn't "officially" supported. These types are Guid, ISet and IDictionary

Fixes #2191 

##  TODO

* [ ] Changelog entry
* [x] Tests (if applicable)

With tests I simply mean that all the warnings are added to `MatchErrorsAndWarnings`.
I don't think we need to mention anything in the changelog. Please, correct me if I'm wrong.


Realm/Realm.Unity/Editor/Common/RealmWeaver.cs
```diff
@@ -47,6 +47,7 @@ internal partial class Weaver
         internal const string DecimalTypeName = "System.Decimal";
         internal const string Decimal128TypeName = "MongoDB.Bson.Decimal128";
         internal const string ObjectIdTypeName = "MongoDB.Bson.ObjectId";
+        internal const string DateTimeTypeName = "System.DateTime";
```
nirinchev, line 50: Let's de-const this to avoid confusing it with supported types.

nirinchev (approved):

PR #2214, opened 2021-02-01
Add support for dictionaries of objects

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2210

##  TODO

* [ ] Changelog entry
* [x] Tests (if applicable)

PR #2230, opened 2021-02-05
Refactor RealmMetadata and remove TableHandle

### LINKED ISSUE #2213: Refactor Realm Metadata
At the moment `Realm.Metadata` is just a `Dictionary<string, RealmObjectBase.Metadata> Metadata`. It would be convenient to refactor it into a full fledged class, so it will be possible to add additional functionality and utility methods. In particular, we need to implement a method that allows to retrieve a `RealmObject`'s metadata from his `table_key`.

papafe:
This PR refactors `RealmMetadata` to create a correspondence between the `RealmObject.Metadata` and the `TableKey`. Given that with the `TableKey` we can get a reference to the table in native, `TableHandle` has been removed too.

Fixes #2213


Realm/Realm/Handles/ObjectHandle.cs
```diff
@@ -66,7 +66,7 @@ private static class NativeMethods
             public static extern IntPtr get_backlinks(ObjectHandle objectHandle, IntPtr property_index, out NativeException nativeException);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_get_backlinks_for_type", CallingConvention = CallingConvention.Cdecl)]
-            public static extern IntPtr get_backlinks_for_type(ObjectHandle objectHandle, TableHandle source_table, IntPtr source_property_index, out NativeException nativeException);
+            public static extern IntPtr get_backlinks_for_type(ObjectHandle objectHandle, TableKey tableKey, IntPtr source_property_index, out NativeException nativeException);
```
nirinchev, line 69:
```suggestion
            public static extern IntPtr get_backlinks_for_type(ObjectHandle objectHandle, TableKey table_key, IntPtr source_property_index, out NativeException nativeException);
```

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -100,8 +100,8 @@ private static class NativeMethods
             [return: MarshalAs(UnmanagedType.U1)]
             public static extern bool refresh(SharedRealmHandle sharedRealm, out NativeException ex);
 
-            [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_get_table", CallingConvention = CallingConvention.Cdecl)]
-            public static extern IntPtr get_table(SharedRealmHandle sharedRealm, [MarshalAs(UnmanagedType.LPWStr)] string tableName, IntPtr tableNameLength, out NativeException ex);
+            [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_get_table_key", CallingConvention = CallingConvention.Cdecl)]
+            public static extern void get_table_key(SharedRealmHandle sharedRealm, [MarshalAs(UnmanagedType.LPWStr)] string tableName, IntPtr tableNameLength, out TableKey tableKey, out NativeException ex);
```
nirinchev, line 104: Try typedefing `realm_class_key` similar to what the C API does: https://github.com/realm/realm-core/pull/3955/files#diff-ac5ca02b88ba9910d618bac409ad7a1e3b281cdaf9f64840469c21870b7114a1R74

and have it as a return value.

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -133,10 +133,10 @@ private static class NativeMethods
             public static extern void write_copy(SharedRealmHandle sharedRealm, [MarshalAs(UnmanagedType.LPWStr)] string path, IntPtr path_len, byte[] encryptionKey, out NativeException ex);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_create_object", CallingConvention = CallingConvention.Cdecl)]
-            public static extern IntPtr create_object(SharedRealmHandle sharedRealm, TableHandle table, out NativeException ex);
+            public static extern IntPtr create_object(SharedRealmHandle sharedRealm, TableKey tableKey, out NativeException ex);
```
nirinchev, line 136: snake_case for argument names

Realm/Realm/Native/TableKey.cs
```diff
@@ -22,27 +22,32 @@
...
 {
     [StructLayout(LayoutKind.Sequential)]
-    internal struct ObjectKey : IEquatable<ObjectKey>
+    internal struct TableKey : IEquatable<TableKey>
     {
-        private Int64 value;
+        private uint value;
 
-        public bool Equals(ObjectKey other) => value.Equals(other.value);
+        public uint Value => value;
```
nirinchev, line 29: Not used - let's remove

Realm/Realm/Realm.cs
```diff
@@ -1364,6 +1359,46 @@ internal void ExecuteOutsideTransaction(Action action)
 
         #endregion Transactions
 
+        internal class RealmMetadata
+        {
+            private readonly Dictionary<string, RealmObjectBase.Metadata> stringToRealmObjectMetadataDict;
+            private readonly Dictionary<TableKey, RealmObjectBase.Metadata> tableKeyToRealmObjectMetadataDict;
+
+            public IEnumerable<RealmObjectBase.Metadata> Values => stringToRealmObjectMetadataDict.Values;
```
nirinchev, line 1367: TODO: check where/why this is used.

Realm/Realm/Realm.cs
```diff
@@ -1364,6 +1359,46 @@ internal void ExecuteOutsideTransaction(Action action)
...
+
+            public bool TryGetValue(string objectType, out RealmObjectBase.Metadata metadata)
+            {
+                return stringToRealmObjectMetadataDict.TryGetValue(objectType, out metadata);
+            }
+
+            public bool TryGetValue(TableKey tablekey, out RealmObjectBase.Metadata metadata)
+            {
+                return tableKeyToRealmObjectMetadataDict.TryGetValue(tablekey, out metadata);
+            }
```
nirinchev, line 1389: These should be expression-bodied

nirinchev: TODO: research the difference between expression bodied methods and regular ones.

Realm/Realm/Realm.cs
```diff
@@ -1364,6 +1359,46 @@ internal void ExecuteOutsideTransaction(Action action)
...
+
+            public RealmObjectBase.Metadata this[string objectType]
+            {
+                get => stringToRealmObjectMetadataDict[objectType];
+            }
+
+            public RealmObjectBase.Metadata this[TableKey tablekey]
+            {
+                get => tableKeyToRealmObjectMetadataDict[tablekey];
+            }
```
nirinchev, line 1399: TODO: check if these are safely used.

wrappers/src/object_cs.cpp
```diff
@@ -154,11 +154,13 @@ extern "C" {
         });
     }
 
-    REALM_EXPORT Results* object_get_backlinks_for_type(Object& object, TableRef& source_table, size_t source_property_ndx, NativeException::Marshallable& ex)
+    REALM_EXPORT Results* object_get_backlinks_for_type(Object& object, TableKey tableKey, size_t source_property_ndx, NativeException::Marshallable& ex)
     {
         return handle_errors(ex, [&] {
             verify_can_get(object);
 
+            auto source_table = object.realm()->read_group().get_table(tableKey);
```
nirinchev, line 162: maybe `const` here?

wrappers/src/shared_realm_cs.cpp
```diff
@@ -247,11 +247,12 @@ REALM_EXPORT void shared_realm_close_realm(SharedRealm& realm, NativeException::
...
 }
 
-REALM_EXPORT TableRef* shared_realm_get_table(SharedRealm& realm, uint16_t* object_type_buf, size_t object_type_len, NativeException::Marshallable& ex)
+REALM_EXPORT void shared_realm_get_table_key(SharedRealm& realm, uint16_t* object_type_buf, size_t object_type_len, TableKey& tableKey, NativeException::Marshallable& ex)
 {
     return handle_errors(ex, [&]() {
         Utf16StringAccessor object_type(object_type_buf, object_type_len);
-        return new TableRef(ObjectStore::table_for_object_type(realm->read_group(), object_type));
+        tableKey = ObjectStore::table_for_object_type(realm->read_group(), object_type)->get_key();
+        return;
```
nirinchev, line 255: unneeded return here

wrappers/src/shared_realm_cs.cpp
```diff
@@ -354,20 +355,23 @@ REALM_EXPORT void shared_realm_write_copy(SharedRealm* realm, uint16_t* path, si
...
 }
 
-REALM_EXPORT Object* shared_realm_create_object_unique(const SharedRealm& realm, TableRef& table, realm_value_t primitive, bool try_update, bool& is_new, NativeException::Marshallable& ex)
+REALM_EXPORT Object* shared_realm_create_object_unique(const SharedRealm& realm, TableKey tableKey, realm_value_t primitive, bool try_update, bool& is_new, NativeException::Marshallable& ex)
 {
     return handle_errors(ex, [&]() {
         realm->verify_in_write();
         realm->read_group();
+
+        auto table = realm->read_group().get_table(tableKey);
```
nirinchev, line 373: Create an inlined helper method like `get_table(SharedRealm&, TableKey)`

wrappers/src/shared_realm_cs.cpp
```diff
@@ -444,4 +448,45 @@ REALM_EXPORT SharedRealm* shared_realm_freeze(const SharedRealm& realm, NativeEx
...
 }
 
+REALM_EXPORT Object* shared_realm_get_object_for_primary_key(SharedRealm& realm, TableKey tableKey, realm_value_t primitive, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() -> Object* {
+        realm->verify_thread();
+
+        auto table = realm->read_group().get_table(tableKey);
+        const std::string object_name(ObjectStore::object_type_for_table_name(table->get_name()));
+        auto& object_schema = *realm->schema().find(object_name);
```
nirinchev, line 458: Add `find(TableKey)` in OS's Schema class and use that instead.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -444,4 +448,45 @@ REALM_EXPORT SharedRealm* shared_realm_freeze(const SharedRealm& realm, NativeEx
...
+        if (object_schema.primary_key.empty()) {
+            const std::string name(table->get_name());
```
nirinchev, line 460: Pass `ObjectStore::object_type_for_table_name(table->get_name())`

wrappers/src/shared_realm_cs.cpp
```diff
@@ -444,4 +448,45 @@ REALM_EXPORT SharedRealm* shared_realm_freeze(const SharedRealm& realm, NativeEx
...
+        return new Object(realm, object_schema, table->get_object(obj_key));
+    });
+}
+
+REALM_EXPORT Results* shared_realm_create_results(SharedRealm& realm, TableKey tableKey, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        realm->verify_thread();
+
+        auto table = realm->read_group().get_table(tableKey);
```
nirinchev, line 487: const

wrappers/src/table_cs.cpp
```diff
@@ -51,57 +51,4 @@ REALM_EXPORT Object* table_add_empty_object(TableRef& table, SharedRealm& realm,
     });
 }
 
-REALM_EXPORT Results* table_create_results(TableRef& table, SharedRealm& realm, NativeException::Marshallable& ex)
```
nirinchev, line 54: This file should be deleted.

Realm/Realm/Realm.cs
papafe: It is used in `realm.RemoveAll()`. It makes sense that it is used there, but probably would be more appropriate to transform the property in a method called `GetAllRealmObjectsMetadata()` or similar

nirinchev: It's probably best to just implement `RemoveAll` natively. It's obviously not a hot path, but it may have peculiar side effects - e.g. if you open the Realm file with a subset of its schema, you likely expect `RemoveAll` to leave the file empty, but it'll remove only some data.

You can either leave it as is and open an issue to refactor `RemoveAll` or go ahead and fix it in this PR 😛

papafe: I have added a new issue (#2233). Too much stuff already in this PR :)

wrappers/src/shared_realm_cs.cpp
papafe: I have just tried to implement this functionality as we discussed with @fealebenpae, by creating a map of `TableKey` to `ObjectSchema` when the `Schema` is created. Unfortunately the `table_key` field of `ObjectSchema` is not assigned yet in the constructor of `Schema`, and so we cannot create the map at that time. 
Given this, probably the fastest option would be to keep the code as it was as two operations are performed:

- `ObjectStore::object_type_for_table_name(table->get_name())` - O(1) time
- `realm->schema().find(object_name)` - O(log n) time (it uses binary search inside).

In alternative we can still implement find(TableKey) method, in two ways:

- We do a linear search - O(n) time
- We build the map the first time the method is called, and then use the map the following times - O(1) time + build time on first call

nirinchev: The big-O estimates for the string case appear to ignore the string operations themselves. Not a C++ expert, but I don't believe the language uses any sort of pooling, which means that constructing the string itself and comparing it is a O(m) operation, which is exacerbated by the fact all table names start with `class_`.

Perhaps ping @fealebenpae for suggestions as that's ultimately code his team will review and maintain 😄

papafe: As discussed with @nirinchev, we have decided to just use the linear search for now, then later we could eventually improve it

wrappers/src/object_cs.cpp
```diff
@@ -154,11 +144,13 @@ extern "C" {
...
 
-    REALM_EXPORT Results* object_get_backlinks_for_type(Object& object, TableRef& source_table, size_t source_property_ndx, NativeException::Marshallable& ex)
+    REALM_EXPORT Results* object_get_backlinks_for_type(Object& object, TableKey table_key, size_t source_property_ndx, NativeException::Marshallable& ex)
     {
         return handle_errors(ex, [&] {
             verify_can_get(object);
 
+            const TableRef source_table = get_table(object.realm(), table_key);
+
             const ObjectSchema& source_object_schema = *object.realm()->schema().find(ObjectStore::object_type_for_table_name(source_table->get_name()));
```
nirinchev, line 154: This can use the new `find(TableKey)` method.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -354,20 +356,22 @@ REALM_EXPORT void shared_realm_write_copy(SharedRealm* realm, uint16_t* path, si
...
+REALM_EXPORT Object* shared_realm_create_object_unique(const SharedRealm& realm, TableKey table_key, realm_value_t primitive, bool try_update, bool& is_new, NativeException::Marshallable& ex)
 {
     return handle_errors(ex, [&]() {
         realm->verify_in_write();
-        realm->read_group();
+
+        const TableRef table = get_table(realm, table_key);
+
         const StringData object_name(ObjectStore::object_type_for_table_name(table->get_name()));
         const ObjectSchema& object_schema = *realm->schema().find(object_name);
```
nirinchev, line 376: `find(table_key)`

wrappers/src/sort_descriptor_cs.cpp
```diff
@@ -34,13 +35,13 @@ REALM_EXPORT void sort_descriptor_destroy(DescriptorOrdering* descriptor)
...
-REALM_EXPORT void sort_descriptor_add_clause(DescriptorOrdering& descriptor, TableRef& table, SharedRealm& realm, size_t* property_chain, size_t properties_count, bool ascending, NativeException::Marshallable& ex)
+REALM_EXPORT void sort_descriptor_add_clause(DescriptorOrdering& descriptor, TableKey table_key, SharedRealm& realm, size_t* property_chain, size_t properties_count, bool ascending, NativeException::Marshallable& ex)
 {
     handle_errors(ex, [&]() {
         std::vector<ColKey> column_keys;
         column_keys.reserve(properties_count);
 
-        const std::string object_name(ObjectStore::object_type_for_table_name(table->get_name()));
+        const std::string object_name(ObjectStore::object_type_for_table_name(get_table(realm, table_key)->get_name()));
         const std::vector<Property>* properties = &realm->schema().find(object_name)->persisted_properties;
```
nirinchev, line 45: This can also use `find(TableKey)`.

nirinchev (commented):
Mostly minor simplification suggestions from me, otherwise should be good to go once the Core PR is merged.

Realm/Realm/Native/TableKey.cs
```diff
@@ -22,27 +22,18 @@
...
-        {
-            switch (obj)
-            {
-                case ObjectKey other:
-                    return value.Equals(other.value);
-                default:
-                    return false;
-            }
-        }
+        public override bool Equals(object obj) => obj is TableKey other && value.Equals(other.value);
```
nirinchev, line 31:
```suggestion
        public override bool Equals(object obj) => obj is TableKey other && Equals(other);
```

nirinchev (approved):
Looks good but probably should be rebased on top of master to get the build fixed.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -362,22 +364,23 @@ REALM_EXPORT void shared_realm_write_copy(SharedRealm* realm, uint16_t* path, si
...
+REALM_EXPORT Object* shared_realm_create_object_unique(const SharedRealm& realm, TableKey table_key, realm_value_t primitive, bool try_update, bool& is_new, NativeException::Marshallable& ex)
 {
     return handle_errors(ex, [&]() {
         realm->verify_in_write();
-        realm->read_group();
-        const StringData object_name(ObjectStore::object_type_for_table_name(table->get_name()));
-        const ObjectSchema& object_schema = *realm->schema().find(object_name);
+
+        const TableRef table = get_table(realm, table_key);
+
```
LaPeste, line 382: Are all these new lines needed?

wrappers/src/shared_realm_cs.cpp
```diff
@@ -452,4 +455,44 @@ REALM_EXPORT SharedRealm* shared_realm_freeze(const SharedRealm& realm, NativeEx
...
+        }
+
+        if (!primitive.is_null() && to_capi(primary_key_property.type) != primitive.type) {
+            throw PropertyTypeMismatchException(object_schema.name, primary_key_property.name, to_string(primary_key_property.type), to_string(primitive.type));
+        }
+
+        const ColKey column_key = object_schema.primary_key_property()->column_key;
+        const ObjKey obj_key = table->find_first(column_key, from_capi(primitive));
+        if (!obj_key)
+            return nullptr;
```
LaPeste, line 482: Maybe this could be wrapped by curly brackets? I generally think it avoids issue.

LaPeste (changes requested):
Beside the 2 minuscule comments I had, one more. ObjectKey is renamed to TableKey and you changed the year on the copyright. I'm not sure what the policy is but I assume all the other files you modified need to be changed, as well? If not, what's the policy about copyrights?

Anyway, beside the very small things mentioned above, PR looks good to me.

papafe:
@LaPeste agree on both points. Regarding ObjectKey renamed to TableKey... it's just git that thinks that I renamed it. I actually deleted ObjectKey and added TableKey, but because they are similar git thinks I have renamed it, that's why the copyright difference.

papafe:
This PR is considered mergeable, once the core changes have been approved (https://github.com/realm/realm-core/pull/4413) and merged to the v11 branch

LaPeste (approved):

nirinchev:
The test failures are caused by a change in the Core query syntax. They'll be fixed once https://github.com/realm/realm-dotnet/pull/2214 is merged.

nirinchev:
Failing tests seem to be my fault - the AOT dynamic runtime doesn't seem to like my code.

PR #2231, opened 2021-02-08
Bump core version

nirinchev:


LaPeste (approved):
Is the v5 the branch for the legacy sdk?

nirinchev:
Yep - a user saw some assertions from Core that were fixed in v10, but they're on legacy cloud, so needed those backported in a v5 release.

papafe (approved):
Looks good to me

PR #2232, opened 2021-02-08
Remove dashes from changelog

nirinchev:


LaPeste (approved):
Is this a new format that was decided upon or what? Beside different aesthetic, does it bring anything else to the table (e.g. some automated tool parsing this file etc)?

nirinchev:
Not really, just noticed that the dashes generated a second separator line and it looked ugly:

Screen Shot 2021-02-08 at 16 43 47

The heading 2 style on Github already has a separator, so no need to get an extra one.

PR #2234, opened 2021-02-08
Updated test projects to .netcore 3.1

### LINKED ISSUE #2117: Upgrade test projects to not use coreapp20
.NET Core 2.0 is EOL-ed, so we should migrate to 2.1 or 3.1.

LaPeste:
## Description
.netcore 2.0 is approaching EOL. This PR bumps up the version of .netcore to 3.1 wherever it's used.

Fixes #2117

##  TODO

* [x] Changelog entry


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 
 ### Internal
 * Using Core 10.3.3.
+* Updated to dotnetcore 3.1.
```
nirinchev, line 32: This should say that only test projects were updated and also make a note that we're no longer testing .NET Core 2.0.

global.json
```diff
@@ -1,5 +1,5 @@
 {
   "msbuild-sdks": {
-    "MSBuild.Sdk.Extras": "2.1.2"
+    "MSBuild.Sdk.Extras": "3.0.23"
```
nirinchev, line 3: I don't mind the change, just curious - why was it necessary? Since we previously targeted .NET 5, it's a bit surprising that the package didn't support 3.1.

LaPeste: There is a problem with importing `Microsoft.WinFX.props` on Linux. Reading [here](https://github.com/dotnet/sdk/issues/11108) there seem to be problems with MSBuildExtras in linux since April 2020 exactly with this `Microsoft.WinFX.props`. Even when the patch was in the sdk it didn't seem to work unless you "patch yourself" some settings. Before doing anything I just gave a shot to the latest version of MSBuild.Sdk.Extras just to see if that would have helped. Which it didn't, so I'll revert.

nirinchev: I see. I guess we can use the workaround from that post then.

Jenkinsfile
```diff
@@ -368,10 +368,14 @@ def NetCoreTest(String nodeName, String targetFramework) {
 
               script += " --baasurl http://mongodb-realm:9090 --baasappid ${appId.trim()}"
               // see https://stackoverflow.com/a/53782505
-              sh """
+              sh '''
+                # see https://github.com/dotnet/sdk/issues/11108
```
nirinchev, line 372: Can we move this comment outside the `sh` invocation and also expand on it a little so people don't have to go to open the linked issue to understand what it means.

nirinchev (commented):
Looks like the renaming workaround should be special-cased for the 3.1 image and you may need to create a custom image that derives from the MS one to solve the permissions issue. Might want to try and ping @fealebenpae as the resident docker guru.

Issue #2198, opened 2021-01-22
Investigate Realm close lifecycle


nirinchev:
Looking at the code for State.RemoveRealm:

https://github.com/realm/realm-dotnet/blob/247a6fcacf757e8187613b54680c441477c25707/Realm/Realm/Realm.cs#L1400-L1415

It looks like we may end up in a situation where we have 2 instances - X and Y; X gets GC-ed and Y - disposed. When `RemoveRealm` is called for Y, it'll see there're still instances in `State._weakRealms` and avoid closing the native Realm.

Issue #2224, opened 2021-02-04
Native libc crash on commit transaction in multithreaded environment


lagmac:
As per discussion https://github.com/realm/realm-dotnet/issues/2207 opened (now closed) a few days ago by my colleague, a similar problem has now also occurred on a Samsung A40 device. This issue is related to a crash during a commit transaction as per the crash report below.
My questions are :
- is there a possibility that the problem is internal to the library?
- is there the possibility that the problem is related to our incorrect implementation? 

## Goals

Successfully commit the transaction

## Expected Results

The transaction commit must complete successfully

## Actual Results

Native libc crash

--------- beginning of crash
01-18 09:01:49.253   917   917 F linker  : CANNOT LINK EXECUTABLE "/system/bin/sec_diag_uart_log": library "libdiag_system.so" not found
01-18 15:37:01.267  9030  3102 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x72616d69725024 in tid 3102 (Thread Pool Wor), pid 9030 (com.vimar.view)
01-18 15:37:01.634  3566  3566 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
01-18 15:37:01.634  3566  3566 F DEBUG   : Build fingerprint: 'samsung/a70qeea/a70q:10/QP1A.190711.020/A705FNXXU5CTK4:user/release-keys'
01-18 15:37:01.634  3566  3566 F DEBUG   : Revision: '14'
01-18 15:37:01.634  3566  3566 F DEBUG   : ABI: 'arm64'
01-18 15:37:01.645  3566  3566 F DEBUG   : Timestamp: 2021-01-18 15:37:01+0100
01-18 15:37:01.645  3566  3566 F DEBUG   : pid: 9030, tid: 3102, name: Thread Pool Wor  >>> com.vimar.view <<<
01-18 15:37:01.645  3566  3566 F DEBUG   : uid: 10319
01-18 15:37:01.645  3566  3566 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x72616d69725024
01-18 15:37:01.645  3566  3566 F DEBUG   :     x0  00000079d5cd1800  x1  0000000000000029  x2  0000000000000000  x3  00000079f9441ed4
[...]

PR #2251, opened 2021-02-15
Remove Realm states dictionary

nirinchev:
## Description

While investigating https://github.com/realm/realm-dotnet/issues/2224, I spent quite some time convincing myself that the Realm lifecycle is correct. While I couldn't find issues with it, I noticed that the `_states` dictionary approach was wrong. It relied on a ThreadLocal dictionary to hold the states for the current thread, but that is no longer safe since a Realm file may be opened/used across different threads as long as the synchronization context is the same. This was only used for the open Realm detection which should be moved to OS once https://github.com/realm/realm-core/issues/4424 is merged.

Fixes https://github.com/realm/realm-dotnet/issues/2198 (doesn't do anything to fix it per se, but adds a thorough explanation why it's not a problem).

##  TODO

* [x] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Realm.cs
+++ b/Realm/Realm/Realm.cs
@@ -48,10 +48,6 @@ namespace Realms
     {
         #region static
 
-        // This is imperfect solution because having a realm open on a different thread wouldn't prevent deleting the file.
-        // Theoretically we could use trackAllValues: true, but that would create locking issues.
-        private static readonly ThreadLocal<IDictionary<string, WeakReference<State>>> _states = new ThreadLocal<IDictionary<string, WeakReference<State>>>(DictionaryConstructor<string, WeakReference<State>>);
-
         // TODO: this will go away once https://github.com/realm/realm-dotnet/pull/2251 is merged, but due to a Mono bug, this needs to be a function rather than a lambda
         private static IDictionary<TKey, TValue> DictionaryConstructor<TKey, TValue>() => new Dictionary<TKey, TValue>();
 
@@ -184,13 +180,6 @@ namespace Realms
             SharedRealmHandle.DeleteFiles(configuration.DatabasePath);
         }
 
-        private static bool IsRealmOpen(string path)
-        {
-            return _states.Value.TryGetValue(path, out var reference) &&
-                   reference.TryGetTarget(out var state) &&
-                   state.GetLiveRealms().Any();
-        }
-
         #endregion static
 
         private State _state;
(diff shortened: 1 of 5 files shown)


DominicFrei:
@nirinchev I've updated this branch by merging in `df/use-delete-realm-from-core`.

Realm/Realm/Realm.cs
```diff
@@ -1448,29 +1430,32 @@ public void AddRealm(Realm realm)
             {
                 // We only want to have each realm once. That should be the case as AddRealm
                 // is only called in the Realm ctor, but let's check just in case.
-                Debug.Assert(!_weakRealms.Any(r =>
-                {
-                    return r.TryGetTarget(out var other) && ReferenceEquals(realm, other);
-                }), "Trying to add a duplicate Realm to the RealmState.");
+                Debug.Assert(!GetLiveRealms().Any(other => ReferenceEquals(realm, other)), "Trying to add a duplicate Realm to the RealmState.");
```
DominicFrei, line 1433: Would it make sense to have this as a test instead of an assert?

Realm/Realm/Realm.cs
```diff
@@ -1448,29 +1430,32 @@ public void AddRealm(Realm realm)
...
+            /// Remove a Realm from the list of Realms tracked by this state. This is only called when
+            /// Dispose is called on the Realm file and will not execute for garbage collected Realm
+            /// instances. This is fine because for GC-ed instances the lifecycle is as:
+            /// 1. Instance is GC-ed, its fields are GC-ed.
+            /// 2. The SharedRealmHandled is GC-ed, which causes Unbind to be called.
+            /// 3. The native pointer is deleted, which calls RealmCoordinator::unregister_realm.
+            /// 4. Once the last instance is deleted, the CSharpBindingContext destructor is called, which frees the state GCHandle.
+            /// 5. The State is now eligible for collection, and its fields will be GC-ed.
+            /// </summary>
+            public void RemoveRealm(Realm realm)
```
DominicFrei, line 1448: According to the documentation this is called when `Dispose` is called:
- So the user does actually not call it themselves?
- Do we need it to be public then?

DominicFrei (changes requested):
Are there any tests we should adjust to account for these changes?

Realm/Realm/Realm.cs
nirinchev: It's an assert because this is currently untestable behavior - there's no way currently to add the same realm twice as adding happens in the constructor. However, because there's nothing in the typesystem or the API design that enforces that, I wanted to add the extra check in case someone accidentally moves things around in the future. And it's a debug assert because the check isn't super cheap, so I didn't want us to do it in production.

Realm/Realm/Realm.cs
nirinchev: The method is public, but the class it is defined in is internal:

https://github.com/realm/realm-dotnet/blob/4b316aaa1eb3b0341c7190269fdffacf73918c61/Realm/Realm/Realm.cs#L1415

So it won't be exposed to end users.

nirinchev:
There are no tests that need to be updated - the whole point of the states dictionary was to prevent deleting the Realm file while it's open. Since that is now achieved thanks to your work to move deletion in OS, the dictionary was actually unused - we were adding and removing elements, but not really using it for anything.

Issue #2205, opened 2021-01-26
RealmValue creates null PrimitiveValue when boxing an unmanaged object


nirinchev:
ToNative of RealmValue boxing an object will produce PrimitiveValue.Null() when the object is unmanaged - this is probably a bug and we should either revisit or plain out throw an exception here.

PR #2252, opened 2021-02-16
Implement Mixed data type

### LINKED ISSUE #2066: Implement Mixed data type

papafe:
Add support for Mixed data type.

Fixes #2066 
Fixes https://github.com/realm/realm-dotnet/issues/2205
Fixes https://github.com/realm/realm-dotnet/issues/2160

##  TODO

* [x] Changelog entry
* [x] Tests


papafe:
Please note that the PR is still in draft, and that it also contains all the changes from PR #2230, because those changes are needed, and can't be merged to master yet

papafe:
I have benchmarked the new approach to value retrieval from the native layer (`TableKey` marshalled back `PrimitiveValue` and used as a key to find `ObjectMetadata`) to what we do now. I have used the `ObjectTests` in `PerformanceTests` to do so, as they are measuring exactly value retrieval. Results:

_Master_
|           Method |       Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|----------------- |-----------:|---------:|---------:|-------:|-------:|------:|----------:|
| **GetPropertyValue** | **1,391.2 ns** | **28.56 ns** | **83.31 ns** | **0.0172** | **0.0076** |     **-** |     **144 B** |
| **SetPropertyValue** |   **289.5 ns** |  **5.57 ns** |  **7.04 ns** |      **-** |      **-** |     **-** |         **-** |


_New approach (this PR)_
|           Method |       Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|----------------- |-----------:|---------:|---------:|-------:|-------:|------:|----------:|
| **GetPropertyValue** | **1,316.5 ns** | **26.91 ns** | **77.22 ns** | **0.0172** | **0.0076** |     **-** |     **144 B** |
| **SetPropertyValue** |   **274.0 ns** |  **0.53 ns** |  **0.47 ns** |      **-** |      **-** |     **-** |         **-** |

The results are very similar, so we can safely continue with the new approach. 

I have run the tests locally on my Windows VM. Also, I have tried to run the benchmarks a couple of times for both branches, but the results do not change much.

Realm/Realm/Linq/RealmResults.cs
```diff
@@ -74,18 +74,19 @@ internal override CollectionHandleBase GetOrCreateHandle()
...
         }
 
-        protected override T GetValueAtIndex(int index) => ResultsHandle.GetValueAtIndex(index, Metadata, Realm).As<T>();
+        protected override T GetValueAtIndex(int index) => ResultsHandle.GetValueAtIndex(index, Realm).As<T>();
 
         public override int IndexOf(T value)
         {
             Argument.NotNull(value, nameof(value));
 
-            if (_argumentType != RealmValueType.Object)
```
papafe, line 83: From what I could see through a search in the codebase, the only case in which we have a `RealmResult<T>` where `T` is not a `RealmValueType.Object`, is when we return the keys of a `RealmDictionary`. Is this the only case?

Realm/Realm/Linq/RealmResults.cs
```diff
@@ -79,14 +79,14 @@ internal override CollectionHandleBase GetOrCreateHandle()
...
-            
-            //TODO What to do here...?
-            //if (_argumentType != RealmValueType.Object) 
-            //{
-            //    throw new NotSupportedException("IndexOf on non-object results is not supported.");
-            //}
 
             var obj = value as RealmObjectBase;
+
+            if (obj == null)
```
nirinchev, line 85: I think that previously we didn't have results that contain non-objects. Is it impossible to support `IndexOf` for non-object values similar to what we do for lists?

Realm/Realm.Unity/Editor/Extensions/PropertyDefinitionExtensions.cs
```diff
@@ -154,6 +154,11 @@ internal static bool IsString(this PropertyDefinition property)
         return property.PropertyType.FullName == StringTypeName;
     }
 
+    internal static bool IsMixed(this PropertyDefinition property)
```
nirinchev, line 157: [nit] Let's rename to `IsRealmValue` - since that's the official name in the SDK, it will likely be a bit confusing for someone new reading the code.

Realm/Realm/Handles/CollectionHandleBase.cs
```diff
@@ -55,20 +55,15 @@ public ResultsHandle Snapshot()
...
             }
 
-            var objectHandle = primitive.AsObject(Root);
-            if (metadata == null)
-            {
-                throw new NotImplementedException("Mixed objects are not supported yet.");
-            }
-
-            return new RealmValue(realm.MakeObject(metadata, objectHandle));
+            (var objectHandle, var tableKey) = primitive.AsObject(Root);
```
nirinchev, line 65: [extreme nit] I prefer the single var syntax for tuple deconstruction: `var (objectHandle, tableKey) = ...` 😄

Realm/Realm/Handles/DictionaryHandle.cs
```diff
@@ -184,24 +184,17 @@ public bool TryGet(string key, RealmObjectBase.Metadata metadata, Realm realm, o
             }
```
nirinchev, line 184: Can't comment on the right line, but you can probably combine both branches of the if statement since ToRealmValue will handle the `Type != Object` case anyway.

Realm/Realm/Helpers/Operator.cs
```diff
@@ -397,7 +397,7 @@ internal static class Operator
                  */
                 if (value is null)
                 {
-                    return Convert<RealmValue, TResult>(RealmValue.Null());
+                    return Convert<RealmValue, TResult>(RealmValue.Null);
```
nirinchev, line 400: This should also be changed in Operator.tt, otherwise the next time the code is regenerated, it will revert to using Null as a method.

Realm/Realm/Schema/PropertyType.cs
```diff
@@ -75,6 +75,11 @@ public enum PropertyType : ushort
         /// </summary>
         LinkingObjects = 8,
 
+        /// <summary>
+        /// A property that can contain RealmValue supported types.
+        /// </summary>
+        Mixed = 9,
```
nirinchev, line 81: Since this is public, it should be `RealmValue`, not `Mixed`. We should never let Core terminology bleed into the public API.

wrappers/src/list_cs.cpp
```diff
@@ -168,7 +168,7 @@ REALM_EXPORT size_t list_find_value(List& list, realm_value_t value, NativeExcep
                 throw ObjectManagedByAnotherRealmException("Can't look up index of an object that belongs to a different Realm.");
             }
 
-            if ((list_type & PropertyType::Flags) == PropertyType::Mixed) {
+            if ((list_type & ~PropertyType::Flags) == PropertyType::Mixed) {
                 return list.find_any(ObjLink(value.link.object->get_object_schema().table_key, value.link.object->obj().get_key()));
```
nirinchev, line 172: I know I probably wrote this code, but would it work if we called `find_any` with `value.link.object.obj()`?

nirinchev (commented):
Haven't reviewed tests yet, but looks good. Mostly nits + fixing `Results.IndexOf`.

wrappers/src/list_cs.cpp
papafe: It seems to work!

wrappers/src/list_cs.cpp
```diff
@@ -168,8 +168,8 @@ REALM_EXPORT size_t list_find_value(List& list, realm_value_t value, NativeExcep
                 throw ObjectManagedByAnotherRealmException("Can't look up index of an object that belongs to a different Realm.");
             }
 
-            if ((list_type & PropertyType::Flags) == PropertyType::Mixed) {
-                return list.find_any(ObjLink(value.link.object->get_object_schema().table_key, value.link.object->obj().get_key()));
+            if ((list_type & ~PropertyType::Flags) == PropertyType::Mixed) {
```
LaPeste, line 171: I'm curious about this: this mistake made sure that any list holding Mixed values were never recognized. Ending up at line 175 where the find, instead, was called. Here it's where I get confused: was that working fine or throwing an error? I can see that List::Find relies on as<T>() which tries to statically convert the list to a list of Mixed in our case. If this conversion was possible then what's the point of find_any? Can I assume JUST performance?
If I have misread all this, then what did I miss?

papafe: This is actually a good question. If we arrive at line 175 with a Mixed property the there's an exception coming from `Collection::validate(const Obj& obj)`, due to the fact that the obj type is not the same of the list type (due to mixed). There is actually even a `// FIXME: This does not work for TypedLink.` there, so I suppose this part here could be fixed, but I am not sure if there are other reasons for not using `find_any` all the time. Maybe it's just a matter of avoiding the conversion from `realm_value_t` to `Mixed`?
Besides, in Results, we do not have this difference between `index_of` and `index_of_any`, so it's a little confusing.
@nirinchev This is probably something we need to talk about :)

nirinchev: Yeah, let's chat about it at the roundtable

papafe:
I converted this to a draft again because I want to do a refactoring of the tests.

papafe:
The PR is in a decent state now, and reviewers can start to have a look.
Missing:
- Test for Sets
- Dynamic access for objects whose class in not in the schema

Tests/Realm.Tests/Database/DynamicAccessTests.cs
```diff
@@ -65,50 +83,111 @@ public void SetAndGetValue<T>(string propertyName, T propertyValue)
...
+
+        private object InvokeGetter(object o, string propertyName)
+        {
+            if (_mode == DynamicTestObjectType.DynamicRealmObject)
+            {
+                var binder = Binder.GetMember(CSharpBinderFlags.None, propertyName, typeof(DynamicAccessTests), new[] { CSharpArgumentInfo.Create(CSharpArgumentInfoFlags.None, null) });
+                var callsite = CallSite<Func<CallSite, object, object>>.Create(binder);
+                return callsite.Target(callsite, o);
+            }
+            else
```
papafe, line 173: remove

papafe:
I have just verified that sync tests are passing. With this there is very little to do, and it depends on bugs/features in Core:
- Validate key in dictionary does not respect size (https://github.com/realm/realm-core/issues/4589)
- `query.equal` does not work with an object parameter (ongoing discussion on Slack)
- Issue with realm in objectStore not containing the full schema There is a ticket (https://github.com/realm/realm-core/issues/4584), but it also seems there is a more complex solution for now that was implemented in Cocoa

nirinchev:
We should also consider exposing `RealmValue.ObjectType` which returns `AsRealmObject().ObjectSchema.Name` - while users can definitely get to it themselves, I'd be quite impressed if they discover it on their own.

PR #2255, opened 2021-02-17
Assorted test-related improvements

nirinchev:
## Description

This is mostly a cleanup PR that achieves the following:
* Fixes `OrderBy(...).OrderBy(...)` query chains by having the last clause replace all preceding ones.
* Reenables all previously disabled Windows tests. They seem to pass locally.
* Reenable all previously explicit tests that have been fixed. Explicit tests that haven't been fixed are [Ignore]-d
* Reenable some previously [Ignore]-d tests that have been fixed.
* Throws `RealmMigrationNeededException` when opening a readonly file that uses an old core format.

##  TODO

* [x] Changelog entry

### DIFF
--- a/Tests/Realm.Tests/Database/LifetimeTests.cs
+++ b/Tests/Realm.Tests/Database/LifetimeTests.cs
@@ -17,6 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 using System;
+using System.Threading.Tasks;
 using NUnit.Framework;
 
 namespace Realms.Tests.Database
@@ -34,8 +35,6 @@ namespace Realms.Tests.Database
         [Test]
         public void RealmObjectsShouldKeepRealmAlive()
         {
-            TestHelpers.IgnoreOnWindows("GC blocks on Windows");
-
             // Arrange
             var realm = GetWeakRealm();
             Person person = null;
@@ -60,8 +59,6 @@ namespace Realms.Tests.Database
         [Test]
         public void FinalizedRealmsShouldNotInvalidateSiblingRealms()
         {
-            TestHelpers.IgnoreOnWindows("GC blocks on Windows");
-
             // Arrange
             using var realm = Realm.GetInstance(RealmConfiguration.DefaultConfiguration.DatabasePath);
             var realmThatWillBeFinalized = GetWeakRealm();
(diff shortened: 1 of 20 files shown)


Tests/Realm.Tests/Database/LifetimeTests.cs
```diff
@@ -34,8 +35,6 @@ private static WeakReference GetWeakRealm()
         [Test]
         public void RealmObjectsShouldKeepRealmAlive()
         {
-            TestHelpers.IgnoreOnWindows("GC blocks on Windows");
```
LaPeste, line 37: How were this test and a few others below fixed when in the test nothing is changed? What am I missing?

nirinchev: Likely fixed when we pulled in the new conditional variable implementation. There's nothing in this PR that fixes them - just noticed they're passing while reenabling all tests in Windows.

Tests/Realm.Tests/Sync/SyncTestBase.cs
```diff
@@ -56,13 +56,13 @@ protected App CreateApp(AppConfiguration config = null)
 
         protected override void CustomTearDown()
         {
-            base.CustomTearDown();
```
LaPeste, line 59: If a Realm is disposed its sync session should be automatically disposed, so does this manual cleanup have the goal of speeding up the process of memory release or is there more to it?

nirinchev: A sync session is independent of the Realm - it may certainly outlive the managed instance. The issue with the order here was that we were calling the base teardown method which disposes and deletes all Realm instances before disposing the session handles. This resulted in race conditions where if a session hasn't been disposed, the deletion would fail.

LaPeste: Then I misinterpreted [this comment](https://github.com/realm/realm-dotnet/blob/35b144d92d866a6e42474ec869de4c520c93c140/Realm/Realm/Sync/Session.cs#L28); I assume it simply meant that the user doesn't have to care to manually deallocating it.
Thank you for the explanation.

LaPeste (approved):
All seems fine to me.

Tests/Realm.Tests/Sync/SyncTestBase.cs
nirinchev: Yes, the automatic lifespan management means that we'll create and terminate sessions based on certain heuristics/internal rules. Typically, this will mean that a session outlives its Realm so that it can upload any remaining changes. This has the unfortunate side effect that deleting a synchronized Realm, even after disposing it, is not exactly safe. This will hopefully be addressed by https://github.com/realm/realm-core/issues/4424.

papafe (approved):
Looks fine

PR #2257, opened 2021-02-17
Try to fix performance issues with weaving

nirinchev:


peppy:
I can see this is unity-specific so apologies for intruding, but I'd like to mention that I also noticed considerable overhead during the compilation of our project, as seen in a non-scientific benchmark of three builds using `echo "// change\r\n" >> osu.Game/OsuGame.cs; dotnet build osu.Desktop`:

20210219 115229 (Numi)

Is it worth creating an issue tracking this separately? Considering we do quite frequent iteration via recompilation it is a minor concern for us.

nirinchev:
For regular .NET projects we rely on Fody to run the code weaver and it's expected that it adds a constant overhead on top of the regular compilation. I expect most of the time is spent in Fody itself and not actually running Realm code, so until we move away from it (which is something we plan to do), there'll be very little we can do to speed up builds. We do have an issue to investigate rolling our own weaving task, but that's not a priority for this quarter 😕

nirinchev:
This was cherry picked in master.

PR #2259, opened 2021-02-19
Pass in correct schema type for explicit and autodiscovered types

### LINKED ISSUE #2193: Pass in correct schema type for autodiscovered types
Once https://github.com/realm/realm-core/pull/4306 goes in, we should amend our schema builder to pass in the correct type, depending on whether the schema was explicitly specified by the user (via ObjectTypes) or we autodiscovered it at build time.

LaPeste:
## Description
Based on `RealmConfigurationBase.ObjectClasses` "AdditiveDiscovered" or "AdditiveExplicit" is now appropriately selected and passed to core.
Currently, because of the way core did it, this is a detail only relevant for synced scenarios. I'll talk to James Stone to understand better the why's'.
This PR also adds an additional exception to have a more precise error category when an issue with the schema validation arises.
 
==================================
UPDATE
I talked with James and he explained that the 2 types of Additive behaviours are purposedly used only for sync-d realms because unreachable embeddedObjects are only problematic for the server. Hence, when the client doesn't sync, there's no reason to check for them as these unreachable objs don't really create an issue; although being logically useless. He added that the only reason why such check had to end up in the client, although only bothering the server, is that when the server received such object,s the message sent back to the client was too obscure for the user to be of help. 

Fixes #2193

##  TODO

* [x] Changelog entry
* [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Fixed
 * Fixed an issue that would result in UWP apps being rejected from the Microsoft Store due to an unsupported API (`__C_specific_handler`) being used. (Issue [#2235](https://github.com/realm/realm-dotnet/issues/2235))
 * Fixed a bug where applying multiple `OrderBy` clauses on a query would result in the clauses being appended to each other as if they were `.ThenBy` rather than the last clause replacing the preceding ones. (PR [#2255](https://github.com/realm/realm-dotnet/issues/2255))
-* Fixed proper selection of Additive mode (Explicit or Discovered) to pass to the schema builder. This selection is based on whether or not the user has provided a subset of classes to the Realm Schema. More info about this subject can be found [here](https://docs.mongodb.com/realm/dotnet/objects/#provide-a-subset-of-classes-to-your-realm-schema). (PR [#2259](https://github.com/realm/realm-dotnet/pull/2259))
+* When explicitly specifying `SyncConfiguration.ObjectTypes`, added a check to validate the schema and ensure all `EmbeddedObject` classes are reachable from a class inheriting from `RealmObject`. More info about this subject can be found [here](https://docs.mongodb.com/realm/dotnet/objects/#provide-a-subset-of-classes-to-your-realm-schema). (PR [#2259](https://github.com/realm/realm-dotnet/pull/2259))
 
 ### Enhancements
 * Add support for the `Guid` data type. It can be used as primary key and is indexable. (PR [#2120](https://github.com/realm/realm-dotnet/pull/2120))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixed
 * Fixed an issue that would result in UWP apps being rejected from the Microsoft Store due to an unsupported API (`__C_specific_handler`) being used. (Issue [#2235](https://github.com/realm/realm-dotnet/issues/2235))
 * Fixed a bug where applying multiple `OrderBy` clauses on a query would result in the clauses being appended to each other as if they were `.ThenBy` rather than the last clause replacing the preceding ones. (PR [#2255](https://github.com/realm/realm-dotnet/issues/2255))
+* Fixed proper selection of Additive mode (Explicit or Discovered) to pass to the schema builder. This selection is based on whether or not the user has provided a subset of classes to the Realm Schema. More info about this subject can be found [here](https://docs.mongodb.com/realm/dotnet/objects/#provide-a-subset-of-classes-to-your-realm-schema). (PR [#2259](https://github.com/realm/realm-dotnet/pull/2259))
 
 ### Enhancements
 * Add support for the `Guid` data type. It can be used as primary key and is indexable. (PR [#2120](https://github.com/realm/realm-dotnet/pull/2120))
--- a/Realm/Realm/Configurations/SyncConfiguration.cs
+++ b/Realm/Realm/Configurations/SyncConfiguration.cs
@@ -122,9 +122,8 @@ namespace Realms.Sync
 
         internal override Realm CreateRealm(RealmSchema schema)
         {
-            var configuration = CreateConfiguration();
-            var syncConfiguration = CreateSyncConfiguration();
-            syncConfiguration.schema_mode = ObjectClasses == null ? SchemaMode.AdditiveDiscovered : SchemaMode.AdditiveExplicit;
+            var configuration = CreateNativeConfiguration();
+            var syncConfiguration = CreateNativeSyncConfiguration();
 
             var srHandle = SharedRealmHandle.OpenWithSync(configuration, syncConfiguration, schema, EncryptionKey);
             if (IsDynamic && !schema.Any())
@@ -138,9 +137,8 @@ namespace Realms.Sync
         [SuppressMessage("Reliability", "CA2000:Dispose objects before losing scope", Justification = "The Realm instance will own its handle")]
         internal override async Task<Realm> CreateRealmAsync(RealmSchema schema, CancellationToken cancellationToken)
         {
-            var configuration = CreateConfiguration();
-            var syncConfiguration = CreateSyncConfiguration();
-            syncConfiguration.schema_mode = ObjectClasses == null ? SchemaMode.AdditiveDiscovered : SchemaMode.AdditiveExplicit;
+            var configuration = CreateNativeConfiguration();
+            var syncConfiguration = CreateNativeSyncConfiguration();
 
             var tcs = new TaskCompletionSource<ThreadSafeReferenceHandle>();
             var tcsHandle = GCHandle.Alloc(tcs);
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Fixed
 * Fixed an issue that would result in UWP apps being rejected from the Microsoft Store due to an unsupported API (`__C_specific_handler`) being used. (Issue [#2235](https://github.com/realm/realm-dotnet/issues/2235))
 * Fixed a bug where applying multiple `OrderBy` clauses on a query would result in the clauses being appended to each other as if they were `.ThenBy` rather than the last clause replacing the preceding ones. (PR [#2255](https://github.com/realm/realm-dotnet/issues/2255))
+* Fixed proper selection of Additive mode (Explicit or Discovered) to pass to the schema builder. This selection is based on whether or not the user has provided a subset of classes to the Realm Schema. More info about this subject can be found [here](https://docs.mongodb.com/realm/dotnet/objects/#provide-a-subset-of-classes-to-your-realm-schema). (PR [#2259](https://github.com/realm/realm-dotnet/pull/2259))
```
LaPeste, line 6: 2 issues with this:

1. I'm not sure this should be considered an issue. After all, it was consciously decided to have a quick change to the breaking changes introduced by the splitting of the schema mode `Additive`. If we agree on this, maybe internal could be a better section. Although, even *internal* means like a change when this must be addressed given the new change in core. So, please advice if this is necessary at all and where it should go.
2. I'm not that happy to have the PR# here just yet. This is because another PR may be needed based on the outcome of the conversation with James.

Realm/Realm/Configurations/SyncConfiguration.cs
```diff
@@ -123,6 +123,7 @@ private SyncConfiguration(object partition, User user, string path)
         internal override Realm CreateRealm(RealmSchema schema)
         {
             var configuration = CreateConfiguration();
+            configuration.additive_discovered = ObjectClasses == null ? true : false;
```
nirinchev, line 126: You don't need the ternary here. `ObjectClasses == null` is already a boolean.

Realm/Realm/Native/Configuration.cs
```diff
@@ -50,6 +50,9 @@ internal string Path
         [MarshalAs(UnmanagedType.U1)]
         internal bool in_memory;
 
+        [MarshalAs(UnmanagedType.U1)]
+        internal bool additive_discovered;
```
nirinchev, line 54: This is not used for non-sync configurations, so we should probably move it to the sync config. Also, let's replace it with a SchemaMode enum instead.

Realm/Realm/Exceptions/RealmSchemaValidationException.cs
```diff
@@ -0,0 +1,31 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Exceptions
+{
+    /// <summary>
+    /// Exception thrown when the current schema doesn't respect Realm's set of rule. The details in the message helps to understand what the broken rule is.
```
nirinchev, line 22: This is a bit ambiguous and somewhat verbose. I would suggest rewording as:

```
<summary>
Exception thrown when the schema specified in your C# models doesn't pass validation for the type of Realm
you're trying to open. The message contains information about the validation errors and how to correct them.
</summary>
<seealso href="https://docs.mongodb.com/realm/dotnet/realms/#std-label-dotnet-realm-schema">
General information about Realm's schema.
</seealso>
```

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -271,6 +272,18 @@ public void InvalidSchemaChange_RaisesClientReset()
             });
         }
 
+        [Test]
+        public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
+        {
+            var conf = new SyncConfiguration("foo-bar", GetFakeUser())
```
nirinchev, line 278: You can use `GetFakeConfig` here.

nirinchev (changes requested):

Realm/Realm/Configurations/SyncConfiguration.cs
LaPeste: 🤦🏻

Realm/Realm/Native/Configuration.cs
LaPeste: I'm in doubt on a thing: should the enum in C# have only the 2 additive options or should it reflect the whole enum that's in the object store? There are pros and cons to both ways, so I can't decide.

nirinchev: Let's only have the two options - if we need more, we can add more - it's an internal API, so we can change it however we please.

CHANGELOG.md
nirinchev, line 6: This message is not something users can understand - none of the mentioned types are publicly exposed. We should instead say something like:

```
When explicitly specifying `SyncConfiguration.ObjectTypes`, added
a check to validate the schema and ensure all `EmbeddedObject`
classes are reachable from a class inheriting from `RealmObject`.
```

Realm/Realm/Configurations/SyncConfiguration.cs
```diff
@@ -187,7 +191,7 @@ internal override async Task<Realm> CreateRealmAsync(RealmSchema schema, Cancell
             }
         }
 
-        internal Native.SyncConfiguration ToNative()
+        internal Native.SyncConfiguration CreateSyncConfiguration()
```
nirinchev, line 194: I liked the previous name better. The new name is confusing because this method is already exposed on a `SyncConfiguration` class. I guess you're modeling it after `CreateConfiguration` but that is also terrible - must have been Yavor who introduced it 😝 If you want to keep `CreateConfiguration`/`CreateSyncConfiguration`, at least throw in `Native` in there.

Realm/Realm/Configurations/SyncConfiguration.cs
```diff
@@ -137,13 +139,15 @@ internal override Realm CreateRealm(RealmSchema schema)
         internal override async Task<Realm> CreateRealmAsync(RealmSchema schema, CancellationToken cancellationToken)
         {
             var configuration = CreateConfiguration();
+            var syncConfiguration = CreateSyncConfiguration();
+            syncConfiguration.schema_mode = ObjectClasses == null ? SchemaMode.AdditiveDiscovered : SchemaMode.AdditiveExplicit;
```
nirinchev, line 143: This can be moved to `CreateSyncConfiguration` instead of repeating it everywhere.

Realm/Realm/Exceptions/RealmSchemaValidationException.cs
```diff
@@ -0,0 +1,34 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Exceptions
+{
+    /// <summary>
+    /// Exception thrown when the schema specified in your C# models doesn't pass validation for the type of Realm you're trying to open.
+    /// The message contains information about the validation errors and how to correct them.
+    /// </summary>
+    /// <seealso href = "https://docs.mongodb.com/realm/dotnet/realms/#std-label-dotnet-realm-schema" >
```
nirinchev, line 25:
```suggestion
    /// <seealso href="https://docs.mongodb.com/realm/dotnet/realms/#std-label-dotnet-realm-schema">
```

Realm/Realm/Sync/SchemaMode.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Realms.Sync
+{
+    /// <summary>
+    /// To see details of each mode check its mirroring definition in <see href="https://github.com/realm/realm-core/blob/0976706c26ce24866e6be6c165b6c6192fb663ed/src/realm/object-store/shared_realm.hpp#L61">core</see>.
+    /// </summary>
+    internal enum SchemaMode : byte
+    {
+        // N.B. the values must match their representation in core!
+        AdditiveDiscovered = 4,
+        AdditiveExplicit
```
nirinchev, line 28: Let's be explicit about all values.

Realm/Realm/Sync/SchemaMode.cs
```diff
@@ -0,0 +1,30 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync
+{
+    /// <summary>
+    /// To see details of each mode check its mirroring definition in <see href="https://github.com/realm/realm-core/blob/0976706c26ce24866e6be6c165b6c6192fb663ed/src/realm/object-store/shared_realm.hpp#L61">core</see>.
+    /// </summary>
+    internal enum SchemaMode : byte
+    {
+        // N.B. the values must match their representation in core!
```
nirinchev, line 26: This is true for all enums being marshaled so doesn't warrant an explicit note.

Realm/Realm/Sync/SchemaMode.cs
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync
+{
+    /// <summary>
+    /// To see details of each mode check its mirroring definition in <see href="https://github.com/realm/realm-core/blob/0976706c26ce24866e6be6c165b6c6192fb663ed/src/realm/object-store/shared_realm.hpp#L61">core</see>.
+    /// </summary>
```
nirinchev, line 23: The enum name matches the Core enum name, so it can be easily searched for with symbol search. Including a github link is prone to link rot, especially when it includes the sha of the commit. Fairly certain when we update the Core dependency, we won't update this link, so in the future we will inevitably end up with the code here and the code in the link being incompatible.

nirinchev (commented):
Mostly documentation nits from me + a suggestion to save a few lines of code

nirinchev (approved):

PR #2267, opened 2021-02-25
Build wrappers on Github Actions

### LINKED ISSUE #2253: Investigate workflow reuse in Github Actions
TODO:

* :white_check_mark: Merge [https://github.com/realm/realm-dotnet/pull/2183](https://github.com/realm/realm-dotnet/pull/2183) or base work on top of it.
* Figure out if there's a way to reuse portions of different workflows. Perhaps start extracting code into separate actions. Ping Brian/Sergey if we need a new repo for those.
* ~~Build macOS wrappers on every PR push.~~

### LINKED ISSUE #2280: Build wrappers for iOS and macOS

### LINKED ISSUE #2281: Build wrappers for Windows and UWP

### LINKED ISSUE #2282: Build wrappers for Linux

### LINKED ISSUE #2283: Build wrappers for Android

### LINKED ISSUE #2294: Create an action that retrieves the content of multiple caches on Github Actions

LaPeste:
## Description
Build macOS wrappers on every PR push and while at it create actions for code reuse.

--------

**To run tests**: from relative folder `realm-dotnet/.github/actions/build_wrappers` run `$npm test`.
**To build**: it depends on where the changes were made

1.  If changes were only made to the action, just open the workspace with VSCode and hit build (or Ctrl+shift+b) and select the "distribute" option.
2. If changes were made also made to the package, run `npm run build` from that project plus point 1

**To lint**: `npm run lint` from the project you want to lint
**To lint and auto fix**: `npm run lint-fix`
**To debug**: in VSCode go under the debug tab and select

- "Launch Program" to debug a normal ts file as long as the break point is put in the transpiled js file
- "mocha tests" to debug test cases where the breakpoint is put in the ts file

A discussion should be held to agree on when and where (locally or remotely) to automatically trigger a test run.

______________________________

Fixes #2294
Fixes #2280
Fixes #2281
Fixes #2282
Fixes #2283
Fixes #2253

##  TODO

* [ ] Changelog entry


.github/actions/build_wrappers/src/build_wrappers.ts
```diff
@@ -0,0 +1,117 @@
...
+import * as cache from "@actions/cache";
+import * as folderHash from "folder-hash";
+import * as crypto from "crypto";
+import * as utils from "./utils/common";
+
+
+async function run(): Promise<void>
+{
+    // TODO to make this general purpose, we need to pass this either from a conf file or as input. Same for the next line
+    const paths = [ "./wrappers/build/" ];
```
nirinchev, line 11: You probably don't want to hash the build folder - that gets generated after you build. Instead, you probably want to just hash the entire `wrappers` folder immediately after checkout.

.github/workflows/main.yml
```diff
@@ -0,0 +1,80 @@
...
+    #   with:
+    #     path: ./wrappers/build/
+    #     key: wrapper-macos-${{hashfiles('./wrappers/**')}}
+    - name: 'Build wrappers'
+      # if: 'steps.wrapper-cache.outputs.cache-hit != true'
+      uses: ./.github/actions/build_wrappers
+      env: 
+        REALM_CMAKE_CONFIGURATION: Release
+      with:
+        cmds: '[ { "cmd": "./wrappers/build-macos.sh", "cmdParams": [] } ]'
```
papafe, line 25: I understand this makes possible to use multiple tuples of (command, parameters), but it looks a little bit clunky to use json here. I suppose in most cases parameters will be embedded in commands already, so I am not sure it's necessary, but that's up for discussion.

Another thing to consider is that in most cases we will have only one command, so it would be nice if the action would just accept a string as an input, without the need to embed it in json.

.gitignore
```diff
@@ -188,7 +188,6 @@ NDependOut
 *.xam
 Resource.designer.cs
 FodyWeavers.xsd
-.vscode
```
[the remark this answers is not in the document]
LaPeste: Yes, but not only. Transpilation from Typescript to JS and "compilation" of the generated js files into 1 only js file thanks to [ncc](https://github.com/vercel/ncc).

.github/workflows/main.yml
LaPeste: I am super open for discussion as I am not sure what the best solution would be about that. However, the reason for commands and their parameters to be diveded instead of being 1 string is the fact that I decided to launch command line command with `child_process.spawn` instead of `child_process.exec`. The latter accepts a string with all in it, the former, as far as I could find in the documentation, wants the arguments passed as list. So, I temporarily took the shortcut of having the JSON library to parse those for me. We can reshape all of that when we've agreed on the final solution.

.github/ga_packages/build_cache_executor/src/build_cache_executor.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @returns CacheKey necessary to recover the cached build later on. Undefined is returned if something went wrong.
+ */
+export async function actionCore(
+    paths: string,
+    cmds: string,
+    oss: utils.outputStream,
+    hashPrefix?: string,
+    hashOptions?: utils.hashOptions,
+    hashFunc?: utils.hashFunc
+): Promise<string | undefined> {
```
nirinchev, line 24: Let's reduce the customizability of this one as well:
* Replace `paths` with `path` - if people want to hash multiple folders, they'll extend it eventually.
* Replace `cmds` with `cmd` - if people want to execute multiple commands, they can combine them in a single one or extend the library.
* Remove `hashOptions` and `hashFunc` - the default implementation should be good enough for now.

.github/ga_packages/build_cache_executor/src/utils/common.ts
```diff
@@ -0,0 +1,71 @@
+import * as folderHash from "folder-hash";
+import * as crypto from "crypto";
+import * as fs from "fs-extra";
+
+export interface outputStream {
```
nirinchev, line 5: That's a weird name for what appears to be essentially a logger - should we name it `logger`?

nirinchev (commented):
Did a very cursory review of the public API of ga_packages/build_cache_executor and think we should simplify it. We shouldn't try and provide too much flexibility in the first iteration because we don't know if people will actually need it. Let's go with the absolute simplest solution and build on top of that rather than the other way around. Happy to pair on some of that so that we align faster.

.github/.vscode/launch.json
```diff
@@ -0,0 +1,37 @@
...
+            "name": "Launch Program",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            //"preLaunchTask": "npm: distribute - actions/build_cache_executor",
+            "program": "${file}",
+            "outFiles": [
+                "${workspaceFolder}/dist/**/*.js"
+            ]
+        },
```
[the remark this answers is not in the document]
LaPeste: I was just a shortcut that would build before starting debugging (if preLaunchTask is uncommented). Otherwise, _in the current state_ it does the default "node debug" behaviour of debugging the current file. More custom things can be done if, let's say you want to spend the time for making the debugger debug on the ts instead of the generate js etc. So it has some usefulness mainly if more time is spent on it, but it can be removed if you want.

.github/templates/process.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
LaPeste, line 1: Do we still need this? After all we know that "on" works.

PR #2269, opened 2021-02-26
Move the netcore image outside of the repo root

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2261


Jenkinsfile
```diff
@@ -453,9 +453,9 @@ def CreateDockerContainer(String targetFramework) {
   def test_runner_image
   switch(targetFramework) {
     case 'netcoreapp3.1':
-      // In .netcore 3.1 using msbuild extras under linux fails because of missing Microsoft.WinFX.props. This is generated by a mismatching lower/upper case for a folder name.
-      // Renaming the folders in questions fixes the issue. This action is achieved through making a custom docker container. More info can be found at https://github.com/dotnet/sdk/issues/11108
-      test_runner_image = docker.build("netcoreapp3.1_custom", "./CustomNetcore31DockerImg")
+      // Using a custom docker image for .NET Core 3.1 because the official has incorrect casing for
+      // Microsoft.WinFX.props. More info can be found at https://github.com/dotnet/sdk/issues/11108
+      test_runner_image = buildDockerEnv("ci/realm-dotnet:netcore31", extra_args: "-f ./Tests/netcore31.Dockerfile")
```
LaPeste, line 458: From what I could understand, this call is to cache the image built on some server and pull from there next times instead of rebuilding if no changes are found. I'm not sure I understand how "changes" are spotted but I guess it's not relevant now.
Did I get it right?

nirinchev: Yes, it's used for caching - I don't think there's a mechanism to detect when the cache should be invalidated, but I added the explicit .NET SDK version in the image tag, so if we decide to update it, we can invalidate the cache.

LaPeste (approved):
Looks good to me.

PR #2276, opened 2021-03-02
Unify the logging API

nirinchev:
## Description

This introduces a new `Logger` class that will be used (eventually) for all Realm-originating messages. Currently it's used for messages from Sync and those from managed code. We'll need to refactor OS to ensure that Core-originating messages also use the SDK-provided logger.

The motivation of this entire thing is twofold:
1. There are some platforms where `stdout`/`stderr` output is swallowed - such as the dotnet cli: https://github.com/realm/realm-core/issues/4485
2. Unity doesn't have a way to output `Console.WriteLine` messages and instead uses its own `Debug.Log` mechanism: https://github.com/realm/realm-dotnet/issues/2277

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)
* [x] Filter messages based on the log level


.editorconfig
```diff
@@ -12,15 +12,14 @@ indent_style = space
 tab_width = 4
 
 # New line preferences
-end_of_line = lf
```
LaPeste, line 15: Why was this removed?

nirinchev: Because I realized this should not be something controlled by the editor, but rather by git itself. You can read on how to configure git line endings [here](https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings). The editor should not be enforcing line endings because this will lead to inconsistencies on Windows, where git checks out files with `crlf`, but the editor uses `lf`.

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -438,7 +449,7 @@ private static unsafe void HandleOpenRealmCallback(IntPtr taskCompletionSource,
             }
         }
 
-        [MonoPInvokeCallbackAttribute(typeof(NativeMethods.OnBindingContextDestructedCallback))]
+        [MonoPInvokeCallback(typeof(NativeMethods.OnBindingContextDestructedCallback))]
```
LaPeste, line 452: I haven't had much luck in finding the difference between these 2. It seems that MonoPInvokeCallback comes up in Unity conversations, while with _Attribute generally leads to Microsoft docs. Could you elaborate on the differences?

Realm/Realm/Native/NativeCommon.cs
```diff
@@ -21,38 +21,15 @@
...
-#endif  // DEBUG
-
         [DllImport(InteropConfig.DLL_NAME, EntryPoint = "delete_pointer", CallingConvention = CallingConvention.Cdecl)]
         public static extern unsafe void delete_pointer(void* pointer);
 
-        [DllImport(InteropConfig.DLL_NAME, EntryPoint = "delete_pointer", CallingConvention = CallingConvention.Cdecl)]
-        public static extern unsafe void delete_pointer(IntPtr pointer);
-
-        [DllImport(InteropConfig.DLL_NAME, EntryPoint = "realm_reset_for_testing", CallingConvention = CallingConvention.Cdecl)]
-        public static extern void reset_for_testing();
```
LaPeste, line 54: Should this deletion and the one above part of this PR?

Realm/Realm/Handles/SharedRealmHandle.cs
nirinchev: There's no difference between these. The compiler allows you to omit the `Attribute` suffix whenever you apply an attribute to reduce the verbosity. Ref https://stackoverflow.com/a/6710431/1649102.

Realm/Realm/Native/NativeCommon.cs
nirinchev: I was cleaning up the debug logger stuff and noticed those are not used, which allowed me to remove debug.cpp.

LaPeste (approved):
It's overall a good idea to have a unified logger whose implementation is supplied by the SDK. It'll make life easier for everyone developing on the SDKs.

Issue #2273, opened 2021-02-26
Investigate how to handle assembly reloads


nirinchev:
Unity hot reloads assemblies, which hangs if there's an active Realm instance. We should hook into the assembly reload events and clean up any open Realms to avoid locking up people's IDEs.

nirinchev:
Depends on https://github.com/realm/realm-core/issues/4486

PR #2295, opened 2021-03-12
More protective measures related to domain unloading

nirinchev:
## Description

More fixes to make the SDK resilient to domain unloads.

Fixes https://github.com/realm/realm-dotnet/issues/2273.

### DIFF
--- a/wrappers/src/app_cs.cpp
+++ b/wrappers/src/app_cs.cpp
@@ -40,16 +40,21 @@ using namespace app;
 using SharedSyncUser = std::shared_ptr<SyncUser>;
 using SharedSyncSession = std::shared_ptr<SyncSession>;
 
+using LogMessageCallbackT = void(void* managed_handler, realm_value_t message, util::Logger::Level level);
+using UserCallbackT = void(void* tcs_ptr, SharedSyncUser* user, MarshaledAppError err);
+using VoidCallbackT = void(void* tcs_ptr, MarshaledAppError err);
+using BsonCallbackT = void(void* tcs_ptr, BsonPayload response, MarshaledAppError err);
+
 namespace realm {
     namespace binding {
         std::string s_platform;
         std::string s_platform_version;
         std::string s_sdk_version;
 
-        void (*s_log_message_callback)(void* managed_handler, realm_value_t message, util::Logger::Level level);
-        void (*s_user_callback)(void* tcs_ptr, SharedSyncUser* user, MarshaledAppError err);
-        void (*s_void_callback)(void* tcs_ptr, MarshaledAppError err);
-        void (*s_bson_callback)(void* tcs_ptr, BsonPayload response, MarshaledAppError err);
+        std::function<void(void* managed_handler, realm_value_t message, util::Logger::Level level)> s_log_message_callback;
+        std::function<void(void* tcs_ptr, SharedSyncUser* user, MarshaledAppError err)> s_user_callback;
+        std::function<void(void* tcs_ptr, MarshaledAppError err)> s_void_callback;
+        std::function<void(void* tcs_ptr, BsonPayload response, MarshaledAppError err)> s_bson_callback;
 
         struct AppConfiguration
         {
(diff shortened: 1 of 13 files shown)


wrappers/src/scheduler_cs.cpp
```diff
@@ -100,22 +96,31 @@ struct ThreadScheduler : public Scheduler {
...
-            return std::make_unique<SynchronizationContextScheduler>(context, post, release, is_on_context);
+            return std::make_unique<SynchronizationContextScheduler>(context);
         }
 
         return std::make_unique<ThreadScheduler>();
     });
 }
 
-REALM_EXPORT void realm_scheduler_invoke_function(void* function_ptr)
+REALM_EXPORT void realm_scheduler_invoke_function(void* function_ptr, bool execute_func)
```
nirinchev, line 116: `execute_func` will be false if the context was invalidated before invoke_function was called. In this case, we simply delete the allocated function.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -39,13 +39,20 @@ using SharedAsyncOpenTask = std::shared_ptr<AsyncOpenTask>;
...
-    void (*s_realm_changed)(void* managed_state_handle);
-    void (*s_get_native_schema)(SchemaForMarshaling schema, void* managed_callback);
-    void (*s_on_binding_context_destructed)(void* managed_handle);
-    void (*s_log_message)(realm_value_t message, util::Logger::Level level);
+    std::function<void(void* task_completion_source, ThreadSafeReference* ref, int32_t error_code, const char* message, size_t message_len)> s_open_realm_callback;
+    std::function<void(void* managed_state_handle)> s_realm_changed;
+    std::function<void(SchemaForMarshaling schema, void* managed_callback)> s_get_native_schema;
+    std::function<void(void* managed_handle)> s_on_binding_context_destructed;
+    std::function<void(realm_value_t message, util::Logger::Level level)> s_log_message;
+    std::atomic<bool> s_can_call_managed;
```
nirinchev, line 55: This will be `false` if the C# domain has been unloaded (thus it's not safe to invoke managed callbacks). It's set to `false` in `shared_realm_close_all_realms` which is invoked on domain unload and back to `true` in `shared_realm_install_callbacks`, which is invoked whenever the SharedRealmHandle is initialized.

nirinchev (commented):
@fealebenpae I've annotated the PR with some motivation for the changes - please let me know if you disagree with something or I've missed some case.

wrappers/src/app_cs.cpp
```diff
@@ -40,16 +40,21 @@ using namespace app;
 using SharedSyncUser = std::shared_ptr<SyncUser>;
 using SharedSyncSession = std::shared_ptr<SyncSession>;
 
+using LogMessageCallbackT = void(void* managed_handler, realm_value_t message, util::Logger::Level level);
```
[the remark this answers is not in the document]
LaPeste: I may be a silly question but, why do they have to be function pointers and not std::function-s? Is it because they have to have C-signature for exposure in shared_app_initialize?

wrappers/src/app_cs.cpp
```diff
@@ -116,19 +121,19 @@ extern "C" {
...
     {
         s_platform = Utf16StringAccessor(platform, platform_len);
         s_platform_version = Utf16StringAccessor(platform_version, platform_version_len);
         s_sdk_version = Utf16StringAccessor(sdk_version, sdk_version_len);
 
-        s_user_callback = user_callback;
-        s_void_callback = void_callback;
-        s_bson_callback = bson_callback;
-        s_log_message_callback = log_message_callback;
+        s_user_callback = wrap_managed_callback(user_callback);
```
LaPeste, line 133: And I assume this and the ones below are `std::function`s because of the issue you describe in `marshalling.hpp`, right?

nirinchev: Yep - I don't believe it's possible to avoid changing the type.

LaPeste (commented):
I can't be of much help in this PR, unfortunately. It was hard to follow without more context, but it surely looked interesting to have a nice Friday conversation on the topic. I look forward to it :)

papafe:
> I can't be of much help in this PR, unfortunately. It was hard to follow without more context, but it surely looked interesting to have a nice Friday conversation on the topic. I look forward to it :)

I definitely agree on this :)

wrappers/src/shared_realm_cs.cpp
```diff
@@ -114,17 +126,18 @@ extern "C" {
...
-    decltype(s_realm_changed) realm_changed, 
-    decltype(s_get_native_schema) get_schema, 
-    decltype(s_open_realm_callback) open_callback, 
-    decltype(s_on_binding_context_destructed) on_binding_context_destructed,
-    decltype(s_log_message) log_message)
+    RealmChangedT* realm_changed, 
+    GetNativeSchemaT* get_schema, 
+    OpenRealmCallbackT open_callback, 
+    OnBindingContextDestructedT on_binding_context_destructed,
+    LogMessageT log_message)
```
fealebenpae, line 133: Why are some of these pointers and the rest not?

PR #2305, opened 2021-03-22
Add e2e tests for Sets

nirinchev:
## Description

* Adds e2e test for Set.
* Applies a workaround and adds tests for the Set methods against non-set native collections.
* Restructures the way errors are reported from GetInstanceAsync to ensure the errors are converted to the correct managed instance.
* Updates the docker image to 2021-03-22 that includes the latest fixes and improvements from Cloud.
* Adds some utility testing functions to allow tests to fail if a session error is reported (e.g. when an invalid changeset is uploaded).
* Ensures that we reset `s_can_call_managed` in the other handle initializers.

Fixes https://github.com/realm/realm-dotnet/issues/2159
Fixes https://github.com/realm/realm-dotnet/issues/2212
Fixes https://github.com/realm/realm-dotnet/issues/2090

##  TODO

* [x] Tests (if applicable)
* [x] Uncomment and add the object tests when https://github.com/realm/realm-core/pull/4551 is done
* [ ] Function tests seem to fail due to an issue with the cloud js engine. Investigating.


LaPeste (approved):
Seems fine to me

Tests/Realm.Tests/Sync/DataTypeTests.cs
```diff
@@ -0,0 +1,146 @@
...
+        public void Set_DateTimeOffset() => TestSetCore(o => o.DateTimeOffsetSet, DateTimeOffset.MinValue, DateTimeOffset.MaxValue);
+
+        [Test]
+        public void Set_String() => TestSetCore(o => o.StringSet, "abc", "cde");
+
+        [Test]
+        public void Set_Binary() => TestSetCore(o => o.ByteArraySet, TestHelpers.GetBytes(5), TestHelpers.GetBytes(6), (a, b) => a.SequenceEqual(b));
+
+        //[Test]
+        //public void Set_Object() => TestSetCore(o => o.ObjectSet, new IntPropertyObject { Int = 5 }, new IntPropertyObject { Int = 456 }, (a, b) => a.Int == b.Int);
```
papafe, line 73: This doesn't work?

Tests/Realm.Tests/Database/TestObjects.cs
```diff
@@ -177,7 +177,7 @@ public class ListsObject : RealmObject
         public IList<RealmInteger<long>?> NullableInt64CounterList { get; }
     }
 
-    public class SetsObject : RealmObject
+    public class CollectionsObject : RealmObject
```
papafe, line 180: Are we going to use this on all other collection tests too?

Tests/Realm.Tests/Sync/DataTypeTests.cs
nirinchev: It didn't, but was fixed in https://github.com/realm/realm-core/pull/4551.

Tests/Realm.Tests/Database/TestObjects.cs
nirinchev: We'll see, I didn't want to refactor a bunch of other tests as that would explode the size of this PR. But I needed the other collections to be able to write tests that take a native collection as an argument.

PR #2313, opened 2021-03-26
Add tests for various partition keys

nirinchev:
## Description

Add additional apps that have different partition key types. It also adds a barebones admin client that fetches all the imported apps rather than have them passed in as command line arguments.

Fixes https://github.com/realm/realm-dotnet/issues/2093


Tests/Realm.Tests/Sync/SyncConfigurationTests.cs
```diff
@@ -30,8 +30,7 @@ public class SyncConfigurationTests : SyncTestBase
         [Test]
         public void SyncConfiguration_WithoutPath()
         {
-            var user = GetFakeUser();
```
nirinchev, line 33: The changes here are somewhat unrelated - I restructured the `SyncTestBase` a little and those were caught up in the refactoring.

Tests/Realm.Tests/Sync/SyncTestHelpers.cs
```diff
@@ -132,5 +162,103 @@ public static string[] ExtractBaasSettings(string[] args)
...
+
+            if (appName == "uuid-partition-key")
+            {
+                return AppConfigType.UUIDPartitionKey;
+            }
+
+            throw new NotSupportedException($"Unexpected appName: {appName}");
+        }
+
+        private class BaasClient : IDisposable
```
nirinchev, line 191: This is a super barebones helper that calls into the admin API and reads all the apps rather than have them passed in as config parameters. It's intentionally not overengineered/safe because it's not intended to be a general-purpose admin api client.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -89,12 +89,10 @@ public void GetInstanceAsync_ShouldDownloadRealm(bool singleTransaction)
         {
             SyncTestHelpers.RunBaasTestAsync(async () =>
             {
-                var user = await GetUserAsync();
```
nirinchev, line 92: This was also caught up in the `SyncTestBase` refactoring.

Tests/Realm.Tests/Sync/SyncTestHelpers.cs
```diff
@@ -75,34 +97,42 @@ public static string[] ExtractBaasSettings(string[] args)
...
                 }
             }
 
-            if (baasUrl != null && baasAppId != null)
+            ExtractBaasApps();
+
+            return result.ToArray();
+        }
+
+        private static void ExtractBaasApps()
```
nirinchev, line 118: This gets called either during initial startup (if run via the command line) or the first time a baas test is run. It's synchronous because I didn't want to refactor a whole lot of code around it to make it async and since it's only called once per process, it's not a huge deal.

Realm/Realm/Helpers/SerializationHelper.cs
```diff
@@ -36,6 +36,7 @@ static SerializationHelper()
         {
             var decimalSerializer = new DecimalSerializer(BsonType.Decimal128, new RepresentationConverter(allowOverflow: false, allowTruncation: false));
             BsonSerializer.RegisterSerializer(decimalSerializer);
+            BsonSerializer.RegisterSerializer(new GuidSerializer(GuidRepresentation.Standard));
```
nirinchev, line 39: This serializes uuids as type 4 rather than type 0. Cloud expects 4 and the MongoDB.Bson package defaults to 0 for legacy reasons.

papafe:
It's quite some changes but it seems to make sense

Issue #2260, opened 2021-02-23
Analytics - differentiate between Unity and regular users


nirinchev:
We can probably report the .NET platform (.NET Core/iOS/Unity/etc.)

PR #2317, opened 2021-04-02
Differentiate between Unity and regular .NET for metrics

nirinchev:
## Description

This PR restructures (again) how Unity-related things are packaged. We're moving the UnityWeaver package from the SetupUnityPackage project to Realm to simplify development. We're also extracting the core weaving logic into a shared project referenced by both the Fody and the Unity weavers. We're also adding a Fody config option to disable sending analytics as well as a Unity Editor menu item. For CI builds, users can disable analytics with an environment variable.

Fixes https://github.com/realm/realm-dotnet/issues/2260

##  TODO

* [ ] Changelog entry
* [x] Allow disabling analytics
* [x] Infer unity target platform for editor builds
* [x] Infer unity target platform for CI builds
* [x] Manually test


Realm/Realm.UnityWeaver/UnityWeaver.cs
```diff
@@ -28,17 +28,40 @@
...
+                Menu.SetChecked(EnableAnalyticsMenuItemPath, value);
+            }
+        }
+
         public int callbackOrder => 0;
 
         [InitializeOnLoadMethod]
         public static void Initialize()
         {
+            // We need to call that again after the editor is initialized to ensure that we populate the checkmark correctly.
```
LaPeste, line 59: Interesting quirk

nirinchev: Yeah, apparently InitializeOnLoad is invoked before the editor has fully rendered, so updating UI elements (like adding a checkmark to the menu item) won't do anything.

LaPeste (approved):
It's good that the structure of the project is now simplified. All looks fine to me.

PR #2333, opened 2021-04-13
Add workflow to verify code todos

nirinchev:
## Description

This is using a very simple action I created during skunkworks - it walks the changed files and emits warnings if there's a `// TODO: blah blah` comment that doesn't have a link to Jira/Github.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,14 @@
+name: lint
+"on":
+  pull_request:
+jobs:
+  verify-todos:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: nirinchev/verify-todo@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          include: "**/*.+(cs|cpp|hpp)"
+          exclude: "wrappers/realm-core/**"
+          pattern: "\\WR[A-Z]+-[0-9]+"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,8 +28,8 @@ stage('Checkout') {
       versionSuffix = "alpha.${env.BUILD_ID}"
     }
     // TODO: temp for beta releases
-    else if (env.CHANGE_BRANCH != null && env.CHANGE_BRANCH == "release/10.2.0-beta.1") {
-      versionSuffix = "beta.1"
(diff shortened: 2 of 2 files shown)


Jenkinsfile
```diff
@@ -28,8 +28,8 @@ stage('Checkout') {
       versionSuffix = "alpha.${env.BUILD_ID}"
     }
     // TODO: temp for beta releases
-    else if (env.CHANGE_BRANCH != null && env.CHANGE_BRANCH == "release/10.2.0-beta.1") {
-      versionSuffix = "beta.1"
+    else if (env.CHANGE_BRANCH != null && env.CHANGE_BRANCH == "release/10.2.0-beta.2") {
```
LaPeste, line 31: This change isn't related. If it's here by mistake, then I made you notice it. Otherwise, disregard this comment.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,14 @@
...
+  verify-todos:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: nirinchev/verify-todo@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          include: "**/*.+(cs|cpp|hpp)"
+          exclude: "wrappers/realm-core/**"
+          pattern: "\\WR[A-Z]+-[0-9]+"
```
LaPeste, line 14: I'm not sure what the pattern of the action is supposed to do, but I assume you tested it and it works.

LaPeste (approved):
Very nice and useful addition. I suppose we should think about moving the action from your personal repo to Realm's CI repo at some point.

Jenkinsfile
nirinchev: Yes - I forgot to include it in my previous PR, so figured might as well add it here 🤦‍♂️

bmunkholm:
Not too helpful I know - but I believe something exists that will check this but also create issues in case they don't exist already. I believe I heard about someone in another team using that. I can ask around if needed. But maybe it's not that useful now :-)?

nirinchev:
I explicitly didn't want to create issues automatically from TODO comments - it's not a massive amount of effort to go ahead and create one manually and judging by our TODOs, a manual ticket is likely to be a lot more meaningful than an automatic one.

PR #2339, opened 2021-04-19
CI - Create nuget packages (no Unity)

### LINKED ISSUE #2284: Create nuget packages (no Unity)

LaPeste:
## Description
Create nuget packages for SDK, excluded the unity package.

Fixes #2284 

## TODO
- [x] need to add version suffix to builds
- [x] ~~try to concat wrapper keys so to write in the resolved logical expression~~ (no need for this as it works already)


.github/pipeline.code-workspace
```diff
@@ -0,0 +1,7 @@
+{
```
papafe, line 1: What does this file do?

papafe, line 1: What does this file do?

LaPeste: It's a VSCode project file. It's just a shortcut to open the project. It can do multiple things, like putting together folders from different paths etc. Anyway, now atm this file just opens the folder where it's place.

papafe:
It looks fine to me, just one note.

Probably the wrapper part is the one that would take the longest, but wouldn't it make sense to use a similar approach for package generation? So, generate the packages only if the code has been modified, otherwise it's not necessary. I could imagine this could happen if we are working just on tests, for instance, and no other code is touched.

LaPeste:
Generally Yavor strongly supports the idea that: if it's fast, don't cache it. I guess @nirinchev could tip in his opinion here.

The numbers are the following:
- The whole nuget pkgs building process from fetching the artifacts of the wrappers to ending all the builds and store the resulting artifacts takes 2 sharp minutes on average.
- At the same time the sum of all the nuget pkgs are below 40 MB, so not much space would be take from the 5GB of cache.

LaPeste:
> 
> 
> It looks fine to me, just one note.
> 
> Probably the wrapper part is the one that would take the longest, but wouldn't it make sense to use a similar approach for package generation? So, generate the packages only if the code has been modified, otherwise it's not necessary. I could imagine this could happen if we are working just on tests, for instance, and no other code is touched.

Generally Yavor strongly supports the idea that: if it's fast, don't cache it. I guess @nirinchev could tip in his opinion here.

The numbers are the following:

    The whole nuget pkgs building process from fetching the artifacts of the wrappers to ending all the builds and store the resulting artifacts takes 2 sharp minutes on average.
    At the same time the sum of all the nuget pkgs are below 40 MB, so not much space would be take from the 5GB of cache.

.github/templates/main.yml
```diff
@@ -72,6 +73,20 @@ steps:
       retention-days: 1
 #@ end
 
+#@ def buildNugetPkgs():
+#@ for pkgName in nugetPackages:
+#@ buildCmd = "powershell MsBuild.exe ${{ github.workspace }}/Realm/" + pkgName + " -t:Pack -p:Configuration=Release -p:PackageOutputPath=${{ github.workspace }}/Realm/packages -restore"
```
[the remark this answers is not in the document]
nirinchev: Do we need to use `powershell` to invoke `MSBuild`? Can't we just execute it as it is?

.github/templates/main.yml
```diff
@@ -72,6 +73,20 @@ steps:
...
+#@ stepName = "Build nuget package " + pkgName
```
nirinchev, line 79: We should just inline this variable.

.github/templates/main.yml
```diff
@@ -72,6 +73,20 @@ steps:
...
+  - name: #@ stepName
+    run: #@ buildCmd
+#@ end
+  - name: Store artifacts
+    uses: actions/upload-artifact@v2
+    with:
+      path: ${{ github.workspace }}/Realm/packages/**
+      retention-days: 1
```
nirinchev, line 87: No need to do it now, but we should perhaps tweak that for master vs PR builds. Maybe something like `{{ github.event_name != 'pull_request' && 30 || 1 }}`

nirinchev:
Oh, and yeah - I agree with Yavor, let's not cache the packages as those are fairly quick to build.

.github/workflows/main.yml
```diff
@@ -205,63 +205,123 @@ jobs:
...
+      uses: microsoft/setup-msbuild@v1.0.2
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        submodules: recursive
+    - name: Set version suffix
+      run: |
+        if [[ $GITHUB_REF == "master" ]]
+        then
+          echo "BUILD_SUFFIX=alpha.$GITHUB_RUN_ID" >> $GITHUB_ENV
```
nirinchev, line 218: Why store those in env variables rather than step outputs? I kind of dislike relying on env variables too much.

LaPeste: It just seemed the consensus on how to do it and if I understand correctly, but I need to investigate further, the same env vars can be used in other jobs within the same runner without the need of: making an output for a step and and output for the job that reads the output of the step; or simply duplicated the step wherever needed.

nirinchev: According to this post: https://github.community/t/sharing-a-variable-between-jobs/16967/2, we can't share environment variables between jobs, so those need to be returned as outputs from the job anyway. Or has that changed since 2019?

LaPeste:
So I added the searching for the nuget version and the version suffix. All of that is exposed also at the job level, it'll be relevant for tests to grab the right nupkg. Plus a little fix for how the nupkg were uploaded.

.github/templates/main.yml
```diff
@@ -124,64 +153,82 @@ jobs:
...
+      - name: Set version suffix
+        id: set-version-suffix
+        run: |
+          if [[ $GITHUB_EVENT_NAME != "pull_request" ]]; then
+            echo "::set-output name=BUILD_SUFFIX::alpha.$GITHUB_RUN_ID."
+          elif [[ GITHUB_EVENT_NAME == "pull_request" ]]; then
+            echo "::set-output name=BUILD_SUFFIX::PR.$GITHUB_RUN_ID."
+          else
+            echo "set-output name=BUILD_SUFFIX::"
+          fi
```
nirinchev, line 172: Do we need bash for that? I thought we could use something like

```
outputs:
  build_suffix: ${{ github.event_name == 'pull_request' && format('PR-{0}.{1}', github.event.number, env. GITHUB_RUN_NUMBER) || format('alpha.{0}', env.GITHUB_RUN_NUMBER)}}
```

.github/templates/main.yml
```diff
@@ -72,6 +73,34 @@ steps:
...
+    id: find-nupkg-version
+    run: |
+      cd Realm/packages
+      tmpName=$(basename Realm.Fody* .nupkg)
+      pkgName=${tmpName#"Realm.Fody."}
+      echo "::set-output name=NUGET_PKG_VERSION::$pkgName"
+    shell: bash
+
+#@ for pkgName in nugetPackages:
+#@ finalPkgName = "${{ steps.set-version-suffix.outputs.BUILD_SUFFIX }}" + pkgName + ".${{ steps.find-nupkg-version.outputs.NUGET_PKG_VERSION }}.nupkg"
```
nirinchev, line 93: The nupkg version should include the suffix, so there's no need to explicitly have it in the finalPkgName.

.github/workflows/main.yml
```diff
@@ -205,15 +205,18 @@ jobs:
     - build-wrappers-linux
     - build-wrappers-windows-uwp
     outputs:
-      nuget_pkg_version: ${{ steps.find-nupkg-version.outpus.NUGET_PKG_VERSION }}
-      build_suffix: ${{ github.event_name == 'pull_request' && format('PR-{0}.{1}', github.event.number, env.GITHUB_RUN_NUMBER) || format('alpha.{0}', env.GITHUB_RUN_NUMBER)}}
+      nuget_pkg_version: ${{ steps.find-nupkg-version.outpus.nuget_pkg_version }}
```
nirinchev, line 208:
```suggestion
      nuget_pkg_version: ${{ steps.find-nupkg-version.outputs.nuget_pkg_version }}
```

LaPeste: 🤦‍♂️

nirinchev: That's why we should have strongly typed languages everywhere 😄

LaPeste: Strongly agree and also stronger glasses for me 😅.
I fixed this, let me know if there is any more feedback.

.github/templates/main.yml
```diff
@@ -124,64 +153,73 @@ jobs:
...
+    outputs:
+      nuget_pkg_version: ${{ steps.find-nupkg-version.outputs.nuget_pkg_version }}
     steps:
-      - name: Fetch Artifacts MacOS
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1.0.2
+      - #@ checkoutCode()
+      - name: Set version suffix
+        id: set-version-suffix
+        run: echo "::set-output name=build_suffix::${{ github.event_name == 'pull_request' && format('PR-{0}.{1}', github.event.number, github.run_number) || format('alpha.{0}', github.run_number)}}"
```
nirinchev, line 164: I wonder if this will work:

```suggestion
        #! Build suffix is PR-1234.5 for PR builds or alpha.123 for branch builds.
        run: echo "::set-output name=build_suffix::${{ github.event_name == 'pull_request' && format('PR-{0}', github.event.number) || 'alpha'}}.${{ github.run_number }}"
```

.github/templates/main.yml
```diff
@@ -72,6 +73,34 @@ steps:
       retention-days: 1
 #@ end
 
+#@ def buildNugetPkgs():
```
nirinchev, line 76: Let's split this into three functions - `buildPackages()` returns the nodes for actually building the packages, `findPackageVersion()` returns the node for looking up the package version, and finally, `uploadPackageArtifacts()` will return the nodes for uploading the artifacts.

.github/templates/main.yml
```diff
@@ -124,64 +153,73 @@ jobs:
       - build-wrappers-android
       - build-wrappers-linux
       - build-wrappers-windows-uwp
+    outputs:
+      nuget_pkg_version: ${{ steps.find-nupkg-version.outputs.nuget_pkg_version }}
```
nirinchev, line 157: We can call this `pacakge_version` as this is the version not only for the NuGet package but the SDK in general.

nirinchev (commented):
Few more nits, otherwise it looks good

.github/templates/main.yml
LaPeste: I would not do `findPackageVersion` a function since it isn't not reused and has "no loop" to perform.

nirinchev: It's more about reducing the complexity of the parent function than code reuse -  since the body of `find-nupkg-version` is a bit of a mouthful and not really interesting, I'd just like to avoid reading it 😅

.github/templates/main.yml
```diff
@@ -72,6 +73,37 @@ steps:
...
+    uses: actions/upload-artifact@v2
+    with:
+      name: #@ finalPkgName
+      path: #@ "${{ github.workspace }}/Realm/packages/" + finalPkgName
+      retention-days: ${{ github.event_name != 'pull_request' && 30 || 1 }}
+#@ end
+#@ end
+
+#@ def findPackageVersion():
+  - name: Find nupkg version
```
nirinchev, line 97: [nit] This can return the node directly rather than array element, similar to `checkoutCode` - then you won't need `template.replace` in the steps array.

nirinchev (approved):
One final nit and it's ready to merge.

Issue #2318, opened 2021-04-03
Undefined symbols error on build iOS Unity project


stdev33:
## Goals
Build Unity application for iOS target platform using Realm

## Expected Results
Finish build process with no linker errors.

## Actual Results
Linker error
`Undefined symbols for architecture arm64:
  "_realm_thread_safe_reference_destroy", referenced from:
      _ThreadSafeReferenceHandle_Unbind_m8B516E342FFE32FA967A0774363A935962952FCD in Realm1.o
      _NativeMethods_destroy_realm_reference_m5953D89668770CFD7399888C4CF3FBCC136A591B in Realm1.o
ld: symbol(s) not found for architecture arm64`

## Steps to Reproduce
1. Create an empty Unity project, switch Build target to iOS;
2. Import `realm.unity.bundle-10.1.2.tgz` using Package Manager;
3. Create sample .cs script that invokes `Realm.GetInstance()` and uses sample `RealmObject`. Attach script to any object in the SampleScene.unity;
4.  Export Xcode project (File --> Build Settings... --> Build);
5. Open Xcode project, setup your `Team`, `Bundle Identifier` and `Provisioning Profile` etc. in `Signing & Capabilities` tab;
6. Set `Enable Bitcode` property to `NO` for both Unity-iPhone and UnityFramework targets (Build Settings tab --> Build Options);
7. Try to build Xcode project (`Command+B` or top menu `Product --> Build`);
8. Linker error will be thrown `Undefined symbols for architecture arm64:
  "_realm_thread_safe_reference_destroy", referenced from:
      _ThreadSafeReferenceHandle_Unbind_m8B516E342FFE32FA967A0774363A935962952FCD in Realm1.o
      _NativeMethods_destroy_realm_reference_m5953D89668770CFD7399888C4CF3FBCC136A591B in Realm1.o
ld: symbol(s) not found for architecture arm64`

## Version of Realm and Tooling

- Realm Object Server Version: ? (you can see this as well in Studio once connected to a Realm instance) 
- Flavor:
  - [ ] Realm Cloud
  - [x] Self-Hosted
- Server OS & Version: ? (e.g. CentOS 6)
- Client SDK Version: realm.unity.bundle-10.1.2
- Client OS & Version: macOS Catalina 10.15.7
- Unity 2020.3.2f1 (also reproduced with Unity 2019.4.20f1)
- Xcode 12.4

nirinchev:
[...]

PR #2340, opened 2021-04-20
Add Unity tests

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2318.

This is a WIP effort to run our automated tests in a Unity context. The list of changes includes:
- [x] Added a Unity project in Tests/Tests.Unity.
- [x] Moved SetupUnityPackage from its own solution to the main one.
- [x] Extended SetupUnityPackage to support setting up the tests project. What it currently does is add the dependencies and then copies the source code files. Open for ideas on how to do it better.
- [x] Changed the weaver so that it tries to weave even editor assemblies (because test assemblies are classified as editor ones).
- [x] Fixed the framework inference for the Unity Weaver. Previously we used .NET Standard 2, but it's Mono, so should use NetFramework.
- [x] Reworked `MongoCollectionHandle.CallNativeMethod` to manage the lifecycle of its task completion source.
- [x] Removed BsonPayload and replaced it with primitive value to avoid a race condition where the native string data may be freed before we get a chance to read it.
- [x] Removed a missing `realm_thread_safe_reference_destroy` method that was causing build errors for iOS.
- [x] Refactored slightly tests dealing with weak references to make it easier to keep track of what's going on.

## TODO
- [x] Resolve conflicts with Realm.Windows.sln
- [x] Add SetupUnityPackage to Realm.sln
- [x] Copy the embedded files to the Unity tests project (needed for migration tests)


Tests/Realm.Tests/Realm.Tests.csproj
```diff
@@ -1,35 +1,60 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
 
   <PropertyGroup>
-    <TargetFrameworks>net461;netcoreapp3.1;monoandroid90;xamarin.ios10;xamarin.mac20</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);uap10.0.19041</TargetFrameworks>
-    <TargetFrameworks Condition="'$(AddNet5Framework)' == 'true'">$(TargetFrameworks);net5.0</TargetFrameworks>
+    <LocalDev>false</LocalDev>
```
nirinchev, line 4: Once this is merged, folks using Windows should be able to set it to `true` to exclude all non-net461 targets, thus speeding up VS builds.

nirinchev:
Thank God for Unity - a few more PRs like this and I'll solidify my #1 contributor spot by LoC committed 😅

PR #2342, opened 2021-04-20
Run tests on Windows .net framework

### LINKED ISSUE #2289: Run tests for .NET Framework on CI

### LINKED ISSUE #2292: Investigate test results reporting

LaPeste:
## Description
Run tests on Windows .net framework

The [chosen plugin/extension](https://github.com/EnricoMi/publish-unit-test-result-action) for reporting the tests is totally up for discussion. Out of the 3 I quickly looked at I just picked the one that seemed the most mature and with the highest downloads/interactions.
Various are the settings for the test result plugin, I just set what looked sensible. So, please, have a look at the list of options and share your thoughts about it.
28/04: I just noticed #2292. I suppose that it goes into this issue. Let me know if you think otherwise.

Fixes #2289
Fixes #2292 (open for discussion as per above)


.github/templates/main.yml
```diff
@@ -93,6 +93,17 @@ steps:
 #@ end
 #@ end
 
+#@ def fetchPackageArtifacts():
+#@ for pkgName in nugetPackages:
```
nirinchev, line 97: We don't really need all artifacts for tests - only Realm and Realm.Fody.

.github/templates/main.yml
```diff
@@ -228,4 +239,24 @@ jobs:
...
+    runs-on: windows-latest
+    name: Run tests for NET framework
+    needs: build-packages
+    steps:
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1.0.2
+      - #@ checkoutCode()
+      - #@ template.replace(fetchPackageArtifacts())
+      - name: Build the tests
+        run: #@ "MSBuild.exe ${{ github.workspace }}/Tests/Realm.Tests -p:RestoreConfigFile=${{ github.workspace }}/Tests/Test.Nuget.Config -p:TargetFramework=net461 -p:Configuration=" + configuration + " -p:UseRealmNupkgsWithVersion=${{ needs.build-packages.outputs.package_version }} -restore"
```
nirinchev, line 253: I think we can move the MSBuild node to a dedicated function.

.github/templates/main.yml
```diff
@@ -228,4 +244,23 @@ jobs:
...
+    name: Run tests for NET framework
+    needs: build-packages
+    steps:
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1.0.2
+      - #@ checkoutCode()
+      - #@ template.replace(fetchPackageArtifacts( [ "Realm", "Realm.Fody" ] ))
+      - #@ buildNetFrameworkTest()
+      - name: Run the tests
+        run: #@ "cd ${{ github.workspace }}/Tests/Realm.Tests/bin/" + configuration + "/net461\n./Realm.Tests.exe --result=TestResults.Windows.xml --labels=After\n"
```
nirinchev, line 259: Why do we need the `cd` here?

.github/templates/main.yml
```diff
@@ -104,6 +115,11 @@ run: |
 shell: bash
 #@ end
 
+#@ def buildNetFrameworkTest():
+name: Build the tests
+run: #@ "MSBuild.exe ${{ github.workspace }}/Tests/Realm.Tests -p:RestoreConfigFile=${{ github.workspace }}/Tests/Test.Nuget.Config -p:TargetFramework=net461 -p:Configuration=" + configuration + " -p:UseRealmNupkgsWithVersion=${{ needs.build-packages.outputs.package_version }} -restore"
```
nirinchev, line 120: My suggestion was to extract the msbuild invocation as a separate function that can be used elsewhere - e.g. in `buildPackages`. It can streamline the execution by e.g. accepting a relative path to the project, inserting the configuration, etc.

.github/templates/main.yml
```diff
@@ -104,6 +115,11 @@ run: |
 shell: bash
 #@ end
 
+#@ def buildNetFrameworkTest():
```
LaPeste, line 118: I deleted your comment by mistake, I meant to delete mine. Sorry about that. If you could re-add it, it'd be helpful for future refs, I suppose.
Anyway, my answer to that comment was the following:
Would you recommend sticking close to [what we do in jenkins](https://github.com/realm/realm-dotnet/blob/cfb988822d647ba3b6e2532cd1a01aaa03da8d39/Jenkinsfile#L411) or a more streamlined version of it?

nirinchev: Yes something similar, but we can definitely simplify some things. For example, it can accept just a relative path to the project as we can always prefix it with `${{ github.workspace }}` - we'll never have to build outside of that. We can also always set the configuration. And the rest will have to be a map of properties that we just concat with `-p:*key*=*value*`.

.github/templates/main.yml
```diff
@@ -104,6 +115,15 @@ run: |
 shell: bash
 #@ end
 
+#@ def buildNetFrameworkTest(relSolutionPath, **properties):
```
nirinchev, line 118: Can we rename this to just `msbuild` and reuse it on line 78 (will likely need to move other stuff as parameters)?

.github/templates/main.yml
```diff
@@ -228,4 +248,23 @@ jobs:
...
+      - #@ buildNetFrameworkTest("Tests/Realm.Tests", TargetFramework="net461" )
+      - name: Run the tests
+        run: #@ "${{ github.workspace }}/Tests/Realm.Tests/bin/" + configuration + "/net461/Realm.Tests.exe --result=TestResults.Windows.xml --labels=After"
+      - name: Publish Unit Test Results
+        uses: EnricoMi/publish-unit-test-result-action/composite@v1
+        if: always()
+        with:
+          files: TestResults.Windows.xml
+          comment_on_pr: false
+          check_name: Windows .NET framework tests results
```
nirinchev, line 270: Is this something we'll need in the future? Might be worth extracting it to a function, but might as well just do it in a follow up PR if we need to.

nirinchev (approved):
Final nits from me - feel free to merge when addressed.

.github/templates/main.yml
LaPeste: It's already done in the other PR, which is a branch of this. You can see it [here](https://github.com/realm/realm-dotnet/blob/4457127e51a7baf7f16945561290ed5184de5f88/.github/templates/main.yml#L107).
I didn't merge it backward.

PR #2347, opened 2021-04-22
Added missing DebuggerDisplay

### LINKED ISSUE #2047: Add [DebuggerDisplay] or a more meaningful ToString for Realm types
The interested types are our `Realm____` types, such as `RealmResult`, `RealmList`, `RealmObject`....

papafe:
This PR adds DebuggerDisplay to some Realm types. In particular debugger display contains:

- `RealmObjectBase`: `IsValid`
- `RealmValueType`: `Type - ToString()`  --> This can be useful to differentiate between different numerical types with similar string representation, for instance

Regarding the rest of the Realm types:
- Collection types (`Set`, `Dictionary`, `List`, `Result`) and their superclass `RealmCollectionBase` don't have anything meaningful to display in my opinion. For the "proper" collection types (`Set`, `Dictionary`, `List`), the parameter `Count` was already in the `DebuggerDisplay` attribute, and that seems fine.
- RealmInteger already prints the containing value in ToString(), and that also seems fine enough.

Fixes #2047

### DIFF
--- a/Realm/Realm/DatabaseTypes/RealmObjectBase.cs
+++ b/Realm/Realm/DatabaseTypes/RealmObjectBase.cs
@@ -41,7 +41,6 @@ namespace Realms
     /// </summary>
     [Preserve(AllMembers = true, Conditional = false)]
     [Serializable]
-    [DebuggerDisplay("{DebuggerDisplay(),nq}")]
     public abstract class RealmObjectBase
         : INotifyPropertyChanged,
           IThreadConfined,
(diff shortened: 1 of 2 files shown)


nirinchev:
I think we can be a bit more precise with these. Currently, this is what I see when I hover over a RealmObject - it's not particularly useful:

Screen Shot 2021-04-22 at 15 16 15

I think we can make it nicer by including the type of the object - perhaps by introducing a private property and using that for the debugger display? I was thinking of something like `Person { Id = 123 }, IsValid = true`. We can get the primary key property name from `ObjectSchema.PrimaryKeyProperty` and the primary key value from `_metadata.Helper.TryGetPrimaryKeyValue(this)`.

For RealmValue, we're currently outputting the namespaced type, which is long and boring. I was thinking we could replace it with something like `RealmValue, Type = String, Value = "abc"` and for objects, just `RealmValue, Person { Id = 123 }, IsValid = true`.

papafe:
I think it makes sense to add the primary key to the `DebugDisplay` for `RealmObjectBase`. I am not sure about adding the `Type` though, I think it would not be a useful information, as the majority of the time the developer would know already that. Besides, it's possible to see it in the "Locals" debug window.

Regarding RealmValue, I agree on adding  `Type = String, Value = "abc"`, but I don't think that adding the string `RealmValue` in front is useful, for the same reason as before. In this case, though, it definitely makes sense to print the object type.

papafe:
I have added some improvements, we can probably discuss about them :)

nirinchev (approved):

LaPeste (approved):

PR #2350, opened 2021-04-23
Disallow counter collections

### LINKED ISSUE #2308: Disallow collections of counters
Those are not currently supported by Core, so we should error out in the weaver.

papafe:
This PR adds an error in the weaver when declaring collections of `RealmInteger`.

Fixes #2308

### DIFF
--- a/Realm/Realm.Weaver/RealmWeaver.cs
+++ b/Realm/Realm.Weaver/RealmWeaver.cs
@@ -105,18 +105,6 @@ namespace RealmWeaver
             RealmValueTypeName,
         };
 
-        public static readonly HashSet<string> _counterTypes = new HashSet<string>
-        {
-            $"Realms.RealmInteger`1<{ByteTypeName}>",
-            $"Realms.RealmInteger`1<{Int16TypeName}>",
-            $"Realms.RealmInteger`1<{Int32TypeName}>",
-            $"Realms.RealmInteger`1<{Int64TypeName}>",
-            $"System.Nullable`1<Realms.RealmInteger`1<{ByteTypeName}>>",
-            $"System.Nullable`1<Realms.RealmInteger`1<{Int16TypeName}>>",
-            $"System.Nullable`1<Realms.RealmInteger`1<{Int32TypeName}>>",
-            $"System.Nullable`1<Realms.RealmInteger`1<{Int64TypeName}>>",
-        };
-
         private static readonly IEnumerable<string> _primaryKeyTypes = new[]
         {
             StringTypeName,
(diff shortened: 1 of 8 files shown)


papafe:
All seems to work, I just need to understand why I get an error during testing with `peverify` active

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -390,10 +402,18 @@ private WeavePropertyResult WeaveProperty(PropertyDefinition prop, TypeDefinitio
             {
                 var genericArguments = ((GenericInstanceType)prop.PropertyType).GenericArguments;
                 var elementType = genericArguments.Last();
-                if (!elementType.Resolve().IsValidRealmObjectBaseInheritor(_references) &&
-                    !_realmValueTypes.Contains(elementType.FullName))
+
+                if (!elementType.Resolve().IsValidRealmObjectBaseInheritor(_references))
                 {
-                    return WeavePropertyResult.Error($"{type.Name}.{prop.Name} is an {collectionType} but its generic type is {elementType.Name} which is not supported by Realm.");
+                    if (_counterTypes.Contains(elementType.FullName))
```
nirinchev, line 408: Rather than having a new list of counters, you can check if the type is realm integer by calling `elementType.Resolve().IsRealmInteger(out _, out _)`.

Tests/Weaver/Realm.Fody.Tests/WeaverTests.cs
```diff
@@ -83,7 +83,7 @@ private TestResult WeavePropertyChanged(string assemblyPath)
                 Config = config
             };
 
-            return weaver.ExecuteTestRun(assemblyPath, ignoreCodes: new[] { "80131869" });
+            return weaver.ExecuteTestRun(assemblyPath, runPeVerify: false, ignoreCodes: new[] { "80131869" }); ;
```
LaPeste, line 86: Double semicolon

nirinchev:
If you get the peverify error on master, it's probably not due to your changes :D

LaPeste:
> 
> 
> All seems to work, I just need to understand why I get an error during testing with `peverify` active

I missed it, has peverify always been there? Or has it been added recently?

papafe:
> > All seems to work, I just need to understand why I get an error during testing with `peverify` active
> 
> I missed it, has peverify always been there? Or has it been added recently?

It has been there for long time, and I think I already disabled it somewhere else in the past.

papafe:
> If you get the peverify error on master, it's probably not due to your changes :D

Yes, I've branched from master :)

nirinchev (approved):

PR #2352, opened 2021-04-27
Sync tests for float

### LINKED ISSUE #2348: Enable float in sync tests
The cloud team recently added float support, so we should enable it in our sync tests and verify it roundtrips as expected.

papafe:
This PR adds a couple of sync tests for float

Fixes #2348


LaPeste (approved):

papafe:
Added tests for properties and list, similarly to what was done for sets

PR #2354, opened 2021-04-27
CI - Add .Netcore, .Net5, Xamarin on MacOs, Weaver tests

### LINKED ISSUE #2287: Run tests for Xamarin.macOS on CI

### LINKED ISSUE #2290: Run tests for .NET Core and .NET 5
On Windows, Linux and macOS

### LINKED ISSUE #2291: Run tests for Weaver

LaPeste:
## Description
As the title says.

UPDATE - I decided to add the tests of xamarin on macOS and the weaver in this PR as too many branches were around with too little code in each.
I also added the conversion of the tests results of the weaver from the `Nunit3` format to the `Junit` one. This allows the tests results to be easily displayed by the majority of the GAs plugins for test reporting .

Fixes #2290 
Fixes #2287 
Fixes #2291


.github/templates/main.yml
```diff
@@ -266,10 +276,40 @@ jobs:
         _: #@ template.replace(msbuild("Tests/Realm.Tests", TargetFramework="net461", RestoreConfigFile="${{ github.workspace }}/Tests/Test.NuGet.Config", UseRealmNupkgsWithVersion="${{ needs.build-packages.outputs.package_version }}" ))
```
nirinchev, line 276: We can modify the `msbuild` function to return the entire node like:

```yaml
#@ def msbuild(projectPath, **properties):
#...
name: #@ "Build " + projectPath
run: #@ "msbuild ${{ github.workspace }}/" + projectPath + " " + target + " -p:Configuration=" + configuration + " -restore" + parsedProps
#@ end
```

which will then simplify the usage to:

```yaml
- #@ msBuild("Tests/Realm.Tests", ...)
```

.github/templates/main.yml
```diff
@@ -103,6 +103,16 @@ steps:
 #@ end
 #@ end
 
+#@ def publishTestsResults(files, test_title):
+  - name: Publish Unit Test Results
```
nirinchev, line 107: This can return a simple object rather than array element. That way we can avoid the `template.replace` when consuming the function.

.github/templates/main.yml
```diff
@@ -266,10 +276,40 @@ jobs:
...
-      - name: Publish Unit Test Results
-        uses: EnricoMi/publish-unit-test-result-action/composite@v1
-        if: always()
-        with:
-          files: TestResults.Windows.xml
-          comment_on_pr: false
-          check_name: Windows .NET framework tests results
+      - #@ template.replace(publishTestsResults("TestResults.Windows.xml", "Windows .NET framework tests results"))
+  run-tests-netcore-net5:
+    name: Run tests for .NetCore and .Net5
```
nirinchev, line 281:
```suggestion
    name: Run tests for .NET Core and .NET 5
```

.github/templates/main.yml
```diff
@@ -266,10 +276,40 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ macos-latest, windows-latest, ubuntu-latest ]
+        targetFramework: [ netcoreapp3.1, net5.0 ]
+    steps:
+    - #@ checkoutCode()
+    - #@ template.replace(fetchPackageArtifacts( [ "Realm", "Realm.Fody" ] ))
+    - name: Build the test
+      run: #@ "dotnet build ${{ github.workspace }}/Tests/Realm.Tests -c " + configuration + " -f ${{ matrix.targetFramework }} -p:RestoreConfigFile=${{ github.workspace }}/Tests/Test.NuGet.Config -p:UseRealmNupkgsWithVersion=${{ needs.build-packages.outputs.package_version }} -p:AddNet5Framework=${{ matrix.targetFramework == 'net5.0' && 'true' || 'false' }}"
```
nirinchev, line 292: Should we extract dotnet build to a function similar to msbuild?

.github/templates/main.yml
```diff
@@ -266,10 +276,40 @@ jobs:
...
+    - name: Run the tests
+      run: #@ "dotnet run -p ${{ github.workspace }}/Tests/Realm.Tests -c " + configuration + " -f ${{ matrix.targetFramework }} --no-build -- --labels=After --result=${{ github.workspace }}/TestResults.NetCore.xml"
```
nirinchev, line 294: Same for dotnet run?

nirinchev: Also, since we already built the binary, can we just run it rather than use `dotnet run`? It's a bit more hassle finding the right location, but it will make it easier to debug things in case of crashes. I had to do this here: https://github.com/realm/realm-dotnet/pull/2355/files.

Tests/Weaver/Realm.Fody.Tests/EmbeddedResources/trusted_ca.pem
```diff
@@ -0,0 +1,123 @@
+Certificate:
```
nirinchev, line 1: Why do we need this certificate?

Tests/Weaver/Realm.Fody.Tests/EmbeddedResources/nunit3-junit.xslt
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0" encoding="utf-8"?>
```
nirinchev, line 1: Let's just add the transform as a link rather than a copy.

Tests/Weaver/Realm.Fody.Tests/Realm.Fody.Tests.csproj
```diff
@@ -7,6 +7,20 @@
...
   </PropertyGroup>
 
+  <ItemGroup>
+    <None Remove="EmbeddedResources\ForMigrationsToCopyAndMigrate.realm" />
+    <None Remove="EmbeddedResources\nunit3-junit.xslt" />
+    <None Remove="EmbeddedResources\trusted_ca.pem" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Include="EmbeddedResources\ForMigrationsToCopyAndMigrate.realm" />
```
nirinchev, line 17: We don't need this in the weaver tests

Tests/Weaver/Realm.Fody.Tests/TestHelpers.cs
```diff
@@ -0,0 +1,62 @@
+﻿// ////////////////////////////////////////////////////////////////////////////
```
nirinchev, line 1: Double comments

Tests/Weaver/Realm.Fody.Tests/TestHelpers.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            var assembly = typeof(Program).Assembly;
+            var resourceName = assembly.GetManifestResourceNames().SingleOrDefault(s => s.EndsWith(realmName));
+            if (resourceName == null)
+            {
+                throw new Exception($"Couldn't find embedded resource '{realmName}' in the RealmTests assembly");
+            }
+
+            using (var stream = assembly.GetManifestResourceStream(resourceName))
+            using (var destination = new FileStream(destPath, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.None))
```
nirinchev, line 54: this will throw if `destPath` is null.

Tests/Weaver/Realm.Fody.Tests/TestHelpers.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Linq;
+using System.Xml;
+using System.Xml.XPath;
+using System.Xml.Xsl;
+
+namespace Realms.Fody.Tests
+{
+    public static class TestHelpers
+    {
+        public static void TransformTestResults(string resultPath)
```
nirinchev, line 30: Rather than copy these methods, let's split TestHelpers into two partial classes and have the weaver reference just one part.

nirinchev (changes requested):

.github/templates/main.yml
```diff
@@ -74,10 +74,12 @@ steps:
 #@ end
 
 #@ def buildPackages():
+#@ finalList = []
 #@ for pkgName in nugetPackages:
-  - name: #@ "Build nuget package " + pkgName
-    _: #@ template.replace(msbuild("Realm/" + pkgName, target="Pack", PackageOutputPath="${{ github.workspace }}/Realm/packages", VersionSuffix="${{ steps.set-version-suffix.outputs.build_suffix }}"))
+#@  finalList.append(msbuild("Realm/" + pkgName, target="Pack", PackageOutputPath="${{ github.workspace }}/Realm/packages", VersionSuffix="${{ steps.set-version-suffix.outputs.build_suffix }}"))
```
nirinchev, line 79: Any benefit of aggregating that in an array vs just returning the nodes as we iterate?

LaPeste: At line 307 I use `template.replace` since I need to replace with a list of objects. `template.replace` wants something returned from the function, that's why I return an array. The modification was made because you expressed preference to use `msbuild` as is without using `temple.replace` for every invocation. This made the function `msbuild` not return anymore an array and so I had to rebuild the array in `buildPackage`.

Tests/Weaver/Realm.Fody.Tests/Realm.Fody.Tests.csproj
```diff
@@ -7,6 +7,18 @@
...
+  <ItemGroup>
+    <EmbeddedResource Include="..\..\Realm.Tests\EmbeddedResources\nunit3-junit.xslt">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </EmbeddedResource>
+
+    <Compile Include="..\..\Realm.Tests\TransformHelpers.cs" Link="TransformHelpers.cs" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <LangVersion>8.0</LangVersion>
```
LaPeste, line 19: This was added to allow C# 8 "using" syntax in fody since it targets also net472, which only supports C# 7.3 unless "forced" to use some parts of C# 8.

Tests/Realm.Tests/TransformHelpers.cs
```diff
@@ -0,0 +1,64 @@
...
+        public static void TransformTestResults(string resultPath, string transformFilePath)
+        {
+            var xpathDocument = new XPathDocument(resultPath);
+            var transform = new XslCompiledTransform();
+            transform.Load(transformFilePath);
+            using var writer = new XmlTextWriter(resultPath, null);
+            transform.Transform(xpathDocument, null, writer);
+        }
+
+        public static string CopyBundledFileToDocuments(string realmName, string destPath)
```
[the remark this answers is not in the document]
nirinchev: Also, let's rename it to `ExtractBundledFile` and change the first parameter to be `fileName` rather than `realmName` since it's not used only for realm files anymore.

nirinchev (approved):
Final nit re: `TransformHelpers.CopyBundledFileToDocuments` and once we remove the commenting functionality from the test reporter, it's ready to merge.

PR #2357, opened 2021-04-28
Remove Lambda.Compile from results visitor

nirinchev:
Replaces `Lambda.Compile` in the query visitor with a newly introduced `Operator.Convert<T>(object)` call. While it's not as versatile as `Lambda.Compile`, it's at least supported in IL2CPP.


nirinchev:
The android tests seem to fail to build due to a recent VS upgrade. Will merge like this and deal with the build failures later.

PR #2358, opened 2021-04-28
Add FileLogger

### LINKED ISSUE #2349: Add a File logger to the list of predefined loggers
We already provide `Console`, `Function`, and `Null` loggers - we should probably add a `File` one as well.

papafe:
This PR adds support for `FileLogger`

Fixes #2349

### DIFF
--- a/Realm/Realm/Logging/Logger.cs
+++ b/Realm/Realm/Logging/Logger.cs
@@ -17,6 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 using System;
+using System.IO;
 using System.Runtime.InteropServices;
 using System.Text;
 using Realms.Sync;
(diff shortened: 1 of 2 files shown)


Realm/Realm/Logging/Logger.cs
```diff
@@ -144,6 +155,23 @@ protected override void LogImpl(LogLevel level, string message)
...
+
+            public FileLogger(string filePath, Encoding encoding = null)
+            {
+                _filePath = filePath;
+                _encoding = encoding ?? Encoding.UTF8;
+            }
+
+            protected override void LogImpl(LogLevel level, string message)
+            {
+                System.IO.File.AppendAllText(_filePath, FormatLog(level, message) + Environment.NewLine, _encoding);
```
nirinchev, line 171: This will be problematic because 1) it requires writing to the filesystem, which may be slow, meaning that the logger will negatively influence the performance of the system and 2) it's not taking any locks, so invoking the method from multiple threads concurrently will result in an IO exception being thrown.

papafe:
I have added a lock to avoid issues with multiple threads. 

Regarding the actual implementation of the file logger, there are essentially two ways:

1. Open the file when we need to log
2. Keep a stream open

The performance of n.2 are better than n.1 because opening a file is an "expensive" operation. Given that though, number 2 would also complicate further the logger, as we would need to take care of disposing the stream (somehow) and give the developer the possibility of doing so, and that's something that would require changing the way that the `Logger` class is implemented. Besides, it gets also a little more complicated to read the log file if we keep the stream open. 
I have also looked around, and some other libraries for logging (such as `Nlog`) give the possibility of choosing between the two options, but they are also much more complicated products.
Given all of that, and also that the developer can provide his own implementation (maybe with a more robust external library), I would prefer to go with the simplest option, eventually putting a note in the documentation about this.

Realm/Realm/Logging/Logger.cs
papafe: We discussed about this offline

LaPeste (approved):
I think that the point you made in your comment about the 2 implementations sounds fair. Maybe if we see a high amount of complains from users, we may rethink the strategy later on. At least they have the possibility to bring their own implementation.
Anyway, nothing to comment on the code itself. All seems fine.

nirinchev (approved):

PR #2374, opened 2021-05-12
Added test for property acess after delete rollback

### LINKED ISSUE #1332: Deleted object is not revived after a transaction rollback
The following snippet fails:
```csharp
var person = new Person();
_realm.Write(() => _realm.Add(person));

using (var transaction = _realm.BeginWrite())
{
    _realm.Remove(person);
    transaction.Rollback();
}

var test = person.FirstName;
```

It looks like the row reference is removed upon deletion and not restored after transaction rollback.

papafe:
This PR adds a test to be sure we don't introduce any regressions with deletion and rollbacks. 

Fixes #1332


nirinchev (approved):
Nice!

PR #2380, opened 2021-05-18
Add an alternative dynamic API

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2379

Addresses some tests from the list in https://github.com/realm/realm-dotnet/issues/2372.

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm/DatabaseTypes/RealmObjectBase.cs
```diff
@@ -501,5 +506,99 @@ public Metadata(TableKey tableKey, IRealmObjectHelper helper, IDictionary<string
                 Schema = schema;
             }
         }
+
+        /// <summary>
+        /// A class that exposes a set of API to access the data in a managed RealmObject dynamically.
+        /// </summary>
+        public struct Dynamic
```
nirinchev, line 513: @papafe @fealebenpae @LaPeste @DominicFrei - what do you folks think about this departure from the proposed design? I moved all the string-based API in a new object so they are "namespaced" together:

```csharp
RealmObject foo = ...;

foo.DynamicApi.Get<int>("MyIntProp");
foo.DynamicApi.Set("MyStringProp", "abc");
```

The main motivation there is that I didn't want those to pollute the user models as these API would have been suggested along with the original user properties even though in 99% of the cases, the user wouldn't need them. The alternatives I considered are:
* Keep them exposed directly on `RealmObjectBase` but annotate with `[EditorBrowsable(Never)]`. The reason why I didn't like this approach was that we would make them quite a bit less discoverable and people would need to consult the docs for every single thing.
* Have RealmObjectBase implement some special interface explicitly and ask people to cast to that. I didn't like this approach a lot because we couldn't declare a method as returning the union of RealmObjectBase, meaning that it would always be one or the other and users would need to cast back and forth depending on whether they need the RealmObject API or the dynamic ones.
* Have just `object Get(string propName)` and `Set(string propName, object)` that would handle lists, sets, singular values and so on. That wasn't great either because it requires quite a bit of casting on the user side and quite a bit of guessing - do I cast to `IList`/`List`/`RealmList` and so on.
* Have some RealmObject inheritor that exposes these methods directly and have the dynamic methods return that. I decided against that because in some cases, the dynamic API are actually returning the user-defined models and that would be impossible if we also needed to return the special type.

DominicFrei: Looking into the options the reasons against them I think the chosen solution seems to be the best way too.
Also, big fan of namespaces as they often improve readability. 👍

Realm/Realm/Schema/PropertyTypeEx.cs
```diff
@@ -174,13 +174,47 @@ public static RealmValueType ToRealmValueType(this PropertyType type)
...
+            else if (propertyType.IsDictionary())
+            {
+                collection = PropertyType.Dictionary;
+            }
+            else
+            {
+                collection = default;
+            }
+
+            return collection != default;
```
DominicFrei, line 202: Would it make sense to name `collection` differently because it can be something that is not a collection?
Or alternatively remove the last `else` and therefore only set it to something that is a `collection` and in the end check for `null`?

Realm/Realm/DatabaseTypes/RealmObjectBase.cs
```diff
@@ -501,5 +506,99 @@ public Metadata(TableKey tableKey, IRealmObjectHelper helper, IDictionary<string
...
+        {
+            private readonly RealmObjectBase _realmObject;
+
+            internal Dynamic(RealmObjectBase ro)
+            {
+                _realmObject = ro;
+            }
+
+            public T Get<T>(string propertyName)
```
DominicFrei, line 522: I do get `Elements should be documented` warnings on this and all other `public` elements in this `struct`.

Realm/Realm/Schema/PropertyTypeEx.cs
nirinchev: We are forced to set `collection` to something - the compiler will slap us otherwise. This approach is more or less consistent with other `TryGetXYZ` methods - if the method returns `true`, the `out` parameter will contain a value that is meaningful. If it's `false`, then the out parameter should not be used. So roughly:

```csharp
PropertyType myProp = ...;
if (myProp.IsCollection(out var collection))
{
    // myProp represents some sort of collection
    // the `collection` variable will be set to either Array/Dict/Set
}

// myProp is not a collection
// the `collection` variable is set to the default value
```

Realm/Realm/DatabaseTypes/RealmObjectBase.cs
nirinchev: Yes, it's not "ready" ready for review but wanted to put it up to discuss the API. Once we're aligned on that, I'll add API docs.

Realm/Realm/Schema/PropertyTypeEx.cs
```diff
@@ -174,13 +174,47 @@ public static RealmValueType ToRealmValueType(this PropertyType type)
...
+        }
+
+        public static string GetDotnetTypeName(this Property property)
+        {
+            _ = property.Type.IsCollection(out var collection);
```
DominicFrei, line 207: Would this be an alternative?

```
public static bool IsCollection(this PropertyType propertyType) => propertyType.IsArray() || propertyType.IsSet() || propertyType.IsDictionary();

public static string GetDotnetTypeName(this Property property)
{
    var format = property.Type switch
    {
        PropertyType.Array => "IList<{0}>",
        PropertyType.Set => "ISet<{0}>",
        PropertyType.Dictionary => "IDictionary<string, {0}>",
        _ => "{0}"
    };

     return string.Format(format, property.ObjectType ?? property.Type.ToType().Name);
}
```

Realm/Realm/Schema/PropertyTypeEx.cs
DominicFrei: Understood. Might be just different conventions then. 👍 
Would something like `tryGetAsCollection` be an alternative?

Realm/Realm/DatabaseTypes/RealmObjectBase.cs
DominicFrei: 👍

Realm/Realm/Schema/PropertyTypeEx.cs
nirinchev: `property.Type` is a flag enum, so it will never be equal exactly to `Array`. It can be equal to e.g. `Array | Double | Required` to denote that this is a list of required double values. So `switch property.Type` would match `_`.

Realm/Realm/Schema/PropertyTypeEx.cs
nirinchev: I considered that but I wanted it to be consistent with the rest of the `IsXXX` extension methods.

DominicFrei: Alright. 👍

Realm/Realm/DatabaseTypes/RealmObjectBase.cs
papafe: I also agree that this seems to be the best option, especially considering it's something that is "rarely" used.

LaPeste: I fully agree with the exposed reasoning. It's nice touch given the rarity of usage.

nirinchev:
This is now ready for review - I'm going to add a few more tests for the new API to test error conditions. The happy path is already tested by the now-refactored dynamic tests.

nirinchev:
Okay, this is now ready and I've added tests for the new API on top of reworking the existing tests to work both with the dynamic and the dictionary-like API.

nirinchev:
The failing tests are unrelated - they are just surfaced because this PR fixes the sync tests code to actually fail the test when there's an exception rather than silently ignore it.

PR #2382, opened 2021-05-20
Fix a few GA nits

nirinchev:
## Description

Fixes a few nits that I noticed when reviewing https://github.com/realm/realm-dotnet/pull/2354, notably:
- `buildPackages` yield returns array nodes while iterating.
- `dotnetPublish` is replaced with `dotnetPublishAndRun` which publishes an executable then runs it.
- `dotnetPublishAndRunTests` is added as a convenience wrapper around `dotnetPublish` for when we need to setup some of the build properties
- Modified the weaver tests to use `dotnetPublish` - there was no fundamental issue with `dotnetRun` but since it was the only job that used it, figured might as well unify things.
- Modified the weaver test project to output executable

### DIFF
--- a/.github/templates/main.yml
+++ b/.github/templates/main.yml
@@ -108,7 +108,7 @@ steps:
 #@   propsArg += " -p:" + prop + "=" + properties[prop]
 #@ end
 #@
-#@ runtime = "${{ (runner.os == 'macOS' && 'osx-x64') || (runner.os == 'Windows' && 'win-x64') || (runmner.os == 'Linux' && 'linux-x64') || '???' }}"
+#@ runtime = "${{ (runner.os == 'macOS' && 'osx-x64') || (runner.os == 'Windows' && 'win-x64') || (runner.os == 'Linux' && 'linux-x64') || '???' }}"
   - name: #@ "Publish " + projectPath
     run: #@ "dotnet publish " + projectPath + " -c " + configuration + " -f " + framework + " -r " + runtime + propsArg
     shell: bash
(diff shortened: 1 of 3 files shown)


.github/templates/main.yml
```diff
@@ -105,6 +102,32 @@ steps:
...
 
+#@ def dotnetPublishAndRunTests(projectPath, framework, executeCommand, addNet5):
+#@ properties = {
+#@   "AddNet5Framework": addNet5,
+#@   "RestoreConfigFile": "Tests/Test.NuGet.Config",
+#@   "UseRealmNupkgsWithVersion": "${{ needs.build-packages.outputs.package_version }}"
+#@ }
+#@
+#@ return dotnetPublishAndRun(projectPath, framework, executeCommand, properties)
+#@
```
LaPeste, line 113: This empty line can be removed.

.github/templates/main.yml
```diff
@@ -74,12 +74,9 @@ steps:
 #@ end
 
 #@ def buildPackages():
-#@ finalList = []
 #@ for pkgName in nugetPackages:
-#@  finalList.append(msbuild("Realm/" + pkgName, target="Pack", PackageOutputPath="${{ github.workspace }}/Realm/packages", VersionSuffix="${{ steps.set-version-suffix.outputs.build_suffix }}"))
+  - #@ msbuild("Realm/" + pkgName, target="Pack", PackageOutputPath="${{ github.workspace }}/Realm/packages", VersionSuffix="${{ steps.set-version-suffix.outputs.build_suffix }}")
```
LaPeste, line 78: Now I understand what you meant in your comment last time. This looks better than having the array. Thanks.

LaPeste (approved):
Thank you for the work.

PR #2384, opened 2021-05-21
Run UWP tests on CI

### LINKED ISSUE #1348: Run tests for UWP on CI
Depends on #1081

LaPeste:
## Description
UWP tests were never run headlessly on CI, hence adjustments had to be made. Namely, make a UWP app launchable from terminal and make such app process the command line input. This is what this PR adds on top of the standard workflow additions that you're used to see.

Fixes #1348 

##  TODO

- [x] Look again into parsing parameters from terminal
- [x] Check if it's possible to quickly estrapolate  package family name for a given uwp app (get-appxpackage -name realm.uwp.tests | select -expandproperty PackageFamilyName)


.github/workflows/main.yml
```diff
@@ -316,3 +316,157 @@ jobs:
...
         retention-days: ${{ github.event_name != 'pull_request' && 30 || 1 }}
+  run-tests-net-framework:
+    runs-on: windows-latest
+    name: Run tests for NET framework
+    needs: build-packages
+    steps:
+    - name: Add msbuild to PATH
+      uses: microsoft/setup-msbuild@v1.0.2
+    - name: Checkout code
+      uses: actions/checkout@v2
```
DominicFrei, line 327: The versioning here (`checkout@v2`, `download-artifact@v2`) appears in multiple locations.
Is this likely to change at some point and if so would it make sense to have the version extracted to one common location?

.github/workflows/main.yml
```diff
@@ -316,3 +316,157 @@ jobs:
...
+      uses: actions/download-artifact@v2
+      with:
+        name: Realm.Fody.${{ needs.build-packages.outputs.package_version }}.nupkg
+        path: ${{ github.workspace }}/Realm/packages/
+    - name: Build the tests
+      run: msbuild ${{ github.workspace }}/Tests/Realm.Tests -p:RestoreConfigFile=${{ github.workspace }}/Tests/Test.NuGet.Config -p:Configuration=Release -p:UseRealmNupkgsWithVersion=${{ needs.build-packages.outputs.package_version }} -restore -p:TargetFramework=net461
+    - name: Run the tests
+      run: ${{ github.workspace }}/Tests/Realm.Tests/bin/Release/net461/Realm.Tests.exe --result=TestResults.Windows.xml --labels=After
+    - name: Publish Unit Test Results
+      uses: EnricoMi/publish-unit-test-result-action/composite@v1
```
DominicFrei, line 345: What happens if this repository disappears at some point?
Does it make sense to just fork it and point to our local copy within the Realm org?

Tests/Tests.UWP/TestHelpersUWP.cs
```diff
@@ -0,0 +1,71 @@
+﻿// ////////////////////////////////////////////////////////////////////////////
+// //
+// // Copyright 2021 Realm Inc.
```
DominicFrei, line 3: There seem to be double comments on this licence part.

Tests/Tests.UWP/TestHelpersUWP.cs
```diff
@@ -0,0 +1,71 @@
...
+        public static string GetResultPath(string cmdParams)
+        {
+            if (!string.IsNullOrEmpty(cmdParams))
+            {
+                var resultStr = "--result=";
+                var indexEndKey = cmdParams.IndexOf(resultStr);
+                if (indexEndKey != -1)
+                {
+                    // TODO unfortunately this assumes that no space exists in the path, find a better way!
+                    var indexEndValue = cmdParams.IndexOf(" ", indexEndKey);
```
DominicFrei, line 55: If there are spaces within the path I assume they are escaped in some way?
Could you search for the first unescaped space instead?

.github/workflows/main.yml
nirinchev: This is the generated workflow file - the template will only specify the versions once: https://github.com/realm/realm-dotnet/blob/6a036617f714bb969bf938588b9fa555abf1adab/.github/templates/main.yml#L11-L16

If we wanted to change it, this is the only place that would need adjustment.

.github/workflows/main.yml
nirinchev: I wouldn't be too worried about it - we don't typically fork repos "just in case" - if it disappears we'll figure it out 😄

.github/workflows/main.yml
DominicFrei: Alright, sounds great. 🚀

.github/workflows/main.yml
DominicFrei: 👍

nirinchev (changes requested):
Looks very nice, a few final comments:
1. Looks like NUnit.Xamarin.dll is built in debug mode - it'll be a good idea to rebuild it in release.
2. We should remove `Dependencies` from the [NuGet.Config](https://github.com/realm/realm-dotnet/blob/67ea45d9ce26d3217078e14ffc1c9af78be6d11b/NuGet.Config#L5) as there are no nuget packages there anymore.
3. We should update [Tests.iOS.csproj](https://github.com/realm/realm-dotnet/blob/67ea45d9ce26d3217078e14ffc1c9af78be6d11b/Tests/Tests.iOS/Tests.iOS.csproj#L134-L136) and Android to not use nunit.xamarin.realm as a package dependency and instead reference the dll directly.

Tests/Tests.UWP/TestHelpersUWP.cs
```diff
@@ -0,0 +1,65 @@
...
+    static class TestHelpersUWP
+    {
+        public static async Task TransformTestResults(string resultsPath)
+        {
+            TestHelpers.CopyBundledFileToDocuments("nunit3-junit.xslt", "nunit3-junit.xslt");
+            var xsltPath = RealmConfigurationBase.GetPathToRealm("nunit3-junit.xslt");
+            var xsltSf = await StorageFolder.GetFolderFromPathAsync(Path.GetDirectoryName(xsltPath));
+            var localFolderSf = ApplicationData.Current.LocalFolder;
+            var xsltFileSf = await xsltSf.GetFileAsync("nunit3-junit.xslt");
+            var xsltFile = await XmlDocument.LoadFromFileAsync(xsltFileSf);
```
nirinchev, line 37: These file calls seem overly complicated - I think you can simply call `StorageFolder.GetFileFromPathAsync` without going through the intermediate steps. It's possible that the UWP API have changed, but this is what I was previously using and it seems a bit more readable:

https://github.com/nirinchev/nunit.xamarin/blob/534d8f31c58dee22aa9e7fa2f3c810931446ccbe/src/runner/nunit.runner/Services/XmlFileTransformer.cs#L53-L61

Tests/Tests.UWP/MainPage.xaml.cs
```diff
@@ -16,25 +16,61 @@
...
+                    {
+                        throw new Exception("You must provide path to store test results with --resultpath path/to/results.xml");
+                    }
+
+                    _nunit.Options.ResultFilePath = Path.Combine(ApplicationData.Current.LocalFolder.Path, resultPath);
+                    _nunit.Options.OnCompletedCallback = async () =>
+                    {
+                        await TestHelpersUWP.TransformTestResults(_nunit.Options.ResultFilePath);
+                        outputWriter.Dispose();
+                        Console.WriteLine("Test finished, reporting results");
```
nirinchev, line 66: I think we can call `CoreApplication.Exit` here to terminate the app in headless mode.

Tests/Tests.UWP/Configuration.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Realms.Tests.UWP
+{
+    internal class Configuration
```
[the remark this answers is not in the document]
LaPeste: A pesky `git add -A` 🤦🏻 
Thanks

Tests/Tests.UWP/MainPage.xaml.cs
LaPeste: I'll try it out. Although I think I remember to have read that Exit should not be called? I'll do some research about that.

Tests/Tests.UWP/Tests.UWP.csproj
```diff
@@ -178,6 +201,14 @@
       <Name>Realm</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <None Remove="Tests.UWP_TemporaryKey.pfx" />
```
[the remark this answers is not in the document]
LaPeste: You're right. .gitignore ignores that, so I'm gonna remove it.

Tests/Tests.UWP/MainPage.xaml.cs
nirinchev: Maybe not in a real app, but in a test one, it should be just fine

LaPeste: Alright, then I'll just add it. Thanks.

nirinchev (approved):

PR #2401, opened 2021-05-27
Use delete_files from core when deleting a Realm.

### LINKED ISSUE #386: Move `DeleteRealm` into object store

### LINKED ISSUE #2455: DeleteRealm wallways say file locked
## Goals

Delete local realm file and logout.

## Expected Results

After dispose Instance, file is free to delete.

## Actual Results

System.IO.IOException
  HResult=0x80070020
  Message=The process cannot access the file 'C:\Users\xxxxxxxxxxxxxxx.realm' because it is being used by another process.
  Source=System.IO.FileSystem
  StackTrace:
   at System.IO.FileSystem.DeleteFile(String fullPath)
   at System.IO.File.Delete(String path)
   at Realms.Realm.DeleteRealm(RealmConfigurationBase configuration)

## Steps to Reproduce

Run debugging and always receive this error. Instance is closed .
image

## Code Sample

```csharp
        public async void Logout()
        {
            if (RealmUser != null)
            {
                if (RealmSyncConfig != null)
                {
                    await RealmInstance.GetSession().WaitForUploadAsync();
                    RealmInstance.Dispose();
                    Realm.DeleteRealm(RealmSyncConfig);
                }
                await RealmUser.LogOutAsync();
                RealmSyncConfig = null;
            }
        }
```

- Flavor:
  - [ x] MongoDB Realm
  - [ ] Local Database only
- Client SDK Version: ? Realm  .Net 10.2.0
- Client OS & Version: ? Windows 10 build 19041
Microsoft Visual Studio Professional 2019
Version 16.10.2
VisualStudio.16.Release/16.10.2+31410.357
Microsoft .NET Framework
Version 4.8.04084

Installed Version: Professional

Visual C++ 2019   00435-60000-00000-AA316
Microsoft Visual C++ 2019

ADL Tools Service Provider   1.0
This package contains services used by Data Lake tools

ASA Service Provider   1.0

ASP.NET and Web Tools 2019   16.10.526.50910
ASP.NET and Web Tools 2019

ASP.NET Web Frameworks and Tools 2019   16.10.526.50910
For additional information, visit https://www.asp.net/

Azure App Service Tools v3.0.0   16.10.526.50910
Azure App Service Tools v3.0.0

Azure Data Lake Node   1.0
This package contains the Data Lake integration nodes for Server Explorer.

Azure Data Lake Tools for Visual Studio   2.6.3000.0
Microsoft Azure Data Lake Tools for Visual Studio

Azure Functions and Web Jobs Tools   16.10.526.50910
Azure Functions and Web Jobs Tools

Azure Logic Apps Tools for Visual Studio   1.0
Add-in for the Azure Resource Group project to support the Logic App Designer and template creation.

Azure Stream Analytics Tools for Visual Studio   2.6.3000.0
Microsoft Azure Stream Analytics Tools for Visual Studio

C# Tools   3.10.0-4.21269.26+029847714208ebe49668667c60ea5b0a294e0fcb
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Dotfuscator Community Edition   6.3.0+8b29d6f4a2
PreEmptive Protection - Dotfuscator CE

Extensibility Message Bus   1.2.6 (master@34d6af2)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

Fabric.DiagnosticEvents   1.0
Fabric Diagnostic Events

GitHub.VisualStudio   2.11.106.19330
A Visual Studio Extension that brings the GitHub Flow into Visual Studio.

IntelliCode Extension   1.0
IntelliCode Visual Studio Extension Detailed Info

Liquid Studio 2021 - Graphical WSDL Editor   19.0.10
Liquid Studio 2021 - Graphical WSDL Editor
    
The WSDL Editor abstracts much of the complexity of the WSDL standard, presenting the structure of the WSDL without showing the unnecessary complexity.

This provides a view that makes understanding and editing your WSDL document simple. Access to the full range of related facets is available via the properties window and the WSDL can be edited directly within the Source view.

Embedded XML Schema within your WSDL can be displayed and edited within an XML Schema Editor View, making defining types for your WSDL simple.

Liquid Studio 2021 - JSON Schema Editor   19.0.10
Liquid Studio 2021 - JSON Schema Editor
    
The JSON Schema Editor abstracts away the complexities of JSON Schema structure by providing an intuitive graphical drag and drop user interface.

The Graphical JSON Schema Designer included in Liquid Studio massively simplifies visualizing, authoring and navigating complex JSON Schemas. The Graphical JSON Schema Editor provides an interactive logical view of the JSON Schema enabling simple and intuitive editing and validation of the JSON Schema, without compromising the ability to work with all the complexities of the Internet Engineering Task Force draft 4.0 JSON Schema Standard.

Liquid Studio 2021 - Web Service Test Client   19.0.10
Liquid Studio 2021 - Web Service Test Client
    
The Web Service Test Client makes it quick and easy to make ad-hoc calls to a web service. Simply select the web service URL, pick the web method from the list, amend any values in the generated SOAP request, and view the results.

Liquid Studio 2021 - XML Schema Editor   19.0.10
Liquid Studio 2021 - XML Schema Editor
    
The XML Schema Editor abstracts away the complexities of XSD structure by providing an intuitive graphical drag and drop user interface.

The Graphical XML Schema Designer included in Liquid Studio massively simplifies visualizing, authoring and navigating complex XML Schemas. The Graphical XSD Editor provides an interactive logical view of the XML Schema enabling simple and intuitive editing and validation of the XSD, without compromising the ability to work with all the complexities of the W3C XML Schema standard.

Liquid Studio 2021 - XPath Expression Builder   19.0.10
Liquid Studio 2021 - XPath Expression Builder
    
The XPath Expression Builder tool allows you to construct and view the results of your query as you type.

It attaches itself to the active XML Editor, so any XML panel in the environment can be queried.

The XPath Query Editor comes with Intellisense, which is intelligently taken from the content of the current XML document. The results of the XPath expression can then be seen in the active XML Editor.

Liquid XML Objects   19.0.10
Generates code from XML Schemas for C# and Visual Basic

Microsoft Azure HDInsight Azure Node   2.6.3000.0
HDInsight Node under Azure Node

Microsoft Azure Hive Query Language Service   2.6.3000.0
Language service for Hive query

Microsoft Azure Service Fabric Tools for Visual Studio   16.10
Microsoft Azure Service Fabric Tools for Visual Studio

Microsoft Azure Stream Analytics Language Service   2.6.3000.0
Language service for Azure Stream Analytics

Microsoft Azure Stream Analytics Node   1.0
Azure Stream Analytics Node under Azure Node

Microsoft Azure Tools   2.9
Microsoft Azure Tools for Microsoft Visual Studio 2019 - v2.9.40423.1

Microsoft Continuous Delivery Tools for Visual Studio   0.4
Simplifying the configuration of Azure DevOps pipelines from within the Visual Studio IDE.

Microsoft JVM Debugger   1.0
Provides support for connecting the Visual Studio debugger to JDWP compatible Java Virtual Machines

Microsoft Library Manager   2.1.113+g422d40002e.RR
Install client-side libraries easily to any web project

Microsoft MI-Based Debugger   1.0
Provides support for connecting Visual Studio to MI compatible debuggers

Microsoft Visual C++ Wizards   1.0
Microsoft Visual C++ Wizards

Microsoft Visual Studio Tools for Containers   1.2
Develop, run, validate your ASP.NET Core applications in the target environment. F5 your application directly into a container with debugging, or CTRL + F5 to edit & refresh your app without having to rebuild the container.

Microsoft Visual Studio VC Package   1.0
Microsoft Visual Studio VC Package

Mono Debugging for Visual Studio   16.10.15 (552afdf)
Support for debugging Mono processes with Visual Studio.

Multilingual App Toolkit   4.0
Multilingual App Toolkit helps you localize your Windows Store app by providing file management, pseudo and machine translation, translation editor, and build integration. http://aka.ms/matinstall

Node.js Tools   1.5.30310.1 Commit Hash:e550dd1b679956810ccace4368dbc7a64443213b
Adds support for developing and debugging Node.js apps in Visual Studio

NuGet Package Manager   5.10.0
NuGet Package Manager in Visual Studio. For more information about NuGet, visit https://docs.nuget.org/

ProjectServicesPackage Extension   1.0
ProjectServicesPackage Visual Studio Extension Detailed Info

Razor (ASP.NET Core)   16.1.0.2122504+13c05c96ea6bdbe550bd88b0bf6cdddf8cde1725
Provides languages services for ASP.NET Core Razor.

SQL Server Data Tools   16.0.62105.04180
Microsoft SQL Server Data Tools

Syntax Visualizer   1.0
An extension for visualizing Roslyn SyntaxTrees.

Test Adapter for Boost.Test   1.0
Enables Visual Studio's testing tools with unit tests written for Boost.Test.  The use terms and Third Party Notices are available in the extension installation directory.

Test Adapter for Google Test   1.0
Enables Visual Studio's testing tools with unit tests written for Google Test.  The use terms and Third Party Notices are available in the extension installation directory.

ToolWindowHostedEditor   1.0
Hosting json editor into a tool window

TypeScript Tools   16.0.30429.2002
TypeScript Tools for Microsoft Visual Studio

Visual Basic Tools   3.10.0-4.21269.26+029847714208ebe49668667c60ea5b0a294e0fcb
Visual Basic components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Visual F# Tools   16.10.0-beta.21262.7+1b23bbeda88ea3cb9be9af777f4c99fa8663df81
Microsoft Visual F# Tools

Visual Studio Code Debug Adapter Host Package   1.0
Interop layer for hosting Visual Studio Code debug adapters in Visual Studio

Visual Studio Container Tools Extensions   1.0
View, manage, and diagnose containers within Visual Studio.

Visual Studio Tools for CMake   1.0
Visual Studio Tools for CMake

Visual Studio Tools for Containers   1.0
Visual Studio Tools for Containers

Visual Studio Tools for Kubernetes   1.0
Visual Studio Tools for Kubernetes

VisualStudio.DeviceLog   1.0
Information about my package

VisualStudio.Mac   1.0
Mac Extension for Visual Studio

Xamarin   16.10.000.231 (d16-10@8111164)
Visual Studio extension to enable development for Xamarin.iOS and Xamarin.Android.

Xamarin Designer   16.10.0.115 (remotes/origin/c750fbf1bde3c720d077f51640fe197c6dac7cbe@c750fbf1b)
Visual Studio extension to enable Xamarin Designer tools in Visual Studio.

Xamarin Templates   16.10.5 (355b57a)
Templates for building iOS, Android, and Windows apps with Xamarin and Xamarin.Forms.

Xamarin.Android SDK   11.3.0.4 (d16-10/ae14caf)
Xamarin.Android Reference Assemblies and MSBuild support.
    Mono: b4a3858
    Java.Interop: xamarin/java.interop/d16-10@f39db25
    ProGuard: Guardsquare/proguard/v7.0.1@912d149
    SQLite: xamarin/sqlite/3.35.4@85460d3
    Xamarin.Android Tools: xamarin/xamarin-android-tools/d16-10@c5732a0

Xamarin.iOS and Xamarin.Mac SDK   14.20.0.10 (0e35b2e0b)
Xamarin.iOS and Xamarin.Mac Reference Assemblies and MSBuild support.

DominicFrei:
## Description

This moves the `DeleteRealm` function into our Core since it's a shared functionality.
The [Core PR](https://github.com/realm/realm-core/pull/4701) enables all SDKs to use the same `DeleteRealm` instead of having duplicated and/or differing implementations of it across the SDKs.

Fixes #386 
Fixes #2455 

##  TODO

* [X] Changelog entry
* [X] Tests (if applicable)

Both not really necessary because the point of moving this does include not changing anything on the user API level.
The expectation would be to have all tests still pass even after deleting the function and just calling it's Core equivalent.


Realm/Realm/Realm.cs
```diff
@@ -171,30 +171,23 @@ public static bool Compact(RealmConfigurationBase config = null)
...
+        /// </remarks>
         /// <param name="configuration">A <see cref="RealmConfigurationBase"/> which supplies the realm path.</param>
+        /// <exception cref="RealmPermissionDeniedException">Thrown if the Realm is still open.</exception>
         public static void DeleteRealm(RealmConfigurationBase configuration)
         {
             Argument.NotNull(configuration, nameof(configuration));
 
-            var fullpath = configuration.DatabasePath;
-            if (IsRealmOpen(fullpath))
+            if (IsRealmOpen(configuration.DatabasePath))
```
nirinchev, line 185: You can remove this call now that Core takes care of performing the check.

DominicFrei: I wasn't sure if we wanted to keep throwing our custom exception here.
Fixed with 84be8d0a1769fed6159dd3cbdcba0ae0dc833d9f.
The test checks for a generic `RealmException` now. Would we want to test the specific exception thrown by core? I guess not.

DominicFrei:
Update on the status here: Core PR is merged, waiting for the `v11` merge now.

Tests/Realm.Tests/Database/InstanceTests.cs
```diff
@@ -141,7 +141,7 @@ public void DeleteRealmFailsIfOpenSameThread()
             using var openRealm = GetRealm();
 
             // Assert
-            Assert.Throws<RealmPermissionDeniedException>(() => Realm.DeleteRealm(RealmConfiguration.DefaultConfiguration));
+            Assert.Throws<RealmException>(() => Realm.DeleteRealm(RealmConfiguration.DefaultConfiguration));
```
nirinchev, line 144: Is Core returning a non-generic exception that we can then wrap in `RealmPermissionDeniedException` or even better - something more concrete, such as `RealmInUseException`? You can see how exceptions are getting converted from Core to a .NET-compatible struct [here](https://github.com/realm/realm-dotnet/blob/fd1daa05d1bbc7e3169663cf9b541d16c6d8e542/wrappers/src/error_handling.cpp#L41) and then from an exception code to an actual managed exception [here](https://github.com/realm/realm-dotnet/blob/fd1daa05d1bbc7e3169663cf9b541d16c6d8e542/Realm/Realm/Exceptions/RealmException.cs#L48).

nirinchev:
One thing I noticed as I was working on https://github.com/realm/realm-dotnet/pull/2423 is that `delete_files` in Core will just go ahead and try to delete the files without first checking if the Realm exists:

https://github.com/realm/realm-core/blob/8765b60e5b4db7bcb4a14d4c22854f12e592b4ec/src/realm/object-store/shared_realm.cpp#L915-L916

It might be a good idea to do the check and perhaps do nothing if the Realm doesn't exist.

Tests/Realm.Tests/Database/InstanceTests.cs
DominicFrei: Added with https://github.com/realm/realm-dotnet/pull/2401/commits/8b93783fba2a17735753fb541b6ffb26404f7ef9.

DominicFrei:
> > One thing I noticed as I was working on #2423 is that `delete_files` in Core will just go ahead and try to delete the files without first checking if the Realm exists:
> > https://github.com/realm/realm-core/blob/8765b60e5b4db7bcb4a14d4c22854f12e592b4ec/src/realm/object-store/shared_realm.cpp#L915-L916
> > It might be a good idea to do the check and perhaps do nothing if the Realm doesn't exist.
> 
> This is what happens at the moment. `call_with_lock` will successfully lock the `lockfile` (because opening seems to also create it if it does not exist) and then tries to delete all files which succeeds if they don't exist.
> 
> Delete on an empty folder will therefore return `true` as well == succeed without doing anything.
> 
> Would that suffice your intention? @nirinchev

@nirinchev Added with https://github.com/realm/realm-dotnet/pull/2401/commits/3dc9251632fe53719115f58fb73c37c425cfd57a. It does not fail locally for me though. Let's see what CI thinks about it. 💪

Realm/Realm/Exceptions/RealmInUseException.cs
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Exceptions
+{
+    /// <summary>
+    /// An exception thrown when an an operation is being performed on an open Realm
+    /// that can only performed on closed Realm (i.e. Realm files).
```
nirinchev, line 23: Not sure I understand the `i.e. Realm files` clarification there - did you mean something like "delete realm files"?

DominicFrei: I think so. I had something in mind but the result was not clear enough. Corrected with https://github.com/realm/realm-dotnet/pull/2401/commits/321d250e5c6935cf2c6388bc0c7314d4842816ee.

DominicFrei:
Summary of the Android crash analysis so far:
https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-dotnet/detail/PR-2401/14/pipeline/445 shows that the process crashes during the Xamarin Android tests (the last successful test being `SynchronizedInstanceTests.GetInstance_WhenDynamic_ReadsSchemaFromDisk` right before the crash).

The `android-logcat.zip` (https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-dotnet/detail/PR-2401/14/artifacts) mentioned in the crash message shows:
```
06-10 17:36:06.097 E/libc++abi(13473): terminating with uncaught exception of type
 realm::util::File::NotFound: open("/data/data/io.realm.xamarintests/cache/rt-$13473/mongodb-realm/
 myapp-123/8e98563c-c188-43d2-9812-49bcd53b28dc/
 s_fe4873eb-028d-4fbf-bcbe-2ec70155ff4a.realm.management/access_control.write.mx") 
failed: No such file or directory 
Path: /data/data/io.realm.xamarintests/cache/rt-$13473/mongodb-realm/
 myapp-123/8e98563c-c188-43d2-9812-49bcd53b28dc/
 s_fe4873eb-028d-4fbf-bcbe-2ec70155ff4a.realm.management/access_control.write.mx
```
`access_control.write.mx` seems to be the write access mutex created in https://github.com/realm/realm-core/blob/master/src/realm/db.cpp#L951 (also see https://github.com/realm/realm-core/blob/master/src/realm/db.cpp#L770).

It is then used in https://github.com/realm/realm-core/blob/master/src/realm/db.cpp#L1774 and https://github.com/realm/realm-core/blob/master/src/realm/db.cpp#L2046.

It's unclear what exactly causes the crash. It does seem to only happen in this PR which is based onhttps://github.com/realm/realm-core/pull/4701.

DominicFrei:
@nirinchev Jenkins is finally happy. 👍 Ready for review again.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -320,6 +320,30 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+        [Test]
+        public void DeleteRealmWorksIfFolderDoesntExist()
+        {
+            var config = GetFakeConfig();
+            var dbFolder = Path.GetDirectoryName(config.DatabasePath);
+            var nonExistingRealm = Path.Combine(dbFolder, "idontexist", "my.realm");
+            var newConfig = new RealmConfiguration(nonExistingRealm);
+
+            Assert.DoesNotThrow(() => Realm.DeleteRealm(newConfig));
+        }
```
nirinchev, line 345: This test doesn't need to be in SynchronizedInstanceTests - it can be rewritten using the local-only API, I just put it in here accidentally 😅

nirinchev (commented):
Looks good, mostly nits from me.

Realm/Realm/Realm.cs
```diff
@@ -169,31 +169,30 @@ public static bool Compact(RealmConfigurationBase config = null)
...
-            var filesToDelete = new[] { string.Empty, ".log_a", ".log_b", ".log", ".lock", ".note" }
-                .Select(ext => fullpath + ext)
-                .Where(File.Exists);
+                // Race condition:
+                // When trying to delete the Realm it can in some occasions (usually when Sync is involved)
+                // still be in use. To make sure other threads that use the same Realm get scheduled again
+                // and can finish their work before we actually delete the Realm files, we have to wait for
+                // a moment here.
+                // This can be removed as soon as https://github.com/realm/realm-core/issues/4762 is resolved.
+                Task.Delay(1).Wait();
```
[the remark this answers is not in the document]
DominicFrei: I had to add this to fix the bug where the lock was still in use. But with the extended wait in the test, it maybe is fixed anyway. I'll push a version without.

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -94,6 +94,11 @@ private static class NativeMethods
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_close_realm", CallingConvention = CallingConvention.Cdecl)]
             public static extern void close_realm(SharedRealmHandle sharedRealm, out NativeException ex);
 
+            [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_delete_files", CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Unicode)]
+            public static extern void delete_files([MarshalAs(UnmanagedType.LPWStr)] string path,
+                IntPtr path_len,
+                out NativeException ex);
```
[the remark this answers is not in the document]
DominicFrei: Done. It fits in one line anyway now. I had it different before but did not update the format. 👍

Realm/Realm/Realm.cs
```diff
@@ -169,31 +169,30 @@ public static bool Compact(RealmConfigurationBase config = null)
         }
 
         /// <summary>
-        /// Deletes all the files associated with a realm.
+        /// Deletes all files associated with a given Realm if the Realm exists and is not open.
         /// </summary>
+        /// <remarks>
+        /// The Realm file must not be open on other threads.
+        /// This method must not be called inside a transaction.
```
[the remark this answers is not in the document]
DominicFrei: Adjusted.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
DominicFrei: Done.

Realm/Realm/Realm.cs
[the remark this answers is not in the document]
DominicFrei: Just double checked. I guess we confused ourselves.

The one in the test code (the base class tear down) was the one necessary for `GetInstance_WhenDynamicAndDoesntExist_ReturnsEmptySchema`.

This one here is the workaround for `DeleteRealmWorksIfCalledMultipleTimes` (also in `SynchronizedInstanceTests`):
https://github.com/realm/realm-core/issues/4762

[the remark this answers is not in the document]
DominicFrei: @nirinchev I've just re-added it. When you compare the two last commits it seems to make the difference, unfortunately. But since it's only 5ms and only on delete it seems to be ok as a workaround until the Core bug is fixed.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -320,6 +320,28 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+            var openRealm = GetRealm(config);
+            openRealm.Dispose();
+            Assert.That(File.Exists(config.DatabasePath));
+
+            // Race condition:
+            // When trying to delete the Realm it can in some occasions (usually when Sync is involved)
+            // still be in use. To make sure other threads that use the same Realm get scheduled again
+            // and can finish their work before we actually delete the Realm files, we have to wait for
+            // a moment here.
+            // This can be removed as soon as https://github.com/realm/realm-core/issues/4762 is resolved.
```
nirinchev, line 336:
```suggestion
            // TODO: remove this when https://github.com/realm/realm-core/issues/4762 is resolved.
```

Adding a TODO will make it easier to come back to it later.

Tests/Realm.Tests/Sync/SyncTestBase.cs
```diff
@@ -57,24 +57,26 @@ protected App CreateApp(AppConfiguration config = null)
...
-                session?.CloseHandle();
-            }
+            _sessions.DrainQueue(session => session?.CloseHandle());
+
+            // Race condition:
+            // When trying to delete the Realm it can in some occasions (usually when Sync is involved)
+            // still be in use. To make sure other threads that use the same Realm get scheduled again
+            // and can finish their work before we actually delete the Realm files, we have to wait for
+            // a moment here.
+            // This can be removed as soon as https://github.com/realm/realm-core/issues/4762 is resolved.
```
nirinchev, line 67:
```suggestion
            // TODO: remove this when https://github.com/realm/realm-core/issues/4762 is resolved.
```

Realm/Realm/Realm.cs
```diff
@@ -169,31 +169,21 @@ public static bool Compact(RealmConfigurationBase config = null)
...
-                .Select(ext => fullpath + ext)
-                .Where(File.Exists);
-
-            foreach (var file in filesToDelete)
-            {
-                File.Delete(file);
-            }
-
-            if (Directory.Exists($"{fullpath}.management"))
+            if (File.Exists(configuration.DatabasePath))
```
nirinchev, line 184: Reminder to move this to OS

nirinchev (approved):
Once the File.Exists check is moved to OS, this should be good to go.

Realm/Realm/Realm.cs
DominicFrei: https://github.com/realm/realm-core/pull/4765

DominicFrei: PR is done, waiting for merge into `v11`.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -320,6 +320,28 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+            openRealm.Dispose();
+            Assert.That(File.Exists(config.DatabasePath));
+
+            // Race condition:
+            // When trying to delete the Realm it can in some occasions (usually when Sync is involved)
+            // still be in use. To make sure other threads that use the same Realm get scheduled again
+            // and can finish their work before we actually delete the Realm files, we have to wait for
+            // a moment here.
+            // TODO: remove this when https://github.com/realm/realm-core/issues/4762 is resolved.
+            Task.Delay(5).Wait();
```
nirinchev, line 337: Can we remove this now?

DominicFrei: I've added the test from https://github.com/realm/realm-dotnet/issues/2455#issuecomment-867641086 into this branch as well (82bae01) which fails. Both tests are similar but the the same.

Testing against `fsa/fix-delete-open-race-v11` (which is https://github.com/realm/realm-core/pull/4768) fixed the problem where we had to add a delay in the `SyncBase` (removed with https://github.com/realm/realm-dotnet/pull/2401/commits/c8b17a4f1182cc308c6ca2b9f3a98c9c7cacbb49) class but it does not fix this specific test.

Only adding the delay fixed it for me. The same with this test case that was already in there. I'm pushing both to CI to verify but it seems like https://github.com/realm/realm-core/pull/4768 does not solve all of those problems yet, as mentioned before. Which makes sense I guess, because https://github.com/realm/realm-core/issues/4762 was the race between deleting the Realm vs. open (which was not related to Sync but a general problem) and on the other side, the test in here is about the Sync part. I guess the misleading part is that I link the same Core issue.

DominicFrei: After further investigation:
- https://github.com/realm/realm-dotnet/blob/df/use-delete-realm-from-core/Tests/Realm.Tests/Database/InstanceTests.cs#L80 succeeds
- https://github.com/realm/realm-dotnet/blob/df/use-delete-realm-from-core/Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs#L324 fails

They are the same test with one difference:
`var config = RealmConfiguration.DefaultConfiguration;`
vs
`var config = GetFakeConfig();`

Maybe it's actually not Core but our test setup. I need to look into this further.

DominicFrei: Interestingly I do see the following result:
image
100% when run solo, 95% when run together (both tests or the whole class). Sometimes both fail.

Adding to the comment prior to this:
`DeleteRealmWorksIfCalledMultipleTimes` in `SynchronizedInstanceTests` actually always succeeds when run solo.

DominicFrei: The parameter is only influencing the call to `AddDummyData` btw.

Now it gets interesting:
When I remove that call completely, i.e. no data in the Realm, both tests always fail.

nirinchev: How do the tests fail?

DominicFrei: The only result I've seen is:
```
Realms.Exceptions.RealmInUseException : Cannot delete files of an open Realm. xxxxx.lock is still in use.
```

nirinchev: Ah, okay, I see - so we should just retry - there's no way to check what the status of the sync thread is, so we should just retry the deletion 10-20 times and assert that it eventually succeeds.

DominicFrei: Alright. Is there any value in keeping them then though? Every other test does that anyway in tear down. I would basically just copy paste tear down (and then call tear down on top of that anyway).

nirinchev: The tests should not depend on code in teardown to actually test the behavior as refactoring that may result in accidentally removing test coverage. So yes, both tests provide value, we should just explain why we're retrying the deletion in the synchronized case.

nirinchev (approved):

PR #2405, opened 2021-05-31
Run iOS tests on CI

### LINKED ISSUE #2286: Run tests for Xamarin.iOS on CI

LaPeste:
## Description
To achieve this a js GH action to run the simulator was created.
To note that GH action should not be here, but rather on the CI repo. This will be done when everything is reviewed and approved, simply because referencing an action from another repo needs a "release" of the action.

Fixes #2286


.github/actions/run-ios-simulator/src/main.ts
```diff
@@ -0,0 +1,87 @@
...
+            const { stdout, stderr } = await childProcess.exec("xcrun simctl list runtimes |  awk '/com.apple.CoreSimulator.SimRuntime.iOS/ { match($0, /com.apple.CoreSimulator.SimRuntime.iOS-[0-9.-]+/); print substr($0, RSTART, RLENGTH); exit }'");
+            runtimeId = stdout?.toString() ?? "";
+            if (stderr) {
+                core.setFailed(stderr.toString());
+            }
+            if (await exec.exec("xcrun", ["simctl", "create", id, "com.apple.CoreSimulator.SimDeviceType." + iphoneToSimulate, runtimeId.toString()]) != 0) core.setFailed(`create simulator failed`);
+        }
+        if (await exec.exec("xcrun", ["simctl", "boot", id]) != 0) core.setFailed(`boot simulator failed`);
+        if (await exec.exec("xcrun", ["simctl", "install", id, appPath]) != 0) core.setFailed(`install app on  simulator failed`);
+        if (await exec.exec("xcrun", ["simctl", "launch", "--console-pty", id, bundleId].concat(args.split(/(?= --)/g))) != 0) core.setFailed(`launch app on simulator failed`);
```
nirinchev, line 66:
```suggestion
        if (await exec.exec(`xcrun simctl launch --console-pty ${id} ${bundleId} ${args}`)) != 0) core.setFailed(`launch app on simulator failed`);
```

.github/actions/run-ios-simulator/src/main.ts
```diff
@@ -0,0 +1,74 @@
...
+            core.info(`runtimeId: ${runtimeId}`);
+        }
+
+        try {
+            await execCmd(`xcrun simctl create ${id} com.apple.CoreSimulator.SimDeviceType.${iphoneToSimulate} ${runtimeId.toString()}`);
+        }
+        catch {
+            // Different combinantions of xcode and macOS versions have shown different syntax acceptance about the runtime, therefore 1 last attempt with a different syntax.
+            // additionally, this uses  childProcess.exec instead of @actions/exec because the latter doesn't properly parse the pipe (|) char
+            const { stdout, stderr } = await childProcess.exec("xcrun simctl list runtimes |  awk '/com.apple.CoreSimulator.SimRuntime.iOS/ { match($0, /com.apple.CoreSimulator.SimRuntime.iOS-[0-9.-]+/); print substr($0, RSTART, RLENGTH); exit }'");
```
nirinchev, line 40: Should this use `execCmd`?

.github/actions/run-ios-simulator/src/main.ts
```diff
@@ -0,0 +1,74 @@
...
+        try {
+            await execCmd(`xcrun simctl shutdown ${id}`);
+            await execCmd(`xcrun simctl delete ${id}`);
+        } catch (error) {
+            core.setFailed(`An error occurred during cleanup: ${error.toString()}`);
+        }
+    }
+}
+
+async function execCmd(cmd: string, options: exec.ExecOptions = {}): Promise<void> {
```
nirinchev, line 64: Can we modify this to return the output from `stdout`?

.github/actions/run-ios-simulator/src/main.ts
nirinchev: Also, we already ran `xcrun simctl list runtimes` - rather than running it a second time and matching with `awk`, we can write a regex that parses both formats for the runtime and try to create a simulator with either one.

.github/actions/run-ios-simulator/src/main.ts
```diff
@@ -0,0 +1,74 @@
...
+            }
+
+            await execCmd(`xcrun simctl create ${id} com.apple.CoreSimulator.SimDeviceType.${iphoneToSimulate} ${runtimeId.toString()}`);
+        }
+
+        await execCmd(`xcrun simctl boot ${id}`);
+        await execCmd(`xcrun simctl install ${id} ${appPath}`);
+        await execCmd(`xcrun simctl launch --console-pty ${id} ${bundleId} ${args}`);
+    } catch (error) {
+        core.setFailed(error.message);
```
nirinchev, line 53: We should enhance error messages - e.g. `An unexpected error occurred: ${error.message} - ${error.stack}`.

.github/actions/run-ios-simulator/src/main.ts
```diff
@@ -0,0 +1,74 @@
...
+    } finally {
+        try {
+            await execCmd(`xcrun simctl shutdown ${id}`);
+            await execCmd(`xcrun simctl delete ${id}`);
+        } catch (error) {
+            core.setFailed(`An error occurred during cleanup: ${error.toString()}`);
```
nirinchev, line 59: I believe `toString()` is automatically called when using string interpolation, but here we probably want `error.message` and `error.stack`.

.github/actions/run-ios-simulator/src/main.ts
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+}
+
+async function execCmd(cmd: string, options: exec.ExecOptions = {}): Promise<void> {
+    const exitCode = await exec.exec(cmd, [], options);
+    if (exitCode != 0) {
+        const msgCmd = cmd.split(" ").slice(0, 3).join(" ");
```
nirinchev, line 67: Why are we removing some of the arguments here - while they'll probably be visible in the logs, it's still useful to have the complete command when an error occurs.

nirinchev (commented):
Needs a rebase on master to get the latest (some renaming around `publishTestsResults`) and I have some suggestions about making `execCmd` more streamlined.

.github/actions/run-ios-simulator/src/main.ts
LaPeste: It doesn't use execCmd because as explained in the comment, the pipe `|` isn't properly parsed by what's at the core of execCmd. The fact that I didn't rework the awk command is simply because I wasn't sure about the usefulness of this "second try" and I wanted to see the opinion of the reviewer. If you think it makes sense to retry with a different runtime syntax, I'll re-write the line using just a regex.

.github/actions/run-ios-simulator/src/main.ts
LaPeste: So would you like to see a tuple like `result: [number, string]` that returns the exit code and the stdout, or do you suggest just the _stdout_?

.github/actions/run-ios-simulator/src/main.ts
LaPeste: Exactly because they are be visible in the logs. Every command executed, regardless of the result, is fully shown in the log. I thought to keep the err message concise since the full cmd is likely a few lines above. But if you want I'll fully print it out.

.github/actions/run-ios-simulator/src/main.ts
LaPeste: You are correct. Somehow I remember to see that the ts transpiler complained. I went to double check and I'm wrong. I'll fix this.

.github/actions/run-ios-simulator/src/main.ts
nirinchev: If I understand the `awk` syntax correctly, it's just a regex match against the runtime string. We can combine both matches (this one and the one in line 28) into a single check:

```js
const match =/(?<runtime1>iOS \d{1,2}(.\d{1,2})?).*(?<runtime2>com\.apple\.CoreSimulator\.SimRuntime\.iOS-[0-9.-]+)/g.exec(runtimeId);

try {
   await execCmd(`xcrun simctl create ${id} com.apple.CoreSimulator.SimDeviceType.${iphoneToSimulate} ${match.groups.runtime1}`);
}
catch {
   await execCmd(`xcrun simctl create ${id} com.apple.CoreSimulator.SimDeviceType.${iphoneToSimulate} ${match.groups.runtime2}`);
}
```

.github/actions/run-ios-simulator/src/main.ts
nirinchev: I think it should internally capture both stdout and stderr and return just stdout. If the exit code is non-zero, it should throw an exception that includes the output of stderr.

.github/actions/run-ios-simulator/src/main.ts
nirinchev: I think it won't become overly verbose - let's print the entire thing so that the error is immediately visible in the status rather than having to go through the logs.

.github/actions/run-ios-simulator/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+    "compilerOptions": {
+      "target": "ES2018",
```
LaPeste, line 3: This was changed to accomodate the interface of `RegExpExecArray` that contains `groups`

.github/actions/run-ios-simulator/src/main.ts
```diff
@@ -12,61 +12,62 @@ async function run(): Promise<void> {
...
         // Sample output: iOS 14.5 (14.5 - 18E182) - com.apple.CoreSimulator.SimRuntime.iOS-14-5
-        // and we want to extract "iOS 14.5"
+        // and we want to extract "iOS 14.5" and "com.apple.CoreSimulator.SimRuntime.iOS-14-5"
         // If we want to allow launching watchOS/tvOS simulators, replace the 'iOS' with an 'os' argument
-        const matches = runtimeId.match(/(iOS \d{1,2}(.\d{1,2})?)/g);
-        if (matches && matches.length > 0) {
-            runtimeId = matches[0].replace(" ", "");
-            core.info(`runtimeId: ${runtimeId}`);
+        const matches =/(?<runtime1>iOS \d{1,2}(.\d{1,2})?).*(?<runtime2>com\.apple\.CoreSimulator\.SimRuntime\.iOS-[0-9.-]+)/g.exec(runtimeId);
+        if (!matches || matches?.length == 0 || matches?.groups || matches?.groups?.runtime1.length == 0 || matches?.groups?.runtime2.length == 0) {
```
nirinchev, line 20: You don't need the question marks if you already checked that the value is not falsy. Additionally, you're entering the if clause if `matches.groups` is truthy, meaning that if you did get some groups  to match, you'll set the result the failed. Finally, `matches.groups.runtime1` being truthy is stronger than `matches.length` being non-zero. So the condition can probably be simplified to something like:

```suggestion
        if (!matches?.groups?.runtime1 || !matches?.groups?.runtime2) {
```

.github/actions/run-ios-simulator/src/main.ts
```diff
@@ -12,61 +12,62 @@ async function run(): Promise<void> {
...
+            core.setFailed(`Impossible to fetch a runtime. Check runtimes and retry.\n${runtimeId}`);
+            return;
         }
+        core.info(`runtimeId: ${runtimeId}`);
 
         try {
-            await execCmd(`xcrun simctl create ${id} com.apple.CoreSimulator.SimDeviceType.${iphoneToSimulate} ${runtimeId.toString()}`);
+            runtimeId = matches?.groups?.runtime1?.replace(" ", "") ?? "";
```
nirinchev, line 27: Those question marks are not necessary since we already checked that the value of `runtime1` is truthy.

LaPeste: Unfortunately, it was a result of the mistake done in the nullity check. Rettified now.

.github/actions/run-ios-simulator/src/main.ts
```diff
@@ -0,0 +1,74 @@
...
+        await execCmd(`xcrun simctl launch --console-pty ${id} ${bundleId} ${args}`);
+    } catch (error) {
+        core.setFailed(`An unexpected error occurred: ${error.message} -\n${error.stack}`);
+    } finally {
+        try {
+            await execCmd(`xcrun simctl shutdown ${id}`);
+            await execCmd(`xcrun simctl delete ${id}`);
+        } catch (error) {
+            core.setFailed(`An error occurred during cleanup: ${error.message} -\n${error.stack}`);
+        }
```
nirinchev, line 47: Since GHA runners are containerized, we can probably not execute the cleanup step.

nirinchev (approved):
Looks good - only questions/nits from me.

.github/actions/run-ios-simulator/src/main.ts
```diff
@@ -0,0 +1,74 @@
+import * as core from "@actions/core";
+import * as exec from "@actions/exec";
+import { v4 as uuidv4 } from "uuid";
+
+async function run(): Promise<void> {
+    const id = uuidv4().split("-").join("");
```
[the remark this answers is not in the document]
LaPeste: There is no need, I just kept it consistent with what was done in Jenkins. I saw no drawback, neither performance, nor readability, in keeping as it was. If you would like that gone, there's no problem with that.

.github/actions/run-ios-simulator/src/main.ts
```diff
@@ -0,0 +1,74 @@
...
+            // Different combinantions of xcode and macOS versions have shown different syntax acceptance about the runtime, therefore 1 last attempt with a different syntax.
+            runtimeId = matches.groups.runtime2;
+            await execCmd(`xcrun simctl create ${id} com.apple.CoreSimulator.SimDeviceType.${iphoneToSimulate} ${runtimeId}`);
+        }
+
+        await execCmd(`xcrun simctl boot ${id}`);
+        await execCmd(`xcrun simctl install ${id} ${appPath}`);
+        await execCmd(`xcrun simctl launch --console-pty ${id} ${bundleId} ${args}`);
+    } catch (error) {
+        core.setFailed(`An unexpected error occurred: ${error.message} -\n${error.stack}`);
```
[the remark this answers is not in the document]
LaPeste: I thought it'd give a better reading experience. I have not look at a stacktrace in this instance, but I expect it to look as usual, a call per line. Because of this, I thought to separate with a new line to clearly show the stack trace, instead of having the first call in the same line with the error message.
If you think I'm wrong, I'll remove the new line.

.github/actions/run-ios-simulator/src/main.ts
LaPeste: Fair point. I mostly did it because it felt good to clean up, but we can save those few seconds it take to clean up. I'll remove it.

.github/actions/run-ios-simulator/src/main.ts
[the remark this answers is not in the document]
nirinchev: No, the newline seems fine, I just thought it'd be weird to have a dash after the message. It's the nittiest of nits, so don't care too much about it, but I'd probably have written it as `An unexpected error occurred: ${error.message}\n${error.stack}`

.github/actions/run-ios-simulator/src/main.ts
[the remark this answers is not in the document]
nirinchev: Yeah, no idea why it was done like that in jenkins which is embarrassing considering I wrote it 😅 anyway, feel free to remove if it serves no purpose.

.github/actions/run-ios-simulator/src/main.ts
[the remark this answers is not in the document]
LaPeste: It felt like a sort of full stop. I won't hurt removing it.

nirinchev (approved):

PR #2421, opened 2021-06-07
Add Unity build to the GHA pipeline

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2285

* Replaces the code to unstash wrapper binaries with a loop
* Adds steps to build the unity package both with and without deps

### DIFF
--- a/.github/templates/main.yml
+++ b/.github/templates/main.yml
@@ -79,6 +79,20 @@ steps:
 #@ end
 #@ end
 
+#@ def buildUnityPackage(includeDependencies):
+#@ finalPkgName = "io.realm.unity-" + ("bundled-" if includeDependencies else "") + "${{ steps.find-nupkg-version.outputs.package_version }}.tgz"
+  - name: #@ "Build Unity" + (" Bundled Dependencies" if includeDependencies else "")
+    run: #@ "dotnet run --project Tools/SetupUnityPackage/ -- realm --packages-path Realm/packages --pack" + (" --include-dependencies" if includeDependencies else "")
+  - name: Store Unity artifacts
+    uses: actions/upload-artifact@v2
+    with:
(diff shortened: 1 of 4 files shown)


.github/templates/main.yml
```diff
@@ -79,6 +79,20 @@ steps:
...
+  - name: #@ "Build Unity" + (" Bundled Dependencies" if includeDependencies else "")
+    run: #@ "dotnet run --project Tools/SetupUnityPackage/ -- realm --packages-path Realm/packages --pack" + (" --include-dependencies" if includeDependencies else "")
+  - name: Store Unity artifacts
+    uses: actions/upload-artifact@v2
+    with:
+      name: #@ finalPkgName
+      path: #@ "${{ github.workspace }}/Realm/Realm.Unity/" + finalPkgName
+      retention-days: ${{ github.event_name != 'pull_request' && 30 || 1 }}
+  - name: Cleanup build folder
+    run: #@ "rm ${{ github.workspace }}/Realm/Realm.Unity/" + finalPkgName
```
LaPeste, line 93: Why do you need to clean the build folder?
I understand that you call this function twice (line 267 and 268) but the final package of each call has a different name. What am I missing?

Jenkinsfile
```diff
@@ -218,6 +218,7 @@ stage('Unity Package') {
   }
 }
 
+
```
LaPeste, line 221: I believe this new line can go.

LaPeste (approved):
Good clean up thanks to the loop to unstash the binaries of the wrappers.

Jenkinsfile
nirinchev: Ah yeah, I wanted to trigger a build to test something 😅

PR #2423, opened 2021-06-07
Make sure we don't accumulate test resources endlessly

nirinchev:
This converts all cleanup lists in tests into queues and we make sure to drain the queues - previously we'd just push stuff to the lists and never empty them.

It resolves the slowing down of test runs with the new delete functionality.

### DIFF
--- a/Tests/Realm.Tests/RealmTest.cs
+++ b/Tests/Realm.Tests/RealmTest.cs
@@ -29,7 +29,7 @@ namespace Realms.Tests
     [Preserve(AllMembers = true)]
     public abstract class RealmTest
     {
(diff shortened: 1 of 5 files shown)


DominicFrei:
@nirinchev Works like a charm, tests are running nicely one my side except for that one failing test. I see that locally too, haven't figured out the reason yet though.

nirinchev:
The failing test is due to a change in Core regarding type coercion of numeric values in a set. Fixed it in https://github.com/realm/realm-dotnet/pull/2425 which is unfortunately blocked on https://github.com/realm/realm-core/pull/4716. Disregard the failure for now.

Issue #2345, opened 2021-04-21
Verify that native callbacks on App/User/?? correctly pass strings


nirinchev:
When native code invokes a managed callback with a string argument, we should read it immediately and convert it to a C# string before invoking any task completion source continuations. Example: https://github.com/realm/realm-dotnet/pull/2340/files#diff-1d9f42def842866bcab25f7a4990ca48dd5e6ffd74a654900e306511ce4537dd.

PR #2435, opened 2021-06-09
Verify that callbacks don't leak strings

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2345


papafe (approved):
Looks fine to me! 
I suppose all those unsafe context were not actually necessary.
Also it would be good to discuss a little bit what we needed this change.

nirinchev:
The unsafes were needed when they operated on byte* pointers to read the strings which is now encapsulated in PrimitiveValue.

We don't need this change per se, but I wanted to clean up places where we're passing byte* pointers and replace those with PrimitiveValue as it's less error prone to pass around.

Issue #2310, opened 2021-03-25
Misleading error message in Unity when failing to add default constructor


DominicFrei:
Hey guys,

I've been talking to @nirinchev about a weaving problem I had with our Unity SDK and we've tried to analyse it.
I've eventually found the problem (which was as usual sitting in front of the PC) but still wanted to open an issue for this to improve the UX a bit.

## Goals

As a developer I'd like to get notified of schema errors and keep this information until it is fixed.

## Expected Results

It would be nice to see the below described error after starting the game.

## Actual Results

I keep getting the following error (and only this one):
```
RealmException: Fody not properly installed. TestObject is a RealmObjectBase but has not been woven.
```

## Steps to Reproduce
(1) Create a new project.
(2) Add Realm as a tarball.
(3) Add a script ('RealmTest') to the 'Main Camera' game obejct.
(4) Replace script content with:
```
using UnityEngine;
using Realms;

class TestObject : RealmObject
{
    [PrimaryKey]
    int PrivateKey { get; set; }
    int Value;

    //TestObject() { }

    TestObject(int value)
    {
        Value = value;
    }
}

public class RealmTest : MonoBehaviour
{
    void Start()
    {
        Debug.Log(RealmConfiguration.GetPathToRealm());
        Realm.GetInstance();
    }
}
```

(5) Go back to Unity and check the console. The following errors shows up. This is the actual problem that would be fixed be uncommenting line 10.
```
E:\Repositories\Playground\PublicInitCrash\Assets\RealmTest.cs(12, 5): Class TestObject must have a public constructor that takes no parameters.
UnityEngine.Debug:LogError (object)
RealmWeaver.UnityWeaver/UnityLogger:Error (string,Mono.Cecil.Cil.SequencePoint)
RealmWeaver.Weaver:WeaveType (Mono.Cecil.TypeDefinition)
RealmWeaver.Weaver:<Execute>b__38_2 (Mono.Cecil.TypeDefinition)
System.Linq.Enumerable:ToArray<RealmWeaver.WeaveTypeResult> (System.Collections.Generic.IEnumerable`1<RealmWeaver.WeaveTypeResult>)
RealmWeaver.Weaver:Execute ()
RealmWeaver.UnityWeaver:WeaveAssemblyCore (string,System.Collections.Generic.IEnumerable`1<string>)
RealmWeaver.UnityWeaver/<>c:<Initialize>b__2_0 (string,UnityEditor.Compilation.CompilerMessage[])
[...]

PR #2446, opened 2021-06-15
Handle weaver errors in Unity 

nirinchev:
## Description

This PR contains two main changes:
* Stops the weaving process on error. This is necessary because Unity will not treat extension-generated error logs as build failures and will happily run the built assembly. This means that we may end up in a situation where classes are partially woven and things fail silently in runtime.
* Modifies the error message for Unity when a RealmConfiguration is created without schema. This is likely due to a build error and we advise users to re-run the weaver and look for error messages in the logs.

Fixes https://github.com/realm/realm-dotnet/issues/2310

##  TODO

* [x] Changelog entry

### DIFF
--- a/Realm/Realm.UnityWeaver/UnityWeaver.cs
+++ b/Realm/Realm.UnityWeaver/UnityWeaver.cs
@@ -216,6 +216,12 @@ namespace RealmWeaver
 
                     var results = weaver.Execute(analyticsConfig);
 
+                    if (results.ErrorMessage != null)
+                    {
+                        UnityLogger.Instance.Error($"[{name}] Weaving failed: {results}");
+                        return false;
+                    }
+
                     // Unity creates an entry in the build console for each item, so let's not pollute it.
                     if (results.SkipReason == null)
                     {
@@ -227,7 +233,7 @@ namespace RealmWeaver
             }
             catch (Exception ex)
             {
-                UnityLogger.Instance.Warning($"[{name}] Weaving failed: {ex}");
+                UnityLogger.Instance.Error($"[{name}] Weaving failed: {ex}");
             }
 
             return false;
--- a/Realm/Realm.UnityWeaver/UnityWeaver.cs
+++ b/Realm/Realm.UnityWeaver/UnityWeaver.cs
@@ -171,7 +171,7 @@ namespace RealmWeaver
             }
             catch (Exception ex)
             {
-                UnityLogger.Instance.Error($"Failed to weave assemblies: {ex}");
+                UnityLogger.Instance.Error($"[Realm] Failed to weave assemblies. If the error persists, please report it to https://github.com/realm/realm-dotnet/issues: {ex}");
             }
             finally
             {
@@ -233,7 +233,7 @@ namespace RealmWeaver
             }
             catch (Exception ex)
             {
-                UnityLogger.Instance.Error($"[{name}] Weaving failed: {ex}");
+                UnityLogger.Instance.Error($"[{name}] Failed to weave assembly. If the error persists, please report it to https://github.com/realm/realm-dotnet/issues: {ex}");
             }
 
             return false;
--- a/Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
+++ b/Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
@@ -34,8 +34,8 @@ internal static class TypeReferenceExtensions
         {
             return @this.GetConstructors()
                 .OrderBy(c => c.Parameters.Count)
-                .Select(c => c.DebugInformation.SequencePoints.FirstOrDefault())
-                .FirstOrDefault(sp => sp != null);
+                .SelectMany(c => c.DebugInformation.SequencePoints)
+                .FirstOrDefault();
         }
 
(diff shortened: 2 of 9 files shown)


Realm/Realm.UnityWeaver/UnityWeaver.cs
```diff
@@ -216,6 +216,12 @@ private static bool WeaveAssemblyCore(string assemblyPath, IEnumerable<string> r
 
                     var results = weaver.Execute(analyticsConfig);
 
+                    if (results.ErrorMessage != null)
```
DominicFrei, line 219: Would it be an option to let the `weaver.Execute` throw to make sure we don't forget that check in other occasions?
The `catch` seems to log the same result then anyway.

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -17,12 +17,35 @@
...
+    public static SequencePoint GetSequencePoint(this TypeDefinition @this)
+    {
+        return GetCtorSequencePoint() ?? GetPropSequencePoint();
+
+        SequencePoint GetCtorSequencePoint()
+        {
+            return @this.GetConstructors()
+                .OrderBy(c => c.Parameters.Count)
+                .Select(c => c.DebugInformation.SequencePoints.FirstOrDefault())
+                .FirstOrDefault(sp => sp != null);
```
DominicFrei, line 38: What's the difference between just calling this `FirstOrDefault` instead of having it twice?

Realm/Realm/Schema/RealmSchema.cs
```diff
@@ -212,12 +210,18 @@ public RealmSchema Build()
             {
                 if (Count == 0)
                 {
-                    throw new InvalidOperationException(
-                        "No RealmObjects. Has linker stripped them? See https://docs.mongodb.com/realm-legacy/docs/dotnet/latest/#linker-stripped-schema");
+                    var message = InteropConfig.Platform switch
+                    {
+                        InteropConfig.UnityPlatform => "Try weaving assemblies again ('Realm' -> 'Weave Assemblies' from the editor or simply make a code change) and make sure you don't have any Realm-related errors in the logs.",
+                        InteropConfig.DotNetPlatform => "Has linker stripped them? See https://docs.mongodb.com/realm-legacy/docs/dotnet/latest/#linker-stripped-schema",
```
DominicFrei, line 216: What is `them` referring to? Is there a message before this that always gets shown?

DominicFrei (approved):
Just a couple questions, not really change requests. 👍

Realm/Realm.UnityWeaver/UnityWeaver.cs
nirinchev: That would be a bad idea for two reasons:
1. Those are different error cases - the one where we have an error message is expected and likely an issue with the user's code. The one being caught is unexpected and likely an issue with our code. We can (and perhaps should) modify the messaging in the `Exception` case to instruct the user to open an issue if they encounter it.
2. The weaver is being used by Fody and there we don't have a `catch` clause - we rely on Fody to handle exceptions in this case and there's a notable difference between returning an error message and throwing an exception.

DominicFrei: Good idea. The modified message would clarify it for the user (and for me 🙃 ).

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
nirinchev: Not sure I understand. If you're talking about the `FirstOrDefault` on line 37, then those are different. On line 37, `FirstOrDefault()` gives us the first element from `DebugInformation.SequencePoints` and puts that into our collection. The `FirstOrDefault` on line 38  gives us the first element from our collection. We could rewrite that as:

```
@this.GetConstructors()
    .OrderBy(c => c.Parameters.Count)
    .SelectMany(c => c.DebugInformation.SequencePoints)
    .FirstOrDefault()
```

DominicFrei: Yes, that's what I meant. With the other example I do understand the difference now. I guess I need to read up on LINQ a bit more so that I get used to it. 👍

Realm/Realm/Schema/RealmSchema.cs
LaPeste, line 216: You may want to link the new documentation about that: [https://docs.mongodb.com/realm/sdk/dotnet/troubleshooting/](https://docs.mongodb.com/realm/sdk/dotnet/troubleshooting/)

LaPeste (approved):

Realm/Realm/Schema/RealmSchema.cs
nirinchev: Yes - on line 220 we throw `No RealmObjects. *this message*`

DominicFrei: Instead of looking into the past I should look into the future. How philosophical. :D 
I was only checking code before that part for some reason. Thanks for the hint! 🚀

PR #2450, opened 2021-06-16
Save xml test results as artifacts

LaPeste:
## Description
Save xml test results as artifacts to inspect their content at a later stage.


nirinchev:
That's no longer needed - it was interesting mostly as a means to extend/improve the unit testing action, but since we're migrating to a different one, we won't need the test result xmls.

PR #2451, opened 2021-06-16
Delete some artifacts from the workflow after it runs

nirinchev:
This deletes the wrappers and the extra packages that only the Unity build uses. Apart from saving space, I'm just annoyed when I have to scroll through a bunch of artifacts to find what I'm looking for 😄

I decided to put it in a separate step even though it can technically be done after packaging because I was thinking we may eventually want to do other things with the artifacts - for example get all the debug symbols and zip them together in a debug package or something similar.

### DIFF
--- a/.github/templates/main.yml
+++ b/.github/templates/main.yml
@@ -6,7 +6,7 @@
 #@ windowsArchs = [ 'Win32', 'x64' ]
 #@ windowsUWPArchs = [ 'Win32', 'x64', 'ARM' ]
 #@ wrappersCacheCondition = "steps.check-cache.outputs.cache-hit != 'true'"
-#@ nugetPackages = [ 'Realm.Fody', 'Realm', 'Realm.UnityUtils', 'Realm.UnityWeaver']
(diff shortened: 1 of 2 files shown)


LaPeste (approved):
I think it's a good idea to remove all that long list of artifacts from the summary. Too bad we can't do some sort of folding, like "wrappers", "nupkg" etc.
The only slightly annoying thing of not having the wrappers artifacts anymore is that you need to build them locally in those few cases you need to do some testing. (I generally build them locally anyway :D) So yes, I think it's for the better.
Also I agree that there's a high chance that we are going to add "extra value" in this separate step.

nirinchev:
The wrapper artifacts are still bundled in the nupkg which is available as an artifact, so we'll be able to download and unpack that.

LaPeste:
> The wrapper artifacts are still bundled in the nupkg which is available as an artifact, so we'll be able to download and unpack that.

Oh right, I didn't think of that. Then I can't really see any cons in removing them from the artifacts.

PR #2452, opened 2021-06-17
Clean up the package.json and add some docs

nirinchev:
## Description

Changes how we distribute the Unity package to [ILRepack](https://github.com/gluck/il-repack) its dependencies and make them internal (except for MongoDB.Bson which we have in the public API, but I don't expect it to conflict with other libraries being used in the Unity project).

ILRepacking our dependencies allows us to distribute a single package and the added managed library should be negligible compared to the native binaries and the actual game assets. If we do run into a situation where we conflict with another library referencing MongoDB.Bson, we'll have to figure out a way to distribute without it.

##  Overview of the changes

* Add a Unity-specific Readme.md to Realm.Unity
* Add Third Party notices for the libraries we're bundling
* Modify SetupUnityPackage to copy Changelog.md  and License.md to the Unity package
* Modify SetupUnityPackage to bundle main package dependencies with ILRepack producing a single package where System.Buffers, Remotion.Linq, and so on are ILMerge-d and internalized with the Realm.dll, while MongoDB.Bson is ILMerged but still exposed.


Realm/Realm.Unity/README.md
```diff
@@ -0,0 +1,42 @@
+![Realm](https://github.com/realm/realm-dotnet/raw/master/logo.png)
+
+Realm is an embedded mobile database that runs directly on phones, tablets, or desktops.
```
nirinchev, line 3: @mongomoe tagging you to review the Readme - this is what's going to be shown on npmjs - e.g. here: https://www.npmjs.com/package/moment as well as if the user clicks on it in the unity package manager.

LaPeste (approved):

Tests/Tests.Unity/Assembly-CSharp-Editor.csproj
```diff
@@ -314,42 +314,18 @@
     <Reference Include="Mono.Cecil">
       <HintPath>C:\work\dotnet\Realm\Realm.Unity\Editor\Cecil\Mono.Cecil.dll</HintPath>
     </Reference>
-    <Reference Include="Nito.AsyncEx.Tasks">
```
LaPeste, line 317: Are these dependences gone because we only do the bundled version of the package for unity, right?

LaPeste (approved):
It's a good amount of changes but all seemed geared toward suppliying ILRepack with the dependencies you want to ILMerge in the final dll.
I don't think I should read all the specifications of ILRepack to verify each single line of the new changes. Despite this, the overall "shape" of the changes look reasonable.

PR #2457, opened 2021-06-21
Automate SDK release from CI

### LINKED ISSUE #2051: Publish package from CI
When building for release, we should add a prompt to publish the package to NuGet.

LaPeste:
## Description
This PR adds a workflow that automates the whole release process of the .NET SDK from Github Actions.

Fixes #2051

##  TODO

- [x] stop generating docs on every run


.github/templates/main.yml
```diff
@@ -290,7 +290,20 @@ jobs:
...
         id: set-version-suffix
         #! Build suffix is PR-1234.5 for PR builds or alpha.123 for branch builds.
-        run: echo "::set-output name=build_suffix::${{ github.event_name == 'pull_request' && format('PR-{0}', github.event.number) || 'alpha'}}.${{ github.run_number }}"
+        run: |
+          $suffix = ""
+          if ($env:GITHUB_EVENT_NAME -eq "pull_request")
+          {
+            if (-Not $env:GITHUB_REF.Contains("release"))
+            {
+              $suffix = "PR-${{ github.event.number }}.$env:GITHUB_RUN_NUMBER"
```
LaPeste, line 299: I'd have like to use the env syntax, since it's easier to read to my eyes, but `$env:GITHUB_EVENT_NUMBER` doesn't seem to produce any output, so I went back to `${{ github.event.number }}`, just because it's the only way to read that value.

.github/workflows/main.yml
```diff
@@ -639,3 +652,32 @@ jobs:
...
+        {
+            New-Item -ItemType directory -Path $resultPath
+        }
+        Compress-Archive -Path Docs/_site -DestinationPath "$resultPath/Docs.zip"
+    - name: Store docs artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        name: Docs.zip
+        path: ${{ github.workspace }}/Realm/packages/Docs.zip
+        retention-days: ${{ github.event_name != 'pull_request' && 30 || 1 }}
```
LaPeste, line 683: I'm not sure the amount of days we'd like to retain the docs. So I just put the standard amount length we always use. Please, advise if 30 days are too many.

.github/workflows/release.yml
```diff
@@ -0,0 +1,54 @@
...
+        echo "RELEASE_NOTES=$output_file" | Out-File $env:GITHUB_ENV -Encoding utf8
+    - name: Extract release name
+      run: |
+        $nupkgName = ls Realm/packages/Realm.Fody.$env:RELEASE_VERSION | select -expandproperty Name | Select-String Fody
+        $nupkgName -match "(?sm)\d{1,2}\.\d{1,2}\.\d{1,2}"
+        $version = $Matches[0]
+        echo "RELEASE_VERSION=$version" | Out-File $env:GITHUB_ENV -Encoding utf8
+    - name: Upload to nuget
+      run: |
+        dotnet nuget push "Realm/packages/Realm.Fody.$env:RELEASE_VERSION/Realm.Fody.$env:RELEASE_VERSION.nupkg" --skip-duplicate --api-key ${{ secrets.NUGET_API_KEY }} --source https://api.nuget.org/v3/index.json
```
LaPeste, line 34: This folder within `packages` is because I decided to download all artifacts. When doing so, by specifications, a folder per artifact is created and each artifact is appropriately placed.

.github/workflows/release.yml
```diff
@@ -0,0 +1,54 @@
...
+        $changelogContent = Get-Content -Raw -Path "./CHANGELOG.md"
+        $output_file = "./RELEASE_NOTES.md"
+        $regex = "(?sm)^(## \d{1,2}\.\d{1,2}\.\d{1,2} \(\d{4}-\d{2}-\d{2}\))(.+?)(\n## \d{1,2}\.\d{1,2}\.\d{1,2} \(\d{4}-\d{2}-\d{2}\))"
+        $changelogContent -match $regex
+        Set-Content -Path $output_file -Value $Matches[2]
+        echo "RELEASE_NOTES=$output_file" | Out-File $env:GITHUB_ENV -Encoding utf8
+    - name: Extract release name
+      run: |
+        $nupkgName = ls Realm/packages/Realm.Fody.$env:RELEASE_VERSION | select -expandproperty Name | Select-String Fody
+        $nupkgName -match "(?sm)\d{1,2}\.\d{1,2}\.\d{1,2}"
```
nirinchev, line 29: Will that regex (and the one above) match beta versions?

.github/workflows/release.yml
```diff
@@ -0,0 +1,54 @@
...
+        $version = $Matches[0]
+        echo "RELEASE_VERSION=$version" | Out-File $env:GITHUB_ENV -Encoding utf8
```
nirinchev, line 31: Rather than use env variables here, I think we should use set-output

.github/workflows/release.yml
LaPeste: Good point. I'm gonna fix that.

.github/workflows/release.yml
LaPeste: I can do that but, what's the advantage? As a con it has longer syntax, what are the pros?

LaPeste:
There's a thing I've been debating about: checking for error/warning in the executed commands. Generally we don't need to do any check on shell command return values, since when a command fails the step automatically fails consequently making the whole workflow fail, unless instructed otherwise. Now this is the problem, the command itself doesn't really fail, otherwise it'd just do what's needed -> exit with failure. But unfortunately the commands doesn't really do well, for example returning warnings.
An example at hand is docfx, it doesn't fail but it reports tons of warnings which one can't see until the logs are checked. The result is a generated documentation half broken/ half fine.
_This is a bad example because I manually set it to be fine with duplicates, but for the point of the conversion it's ok_: Another example could be in nuget, when uploading a package that's already there and the option for skipping duplicate is set then a warning is given but GH Actions doesn't understand it and it goes [baried in the logs](https://github.com/realm/test/runs/3000327690?check_suite_focus=true).

So, the conclusion is: just sometimes GH runners don't understand warnings. What should we do about this?

1. We could check on each single error of each command **(bad idea in my opinion as it'll make the code unreadable)**
2. We could parse the output of each command in a custom function looking for a few keywords: E/e-rror-s, W/warning-s etc and release a general warning from the step so that it appears in the summary of the run workflow **(maybe the best solution)**
3. Do nothing and manually check the logs and results of each step we know it's delicate, e.g. so far it's been only a problem for release workflow to my knowledge) **(I suppose this is a still acceptable solution)**.

What's your opinion on the matter?

.github/workflows/release.yml
nirinchev: I generally dislike env variables as a way of passing data between steps. The step outputs are designed precisely for that and we should prefer to use them.

nirinchev (commented):
Added 2 minor comments and some explanation for the changes I made to get the generated docs to resolve references correctly.

.github/templates/Readme.md
```diff
@@ -4,9 +4,16 @@ We're using [ytt](https://github.com/vmware-tanzu/carvel-ytt) to generate the gi
...
+If instead ytt isn't in your packager manager or you don't use one you can manually get ytt as follows:
+
 1. Download the executable for your platform from the [releases page](https://github.com/vmware-tanzu/carvel-ytt/releases).
 1. Rename it to `ytt`.
 1. Place it in your PATH.
 
 ## Building the docs
-1. In `$SolutionDir/.github/templates` run `ytt -f main.yml --output-files ../workflows/`.
+1. `cd $SolutionDir/.github/templates`
+1. `ytt -f . --output-files ../workflows/` ==> to target all files<br/>
```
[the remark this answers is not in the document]
LaPeste: In markdown we can insert a line break in 2 ways:
1. Specifically adding a line break tag (like a did)
2. Ending a line with 2 spaces and a new line char

I thought that in the option number 2 it would be harder to notice the new line. But maybe it's nicer to read? Which approach do you recommend?

.github/templates/main.yml
```diff
@@ -288,10 +315,24 @@ jobs:
...
+          }
+          else
+          {
+            $suffix = "alpha.$env:GITHUB_RUN_NUMBER"
+          }
+          echo "::set-output name=build_suffix::$suffix"
       - #@ template.replace(fetchWrapperBinaries())
       - #@ template.replace(buildPackages())
       - #@ findPackageVersion()
+      - #@ template.replace(buildDocs())
```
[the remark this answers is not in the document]
LaPeste: 🤦🏻

.github/templates/main.yml
```diff
@@ -66,11 +66,12 @@ steps:
       retention-days: 1
 #@ end
 
-#@ def checkoutCode():
+#@ def checkoutCode(submodules=False):
```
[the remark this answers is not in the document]
LaPeste: Considering that the submodules update is needed only for the wrappers, it totally makes sense to skip them in any other case. In fact, in the release workflow I was not fetching the submodules as it was only needed the CHANGELOG.
Good change.

.github/templates/main.yml
```diff
@@ -66,11 +66,12 @@ steps:
...
   - name: Checkout code
     uses: actions/checkout@v2
     with:
-      submodules: recursive
+      submodules: #@ submodules
+      ref: ${{ github.event.pull_request.head.sha }}
```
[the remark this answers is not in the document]
LaPeste: I want to double check that I got this right:
Our main workflow is triggered with push on master/main and on pull_request (with events _opened, synchronize, or reopened_). Since `ref` defaults to the reference or SHA of the event that triggered the workflow, you're saying that we need to avoid to fetch a push to master, which generally happens to be a merge of a pull request, but we always need to carefully fetch the head of the current PR which could be lacking the latest merged PRs that master has.
Did I get this right?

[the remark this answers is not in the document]
nirinchev: The default behavior for checkout on a PR trigger is to checkout the merge commit (i.e. have the PR branch merged into the base branch and use that). This is why when there are merge conflicts, our GHA workflows don't run. This is the change suggested in the [checkout action docs](https://github.com/actions/checkout#Checkout-pull-request-HEAD-commit-instead-of-merge-commit) to make sure that PR builds actually build and test the code that is in the PR itself.

While for PR builds against master, checking out the merge commit is not a big deal - and in some cases may even be useful - for release PRs it's plain wrong and may result in very surprising behavior where we actually push a nuget package containing code that was not present in the release branch.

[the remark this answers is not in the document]
LaPeste: Now it all makes sense. I wasn't fully aware of the default behaviour for the checkout on PRs, that's why I got confused. Thank you for the explanation. On top of this, the change of line 74 will stop "there are merge conflicts with master" that prevent the workflow to run.

PR #2461, opened 2021-06-22
Fixes for the sync datatype tests

nirinchev:
## Description

Fixes a few issues that I came across when running the datatype tests on macOS - most notably, we were trying to add an object belonging to a different Realm to a dictionary which was failing in all sorts of weird ways. Additionally, it adds adds equals overrides for the RealmValue tests as `RealmValue.Equals` will return `false` for `RealmObject` instances belonging to different Realms.

### DIFF
--- a/Tests/Realm.Tests/Sync/DataTypeTests.cs
+++ b/Tests/Realm.Tests/Sync/DataTypeTests.cs
@@ -24,6 +24,7 @@ using System.Threading.Tasks;
 using MongoDB.Bson;
 using NUnit.Framework;
 using Realms.Helpers;
+using Realms.Logging;
 
 namespace Realms.Tests.Sync
 {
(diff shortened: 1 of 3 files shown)


Tests/Realm.Tests/Sync/DataTypeTests.cs
```diff
@@ -24,6 +24,7 @@
 using MongoDB.Bson;
 using NUnit.Framework;
 using Realms.Helpers;
+using Realms.Logging;
```
papafe, line 27: Is this used?

papafe (approved):
Looks good. Interesting stuff...

Tests/Realm.Tests/Sync/DataTypeTests.cs
```diff
@@ -570,6 +569,65 @@ private static RealmValue Clone(RealmValue original)
...
+            // Special handling the object case as they might be "equivalent" but belonging to different realms
+            if (a.Type != RealmValueType.Object || b.Type != RealmValueType.Object)
+            {
+                return false;
+            }
+
+            var objA = a.AsRealmObject();
+            var objB = b.AsRealmObject();
+
+            if (objA.GetType() != objB.GetType())
```
DominicFrei, line 615: How is this different from line `607`? Don't we already know they are both `RealmValueType.Object`?

nirinchev: This is not as obvious in the github diff view as it would have been in VS, but 607 is checking `RealmValue.Type` which is an enum of `int`, `string`, `object`, etc.

Line 617 is using reflection to get the .NET type of the `RealmObject` instance that we obtained. The [System.Type](https://docs.microsoft.com/en-us/dotnet/api/system.type?view=net-5.0) being returned here contains information about the actual .NET type - e.g. the class `IntPropertyObject` in the namespace `Realms.Tests` from the assembly XYZ, etc.

PR #2464, opened 2021-06-23
Add tests and docs for Realm.WriteCopy with Sync

### LINKED ISSUE #2417: Add tests and docs for Realm.WriteCopy with Sync
Core changed the behavior of WriteCopy for synchronized Realms to remove the client file identifier. This means that we can now WriteCopy a Realm owned by user A and open it with user B and not have to redownload the data/experience a client reset. We should get that for free since we already use WriteCopy, but we should:
1. Write tests that verify that this is indeed the behavior, including error cases where the user tries to copy a Realm that hasn't finished uploading all its changes.
2. Update the API docs to clarify what the behavior for sync configurations is.

DominicFrei:
## Description

Fixes #2417 

##  TODO

* [X] Changelog entry
* [X] Tests (if applicable)


wrappers/src/shared_realm_cs.cpp
```diff
@@ -410,6 +410,16 @@ REALM_EXPORT void shared_realm_write_copy(SharedRealm* realm, uint16_t* path, si
     });
 }
 
+REALM_EXPORT void shared_realm_write_copy_without_client_file_id(SharedRealm* realm, uint16_t* path, size_t path_len, char* encryption_key, bool allow_overwrite, NativeException::Marshallable& ex)
+{
+    handle_errors(ex, [&]() {
+        Utf16StringAccessor pathStr(path, path_len);
+
+        // by definition the key is only allowed to be 64 bytes long, enforced by C# code
+        realm->get()->write_copy_without_client_file_id(pathStr, BinaryData(encryption_key, encryption_key ? 64 : 0), allow_overwrite);
```
nirinchev, line 419: Why is this exposed as a different method in Core? I would very much prefer that they consolidate `write_copy` and `write_copy_without_client_file_id` as I don't see a point in having two different things here.

Realm/Realm/Realm.cs
```diff
@@ -1367,6 +1367,24 @@ public void WriteCopy(RealmConfigurationBase config)
...
+        /// non-null <see cref="RealmConfigurationBase.EncryptionKey"/>, the copy will be encrypted with that key.
+        /// </summary>
+        /// <remarks>
+        /// If this is called from within a transaction it writes the current data, and not the data as it was when
+        /// the last transaction was committed.
+        /// </remarks>
+        /// <param name="config">Configuration, specifying the path and optionally the encryption key for the copy.</param>
+        /// <param name="allowOverwrite">If true, an already existing Realm file will be overwritten. If false and the
+        /// file already exists, this will throw an error.</param>
+        public void WriteCopyWithoutClientFileId(RealmConfigurationBase config, bool allowOverwrite)
```
nirinchev, line 1381: Regardless of what Core decides about the methods internally, we should not expose an additional method here. If they are very much against consolidating the two, we can push the distinction further down our stack.

nirinchev (changes requested):

wrappers/src/shared_realm_cs.cpp
DominicFrei: ^ @jedelbo I had the same thought. What's your opinion about having the `copy client file id` part as a parameter to that function instead of having two functions?

nirinchev: It shouldn't even be a parameter - there's no reason to ever copy the client file Id.

DominicFrei: Fixed with the latest commit.

DominicFrei:
@nirinchev The CI fails in the most recent build (https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-dotnet/detail/PR-2464/2/pipeline/450) are based on the fact that copying inside a transaction does work differently when using `write_copy_without_file_id` which fails this test:
https://github.com/realm/realm-dotnet/blob/df/add-write-copy-without-file-id-tests/Tests/Realm.Tests/Database/InstanceTests.cs#L569

If have replicated / verified this in Core via two new tests:
https://github.com/realm/realm-core/commit/0d979c2c67c5e0462e84bb56b41a2bbb5cff24d9

The question that arises here:
What is the desired behavior of the new functionality? A.k.a. do we have to adjust our test or Core's implementation?

@jedelbo ^ FYI

nirinchev:
There should only be 1 method in Core, so `write_copy_without_file_id` should be folded into `write_copy` and adopt its behavior unless there are technical challenges that are not obvious.

jedelbo:
`write_copy_without_file_id` will not write uncommitted data into the copy. The whole idea is that we write a copy with content that has been integrated in the server. So in that way it behaves different than current `Realm::write_copy`.

Realm/Realm/Realm.cs
```diff
@@ -1324,6 +1323,11 @@ public void RemoveAll()
         /// </summary>
         /// <remarks>
         /// The destination file cannot already exist.
+        ///
+        /// For synchronized realms, the file written will have the client file identifier removed.
+        /// Furthermore it is required that all local changes are synchronized with the server before the copy can be written.
+        /// This is to be sure that the file can be used as a starting point for a newly installed application.
+        /// The function will throw if there are pending uploads.
         /// <para/>
         /// If this is called from within a transaction it writes the current data, and not the data as it was when
```
papafe, line 1332: I know it's not in your PR, but I think the text here in the remark is a little confusing. Does it mean that it first will write the current data and only then do the copy? (So that the new changes are inlcuded)

papafe: Also because of what you have added I suppose that this will throw an exception, right?

DominicFrei:
Updated with the new Core PR. Also updated description. The failing tests look like known fails to me.

wrappers/src/sync_user_cs.cpp
```diff
@@ -179,31 +179,31 @@ extern "C" {
             switch (profile_field)
             {
             case UserProfileField::name:
-                field = user->user_profile().name;
+                field = user->user_profile().name();
```
papafe, line 182: Was this not working before?

papafe (commented):
Overall it seems fine to me

wrappers/src/sync_user_cs.cpp
nirinchev: I think that's due to https://github.com/realm/realm-core/pull/4244 changing the Core API.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -320,6 +320,74 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+            SyncTestHelpers.RunBaasTestAsync(async () =>
+            {
+                var originalPartition = Guid.NewGuid().ToString();
+                var originalConfig = await GetIntegrationConfigAsync(originalPartition);
+                if (originalEncrypted)
+                {
+                    originalConfig.EncryptionKey = TestHelpers.GetEncryptionKey(42);
+                }
+
+                var copiedPartition = Guid.NewGuid().ToString();
```
nirinchev, line 336: The copy should be synchronizing on the same partition - we can't change the partitions, so if this test passes, let's add a check `WriteCopy` to make sure the partition is the same.

Realm/Realm/Realm.cs
DominicFrei: It doesn't matter if I changed it or not. We want to make it as accurate and easy to understand as possible hence we need to update this. 👍 

The more I think about it and read the discussions in Slack about we should probably not state this part at all and rather make it clear that Realms should not be copied while anything is going on since that can lead to all kinds of problems (especially with Sync).

I'll re-write that part and then we can think about it and see if everything necessary is included.

DominicFrei: @papafe Update in bcbb402. What do you think?

wrappers/src/sync_user_cs.cpp
DominicFrei: This PR needed the newest Core version to use the new `write_copy` released with https://github.com/realm/realm-core/pull/4827.

https://github.com/realm/realm-core/pull/4244 is part of this release which includes changing the properties to functions, see https://github.com/realm/realm-core/pull/4244/files#diff-a3fd6f3088dd3faa41d4cbffd668496ceb272c4937cdce9fc4919a9acbc4a1a7R77.

Realm/Realm/Realm.cs
papafe: Seems fine, but I'm not sure about where the part with "If this is called from within a transaction it writes the current data, and no..." went. Is it nr.4 in the commit?

Realm/Realm/Realm.cs
```diff
@@ -1323,10 +1322,12 @@ public void RemoveAll()
         /// non-null <see cref="RealmConfigurationBase.EncryptionKey"/>, the copy will be encrypted with that key.
         /// </summary>
         /// <remarks>
-        /// The destination file cannot already exist.
-        /// <para/>
-        /// If this is called from within a transaction it writes the current data, and not the data as it was when
-        /// the last transaction was committed.
+        /// 1. The destination file cannot already exist.
+        /// 2. For synchronized realms, the file written will have the client file identifier removed.
```
nirinchev, line 1326: I don't think we need to explicitly mention this - the client file identifier is an implementation detail and not something exposed to users, so the fact that we're removing it is kind of irrelevant.

Realm/Realm/Realm.cs
```diff
@@ -1323,10 +1322,12 @@ public void RemoveAll()
...
+        /// 3. Furthermore it is required that all local changes are synchronized with the server before the copy can be written.
+        ///    This is to be sure that the file can be used as a starting point for a newly installed application.
+        ///    The function will throw if there are pending uploads.
+        /// 4. It is advised to only copy a Realm file while it is not in use. Copying open Realms can lead to unexpected behvaiour.
```
nirinchev, line 1330: I don't think this comment is correct - copying an open file should be defined behavior, even if that behavior is to throw an error. Additionally, the way it's worded is quite confusing as `WriteCopy` is an instance method that can only be called on an open Realm. If the comment was taken verbatim, it would imply that users need to open a file, close it, then call `WriteCopy`, which will not work.

nirinchev (changes requested):

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
DominicFrei: 90d3d8e adds a test to check for that behaviour and also adjusts the other tests to use the same partition.

Realm/Realm/Realm.cs
DominicFrei: Removed with 192bcec.

Realm/Realm/Realm.cs
DominicFrei: https://github.com/realm/realm-dotnet/pull/2464#discussion_r682514879 addresses the same question. Let's continue there.

Realm/Realm/Realm.cs
DominicFrei: 6ec8b1f clarifies this difference and points out what happens for local vs. synced Realms.

Realm/Realm/Realm.cs
```diff
@@ -1323,16 +1322,28 @@ public void RemoveAll()
...
         /// </remarks>
         /// <param name="config">Configuration, specifying the path and optionally the encryption key for the copy.</param>
         public void WriteCopy(RealmConfigurationBase config)
         {
             Argument.NotNull(config, nameof(config));
 
+            if (config is SyncConfiguration && Config is SyncConfiguration)
+            {
+                SyncConfiguration originalConfig = (SyncConfiguration)Config;
+                SyncConfiguration copiedConfig = (SyncConfiguration)config;
```
nirinchev, line 1340: This can be rewritten using pattern matching `is` expression:

```suggestion
            if (config is SyncConfiguration copiedConfig && Config is SyncConfiguration originalConfig)
            {
```

Realm/Realm/Realm.cs
```diff
@@ -1323,16 +1322,28 @@ public void RemoveAll()
...
+                if (originalConfig.Partition != copiedConfig.Partition)
+                {
+                    throw new NotSupportedException("Copying a Realm to another partition is not supported.");
```
nirinchev, line 1343: Can we add the two partitions to the error message? Something like:

```csharp
throw new NotSupportedException($"Changing the partition to synchronize on is not supported when writing a Realm copy. Original partition: {originalConfig.Partition}, passed partition: {copiedConfig.Partition}");
```

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -320,6 +320,121 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+                AddDummyData(originalRealm, true);
+
+                await WaitForUploadAsync(originalRealm);
+                await WaitForDownloadAsync(originalRealm);
+
+                originalRealm.WriteCopy(copyConfig);
+
+                using var copiedlRealm = GetRealm(copyConfig);
+
+                Assert.AreEqual(copiedlRealm.All<ObjectIdPrimaryKeyWithValueObject>().Count(), DummyDataSize / 2);
```
nirinchev, line 357:
```suggestion
                using var copiedRealm = GetRealm(copyConfig);

                Assert.AreEqual(copiedRealm.All<ObjectIdPrimaryKeyWithValueObject>().Count(), DummyDataSize / 2);
```

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -320,6 +320,121 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+                Assert.That(originalConfig.Partition, !Is.EqualTo(copyConfig.Partition));
+
+                File.Delete(copyConfig.DatabasePath);
+
+                using var originalRealm = GetRealm(originalConfig);
+
+                AddDummyData(originalRealm, true);
+
+                await WaitForUploadAsync(originalRealm);
+                await WaitForDownloadAsync(originalRealm);
```
nirinchev, line 389: We don't really need to add data and wait for it to upload to verify the error happens.

nirinchev (approved):
Final nits from me

Realm/Realm/Realm.cs
DominicFrei: Learned something new today. 👍 
95f232c incorporates that and simplifies the whole check very nicely.

Realm/Realm/Realm.cs
DominicFrei: c6632b2 clarifies that.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
DominicFrei: Fixed with e9c5580.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
DominicFrei: Removed with 1b3f1cd.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -320,6 +320,117 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+                }
+
+                File.Delete(copyConfig.DatabasePath);
+
+                using var originalRealm = GetRealm(originalConfig);
+
+                AddDummyData(originalRealm, true);
+
+                // The error is thrown as a generic `RealmError` by Core which translates to a generic `RealmException` on our side.
+                Assert.Throws<RealmException>(() => originalRealm.WriteCopy(copyConfig));
```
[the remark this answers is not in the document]
DominicFrei: The message is:
```
Could not write file as not all client changes are integrated in server
```

Considering the test being `WriteCopy_FailsWhenNotFinished` I couldn't think of a better way to describe it. 👍

papafe (approved):
I think everything makes sense sense :)

bmunkholm:
@DominicFrei Just wondering how much of these tests could be in OS, and much is testing the code in .NET ?

PR #2465, opened 2021-06-23
Verify that objects do not belong to different realm when added to collection

### LINKED ISSUE #2462: Adding objects to collections across Realms fails in awkward ways
While working on https://github.com/realm/realm-dotnet/pull/2461 I noticed something like this fails with a pretty obscure exception from Core:

```csharp
var realm1 = await GetIntegrationRealmAsync(partition);
var realm2 = await GetIntegrationRealmAsync(partition);

var obj1 = realm1.Find<SyncCollectionsObject>(...);
var obj2 = realm2.Find<SyncCollectionsObject>(...);

var dict1 = obj1.ObjectDict;
var dict2 = obj2.ObjectDict;

var item = new IntPropertyObject();
realm1.Write(() =>
{
    dict1["foo"] = item;
});

realm2.Write(() =>
{
    dict2["bar"] = item;
});
```

The issue here is that `item` is already managed by a different Realm but we don't appear to be doing the check and happily pass it down to Core which gets very confused by what we're asking it to do.

papafe:
Added to checks to verify that objects do not belong to a different realm when being added to collection

Fixes #2462

### DIFF
--- a/Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
+++ b/Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
@@ -196,6 +196,12 @@ namespace Realms
             }
 
             var robj = value.AsRealmObject<RealmObject>();
+
+            if (robj.IsManaged && robj.Realm != Realm)
+            {
+                throw new RealmException("Can't add to the collection an object that is already in another realm.");
+            }
+
             if (!robj.IsManaged)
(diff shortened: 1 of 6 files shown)


nirinchev (commented):
Good stuff! Can you also add a test for lists?

nirinchev (approved):

PR #2467, opened 2021-06-23
Clean up native resources on Unity application exit

nirinchev:
## Description

When Unity closes the application, it will suspend garbage collection and wait for all native threads to exit (or at least stop doing work - threads that are waiting on conditional variables are not an issue). This is problematic because the managed AppHandle holds a shared_ptr to the native App, which in turn holds a shared_ptr to the SyncManger. This prevents the destruction of the SyncManager and this is where we actually stop the sync client. Not stopping the sync client means that it'll keep polling the socket in a loop, forcing Unity to wait forever.

The fix is to keep a collection of weak references to each AppHandle we create and force close all of them when the application exits.

##  TODO

* [x] Changelog entry

### DIFF
--- a/Realm/Realm/Handles/AppHandle.cs
+++ b/Realm/Realm/Handles/AppHandle.cs
@@ -219,10 +219,8 @@ namespace Realms.Sync
             {
                 foreach (var weakHandle in _appHandles)
                 {
-                    if (weakHandle.Target is AppHandle handle)
-                    {
-                        handle?.Close();
-                    }
+                    var appHandle = (AppHandle)weakHandle.Target;
+                    appHandle?.Close();
                 }
 
                 _appHandles.Clear();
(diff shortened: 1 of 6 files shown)


fealebenpae (approved):

Realm/Realm/Handles/AppHandle.cs
```diff
@@ -205,6 +213,22 @@ public static AppHandle CreateApp(Native.AppConfiguration config, byte[] encrypt
             return new AppHandle(handle);
         }
 
+        public static void ForceCloseHandles()
+        {
+            lock (_appHandles)
+            {
+                foreach (var weakHandle in _appHandles)
+                {
+                    if (weakHandle.Target is AppHandle handle)
```
papafe, line 222: Are we only being overly cautious here with the is operator and then then the null conditional operator? It seems it can't be anything else than AppHandle

papafe (approved):
Looks good

LaPeste (approved):
Beside the part that Ferdinando pointed out, the solution looks good. Most importantly, good stuff in discovering something that looks tedious to figure out.

PR #2477, opened 2021-06-28
Fix TableKey / ObjKey conversion on Android x86

### LINKED ISSUE #2456: Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR)
## Goals

After upgrading an app to Realm 10.x it stopped to launch in Android... but only for new deployments. Those users that already had an old version of the app installed didn't notice anything (neither did the developers), which has made everything so much fun...

## Expected Results

The app running and users happy.

## Actual Results

The app failing to launch and new users lost.

## Steps to Reproduce

- Download the QuickJournal example from this repository.
- Upgrade the nuget packages to their latest versions: Realm 10.2.0, Xamarin.Forms 5.0.0.2012, Xamarin.Essentials 1.6.1.
- Remove the EmbedAssembliesIntoApk from the Android Debug config (you can just delete the line from QuickJournal.Android.csproj)
- Try to debug the app (either in an emulator or a device). **Important: no data must exist for the app** (either delete the app first or delete its data).
- It will crash with the message in the title.

I haven't tested it in detail, but I think the crash also happens if the app is compiled for x86 (and deployed in an Intel Android image), regardless of the value of EmbedAssembliesIntoApk.

Note that it also works if AndroidFastDeploymentType is enabled.

Some extra info: this was a project that was created a few years ago, and either the EmbedAssembliesIntoApk did not exist then, or was not preconfigured by default (this was not something that we had changed since then). That's the reason why it was not set, and the fact that it actually worked perfectly on those devices that just upgraded the version totally caught us by surprise.

- Flavor:
  - [ ] MongoDB Realm
  - [x] Local Database only
- Client SDK Version: Realm .Net 10.2.0
- Client OS & Version: Android (crashed on all versions of Android 10.x and Android 11.x that we tested).

### LINKED ISSUE #2473: GetHashCode on Object should use TableKey in addition to ObjKey
ObjKey is not unique across tables so we should use TableKey in combination with it.

DominicFrei:
## Description
When retrieving `TableKey` and `ObjKey` from Core we interpret the returned values of type `UInt32` and `Int64` as `TableKey` and `ObjKey` without any conversion.

On Android x86 this leads to a misalignment in memory though since `TableKey` and `ObjKey` are structs which leads to a crash as described in https://github.com/realm/realm-dotnet/issues/2456.

This PR reads the return values correctly and converts them in their corresponding Handle into the structs used on the .NET side.

N.B.: No further tests added. The problem that this was not caught by our tests is not the test suite itself but the fact that we do not test on Android x86 at the moment. As soon as such a device is added and run the current tests will implicitly apply.

In addition to that the hash code of objects is now calculated using table key in addition to just the object key (see #2473).

Fixes #2456
Fixes #2473 

##  TODO

* [X] Changelog entry
* [X] Tests (if applicable)


Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -115,7 +115,7 @@ private static class NativeMethods
             public static extern bool refresh(SharedRealmHandle sharedRealm, out NativeException ex);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_get_table_key", CallingConvention = CallingConvention.Cdecl)]
-            public static extern TableKey get_table_key(SharedRealmHandle sharedRealm, [MarshalAs(UnmanagedType.LPWStr)] string tableName, IntPtr tableNameLength, out NativeException ex);
+            public static extern UInt32 get_table_key(SharedRealmHandle sharedRealm, [MarshalAs(UnmanagedType.LPWStr)] string tableName, IntPtr tableNameLength, out NativeException ex);
```
nirinchev, line 118: This adds a build warning - you'll need to add a pragma disable similar to what we have elsewhere (I think we have one in ObjectHandle).

Realm/Realm/Handles/ObjectHandle.cs
```diff
@@ -67,7 +67,7 @@ private static class NativeMethods
             public static extern bool equals_object(ObjectHandle handle, ObjectHandle otherHandle, out NativeException ex);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_get_obj_key", CallingConvention = CallingConvention.Cdecl)]
-            public static extern ObjKey get_obj_key(ObjectHandle handle, out NativeException ex);
+            public static extern Int64 get_obj_key(ObjectHandle handle, out NativeException ex);
```
nirinchev, line 70: This is marginally related to this PR, but let's remove this API altogether. It's only used to calculate the hash code of `RealmObjectBase`. Instead, let's replace it with a `object_get_hashcode` and implement it natively. That way we can remove ObjKey from the .NET SDK entirely. Additionally, it gives us an opportunity to include the TableKey in the hashcode calculation.

Realm/Realm/Handles/SortDescriptorHandle.cs
```diff
@@ -30,7 +30,7 @@ private static class NativeMethods
             public static extern void destroy(IntPtr queryHandle);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "sort_descriptor_add_clause", CallingConvention = CallingConvention.Cdecl)]
-            public static extern void add_clause(SortDescriptorHandle descriptor, TableKey table_key, SharedRealmHandle realm,
+            public static extern void add_clause(SortDescriptorHandle descriptor, UInt32 table_key, SharedRealmHandle realm,
```
nirinchev, line 33: Need a pragma disable here.

Realm/Realm/Handles/ObjectHandle.cs
```diff
@@ -66,8 +66,8 @@ private static class NativeMethods
             [return: MarshalAs(UnmanagedType.U1)]
             public static extern bool equals_object(ObjectHandle handle, ObjectHandle otherHandle, out NativeException ex);
 
-            [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_get_obj_key", CallingConvention = CallingConvention.Cdecl)]
-            public static extern ObjKey get_obj_key(ObjectHandle handle, out NativeException ex);
+            [DllImport(InteropConfig.DLL_NAME, EntryPoint = "object_get_hashcode", CallingConvention = CallingConvention.Cdecl)]
+            public static extern int get_hashcode(ObjectHandle handle, out NativeException ex);
```
nirinchev, line 70:
```suggestion
            public static extern Int32 get_hashcode(ObjectHandle handle, out NativeException ex);
```

Realm/Realm/Handles/ObjectHandle.cs
```diff
@@ -138,9 +138,9 @@ public override bool Equals(object obj)
             return result;
         }
 
-        public ObjKey GetObjKey()
+        public Int32 GetObjHash()
```
nirinchev:
```suggestion
        public int GetObjectHashCode()
```

wrappers/src/object_cs.cpp
```diff
@@ -241,11 +241,18 @@ extern "C" {
...
     }
 
-    REALM_EXPORT int64_t object_get_obj_key(const Object& object, NativeException::Marshallable& ex)
+    REALM_EXPORT int32_t object_get_hashcode(const Object& object, NativeException::Marshallable& ex)
     {
         return handle_errors(ex, [&]() {
-            // ObjKey is incompatible with C, so we return just the value.
-            return object.obj().get_key().value;
+            auto table_key_value = object.obj().get_table()->get_key().value;
+            auto table_key_hash = std::hash<int>{}(table_key_value);
```
nirinchev, line 248: Do we need these `std::hash`-es here? Can we use `table_key_value` directly - maybe with a cast?

nirinchev (approved):
Final nits, otherwise it looks good.

wrappers/src/object_cs.cpp
DominicFrei: I guess the normal way would be to only use them since they are already provided by the standard libary but the other way round should be fine as well. Removed them.

nirinchev: Yeah, I'm not familiar with how C++ does things, so happy to use the idiomatic way there - it's just coming from C#, you would rarely use `GetHashCode` on a primitive type when you can just use the value itself.

nirinchev (approved):

wrappers/src/object_cs.cpp
DominicFrei: I did some more research now that you mention it. Good point about the primitives:
https://stackoverflow.com/questions/38304877/why-stdhashint-seems-to-be-identity-function
It makes sense but means that I can just remove it since `std::hash<int>(i) == i` anyway.

I'll just keep it our way then. There are certainly many more ways but if the current implementation is fine I guess we can just keep it.
I thought about just concatenating table key and object key since that would always lead to a unique 'hash':
```
auto table_key_value = object.obj().get_table()->get_key().value;
auto object_key_value = object.obj().get_key().value;

std::string concatenation = "" + table_key_value + object_key_value;
int hashCode = std::stoi(concatenation);

return hashCode;
```

DominicFrei:
I had to add c249f2551041f927b908bd03b6313ce3792b3c32 since there was a valid warning on UWP:
```
'return': conversion from '__int64' to 'int32_t', possible loss of data
```

@nirinchev fyi + above comment about hashes.

It's just a tiny change, but I don't like merging unseen commits.

wrappers/src/object_cs.cpp
```diff
@@ -241,11 +241,16 @@ extern "C" {
...
-    REALM_EXPORT int64_t object_get_obj_key(const Object& object, NativeException::Marshallable& ex)
+    REALM_EXPORT int32_t object_get_hashcode(const Object& object, NativeException::Marshallable& ex)
     {
         return handle_errors(ex, [&]() {
-            // ObjKey is incompatible with C, so we return just the value.
-            return object.obj().get_key().value;
+            auto table_key_value = object.obj().get_table()->get_key().value;
+            auto object_key_value = object.obj().get_key().value;
+
+            int32_t hashCode = -9999999769;
```
nirinchev, line 250: That warning is correct - `-9999999769` is beyond what `int32` can hold. This is what VS generated for a simple struct with table key and object key if you want to use it:

```csharp
            public override int GetHashCode()
            {
                var hashCode = -986587137;
                hashCode = hashCode * -1521134295 + ObjKey.GetHashCode();
                hashCode = hashCode * -1521134295 + TableKey.GetHashCode();
                return hashCode;
            }
```

PR #2518, opened 2021-07-12
LINQ Provider Rewrite

joakim-olsson:
## Description

I'm experimenting with ways to rewrite the way the LINQ provider works to a more object-oriented structure. This is just a rough sketch so this is not planned to be merged and might not even be close to the end product.

The idea is to have it as objects so that it can be easily transferred to a JSON structure if that is the case. E.g a query like:
`var query = _realm.All<Person>().Where(p => p.Score == 5 && p.FirstName == "Test" && p.LastName == "Testersson");`

should give the ability to create a JSON structure like:

```json
{
    "WhereClause":
    [
        {
            "property": "Score",
            "value": 5,
            "op": "eq"
        },
        {
            "property": "FirstName",
            "value": "Test",
            "op": "eq"
        },
        {
            "property": "LastName",
            "value": "Testersson",
            "op": "eq"
            
        }
    ]
}
```


Realm/Realm/Linq/QueryObjects.cs
```diff
@@ -0,0 +1,150 @@
+﻿using System;
+
+namespace Realms
+{
+    public class QueryModel
```
papafe, line 5: This should probably be an internal class

Realm/Realm/Linq/QueryObjects.cs
```diff
@@ -0,0 +1,150 @@
...
+    public abstract class BooleanBinaryNode : ExpressionNode
+    {
+        public ExpressionNode Left { get; set; }
+
+        public ExpressionNode Right { get; set; }
+
+        public abstract string Operator { get; }
+    }
+
+    public class AndNode : BooleanBinaryNode
```
papafe, line 68: Do we need to keep both the operator and the kind for these kind of nodes? It seems just one would be enough

Realm/Realm/Linq/QueryObjects.cs
```diff
@@ -0,0 +1,150 @@
...
+    public class ThenByDescendingNode : OrderingClause
+    {
+        public override string Kind => "thenbydescclause";
+    }
+
+    public class WhereClause
+    {
+        public ExpressionNode ExpNode { get; set; }
+
+        public string Kind { get; } = "whereclause";
```
papafe, line 51: Here you are using an auto property, while later you are using an expression-bodied property (like on line 70). It should be uniformed (probably with the latter) :)

Realm/Realm/Linq/RealmResultsVisitor.cs
```diff
@@ -188,6 +189,10 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
                 {
                     Visit(node.Arguments[0]);
                     var lambda = (LambdaExpression)StripQuotes(node.Arguments[1]);
+                    var ans = new QueryModel();
```
papafe, line 192: I suppose this part here can be removed, right?

Tests/Tests.XamarinMac/Tests.XamarinMac.csproj
```diff
@@ -57,7 +57,7 @@
     <AOTMode>None</AOTMode>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(RunConfiguration)' == 'Default' ">
-    <StartArguments>--headless --labels=After</StartArguments>
+    <StartArguments>--labels=After --where=test=~AAAATests</StartArguments>
```
papafe, line 60: This is something you need to run tests, but you should not push it, because we actually use it also to run tests automatically.

Realm/Realm/Realm.csproj
```diff
@@ -49,61 +53,61 @@
     <None Include="..\..\wrappers\build\Darwin\$(Configuration)\librealm-wrappers.dylib">
       <Pack>true</Pack>
       <PackagePath>runtimes\osx-x64\native</PackagePath>
-      <Link>wrappers\Darwin\%(Filename)%(Extension)</Link>
```
papafe, line 52: This needs to be reverted to how it was before :)

Realm/Realm/Realm.csproj
```diff
@@ -29,6 +29,10 @@
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.3" />
     <PackageReference Include="Remotion.Linq" Version="2.2.0" />
     <ProjectReference Include="..\Realm.Fody\Realm.Fody.csproj" PrivateAssets="None" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
papafe, line 32: We need to check if we can create JSON strings without adding new dependencies. In general this is not an issue, but it's better to try to reduce, especially if we use the dependency in very few places.

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,329 @@
...
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using Newtonsoft.Json;
+using Realms.Schema;
+using LazyMethod = System.Lazy<System.Reflection.MethodInfo>;
+
+namespace Realms
+{
+    internal class RealmLinqExpression : Expression
```
papafe, line 13: Obviously this needs to be moved somewhere else

Realm/Realm/Linq/RealmResultsVisitor2.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace Realms
+{
+    internal class RealmResultsVisitor2 : ExpressionVisitor
+    {
+        private readonly Realm _realm;
+        private readonly RealmObjectBase.Metadata _metadata;
+        private QueryModel _query;
+
+        private QueryHandle _coreQueryHandle;
+        private SortDescriptorHandle _sortDescriptor;
```
papafe, line 15: I think that we are not using this

Realm/Realm/Linq/SortClause.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+            else
+            {
+                throw new NotSupportedException("Orderclause must have memberexpression");
+            }
+
+            return _orderingClause;
+        }
+
+        private static Expression StripQuotes(Expression e)
```
papafe, line 57: This method is the same in the WhereClauseVisitor. It would be convenient to have a base class where we can have all those common methods.

Realm/Realm/Linq/SortClause.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Realms
+{
+    internal class SortClause
+    {
+        public OrderingClause _orderingClause;
+
+        public OrderingClause VisitOrderClause(MethodCallExpression orderClause)
+        {
+            if (orderClause.Method.Name == nameof(Queryable.OrderBy))
```
papafe, line 13: As I have said this is something that needs to be discussed with Jakob, about how exactly this clause needs to look like, especially in regards to isAcending / isReplacing.

Realm/Realm/Linq/SortClause.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+            var lambda = (LambdaExpression)StripQuotes(orderClause.Arguments[1]);
+            if (lambda.Body is MemberExpression me)
+            {
+                _orderingClause.Property = me.Member.Name;
+            }
+            else
+            {
+                throw new NotSupportedException("Orderclause must have memberexpression");
```
papafe, line 51: Let's not forget to correct the spelling here. Also we need to consider that his is an expression that will be thrown to the user, so it should be more descriptive and less connected to the internals of Realm. The user probably does not know that we call this thing an OrderClause, for instance.

Realm/Realm/Linq/QueryObjects.cs
```diff
@@ -0,0 +1,150 @@
+﻿using System;
+
+namespace Realms
+{
+    public class QueryModel
+    {
+        public WhereClause WhereClause { get; set; }
```
papafe, line 7: As we said probably those needs to be lists, and not single values

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,329 @@
...
+                        method = method.GetGenericMethodDefinition();
+                    }
+
+                    return method;
+                });
+            }
+
+            internal static class String
+            {
+                internal static readonly LazyMethod Contains = Capture<string>(s => s.Contains(string.Empty));
```
papafe, line 43: Let's remove all the things that are not working or that are not being used, including comments. This will make the code easier to read for now

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,329 @@
...
+        {
+            _metadata = metadata;
+            _whereClause = new WhereClause();
+        }
+
+        public WhereClause VisitWhere(LambdaExpression whereClause)
+        {
+            _whereClause.ExpNode = Extract(whereClause.Body);
+            var json = JsonConvert.SerializeObject(_whereClause, formatting: Formatting.Indented);
+            Visit(whereClause.Body);
```
papafe, line 81: You don't need the Visit method here, as it's done already in Extract

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,329 @@
...
+                if (be.Left is MemberExpression me)
+                {
+                    if (me.Expression != null && me.Expression.NodeType == ExpressionType.Parameter)
+                    {
+                        var leftName = GetColumnName(me, me.NodeType);
+
+                        comparisonNode.Left.Value = leftName;
+                        comparisonNode.Left.Kind = "property";
+                        // TODO: Possible with switch statment in our current .NET version?
+                        if (me.Type == typeof(float))
```
papafe, line 144: This code is repeated in the same way some lines down. This is generally a sign that it could be extracted to another method.

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,329 @@
...
+                    }
+                    else
+                    {
+                        throw new NotSupportedException(valueType + "is not a supported type.");
+                    }
+                    // Add all types
+                }
+                returnNode = comparisonNode;
+            }
+            //else if (exp is MemberExpression)
```
papafe, line 193: Let's remove all of this for now.

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,329 @@
...
+                        comparisonNode = new GtNode();
+                        break;
+                    case ExpressionType.GreaterThanOrEqual:
+                        comparisonNode = new GteNode();
+                        break;
+                    default:
+                        throw new NotSupportedException($"The binary operator '{be.NodeType}' is not supported");
+                }
+
+                if (be.Left is MemberExpression me)
```
papafe, line 135: Here we are considering that in the expression the property is on the left and the value is on the right, like "p.Name == "alice". What would happen if they're inverted, for example "alice == p.Name"? 
This is also something that needs to be discussed with Jakob, in case it makes a difference

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,329 @@
...
+            if (parentType.HasValue)
+            {
+                if (name == null ||
+                    memberExpression.Expression.NodeType != ExpressionType.Parameter ||
+                    !(memberExpression.Member is PropertyInfo) ||
+                    !_metadata.Schema.TryFindProperty(name, out var property) ||
+                    property.Type.HasFlag(PropertyType.Array) ||
+                    property.Type.HasFlag(PropertyType.Set))
+                {
+                    throw new NotSupportedException($"The left-hand side of the {parentType} operator must be a direct access to a persisted property in Realm.\nUnable to process '{memberExpression}'.");
```
papafe, line 290: For the same reason I've said before, I think that this is not necessarily the left-hand side

Realm/Realm/Linq/RealmResultsVisitor2.cs
lomo133: I'll keep those for compiling purposes since MakeResultsForQuery method uses these. I'll remove them when the CreateResults method is rewritten

papafe: Now it's probably the right time to remove them :)

Realm/Realm/Linq/QueryObjects.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        public ComparisonNode()
+        {
+            Left = new StandardNode();
+
+            Right = new StandardNode();
+        }
+    }
+
+    public class StandardNode
```
papafe, line 72: What is a StandardNode?

Realm/Realm/Linq/RealmResultsVisitor2.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        IQueryableCollection results;
+
+        internal RealmResultsVisitor2(Realm realm, RealmObjectBase.Metadata metadata)
+        {
+            _realm = realm;
+            _metadata = metadata;
+            _query = new QueryModel();
+            _query.WhereClauses = new List<WhereClause>();
```
papafe, line 26: You can do this directly in the class

Realm/Realm/Linq/RealmResultsVisitor2.cs
```diff
@@ -0,0 +1,102 @@
...
+                    var whereClause = new WhereClauseVisitor(_metadata);
+                    var lambda = (LambdaExpression)StripQuotes(node.Arguments[1]);
+                    _query.WhereClauses.Add(whereClause.VisitWhere(lambda));
+                    return node;
+                }
+
+                if (IsSortClause(node.Method.Name))
+                {
+                    Visit(node.Arguments[0]);
+                    // Fix SOrtClause naming
```
papafe, line 46: For all these things that are "TODO" more than comments, I would suggest to write //TODO. This way they're more visible and also can be found automatically in Visual Studio

Realm/Realm/Linq/RealmResultsVisitor2.cs
```diff
@@ -0,0 +1,102 @@
...
+                    throw new NotSupportedException($"The method call '{node.Method.Name}' is not supported");
+                }
+            }
+            else
+            {
+                throw new NotSupportedException($"The method '{node.Method.Name}' is not supported");
+            }
+        }
+
+        // strange as it may seem, this is also called for the LHS when simply iterating All<T>()
```
papafe, line 62: I know I added this, but please remove this comment xD

Realm/Realm/Linq/RealmResultsVisitor2.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            return methodName == nameof(Queryable.OrderBy)
+                || methodName == nameof(Queryable.ThenBy)
+                || methodName == nameof(Queryable.OrderByDescending)
+                || methodName == nameof(Queryable.ThenByDescending);
+        }
+
+        public ResultsHandle MakeResultsForQuery()
+        {
+            var json = JsonConvert.SerializeObject(_query, formatting: Formatting.Indented);
```
papafe, line 97: As we discussed, we'll make this in a more elegant way, but this will work for now

Realm/Realm/Linq/SortClause.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Linq;
+using System.Linq.Expressions;
+
+namespace Realms
+{
+    internal class SortClause
```
papafe, line 7: This should be something like SortClauseVisitor

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,171 @@
...
+
+                returnNode = comparisonNode;
+            }
+
+            return RealmLinqExpression.Create(returnNode);
+        }
+
+        private static string GetKind(object valueType)
+        {
+            // TODO: Possible with switch statment in our current .NET version?
```
papafe, line 132: It's probably the time to check :)

papafe (commented):
Added some comments!

Realm/Realm/Linq/WhereClauseVisitor.cs
lomo133: Complains about version should be C# 9.0 or better, so I guess not :)

Realm/Realm/Linq/RealmResultsVisitor2.cs
lomo133: Discuss this today after Proj meeting

Realm/Realm/Linq/WhereClauseVisitor.cs
nirinchev: Looks like pattern matching doesn't require a new runtime: https://sergiopedri.medium.com/enabling-and-using-c-9-features-on-older-and-unsupported-runtimes-ce384d8debb

This means we can just bump the C# version and use it.

Realm/Realm/Linq/SortClauseVisitor.cs
```diff
@@ -0,0 +1,43 @@
...
+                || orderClause.Method.Name == nameof(Queryable.ThenByDescending))
+            {
+                _orderingClause.IsAscending = false;
+            }
+            else
+            {
+                throw new NotSupportedException(orderClause.Method.Name + " is not a supported ordering method");
+            }
+
+            var lambda = (LambdaExpression)RealmResultsVisitor2.StripQuotes(orderClause.Arguments[1]);
```
papafe, line 30: I am not a fan of this. I don't think that `StripQuotes` is in the right place at the moment.

Realm/Realm/Linq/RealmResultsVisitor2.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        protected override Expression VisitMethodCall(MethodCallExpression node)
+        {
+            if (node.Method.DeclaringType == typeof(Queryable))
+            {
+                if (node.Method.Name == nameof(Queryable.Where))
+                {
+                    Visit(node.Arguments[0]);
+                    var whereClause = new WhereClauseVisitor(_metadata);
+                    var lambda = (LambdaExpression)StripQuotes(node.Arguments[1]);
```
papafe, line 33: I would prefer to move this line and the following one to the `WhereClauseVisitor`. It will allow to have the code look very similar for all the clauses. Also, as we use `StripQuotes` in both ClauseVisitors, it would make sense to create a super class for those where we can add these common methods

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,252 @@
...
+                else if (test.Equals("EndsWith"))
+                {
+                    result = new EndsWithNode();
+                }
+                else if (test.Equals("Contains"))
+                {
+                    result = new ContainsNode();
+                }
+
+                // TODO Undecided implementation
```
papafe, line 55: Is this undecided because it was not discussed with Jacob?

papafe (changes requested):

Realm/Realm/Linq/WhereClauseVisitor.cs
papafe: Yep, so let's do it :)

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,252 @@
...
+            {
+                var orNode = new OrNode();
+                orNode.Left = Extract(be.Left);
+                orNode.Right = Extract(be.Right);
+                returnNode = orNode;
+            }
+            else
+            {
+                ComparisonNode comparisonNode;
+                switch (be.NodeType)
```
papafe, line 117: When we bump the C# version, we can use the simplified switch expression here too

Realm/Realm/Linq/QueryObjects.cs
```diff
@@ -0,0 +1,136 @@
+﻿using System.Collections.Generic;
+
+namespace Realms
```
papafe, line 3: Let's rename the file to QueryMode, it makes more sense

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,358 @@
...
+                        comparisonNode = new GtNode();
+                        break;
+                    case ExpressionType.GreaterThanOrEqual:
+                        comparisonNode = new GteNode();
+                        break;
+                    default:
+                        throw new NotSupportedException($"The operator '{binaryExpression.NodeType}' is not supported");
+                }
+
+                if (binaryExpression.Left is MemberExpression binaryLeftMemberExpression)
```
papafe, line 188: You have a lot of repeated code here, try to avoid this. There is no difference between left and right here

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,358 @@
...
+
+                returnNode = comparisonNode;
+            }
+
+            return RealmLinqExpression.Create(returnNode);
+        }
+
+        protected override Expression VisitUnary(UnaryExpression node)
+        {
+            NegationNode returnNode = new NegationNode();
```
papafe, line 268: Here you are creating already a node that is very specific. We should keep it as general until we can get the info on the specific type. I know we have only one at the moment, but this could grow

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,358 @@
...
+            }
+
+            return RealmLinqExpression.Create(returnNode);
+        }
+
+        protected override Expression VisitMember(MemberExpression node)
+        {
+            if (node.Expression != null && node.Expression.NodeType == ExpressionType.Parameter)
+            {
+                ComparisonNode comparisonNode = new EqualityNode();
```
papafe, line 285: When is this used?

Realm/Realm/Linq/WhereClauseVisitor.cs
```diff
@@ -0,0 +1,358 @@
...
+
+                return RealmLinqExpression.Create(comparisonNode);
+            }
+
+            throw new NotSupportedException($"The member '{node.Member.Name}' is not supported");
+        }
+
+        private static string GetKind(object valueType)
+        {
+            if (valueType == typeof(float))
```
papafe, line 308: Can we do this with a switch statement? Also, does valueType need to be object? Can we be more specific with the type?

Tests/Realm.Tests/Database/AAAATests.cs
```diff
@@ -0,0 +1,245 @@
...
+            && p.Score < 48).ToList();
+            Assert.That(scoreGtAndLt.Count, Is.EqualTo(1));
+            Assert.That(scoreGtAndLt[0].FullName, Is.EqualTo("Peter Jameson"));
+        }
+
+        [Test]
+        public void WhereFloatGteAndLteThanTest()
+        {
+            var scoreGteAndLte = _realm.All<Person>().Where(p => p.Score >= 100.0f
+            && p.Score <= 100.0f).ToList();
```
papafe, line 110: We try not to break the body of methods if the line is too long. You can put the whole Where on the next line if you eed to

Tests/Realm.Tests/Database/AAAATests.cs
```diff
@@ -0,0 +1,245 @@
...
+        {
+            var scoreEqAndLastNameEq = _realm.All<Person>().Where(p => p.Score == 100 && p.LastName == "Doe").ToList();
+            Assert.That(scoreEqAndLastNameEq.Count, Is.EqualTo(1));
+            Assert.That(scoreEqAndLastNameEq[0].FullName, Is.EqualTo("John Doe"));
+        }
+
+        [Test]
+        public void WhereBooleanOrTest()
+        {
+            var firstNameEqOrScoreEq = _realm.All<Person>().Where(p => p.FirstName == "NonExistant"
```
papafe, line 151: Same comment as before

Tests/Tests.XamarinMac/Tests.XamarinMac.csproj
papafe: Remember about this

Realm/Realm/Linq/WhereClauseVisitor.cs
lomo133: boolean statements with no equality sign (e.g .Where(p => !p.IsInteresting))

Realm/Realm/Linq/WhereClauseVisitor.cs
lomo133: Did some research on this and it does not seem like there is a good way to do this, we can discuss this later

papafe:
This is still a draft, but I've cleaned it out a little, so it should be readable enough.

PR #2519, opened 2021-07-13
Add EnumerableInterfaceImplementerSerializer to preserved BSON type

LaPeste:
## Description

Fixes #

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Fixed
 * Fixed an issue that would cause `Logger.Default` on Unity to always revert to `Debug.Log`, even when a custom logger was set. (Issue [#2481](https://github.com/realm/realm-dotnet/issues/2481))
 * Fixed an issue where `Logger.Console` on Unity would still use `Console.WriteLine` instead of `Debug.Log`. (Issue [#2481](https://github.com/realm/realm-dotnet/issues/2481))
+* [Unity] Preserved additional constructors necessary to serialize and deserialize Custom User Data. (PR [#2519](https://github.com/realm/realm-dotnet/pull/2519))
```
DominicFrei, line 6: The read me entry leads to this PR. Do we prefer to link to the PR or the issue itself (like it was done in the lines above)?

LaPeste: As written in the PR itself, unfortunately no issue was generated for that. I don't know if I should have done that. It just encountered the issue and opened a PR right away to fix it. Maybe this isn't the way it's done, I'm not sure.

Realm/Realm/Helpers/SerializationHelper.cs
```diff
@@ -99,6 +100,9 @@ internal static void PreserveSerializers()
             _ = new BsonArraySerializer();
 
             _ = new ObjectSerializer();
+
+            _ = new EnumerableInterfaceImplementerSerializer<IList<object>, object>();
+            _ = new ExpandoObjectSerializer();
```
nirinchev, line 105: What code did you write where the `ExpandoObjectSerializer` was needed? As far as I know, `ExpandoObject` is not supported on Unity, so I'm kind of surprised we ended up with a code path that reaches its serializer, although it's possible that it's just a fallback and the bson library uses it for non-expando objects as well.

Realm/Realm/Helpers/SerializationHelper.cs
```diff
@@ -99,6 +100,9 @@ internal static void PreserveSerializers()
             _ = new BsonArraySerializer();
 
             _ = new ObjectSerializer();
+
+            _ = new EnumerableInterfaceImplementerSerializer<IList<object>, object>();
```
nirinchev, line 104: Can this be `EnumerableInterfaceImplementerSerializer<IEnumerable<object>, object>` instead?

Realm/Realm/Helpers/SerializationHelper.cs
LaPeste: I'll put the code in a repo for you to take a look.

Realm/Realm/Helpers/SerializationHelper.cs
LaPeste: I'll try it out and report back.

LaPeste: You're correct. It works with IEnumerable as well.

Issue #2499, opened 2021-07-07
Release build picks up the wrong artifacts


nirinchev:
Seems like when downloading artifacts, we pass in only the branch name:

https://github.com/realm/realm-dotnet/blob/bd60f1701eca431c7a44a294de45e829c33c5bde/.github/workflows/release.yml#L16

which means that in cases where we have multiple builds from the same branch, we'll get a random set of artifacts. Instead, we should pass in the commit.

nirinchev:
Fixed in https://github.com/realm/realm-dotnet/pull/2524

PR #2524, opened 2021-07-14
Fix some CI publish workflow

### LINKED ISSUE #2498: Incorrect condition for building docs
The check for release branch:

https://github.com/realm/realm-dotnet/blob/bd60f1701eca431c7a44a294de45e829c33c5bde/.github/templates/main.yml#L103

appears to be wrong - docs are not built in release branches - e.g.

https://github.com/realm/realm-dotnet/runs/3013303618?check_suite_focus=true

Perhaps we should change it to `github.event.pull_request.head.ref`, but that should be tested.

LaPeste:
## Description
Fix 2 issues related to release workflow:
1- condition for when to build the docs from the main workflow. The problem was that the var used didn't have the name of the branch in it.
2- The artifacts fetched by the release workflow weren't specified with an exact commit number, so random artifact versions were being fetched

Fixes #2498 , #2499

### DIFF
--- a/.github/templates/main.yml
+++ b/.github/templates/main.yml
@@ -103,7 +103,7 @@ steps:
 #@ end
 
 #@ def buildDocs():
-#@ docsCondition = "${{ contains(github.ref, 'release') }}"
+#@ docsCondition = "${{ contains(github.head.ref, 'release') }}"
   - name: Check Docfx cache
     id: check-docfx-cache
     if: #@ docsCondition
--- a/.github/templates/main.yml
+++ b/.github/templates/main.yml
@@ -103,7 +103,7 @@ steps:
 #@ end
 
 #@ def buildDocs():
-#@ isRelease = "contains(github.head.ref, 'release')"
+#@ isRelease = "contains(github.head_ref, 'release')"
 #@ docsCondition = "${{ " + isRelease + " }}"
   - name: Check Docfx cache
     id: check-docfx-cache
--- a/.github/templates/main.yml
+++ b/.github/templates/main.yml
@@ -103,7 +103,8 @@ steps:
 #@ end
 
 #@ def buildDocs():
-#@ docsCondition = "${{ contains(github.head.ref, 'release') }}"
+#@ isRelease = "contains(github.head.ref, 'release')"
+#@ docsCondition = "${{ " + isRelease + " }}"
   - name: Check Docfx cache
     id: check-docfx-cache
     if: #@ docsCondition
@@ -112,7 +113,7 @@ steps:
       path: 'C:\docfx'
       key: docfx
   - name: Download docfx
-    if: ${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && contains(github.ref, 'release') }}
+    if: #@ "${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && " + isRelease + " }}"
     run: |
       Invoke-WebRequest -Uri https://github.com/dotnet/docfx/releases/download/v2.58/docfx.zip -OutFile C:\docfx.zip
       Expand-Archive -Path C:\docfx.zip -DestinationPath C:\docfx
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -353,6 +353,7 @@ jobs:
         echo ${{ github.event.pull_request.head.ref }}
         echo $GITHUB_HEAD_REF
         echo $GITHUB_BASE_REF
+      shell: bash
     - name: Find nupkg version
       id: find-nupkg-version
       run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -357,24 +357,24 @@ jobs:
       shell: bash
     - name: Check Docfx cache
       id: check-docfx-cache
-      if: ${{ contains(github.head.ref, 'release') }}
+      if: ${{ contains(github.head_ref, 'release') }}
       uses: actions/cache@v2
       with:
         path: C:\docfx
         key: docfx
     - name: Download docfx
-      if: ${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && contains(github.head.ref, 'release') }}
+      if: ${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && contains(github.head_ref, 'release') }}
       run: |
         Invoke-WebRequest -Uri https://github.com/dotnet/docfx/releases/download/v2.58/docfx.zip -OutFile C:\docfx.zip
         Expand-Archive -Path C:\docfx.zip -DestinationPath C:\docfx
       shell: powershell
     - name: Build docs
-      if: ${{ contains(github.head.ref, 'release') }}
+      if: ${{ contains(github.head_ref, 'release') }}
       run: |
         C:\docfx\docfx Docs/docfx.json
         Compress-Archive -Path Docs/_site -DestinationPath "Realm/packages/Docs.zip"
     - name: Store docs artifacts
-      if: ${{ contains(github.head.ref, 'release') }}
+      if: ${{ contains(github.head_ref, 'release') }}
       uses: actions/upload-artifact@v2
       with:
         name: Docs.zip
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -350,7 +350,7 @@ jobs:
     - name: Find nupkg version
       id: find-nupkg-version
       run: |
-        echo $github_ref
+        echo $GITHUB_REF
         cd Realm/packages
         tmpName=$(basename Realm.Fody* .nupkg)
         pkgName=${tmpName#"Realm.Fody."}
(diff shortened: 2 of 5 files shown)


.github/templates/main.yml
```diff
@@ -103,7 +103,7 @@ steps:
 #@ end
 
 #@ def buildDocs():
-#@ docsCondition = "${{ contains(github.ref, 'release') }}"
+#@ docsCondition = "${{ contains(github.head.ref, 'release') }}"
```
LaPeste, line 106: Just for context, in this branch `github.head.ref` = `ac/ci-fix-publish-workflow`. So this is what we want.

DominicFrei: If I understand correct, the docs will be created whenever a branch contains the word `release`.

Could you rename your branch to something that contains `release` so that we can see this effect?
A.k.a. how would I test this? :)

DominicFrei (commented):
Does line 366 in `main.yml` need to be changes too?
```
if: ${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && contains(github.ref, 'release') }}
```

.github/templates/main.yml
LaPeste: You can just print the content of the var, like I did. That's why I wrote the comment. You can modify the workflow and add:
1. in powershell: `$env:GITHUB_HEAD_REF`
2. in bash: `echo $GITHUB_HEAD_REF`

To any of the steps that use run, or make your own run step. Then, you can see in the workflow output its content.

For an explanation of what all github's env vars contain, you can look [here](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables).

.github/workflows/main.yml
```diff
@@ -357,24 +357,24 @@ jobs:
...
       id: check-docfx-cache
-      if: ${{ contains(github.ref, 'release') }}
+      if: ${{ contains(github.head.ref, 'release') }}
       uses: actions/cache@v2
       with:
         path: C:\docfx
         key: docfx
     - name: Download docfx
-      if: ${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && contains(github.ref, 'release') }}
+      if: ${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && contains(github.head.ref, 'release') }}
```
LaPeste, line 366: @DominicFrei good catch. I didn't notice because of the structure of the templated `main.yml`. I rearranged the template code. How does it look (line 106-107, 116 in `templates/main.yml`)?

.github/templates/main.yml
LaPeste: image
You can see here. The first echo was a test that was not correct, hence a blank line after the new line and then the 3 vars are shown below.

DominicFrei: Gotcha! Thanks for clarifying.

.github/workflows/main.yml
DominicFrei: Look good to me. Would you then want to use this substitution for `contains(github.head.ref, 'release')` in line 366 as well?

LaPeste: I'm not sure I follow. It's done in line 366. What do you mean?

DominicFrei: No worries.

Here is what I meant:
Line 106 was added:
```#@ isRelease = "contains(github.head.ref, 'release')"```

It was then used in line 116:
```    if: #@ "${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && " + isRelease + " }}"```

The same could be done in line 366:
```      if: ${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && contains(github.head.ref, 'release') }}```
could be
```      if: ${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && isRelease }}```

The `contains(github.head.ref, 'release')` could also be substituted with `isRelease` since you created it for that purpose, right?

In fact, you could also use it in line 360, 372 and 377:
```      if: ${{ contains(github.head.ref, 'release') }}```
could be 
```      if: ${{ isRelease }}```

File: .github/workflows/main.yml [hunk 3]
```diff
@@ -357,24 +357,24 @@ jobs:
       shell: bash
     - name: Check Docfx cache
       id: check-docfx-cache
-      if: ${{ contains(github.ref, 'release') }}
+      if: ${{ contains(github.head.ref, 'release') }}
```
DominicFrei, line 360:
```suggestion
      if: ${{ isRelease }}
```

.github/workflows/main.yml
DominicFrei:
```suggestion
      if: ${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && isRelease }}
```

.github/workflows/main.yml
```diff
@@ -357,24 +357,24 @@ jobs:
...
     - name: Download docfx
-      if: ${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && contains(github.ref, 'release') }}
+      if: ${{ steps.check-docfx-cache.outputs.cache-hit != 'true' && contains(github.head.ref, 'release') }}
       run: |
         Invoke-WebRequest -Uri https://github.com/dotnet/docfx/releases/download/v2.58/docfx.zip -OutFile C:\docfx.zip
         Expand-Archive -Path C:\docfx.zip -DestinationPath C:\docfx
       shell: powershell
     - name: Build docs
-      if: ${{ contains(github.ref, 'release') }}
+      if: ${{ contains(github.head.ref, 'release') }}
```
DominicFrei, line 372:
```suggestion
      if: ${{ isRelease }}
```

.github/workflows/main.yml
```diff
@@ -357,24 +357,24 @@ jobs:
...
       run: |
         C:\docfx\docfx Docs/docfx.json
         Compress-Archive -Path Docs/_site -DestinationPath "Realm/packages/Docs.zip"
     - name: Store docs artifacts
-      if: ${{ contains(github.ref, 'release') }}
+      if: ${{ contains(github.head.ref, 'release') }}
```
DominicFrei, line 377:
```suggestion
      if: ${{ isRelease }}
```

.github/workflows/main.yml
DominicFrei: I've added them all as change suggestions. Let me know if that makes sense. :)

DominicFrei (changes requested):

.github/workflows/main.yml
LaPeste: Now I understand what you mean. You can't do it that way. `isRelease` is a var that exist only for ytt. That has no meaning in the outputted yaml file. It's only living in the template accepted by ytt.

DominicFrei: I knew it will be something like that. Thanks for clarifying! Those CI configs are still quite confusing when you're not used to working with them. 🙈

LaPeste: No problem. Yes yes, we know it looks a bit weird at first. We all felt that way at the beginning 😛

papafe:
Looks good for my limited understanding 😁

.github/templates/main.yml
```diff
@@ -103,7 +103,8 @@ steps:
 #@ end
 
 #@ def buildDocs():
-#@ docsCondition = "${{ contains(github.ref, 'release') }}"
+#@ isRelease = "contains(github.head.ref, 'release')"
```
papafe, line 106: I actually have a question. In the github context docs (https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) I can see that there is `github.head_ref`, but not `github.head.ref`. Is it documented somewhere else?

DominicFrei (approved):

.github/templates/main.yml
LaPeste: That is, unfortunately, a typo, so it's a really good catch.
Working with release workflows is tricky because I can't really release, so I can only use another test repo for these release tests. For a modification so small I didn't use that repo, as I only had to change the name of a var in a condition and the typo happened. So, thank you.

I wish github could somehow warn of "unrecognized variables" in some way.

papafe: Yeah, unfortunately it's almost testing in production with those kind of things 🥲

.github/workflows/release.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       uses: dawidd6/action-download-artifact@v2
       with:
         workflow: main.yml
-        branch: ${{ github.ref }}
+        commit: ${{ github.event.pull_request.head.sha }}
```
nirinchev, line 16: Is this available for `workflow_dispatch` runs? I'd imagine `github.event.pull_request` will be `null` then. Maybe just use `github.sha`?

LaPeste: Thank you for spotting that. I verified in the test branch and `github.sha` in a `workflow_dispatch` rightly points to the latest commit.

LaPeste:
I'll make PR a draft again so that I can add the proper changelog for nuget.

nirinchev (approved):

PR #2525, opened 2021-07-15
Update repo examples

### LINKED ISSUE #2520: Update examples in repository
We need to check if the examples we have in the repository are still relevant. If so update them and make them more actual.

papafe:
This PR removed one of the examples in the repository and updated the other one (QuickJournal). This was updated to use the latest version of Realm, as well as was refactored and improved to use a better structure for the project

Fixes #2520


examples/QuickJournal/QuickJournal.Android/QuickJournal.Android.csproj
```diff
@@ -53,10 +53,10 @@
     <Reference Include="System.Numerics.Vectors" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Xamarin.Forms" Version="4.8.0.1451" />
-    <PackageReference Include="Xamarin.Essentials" Version="1.5.3.2" />
+    <PackageReference Include="Xamarin.Forms" Version="5.0.0.2083" />
+    <PackageReference Include="Xamarin.Essentials" Version="1.7.0" />
     <PackageReference Include="Realm">
-      <Version>5.1.1</Version>
+      <Version>10.3.0</Version>
```
DominicFrei, line 59: Would it make sense (if feasible) to just use the surrounding Realm version? User will most likely just check out the whole repo itself right?

DominicFrei: Do we build and run our examples as part of CI to make sure they always work?

examples/QuickJournal/QuickJournal/Services/NavigationService.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace QuickJournal.Services
+{
+    public static class NavigationService
+    {
+        static INavigation Navigation => Application.Current.MainPage?.Navigation;
+
+        public static async Task NavigateTo(BaseViewModel viewModel)
+        {
+            if (Navigation == null)
+                throw new InvalidOperationException("Main page does not support navigation");
```
DominicFrei, line 16: I remember seeing a comment in one of my PRs about brackets. Do we have a consistent opinion about that?

papafe: So.... My style was to avoid brackets for one liners, and so I went with my style because I took inspiration from another project I wrote ;)
But you're right, that in Realm we generally put brackets everywhere

examples/QuickJournal/QuickJournal.Android/QuickJournal.Android.csproj
papafe: I think people look at examples so that they can take inspiration and eventually play with them. If they have to get the whole repo then they'll need to build wrappers and it gets messy, so it's better to have them as separate projects in my opinion. 
Regarding building them, so far it wasn't done. Also examples were not even mentioned at all anywhere

papafe:
@LaPeste @DominicFrei I am not sure about the modifications to the main readme file, as well as the readme file for the QuickJournal example. Let me know what you think

examples/QuickJournal/QuickJournal.Android/QuickJournal.Android.csproj
DominicFrei: Fully agree. They should not need to build wrappers and everything themselves. It should 'just work'.
What I am trying to point at:
We seem to have not updates those examples in while because other tickets were more important. How do we make sure those examples do not get outdated again and keep working with newer versions too?
Someone might just have a look into the examples, copy some code and then it doesn't work with their code because they are using the newest version, not the one the example was built with.

In Cocoa this is part of the test process. We do build all Cocoa examples and if they don't build anymore CI turns red.

Maybe this is out of scope for this ticket and we should create a new one to take care of it. Let me know. :)

examples/QuickJournal/QuickJournal/Services/NavigationService.cs
DominicFrei: Not trying to bring in too much of my opinion, since consistency is the most important part for me here but I'm exactly the other way round because brackets make sure that when you add another statement and forget to add the brackets you don't alter the whole thing.

But as I said, my recommendation is to go with whatever our usual, recommended, consistent code style would be. :)

papafe: That is also Nikola's opinion 😁
Anyways to be consistent I also put brackets everywhere, so we follow the style of the main project

README.md
```diff
@@ -57,6 +57,10 @@ Prerequisites:
 
 If you are actively testing code against the Realm source, see also the unit test projects and other tests under the Tests folder.
 
+## Examples
+
+Some minimal examples of Realm use can be found in the `examples` folder.
```
DominicFrei, line 62: You could maybe specifically list them (might not make sense now because there is just one but there will be more in the future) and say what they are about.

Since I am working on a Unity example at the moment there might be quite some differences in those examples.

Which platform are they for?
Which Realm version were they built with (especially if we keep it fixes this could be important to point out early)?
What use cases can be found in this example? Just a simple local Realm, notifications, Sync, etc.
Are there any requirements to use this example? Xamarin v42.3, .NET 23.6
Maybe a summary of what this example is?

edit:
Just saw the summary already being part of the example itself.

I guess the question is: Do we want to keep everything in here in one place to give an overview or is it better placed in each example to only provide information where it is needed?

examples/QuickJournal/QuickJournal.Android/QuickJournal.Android.csproj
papafe: What you said makes sense, and I think we should address it in a new ticket. Probably it would make sense to update at least the Realm version to every major release we do

README.md
papafe: Yes, that's the point. I think maybe just a one liner for each example in the main readme will help, so that they don't need to open each example.

examples/QuickJournal/QuickJournal.Android/QuickJournal.Android.csproj
DominicFrei: Would you mind opening a new ticket adding anything that will not be done within this ticket/PR?

papafe: Sure

examples/QuickJournal/QuickJournal/ViewModels/JournalEntryDetailsViewModel.cs
```diff
@@ -1,36 +1,52 @@
...
-            Navigation.PopAsync(true);
+            if (string.IsNullOrEmpty(Entry.Title))
+            {
+                await DialogService.ShowAlert("Error", "You cannot save a note with an empty title");
+                return;
+            }
+
+            Entry.Metadata.LastModifiedDate = DateTimeOffset.Now;
+            transaction.Commit();
+            transaction = null;
```
LaPeste, line 42: What advantages do you gain in setting the transaction to null and checking OnDisappearing if it's still around to dispose of it?

README.md
DominicFrei: Looks good. :)

DominicFrei (approved):

examples/QuickJournal/QuickJournal/ViewModels/JournalEntryDetailsViewModel.cs
papafe: `OnDisappearing` gets called every time the page disappears, both when the user just presses the back button and when we programmatically go back to the previous page because the user wants to save the note. 
By setting it to null after saving we avoid disposing of a transaction that was already committed. It actually works anyways (no exceptions or anything) but it seems like terrible style 😁

DominicFrei: Do we need to dispose it at all? In other words (not a Xamarin lifecycle expert here): Could we just avoid `null` and `Dispose` and let the garbage collection take care of it?

LaPeste: I wasn't imply it was a mistake or anything. It was pure curiosity. Thank you for the explanation.

examples/QuickJournal/QuickJournal/ViewModels/JournalEntryDetailsViewModel.cs
```diff
@@ -1,36 +1,52 @@
...
 using Xamarin.Forms;
 
-namespace QuickJournal
+namespace QuickJournal.ViewModels
 {
-    public class JournalEntryDetailsViewModel
+    public class JournalEntryDetailsViewModel : BaseViewModel
     {
-        private Transaction _transaction;
+        private Transaction transaction;
```
DominicFrei, line 13: Nit pick here: I'd recommend sorting fields the same way you sort functions: public first, then private.

LaPeste (approved):

DominicFrei (approved):
Just a couple questions, no real change requests.

examples/QuickJournal/QuickJournal/ViewModels/JournalEntryDetailsViewModel.cs
LaPeste: @DominicFrei I think you gain that the GC doesn't get in that condition where it has to stop everything and do a round of cleaning, because you clean when you know it's time to. I assume this is the right way to do it in any language with automatic memory management.
@papafe am I right?

papafe: So... Here more than the memory management I'm worrying about leaving the db in a "valid" state. When we dispose of the transaction it also rollbacks all the changes that have been done since the transaction has been opened. If we leave it to the GC, we wouldn't know when this happens. 
Looking at the source code I can also see that it's never an issue to dispose of a transaction after it was committed (or viceversa actually).

DominicFrei: Fair point. Choosing the safe approach is surely the right way to go. Is this also documented so that users know about it? Do we always follow this rule consistently and how can we make sure we do?

papafe: I was just reading the new documentation, and it seems it does not say anywhere about the possibility of opening a transaction explicitly (not using `realm.Write`). It's definitely something that should be added

examples/QuickJournal/QuickJournal/ViewModels/JournalEntriesViewModel.cs
```diff
@@ -1,66 +1,77 @@
...
-            _realm = Realm.GetInstance();
+            realm = Realm.GetInstance();
+            Entries = realm.All<JournalEntry>();
 
-            Entries = _realm.All<JournalEntry>();
+            InitCommands();
+        }
 
-            AddEntryCommand = new Command(AddEntry);
+        private void InitCommands()
```
nirinchev, line 44: [nit] I'm not sure how I feel about this `InitCommands` method. Generally, I'd prefer stuff that only ever gets called once to be in the ctor rather than having a dedicated method for it. While I understand the sentiment of breaking up a large method into smaller components, it does feel a  little clunky and forces us to have a setter for every command. If we folded it into the ctor, we could have a getter-only property instead.

examples/QuickJournal/QuickJournal/ViewModels/JournalEntryDetailsViewModel.cs
nirinchev: I wonder if we should add `State` property to the `Transaction` type. Seems a bit silly that we're throwing exceptions if the user tries to double commit or rollback after a commit, but we don't offer an API for them to check what the state of the transaction is and if they can call methods on it. If we had a `State` property, we could then check it in `OnDisappearing` and have something like `if (transaction.State == TransactionState.Active) { rollback() }`.

examples/QuickJournal/QuickJournal/Views/JournalEntriesPage.xaml
```diff
@@ -1,24 +1,30 @@
...
-	</ContentPage.ToolbarItems>
-	<ContentPage.Content>
-		<ListView ItemsSource="{Binding Entries}" x:Name="EntriesListView" ItemTapped="OnItemTapped" ItemSelected="OnItemSelected">
-			<ListView.ItemTemplate>
-				<DataTemplate>
-					<TextCell Text="{Binding Title}" Detail="{Binding Metadata.Date, StringFormat='{0:dddd, MMMM d yyyy}'}">
-						<TextCell.ContextActions>
-							<MenuItem Text="Delete" IsDestructive="true" 
-									  Command="{Binding BindingContext.DeleteEntryCommand, Source={x:Reference EntriesListView}}" 
+        xmlns:d="clr-namespace:QuickJournal.Views"
```
nirinchev, line 4: should we call that `local` rather than `d`?

examples/QuickJournal/QuickJournal/ViewModels/JournalEntriesViewModel.cs
papafe: I agree, probably for such a simple app all of this small methods are unnecessary

examples/QuickJournal/QuickJournal/Services/DialogService.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Threading.Tasks;
+using Xamarin.Forms;
+
+namespace QuickJournal.Services
+{
+    public static class DialogService
```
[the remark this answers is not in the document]
papafe: Given that it's just a simple example app, I don't think it's necessary

examples/QuickJournal/QuickJournal/App.cs
```diff
@@ -1,12 +1,18 @@
...
 {
     public class App : Application
     {
         public App()
         {
-            MainPage = new NavigationPage(new JournalEntriesPage());
+            var page = new JournalEntriesPage();
+            {
+                BindingContext = new JournalEntriesViewModel();
+            };
```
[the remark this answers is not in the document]
papafe: The main idea here would be to have the `Pages` completely stupid and devoid of anything related to the logic of the app. I have decided to make this actually part of `NavigationService`, so it's more uniform.

examples/QuickJournal/QuickJournal/ViewModels/JournalEntriesViewModel.cs
```diff
@@ -1,66 +1,77 @@
...
 {
-    public class JournalEntriesViewModel
+    public class JournalEntriesViewModel : BaseViewModel
     {
-        // TODO: add UI for changing that.
-        private const string AuthorName = "Me";
-
-        private Realm _realm;
+        readonly JournalEntry selectedEntry = null;
+        readonly Realm realm;
```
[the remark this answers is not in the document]
papafe: It is always null, this is actually a "trick" to handle the click in the view. There's actually no command associated with the click event, so either we pass though the view, or we bind the `SelectedItem` like here. `selectedEntry` is always null so the clicked element does not get a different color. 
It's also possible to use `Behaviours`, but I didn't want to make it more complicated.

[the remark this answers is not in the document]
nirinchev: Do we need to have a local field for that then? Can we just modify the property to return `null` rather than the field?

examples/QuickJournal/QuickJournal/QuickJournal.csproj
```diff
@@ -11,8 +11,18 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <None Remove="Services\" />
+    <None Remove="Models\" />
+    <None Remove="Xamarin.CommunityToolkit" />
+  </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Services\" />
+    <Folder Include="Models\" />
   </ItemGroup>
```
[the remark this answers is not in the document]
papafe: I am not sure what happened here. I have removed everything

papafe:
Apart from all the different comments, I think we should all first agree on the way we want to handle the Add/Edit case, so that we can be consistent from now on.

examples/QuickJournal/QuickJournal/ViewModels/JournalEntriesViewModel.cs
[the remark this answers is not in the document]
papafe: Definitely

papafe:
So... regarding this example. This probably will be the simplest example we will have in our repository, so I am not totally sure about how we want to deal with the AddEditPage:

1. We remove the save button, use "live editing", so everything gets saved as it goes, easy flow
2. We keep the save button, and need to decide which kind of pattern we recommend in this case
3. We show both, maybe having two different pages implemented in different ways

In my opinion we should go with nr.1, given that it's probably the more "realm" way, and we want to make a simple example. 
Opinions on this? @nirinchev @LaPeste @DominicFrei

nirinchev:
I think for this example we should keep the open transaction approach. This allows us to rollback object creation, which isn't something that's trivial to do with approach 1. It's also likely to be more in line with people's expectations, even if it's not a best practice, especially when using Sync. We can create another tutorial/walkthrough that talks specifically about editing forks and explain the trade-offs of the various approaches there.

examples/QuickJournal/QuickJournal/ViewModels/JournalEntryDetailsViewModel.cs
papafe: Agree, opened #2551

papafe:
@nirinchev Let's do it like this. So for me the PR is ready as it is. 
I re-requested your review because Github was showing you proposed some changes but there were no changes actually it seems

examples/QuickJournal/QuickJournal/Services/NavigationService.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                throw new InvalidOperationException("Main page does not support navigation");
+            }
+        }
+
+        private static Page GetPageForViewModel(BaseViewModel viewModel)
+        {
+            return viewModel switch
+            {
+                JournalEntryDetailsViewModel _ => new JournalEntryDetailsPage(),
```
nirinchev, line 48:
```suggestion
                JournalEntryDetailsViewModel => new JournalEntryDetailsPage(),
```

examples/QuickJournal/QuickJournal/Views/JournalEntriesPage.xaml
```diff
@@ -1,24 +1,29 @@
...
+        <ListView ItemsSource="{Binding Entries}"
+                  SelectedItem="{Binding SelectedEntry, Mode=TwoWay}"
+                  SelectionMode="Single">
+            <ListView.ItemTemplate>
+                <DataTemplate>
+                    <TextCell Text="{Binding Title}"
+                              Detail="{Binding Metadata.CreatedDate, StringFormat='{0:dddd, MMMM d yyyy}'}">
+                        <TextCell.ContextActions>
+                            <MenuItem Text="Delete" IsDestructive="true" 
+									  Command="{Binding Path=BindingContext.DeleteEntryCommand, Source={x:Reference page}}"
```
nirinchev, line 21: This line seems to be using tabs instead of spaces

examples/QuickJournal/QuickJournal/Views/JournalEntryDetailsPage.xaml
```diff
@@ -1,17 +1,14 @@
...
-	<Grid>
-		<Grid.RowDefinitions>
-			<RowDefinition Height="35" />
-			<RowDefinition Height="*" />
-		</Grid.RowDefinitions>
-		<Entry Text="{Binding Entry.Title}" Placeholder="Title" Grid.Row="0" />
-		<Editor Text="{Binding Entry.BodyText}" Grid.Row="1" />
-	</Grid>
-</ContentPage>
+<local:BasePage  xmlns="http://xamarin.com/schemas/2014/forms"
```
nirinchev, line 2:
```suggestion
<local:BasePage xmlns="http://xamarin.com/schemas/2014/forms" 
```

nirinchev (approved):
LGTM - final nits, but it should be ready to go

PR #2552, opened 2021-08-10
Run the benchmark tests on PR builds

### LINKED ISSUE #2226: Use Charts for performance tests
The sync team has prepared some really nice charts tracking performance data - we should use those as inspiration and start tracking our performance test results in a similar fashion.

nirinchev:
## Description

This adds an action that takes the results.json file from a Benchmark.NET run and uploads it to MongoDB Atlas using the realm-web SDK. In the process, it does a couple of things:
1. It enhances the results with git information (i.e. commit, branch, etc.). This is not currently visualized in the Charts dashboard, but I have submitted [a request](https://feedback.mongodb.com/forums/923524-charts/suggestions/43975014-ability-to-customize-label-data) to allow us to visualize it, as well as [upvoted an existing one](https://feedback.mongodb.com/forums/923524-charts/suggestions/38670898-clickable-values) that will allow us to link to the actual commit on github.
2. It generates a charts dashboard that we can then import. This is only actually necessary when the benchmark tests change in any way, but there wasn't an easy way to do it conditionally.
3. It uploads the enhanced results to MongoDB Atlas using MongoDB Realm.

A couple of caveats:
1. The dashboard uses github's RunId instead of commit shas because [Charts doesn't allow sorting by another field](https://feedback.mongodb.com/forums/923524-charts/suggestions/43922754-allow-to-sort-series-by-a-field-that-is-not-in-a-s).
2. We're generating and saving the charts dashboard file because Charts doesn't have an API to programmatically import a dashboard. Since the expectation is for the dashboard to only change rarely, I think that's alright and we can revisit if it becomes a nuisance.
3. Charts [doesn't currently support](https://feedback.mongodb.com/forums/923524-charts/suggestions/43974894-add-error-bars-to-bar-line-charts) error bars, which means we can't plot the errors calculated by Benchmark.NET but we're storing those nonetheless so eventually we might be able to generate better graphs.
4. Right now we're plotting each benchmark on a separate graph with different arguments represented by different series. We may probably enhance that, but I didn't want to customize the dashboard too much as that would complicate generating it. If we notice we can significantly improve readability by doing something custom, we can definitely do it, but the effort needs to be justified.
5. Every time we import a dashboard, a new id gets generated, so we need to update links. Perhaps it's worth setting up a vanity url that will redirect to the correct current dashboard.

Fixes #2226

Benchmark dashboard: https://charts.mongodb.com/charts-realm-sdk-metrics-yxjvt/public/dashboards/6115babd-c7fe-47ee-836f-efffd92ffae3

##  TODO

* [ ] Delete the two dummy entries in the dashboard once we do a couple of merges to master.


papafe (commented):
Seems ok for the parts I can understand :)

LaPeste (approved):
Very good stuff! Thank you for helping with the epic.

PR #2558, opened 2021-08-17
Add MSBuild task to extract changelog at build time

### LINKED ISSUE #2050: Upload release notes in NuGet package
When building for release, we should update the nuget package with the release notes from the changelog.

LaPeste:
## Description
Add MSBuild task to extract changelog at build time. There is some reshuffling in the template of `main.yml` to accommodate the addition of the changelog to the artifacts to upload.

Fixes #2050


Tools/MSBuildTaskExtractChangelog/ExtractChangelog.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public override bool Execute()
+        {
+            try
+            {
+                var textToParse = File.ReadAllText(FilePath);
+                var regex = new Regex("(?sm)^(## \\d{1,2}\\.\\d{1,2}\\.\\d{1,2}(?:-beta\\.\\d{1,2})? \\(\\d{4}-\\d{2}-\\d{2}\\))(.+?)(?=\n## \\d{1,2}\\.\\d{1,2}\\.\\d{1,2}(?:-beta\\.\\d{1,2})? \\(\\d{4}-\\d{2}-\\d{2}\\))");
+                var matches = regex.Matches(textToParse);
+                ExtractedText = matches[0].Groups[0].Value;
+                File.WriteAllText("./ExtractedChangelog.md", ExtractedText);
```
nirinchev, line 25: Instead of writing to the file directly from the MSBuild task, a better approach would be to use another task to do that. You can use [WriteLinesToFile](https://docs.microsoft.com/en-us/visualstudio/msbuild/writelinestofile-task?view=vs-2019) in the Realm.csproj to stream the changelog contents to a file.

Realm/Realm/Realm.csproj
```diff
@@ -1,4 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <UsingTask TaskName="MSBuildReadTask.ExtractChangelog" AssemblyFile="$(ProjectDir)\..\..\Tools\MSBuildTaskExtractChangelog\MSBuildTaskExtractChangelog.dll" />
+
+  <Target Name="PrepareForRelease" BeforeTargets="GenerateNuspec">
```
nirinchev, line 5: This should probably only be done on CI, not locally.

Tools/MSBuildTaskExtractChangelog/ExtractChangelog.cs
LaPeste: The reason for this being that it's better to keep the task as focused as possible? Or is there something else?

Tools/MSBuildTaskExtractChangelog/MSBuildTaskExtractChangelog.csproj
```diff
@@ -0,0 +1,16 @@
...
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Build.Framework" Version="16.10.0" />
+    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="16.10.0" />
+  </ItemGroup>
+
+  <Target Name="CopyDllToProjDir" AfterTargets="AfterBuild">
```
LaPeste, line 12: This is simply done so that it is easier to read when imported (see line 3 of Realm/Realm/Realm.csproj)

LaPeste:
@nirinchev I converted it back to draft to add the inline msbuild task. I'll reopen when done.

.github/templates/main.yml
```diff
@@ -115,14 +115,24 @@ steps:
...
+#@ for pkgName in pkgNames:
+#@  finalPkgName = pkgName + ".${{ steps.find-nupkg-version.outputs.package_version }}" 
+#@  relativePkgPaths.append("Realm/packages/" + finalPkgName + ".*nupkg")
+#@ end
+#@ pkgNames.append("ExtractedChangelog.md")
+#@ relativePkgPaths.append("Realm/Realm/ExtractedChangelog.md")
+#@ for i in range(0, len(pkgNames)):
+#! get the filename with no extension
+#@ finalPkgName = relativePkgPaths[i][(relativePkgPaths[i].rfind("/") + 1) : len(relativePkgPaths[i])]
+#@ finalPkgName = finalPkgName[0 : finalPkgName.rfind(".")]
```
nirinchev, line 129: This seems like quite a lot of complexity to just add an extra step. Since `uploadPackageArtifacts` is only used once, it's better to just add a step after it to upload the release notes.

LaPeste: Do you prefer it in the fuction or right after it's used? I'd expect in the function.

nirinchev: Since the function is called `uploadPackageArtifacts`, I'd keep it focused on uploading only the package artifacts. We can have a final step, after `template.replace(buildUnityPackage())` that uploads the release notes. We can also extract a function for uploading artifacts - something like:

```yaml
#@ def uploadArtifacts(name, path):
  - name: #@ "Store artifacts: " + name
    uses: #@ actionUploadArtifact
    with:
      name: #@ name
      path: #@ "${{ github.workspace }}/" + path
      retention-days: ${{ github.event_name != 'pull_request' && 30 || 1 }}
#@ end
```

We can then use it in `buildUnityPackage`, `uploadPackageArtifacts` and all the other places where we upload artifacts. That can be done in a follow-up PR though, let's not bloat this one with unrelated improvements.

Realm/Realm/Realm.csproj
```diff
@@ -1,4 +1,50 @@
...
+    <Task>
+      <Using Namespace="System"/>
+      <Using Namespace="System.IO"/>
+      <Using Namespace="System.Text.RegularExpressions"/>
+      <Code Type="Fragment" Language="cs">
+<![CDATA[
+try
+{
+    var textToParse = File.ReadAllText(FilePath);
+    var regex = new Regex("(?sm)^(## \\d{1,2}\\.\\d{1,2}\\.\\d{1,2}(?:-beta\\.\\d{1,2})? \\(\\d{4}-\\d{2}-\\d{2}\\))(.+?)(?=\n## \\d{1,2}\\.\\d{1,2}\\.\\d{1,2}(?:-beta\\.\\d{1,2})? \\(\\d{4}-\\d{2}-\\d{2}\\))");
```
nirinchev, line 22: This regex is slightly different from the one in `extract-release-notes`. Can you modify it to handle other prerelease types, such as `-alpha.2`, `-pre.123`, or `-flexsync.1`.

Realm/Realm/Realm.csproj
```diff
@@ -1,4 +1,50 @@
...
+    var matches = regex.Matches(textToParse);
+    ExtractedText = matches[0].Groups[0].Value;
+}
+catch (Exception ex)
+{
+    Log.LogError($"The changelog could not be extracted because: {ex.Message}\nAborting the build.");
```
nirinchev, line 28: Might be useful to also print the stacktrace here.

nirinchev (commented):
Good stuff - two minor comments from me + can you move that task toward the bottom of `Realm.csproj` - since it's code in a CDATA, it's a bit annoying to read and is unlikely to change often, if ever, so let's shove it after the assembly info props.

Realm/Realm/Realm.csproj
```diff
@@ -1,4 +1,16 @@
...
+  <Target Name="ExtractChangelogToPack" BeforeTargets="GenerateNuspec" Condition="'$(CI)' == 'true'">
+    <Message Text="Extracting release note" />
+    <ExtractChangelog FilePath="$(ProjectDir)\..\..\CHANGELOG.md">
+      <Output TaskParameter="ExtractedText" PropertyName="PackageReleaseNotes" />
+    </ExtractChangelog>
+  </Target>
+
+  <Target Name="WriteChangelogToFile" AfterTargets="ExtractChangelogToPack">
+    <WriteLinesToFile File="ExtractedChangelog.md" Lines="$(PackageReleaseNotes)" Overwrite="true" Encoding="Unicode" />
+  </Target>
```
nirinchev, line 12: Can we move this toward the bottom as well or is it important that they're at the top?

nirinchev (approved):
LGTM - final comment from me and should be good to go after a rebase.

LaPeste:
> 
> 
> LGTM - final comment from me and should be good to go after a rebase.

A second eye on the merge from master would be appreciated.

nirinchev (approved):
Merge seems fine to me

PR #2565, opened 2021-08-18
Close realm in case of exceptions while creating schema

### LINKED ISSUE #1989: Opening realm with IsDynamic=true results in hung file when exception building schema.
## Goals

To be able to open an existing realm file in dynamic mode without the file being left open with no way to delete it in the event of an exception is thrown when building the schema dynamically.

Essentially:

	try
	{
	    var config = new RealmConfiguration("C:\path\to\realmfile") { IsDynamic = true };
	    var instance = Realms.Realm.GetInstance(config);
	}
	catch (InvalidOperationException ex)
	{
	    File.Delete("C:\path\to\realmfile");
	}

## Expected Results

When the exception from ObjectSchema.cs Build() gets thrown: 

	if (Count == 0)
	{
	    throw new InvalidOperationException(
	        "No RealmObjects. Has linker stripped them? See https://realm.io/docs/xamarin/latest/#linker-stripped-schema");
	}

I should be able to delete the file `"C:\path\to\realmfile"`

## Actual Results

Deleting results in exception due to the file still being open:

	System.IO.IOException: The process cannot access the file 'C:\path\to\realmfile' because it is being used by another process

## Steps to Reproduce

Use any realm that has an object without properties (possible with older versions of realm), then try deleting after failing to get realm instance.

## Code Sample

	try
	{
		var config = new RealmConfiguration("C:\path\to\realmfile") { IsDynamic = true };
		var instance = Realms.Realm.GetInstance(config);
	}
	catch (InvalidOperationException ex)
	{
		File.Delete("C:\path\to\realmfile");
	}

It looks like the code that needs to handle closing the file is at

`RealmConfiguration.cs:`

        internal override Realm CreateRealm(RealmSchema schema)
        {

	        // ...

            var srHandle = new SharedRealmHandle(srPtr);
            if (IsDynamic && !schema.Any())
            {
                // NOTE: exception here will leave realm file open--srHandle needs to close/dispose resources?
                srHandle.GetSchema(nativeSchema => schema = RealmSchema.CreateFromObjectStoreSchema(nativeSchema));
            }

            return new Realm(srHandle, this, schema);
        }

Visual Studio Professional 2019
16.5.4

OS: Windows 10

TargetFramework: netstandard2.0
Nuget package version: 4.3.0

papafe:
This fix closes the realm if there are exceptions while retrieving the schema from disk when opening the realm in dynamic mode. At the moment it seems that it's actually not possible to make it throw an exception like is shown in the ticket or even in other ways (at least I have not found a way). 
Given this... this is more of a protection in case something changes in the future, more than fixing an actual bug, so I am not sure if I should continue with this. 
If we decide to continue, the same try-catch needs to be put in another two places.

Fixes #1989

### DIFF
--- a/Realm/Realm/Configurations/RealmConfigurationBase.cs
+++ b/Realm/Realm/Configurations/RealmConfigurationBase.cs
@@ -170,7 +170,7 @@ namespace Realms
                 {
                     schema = sharedRealmHandle.GetSchema();
                 }
-                catch (Exception)
+                catch
                 {
                     sharedRealmHandle.Close();
                     throw;
--- a/Realm/Realm/Configurations/RealmConfigurationBase.cs
+++ b/Realm/Realm/Configurations/RealmConfigurationBase.cs
@@ -18,6 +18,7 @@
 
 using System;
 using System.IO;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Realms.Schema;
@@ -159,6 +160,24 @@ namespace Realms
                 enable_cache = EnableCache,
                 max_number_of_active_versions = MaxNumberOfActiveVersions
             };
+        }
+
+        internal Realm GetRealm(SharedRealmHandle sharedRealmHandle, RealmSchema schema)
+        {
+            if (IsDynamic && !schema.Any())
+            {
+                try
+                {
+                    schema = sharedRealmHandle.GetSchema();
+                }
+                catch (Exception)
+                {
+                    sharedRealmHandle.Close();
+                    throw;
+                }
+            }
+
+            return new Realm(sharedRealmHandle, this, schema);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


nirinchev (commented):
This makes sense to me. There are a few QoL improvements I'd like to suggest:

1. Let's have `SharedRealmHandle.Open` and `SharedRealmHandle.ResolveFromReference` return `SharedRealmHandle` rather than the raw pointer.
2. Let's move the schema lookup and `new Realm` call to a common method so that it can be reused in all 3 places where we open the Realm. I think it's all in `RealmConfigurationBase` inheritors, so we should be able to move it there.

Realm/Realm/Configurations/RealmConfigurationBase.cs
```diff
@@ -160,5 +161,23 @@ internal Native.Configuration CreateNativeConfiguration()
...
+
+        internal Realm GetRealm(SharedRealmHandle sharedRealmHandle, RealmSchema schema)
+        {
+            if (IsDynamic && !schema.Any())
+            {
+                try
+                {
+                    schema = sharedRealmHandle.GetSchema();
+                }
+                catch (Exception)
```
nirinchev, line 173: You don't need `(Exception)` here.

nirinchev (approved):
Good stuff, a minor nit from me, otherwise, it's good to go.

PR #2568, opened 2021-08-18
Enhance benchmark results with information about package sizes

### LINKED ISSUE #2225: Track binary size changes
We can use charts to upload data about the binary sizes (ideally broken down by platform/architecture). This is what Java is using: https://stats.realmlab.net/d/SiHPF44mk/java?orgId=1, but we should probably just reuse the platform we go for with the performance tests.

nirinchev:
## Description

Fixes #2225

##  TODO

* [x] Changelog entry


papafe (approved):
Looks fine

Issue #2155, opened 2020-12-08
Add code scanning to project


nirinchev:
It might be interesting to see if [CodeQL](https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/enabling-code-scanning-for-a-repository) finds any issues with our code. Unfortunately, the default Actions workflow doesn't work because it can't build the wrappers automatically. Once we move building things on GA, we should look into enabling it.

PR #2569, opened 2021-08-18
Try to setup codeql

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2155

##  TODO

* [x] Changelog entry
* [ ] Tests (if applicable)


DominicFrei:
Is there a way to re-use one of the builds that we use to run tests to also run the analysis? I noticed the CodeQL steps take quite long and most of it is actually building the wrappers / project itself.

And what are our plans on how to use that? Will it fail the CI build if there is a warning?

nirinchev:
The reason I set it up as a separate workflow is that code analysis significantly increases build times - it's not that wrapper builds are slow, it's wrapper builds being analyzed for quality that are. If we did it as part of the regular PR workflow, it would roughly triple the time it takes to build wrappers, which will then delay the package build and tests.

Regarding plans for using it - we'll see, I don't have a plan yet, but ideally, we should surface code quality issues and address them as part of the PR review process. A lot of these are either false positives or things happening in Core, so we'll just dismiss them, but if there are issues uncovered in our code, we should probably fix them.

papafe (approved):

PR #2572, opened 2021-08-19
Fix some CodeQL warnings

nirinchev:
## Description

This fixes all not ignored CodeQL warnings. They are mostly trivial things and not something we'd hit in reality, but it's still a good idea to follow good practices + they did force me to redo the code for snapshotting collections which was pretty obnoxious previously.


Realm/Realm/Configurations/SyncConfiguration.cs
```diff
@@ -34,6 +33,10 @@ namespace Realms.Sync
     /// <seealso href="https://docs.mongodb.com/realm/sync/overview/">Sync Overview Docs</seealso>
     public class SyncConfiguration : RealmConfigurationBase
     {
+        // TODO: https://github.com/realm/realm-dotnet/issues/2571 - we should update the public property type to be RealmValue and
+        // get rid of this field.
+        internal readonly RealmValue _partition;
```
nirinchev, line 38: We were storing Partition as an object, making it harder to do comparisons between partitions of different types. This addresses https://github.com/realm/realm-dotnet/security/code-scanning/621?query=ref%3Arefs%2Fheads%2Fmaster

Realm/Realm/Configurations/SyncConfiguration.cs
```diff
@@ -125,17 +128,17 @@ public SyncConfiguration(ObjectId? partition, User user, string optionalPath = n
         /// </param>
         [SuppressMessage("Design", "CA1062:Validate arguments of public methods", Justification = "Arguments are validated in the private ctor.")]
         public SyncConfiguration(Guid? partition, User user, string optionalPath = null)
-            : this((object)partition, user, optionalPath)
+            : this(user, partition, optionalPath)
         {
         }
 
-        private SyncConfiguration(object partition, User user, string path)
+        private SyncConfiguration(User user, RealmValue partition, string path)
```
nirinchev, line 135: The change in ordering was done so that we could use the implicit conversion operators from the other constructors.

Realm/Realm/Handles/CollectionHandleBase.cs
```diff
@@ -24,13 +24,9 @@ namespace Realms
 {
     internal abstract class CollectionHandleBase : NotifiableObjectHandleBase
     {
-        protected delegate IntPtr SnapshotDelegate(out NativeException ex);
-
         public abstract bool IsValid { get; }
 
-        public bool CanSnapshot => SnapshotCore != null;
-
-        protected virtual SnapshotDelegate SnapshotCore => null;
```
nirinchev, line 33: This refactors `SnaphotCore` so that it's not a virtual property but rather a method being called by `Snapshot`. Addresses https://github.com/realm/realm-dotnet/security/code-scanning/606?query=ref%3Arefs%2Fheads%2Fmaster

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -233,7 +234,7 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
             }
             catch (Exception ex)
             {
-                Console.WriteLine(ex.Message);
+                Logger.Default.Log(LogLevel.Warn, $"An error has occurred while handling a session error: {ex}");
```
nirinchev, line 237: This is a drive-by - it should have used the logger rather than Console.WriteLine-ing.

Realm/Realm/Linq/RealmResultsVisitor.cs
```diff
@@ -553,14 +553,19 @@ protected override Expression VisitBinary(BinaryExpression node)
                     memberExpression = leftExpression as MemberExpression;
                 }
 
+                if (memberExpression == null)
```
nirinchev, line 556: We've never encountered this in practice, but technically, there was no code preventing a case where `memberExpression` could be `null`.

Realm/Realm/Thread Handover/ThreadSafeReference.cs
```diff
@@ -104,7 +106,8 @@ public static Object<T> Create<T>(T value)
         /// </param>
         /// <typeparam name="T">The type of the objects contained in the list.</typeparam>
         /// <returns>A <see cref="ThreadSafeReference"/> that can be passed to <see cref="Realm.ResolveReference{T}(List{T})"/> on a different thread.</returns>
-        public static List<T> Create<T>(IList<T> value) => new List<T>(value);
+        public static List<T> Create<T>(IList<T> value)
+            => new List<T>(Argument.EnsureType<RealmList<T>>(value, "value must be a managed List (i.e. property of a RealmObject).", nameof(value)));
```
nirinchev, line 110: This change replaces an InvalidCastException that would have been thrown when we attempt to cast `value` to the correct type with an `ArgumentException` that contains more actionable message.

.github/templates/main.yml
```diff
@@ -44,21 +44,10 @@ steps:
 #@ end
 
 #@ def setupWin81SDK():
-  - name: Check Win8.1 SDK Cache
```
nirinchev, line 47: This is unrelated to the fixes - I just noticed that downloading the SDK takes a few seconds, so caching is not really helpful

Realm/Realm/DataBinding/WovenSetterMethodInfo.cs
```diff
@@ -60,23 +60,19 @@ public WovenSetterMethodInfo(MethodInfo mi)
...
-            }
 
-            var result = _mi.Invoke(obj, invokeAttr, binder, parameters, culture);
-
-            if (writeTransaction != null)
+            // If managingRealm is null or is currently in transaction, just set the property
+            if (managingRealm?.IsInTransaction == false)
             {
-                writeTransaction.Commit();
-                writeTransaction.Dispose();
```
nirinchev, line 74: There was a small risk that something before the dispose would throw, thus never reaching the actual dispose call.

Realm/Realm/Handles/AppHandle.cs
```diff
@@ -256,18 +252,18 @@ public void Reconnect()
             NativeMethods.reconnect(this);
         }
 
-        public bool TryGetCurrentUser(out SyncUserHandle handle)
```
nirinchev, line 259: Changes like this one are to avoid shadowing the private `handle` field.

Realm/Realm/Handles/ObjectHandle.cs
```diff
@@ -115,26 +113,9 @@ public ObjectHandle(RealmHandle root, IntPtr handle) : base(root, handle)
         {
         }
 
-        public override bool Equals(object obj)
+        public bool ObjEquals(ObjectHandle other)
```
nirinchev, line 116: This was replaced as it wasn't correct to override `ObjectHandle.Equals` as we're not really checking for equality of the ObjectHandle struct, but rather the value that it represents (e.g. two different ObjectHandle instances pointing to the same underlying object are not necessarily "equal" as they may have different unbind lists and parents).

Realm/Realm/Realm.cs
```diff
@@ -1334,9 +1332,9 @@ public void WriteCopy(RealmConfigurationBase config)
         {
             Argument.NotNull(config, nameof(config));
 
-            if (Config is SyncConfiguration originalConfig && config is SyncConfiguration copiedConfig && originalConfig.Partition != copiedConfig.Partition)
+            if (Config is SyncConfiguration originalConfig && config is SyncConfiguration copiedConfig && originalConfig._partition != copiedConfig._partition)
```
papafe, line 1335: I understand why it was done, it just looks a little funny with the underscore in front of the property.

papafe (approved):
Overall it looks good, and it's always nice to have a little bit less lines of code ;)

Realm/Realm/Realm.cs
nirinchev: It's because that's a field, not the actual property. Once we release a new major version, we can replace the old property with the field and that will look a lot nicer.

PR #2575, opened 2021-08-23
Fixed InvaliOperationException when working with WPF

### LINKED ISSUE #1903: InvalidOperationException when removing databound object
## Goals
I have a WPF application in which I am binding a ListView to a collection of RealmObjects. My goal is to remove one of these objects from the database

## Expected Results
I expected the object to be successfully removed and the bound ListView to update.

## Actual Results
System.InvalidOperationException: 'Added item does not appear at given index '1'.'. No stack trace is given. On enabling native code debugging the exception seems to be occuring in realm-wrappers.dll. Slightly misleading as the debug output shows the exception occurs in PresentationFramework.dll

## Steps to Reproduce
Bind a ListView to a collection of RealmObjects. Remove an object from the realm instance and you will cause an InvalidOperationException

## Code Sample
I've created a repository with a couple of [test projects](https://github.com/carlst99/RealmTester) highlighting the issue. Add a few blogs and then select and try to delete one; it should crash with the error. 

Note that the first test project mimics my use-case scenario, hence the libraries Stylet and MaterialDesignThemes

* realm-dotnet 4.1.0, local DB
* Have tested on .NET Framework 4.6.1, 4.8
* Windows 10

papafe:
This PR fixed an issue where WPF throws an exception while a collection is bound the the UI and an object is removed from realm. This is due to the [internals](https://github.com/dotnet/wpf/blob/3a7e2a77d41d2e3ec68dc7ad32c60d7df88ab49f/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/EnumerableCollectionView.cs#L430-L431) of WPF, where it checks for equality between the OldItem in the notification and something else. In our case, because the object does not exist anymore, OldItem is null, so the equality is false and it throws an exception (with the wrong text too though).
In order to solve this issue we decided to create an InvalidRealmObject, that will be used to simulate the removed object that we do not have anymore. This class has an overridden Equals method that will always return true in order to go around the issue with WPF.

Fixes #1903 

##  TODO

* [ ] Changelog entry


papafe:
This is still a draft because I want to see if there are issues with CI

Realm/Realm/DatabaseTypes/InvalidRealmObject.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Special invalid realm object that is used to avoid an exception in WPF
+    /// when deleting an element from a collection bound to UI (<see href="https://github.com/realm/realm-dotnet/issues/1903">#1903</see>).
+    /// </summary>
+    internal class InvalidRealmObject : RealmObjectBase
+    {
+        private static readonly InvalidRealmObject _instance = new InvalidRealmObject();
+
+        private InvalidRealmObject() { }
+
+        public static InvalidRealmObject Instance
```
DominicFrei, line 31: Would

```
public static InvalidRealmObject Instance => _instance;
```

work here too?

DominicFrei (commented):
I added a question more out of interest than anything.

Can't see anything wrong with it. 👍

Realm/Realm/DatabaseTypes/InvalidRealmObject.cs
papafe: Absolutely, don't know why I wrote it like that :)

Realm/Realm/DatabaseTypes/InvalidRealmObject.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        private InvalidRealmObject() { }
+
+        public static InvalidRealmObject Instance => _instance;
+
+        // The method is overriden to avoid the bug in WPF
+        public override bool Equals(object obj)
+        {
+            return true;
+        }
```
nirinchev, line 37: Let's be a bit more restrictive with the equality check here:

```suggestion
        public override bool Equals(object obj) => obj is RealmObjectBase robj && !robj.IsValid;
```

Realm/Realm/DatabaseTypes/InvalidRealmObject.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// Special invalid realm object that is used to avoid an exception in WPF
+    /// when deleting an element from a collection bound to UI (<see href="https://github.com/realm/realm-dotnet/issues/1903">#1903</see>).
+    /// </summary>
+    internal class InvalidRealmObject : RealmObjectBase
+    {
+        private static readonly InvalidRealmObject _instance = new InvalidRealmObject();
+
+        private InvalidRealmObject() { }
+
+        public static InvalidRealmObject Instance => _instance;
```
nirinchev, line 31: This is a prop, so should be above the ctor. You also don't need a backing field for a getter-only property:

```suggestion
        public static InvalidRealmObject Instance { get; } = new InvalidRealmObject();
```

Realm/Realm/DatabaseTypes/InvalidRealmObject.cs
```diff
@@ -0,0 +1,39 @@
...
+// //
+// ////////////////////////////////////////////////////////////////////////////
+
+namespace Realms
+{
+    /// <summary>
+    /// Special invalid realm object that is used to avoid an exception in WPF
+    /// when deleting an element from a collection bound to UI (<see href="https://github.com/realm/realm-dotnet/issues/1903">#1903</see>).
+    /// </summary>
+    internal class InvalidRealmObject : RealmObjectBase
```
nirinchev, line 25: The GetHashCode warning is legit - I'm not sure we have good options here. The `GetHashCode` for `RealmObjectBase` will cache the value for the hash code, meaning it'll have some non-zero value. Let's hope there's no UI framework out there that's putting objects in a hashset or we'd be in trouble. For now, what you can do to dismiss the warning is to just return `0` from `GetHashCode`.

Realm/Realm/DatabaseTypes/InvalidRealmObject.cs
```diff
@@ -0,0 +1,39 @@
+﻿// ////////////////////////////////////////////////////////////////////////////
```
nirinchev, line 1: Double commenting here.

nirinchev (changes requested):

nirinchev:
On a second look, it seems like `InvalidRealmObject.IsValid` will return `true`, although we probably want it to return `false`. It will likely require a bit of fiddling with the internals of `RealmObjectBase` - likely we'd need to introduce a separate internal virtual property to include in the `IsValid` check as we definitely don't want to turn `IsValid` virtual.

Realm/Realm/DatabaseTypes/InvalidRealmObject.cs
papafe: Regarding this, I was trying to be less restrictive because we use this in `RealmCollectionBase<T>`, and we don't know if `T` is actually a `RealmObjectBase`. Probably, because of this, we don't even need to have this fake object derive from RealmObjectBase and just use a completely independent filler object. What do you think?

nirinchev: Ugh... I see. Yes, then let's make it a completely independent object. That being said, we should probably check customize `Equals` on `RealmObjectBase` and make sure to return `true` if `other` is `InvalidRealmObject` and `this.IsValid == false`.

papafe: I have just pushed a commit that makes this a separate object. Regarding Equals... I understand the sentiment, but do you think we should go the extra mile for something that is a patch for a bug in WPF and has a very limited scope?

nirinchev: I guess the issue is that there might be other controls out there that do the inverse check - i.e. `original.Equals(deleted)` which will return `false`. While there's not much we can do about non-object types, we can still make objects behave as expected.

papafe:
Finished with the requested changes, this is mergeable for me.

nirinchev (approved):

PR #2577, opened 2021-08-23
Add schema cache

### LINKED ISSUE #2194: Consider caching the schema
We're using reflection to build the schema every time a Realm is opened. We should consider keeping a `<Type, Schema>` dictionary to avoid paying the reflection penalty. We should add a benchmark test and see if caching the schema has a "significant enough" effect to warrant doing it.

nirinchev:
## Description

Fixes #2194

This will not have massive performance effect for most users as we expect most people to just use the default schema, which is already cached. But for people who do set `ObjectClasses`, it should speed things up nicely.

### Before the change

|                                                                              Method |      Mean |     Error |    StdDev |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------------------------------------------------------------------------------ |----------:|----------:|----------:|--------:|-------:|------:|----------:|
| CreateSchema |  3.897 ms | 0.0313 ms | 0.0261 ms | 78.1250 | 7.8125 |     - |    859 KB |
| OpenRealm | 16.567 ms | 0.2680 ms | 0.2375 ms | 62.5000 |      - |     - |    898 KB |

### After the change

|                                                                              Method |          Mean |       Error |      StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------------------------------------------------------------------ |--------------:|------------:|------------:|-------:|------:|------:|----------:|
| CreateSchema |      2.128 us |   0.0090 us |   0.0075 us | 0.1793 |     - |     - |      2 KB |
| OpenRealm | 12,254.235 us | 230.4506 us | 215.5636 us |      - |     - |     - |     40 KB |

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   * `list.Filter(query, arguments)` will filter the list and return the filtered collection. It is roughly equivalent to `list.AsRealmQueryable().Filter(query, arguments)`.
 
   The resulting queryable collection will behave identically to the results obtained by calling `realm.All<T>()`, i.e. it will emit notifications when it changes and automatically update itself. (Issue [#1499](https://github.com/realm/realm-dotnet/issues/1499))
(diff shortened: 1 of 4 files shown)


papafe (approved):
Looks good

PR #2578, opened 2021-08-23
Don't try to serialize the `DynamicApi` property

nirinchev:
## Description

This replaces `[XmlIgnore]` with `[IgnoreDataMember]` as that works for both the Xml and Json serializer. It removes `[Serializable]` as that was only used in the `BinarySerializer` which is very strongly obsoleted.

Finally, it annotates the RealmCollectionBase properties with `[IgnoreDataMember]` to make sure a json serializer doesn't try to touch them and adds tests for both Xml and Json serialization.

A couple of caveats:
1. Newtonsoft.Json doesn't work on Unity out of the box, so we only test on the other platforms.
2. MongoDB.Bson and XmlSerializer don't serialize collections by default, so we're not validating it.

Fixes https://github.com/realm/realm-dotnet/issues/2579

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
+++ b/Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
@@ -25,6 +25,8 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Runtime.Serialization;
+using System.Xml.Serialization;
 using Realms.Exceptions;
 using Realms.Helpers;
 using Realms.Schema;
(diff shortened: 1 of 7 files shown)


Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -39,17 +41,20 @@ public abstract class RealmCollectionBase<T>
           IThreadConfined,
           IMetadataObject
     {
+        private readonly List<NotificationCallbackDelegate<T>> _callbacks = new List<NotificationCallbackDelegate<T>>();
```
nirinchev, line 44: Unfortunately, this change also reorders the members of `RealmCollectionBase` because they were all over the place. Apart from the reordering, the only other change is the addition of the attributes.

Realm/Realm/DatabaseTypes/RealmObjectBase.cs
```diff
@@ -40,29 +41,22 @@ namespace Realms
...
     /// </summary>
     [Preserve(AllMembers = true)]
-    [Serializable]
     public abstract class RealmObjectBase
         : INotifyPropertyChanged,
           IThreadConfined,
           INotifiable<NotifiableObjectHandleBase.CollectionChangeSet>,
           IReflectableType
     {
-        [NonSerialized, XmlIgnore]
```
nirinchev, line 50: The Xml serializer ignores the private/internal fields anyway.

fealebenpae (approved):

PR #2589, opened 2021-08-28
Upgrade to latest Core

nirinchev:
Pulls in https://github.com/realm/realm-core/pull/4839 and makes a minor change in shared_realm_cs.cpp to fix the compilation.


Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -205,7 +205,7 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
         {
             try
             {
-                var handle = new SessionHandle(sessionHandlePtr);
+                using var handle = new SessionHandle(sessionHandlePtr);
```
nirinchev, line 208: This was causing issues since it was extending the lifetime of the session in the event of an error, thus preventing the Realm from being deleted. It has always been a problem, but the Core changes exposed altered the timing, exposing it in our unit tests.

Realm/Realm/Handles/AppHandle.cs
```diff
@@ -232,7 +232,7 @@ public string GetRealmPath(User user, string partition)
             });
         }
 
-        public SessionHandle GetSessionForPath(SharedRealmHandle realm)
+        public SessionHandle GetSessionForRealm(SharedRealmHandle realm)
```
nirinchev, line 235: I noticed this while looking for other cases where we might not be disposing properly of the sync session. The name was a leftover from when it had a string argument and figured it no longer reflected the current signature.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -444,21 +444,25 @@ public void DeleteRealmWorksIfCalledMultipleTimes()
         }
 
         [Test]
-        public void DeleteRealm_afterSessionDisposed([Values(true, false)] bool singleTransaction)
+        public void DeleteRealm_AfterDispose_Succeeds([Values(true, false)] bool singleTransaction)
```
nirinchev, line 447: This test name was not following our naming guidelines and was a bit unclear. The new name is not much better, so I added some comments in the test body to clarify it.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -117,7 +117,6 @@ Realm::Config get_shared_realm_config(Configuration configuration, SyncConfigura
         auto& key = *reinterpret_cast<std::array<char, 64>*>(encryption_key);
 
         config.encryption_key = std::vector<char>(key.begin(), key.end());
-        config.sync_config->realm_encryption_key = key;
```
nirinchev, line 120: Core removed `realm_encryption_key` from `sync_config`, so this is just getting the build to pass.

nirinchev (commented):
@papafe I know how much you enjoy reviewing PRs with a bunch of disconnected changes, but in my mind they were vaguely related 😅

papafe (approved):
Looks good to me!

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
DominicFrei: Is the `succeeds` part here redundant? I assume it refers to the deletion, not the test itself but would that mean we should always add `succeeds` and `fails` to all tests to indicate of they test a positive or a negative case?

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -444,21 +444,25 @@ public void DeleteRealmWorksIfCalledMultipleTimes()
...
 
-                using var realm = GetRealm(config);
+                var realm = GetRealm(config);
                 AddDummyData(realm, singleTransaction);
 
                 await WaitForUploadAsync(realm);
                 realm.Dispose();
 
+                // Ensure that the Realm can be deleted from the filesystem. If the sync
+                // session was still using it, we would get a permission denied error.
```
DominicFrei, line 465: Would we want to test for this `PermissionDeniedError` somewhere?

DominicFrei (commented):
No changes for me, just added some questions I was curious about.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
nirinchev: Generally, I've been trying to name tests like *method*_*when-something*_*result*. I guess here a more precise name would have been `DeletesFiles` rather than `Succeeds` - I guess I was a little lazy when coming up with the name. That being said, our tests are a bit of a mixed bag when it comes to naming conventions, so it doesn't matter that much.

DominicFrei: 👍

PR #2590, opened 2021-08-30
Disable analytics on CI

nirinchev:
## Description

Most CI services will set the environment variable `CI` to some value, so it's a good idea to check for it and make sure we don't count CI builds as actual developers.

* [Github Actions](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables)
* [Travis CI](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables)
* [Appveyor](https://www.appveyor.com/docs/environment-variables/)
* [Jenkins](https://github.com/jenkinsci/jenkins/pull/5370)

Additionally, this cleans up our workflow to set `REALM_DISABLE_ANALYTICS` statically on the entire workflow rather than per at runtime for each job, similarly to https://github.com/realm/realm-js/pull/3918.

### DIFF
--- a/Tests/Tests.UWP/FodyWeavers.xml
+++ b/Tests/Tests.UWP/FodyWeavers.xml
@@ -1,3 +1,3 @@
 ﻿<Weavers xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="FodyWeavers.xsd">
-  <Realm />
+  <Realm DisableAnalytics="true"/>
 </Weavers>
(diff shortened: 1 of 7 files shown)


Tests/Tests.UWP/FodyWeavers.xml
```diff
@@ -1,3 +1,3 @@
 ﻿<Weavers xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="FodyWeavers.xsd">
```
LaPeste, line 1: Can you refresh my memory on what this file is supposed to be for and why only the UWP tests have it?

nirinchev: All tests have it, we must have forgotten to disable analytics on UWP which is why this is the only one that shows up as changed. The other tests pull the FodyWeavers.xml from the common Realm.Tests project because they're using the new csproj format, whereas the uwp project hasn't been upgraded or just doesn't support it.

PR #2595, opened 2021-08-31
Change test reporting gh action to the chosen one

### LINKED ISSUE #2531: Look for another Github action for reporting test results
The current github action used to visualize the report from the test results from the github workflow is not very good:

1. It's written in python, so it's more annoying than JS/TS to modify, given the team's knowledge
1. It keeps having a weird where the test result is thrown around in random workflows
1. no stack trace
1. and a few more...

It's a good idea to search for a better replacement. The final decision will be made by the team as a whole.

LaPeste:
## Description
Given the conversation had around the shortcomings of the [current](https://github.com/EnricoMi/publish-unit-test-result-action) gh action to report the test results, this PR is trying out a [different one](https://github.com/dorny/test-reporter) which was agreed upon.
Unfortunately, the chosen action showed to have some bugs around handling `<failure>`s without a stacktrace and the processing of the `message` attribute of `<failure>`. This prompt me to fork the project where to make the necessary fixes. Because of this, our current `main.yml` workflow points to the forked version instead of the original one. If and when we decide to publish a PR in the [original repo](https://github.com/dorny/test-reporter), we'll change what our main workflow points to.
The forked version can be found in [my gh repos](https://github.com/LaPeste/test-reporter).

Fixes #2531


.github/templates/main.yml
```diff
@@ -187,12 +187,14 @@ steps:
 
 #@ def publishTestsResults(files, test_title):
 name: Publish Unit Test Results
-uses: EnricoMi/publish-unit-test-result-action/composite@v1
+uses: dorny/test-reporter@v1
```
DominicFrei, line 190: Would it make sense to fork it if it's a fixed version anyway to make sure this doesn't start failing if the repo doesn't exist anymore?

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -46,6 +46,7 @@ public void AsyncWrite_ShouldExecuteOnWorkerThread()
 
                 Assert.That(_realm.All<Person>().Count(), Is.EqualTo(1));
                 Assert.That(otherThreadId, Is.Not.EqualTo(currentThreadId));
+                Assert.That(false);
```
DominicFrei, line 49: Is this PR still a draft and you use this to test the new GH action or is there some deeper meaning behind that line that yet have to understand? :)

DominicFrei (commented):
No changes, just questions.

.github/templates/main.yml
LaPeste: That's not a fixed version. It'll take any 1.x.x as soon as it's updated. Hence, forking that would mean we manually integrate all the updates. I think that forking that makes sense only the moment that it disappears, with say something like wayback machine. If that would not be possible, there surely are at least 3 to 4 more gh actions for that specific  task. And lastly, I'm sure someone else has already forked it and will likely become the "new de-facto" replacement to the one taken down.

Tests/Realm.Tests/Database/AsyncTests.cs
LaPeste: It's still a draft but I opened it by mistake as "ready for review" and until you push again you can't set it as draft (at least so I think). Then after pushing, I forgot to set it to "draft". I am just testing how error reporting looks.

.github/templates/main.yml
nirinchev: I don't think we should be forking 3rd party actions. However, we should fix the commit sha rather than the version to protect against a malicious user taking over the 3rd party repo.

Tests/Realm.Tests/Database/AsyncTests.cs
DominicFrei: Ahhh, gotcha! Just wanted to make sure I understand everything that's going on. :)

.github/templates/main.yml
DominicFrei: > 
> 
> That's not a fixed version. It'll take any 1.x.x as soon as it's updated. Hence, forking that would mean we manually integrate all the updates. I think that forking that makes sense only the moment that it disappears, with say something like wayback machine. If that would not be possible, there surely are at least 3 to 4 more gh actions for that specific task. And lastly, I'm sure someone else has already forked it and will likely become the "new de-facto" replacement to the one taken down.

Totally agree. I only brought it up because it happened to a project I've been part of in the past and it happened during a time where we had a lot of other stress going on when all of a sudden a library that we used to use disappeared. That's where we started to fork things we use to make sure we don't loose dependencies the day before the release. :D

nirinchev:
@LaPeste I pushed a commit that might fix the message parsing for junit as that has annoyed me both with the previous and this action. If it works, you might want to submit an upstream PR with that change as well (or I can do it).

Tests/Tests.UWP/MainPage.xaml.cs
```diff
@@ -60,6 +60,8 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
                     }
 
                     _nunit.Options.ResultFilePath = Path.Combine(ApplicationData.Current.LocalFolder.Path, resultPath);
+                    var sfFileToCopy = await ApplicationData.Current.LocalFolder.GetFileAsync(resultPath);
+                    var sfOriginalFile = await sfFileToCopy.CopyAsync(ApplicationData.Current.LocalFolder, "nonConvertedTestResults.xml", NameCollisionOption.ReplaceExisting);
```
nirinchev, line 64: Do we need this anymore?

.github/templates/main.yml
```diff
@@ -193,12 +193,15 @@ steps:
...
 with:
-  files: #@ files
-  comment_mode: "off"
-  check_name: #@ "Results " + test_title
+  name: #@ "Results " + test_title
+  path: #@ files
+  reporter: java-junit
+  list-suites: failed
+  list-tests: failed
+  path-replace-backslashes: true
```
nirinchev, line 204: Can we set `fail-on-error: false` to make sure we don't get 2 failures per failed test.

Tests/Tests.UWP/MainPage.xaml.cs
LaPeste: I actually forgot to fix this. This make the test fail because it should be on line 67. Considering that these 2 lines of code aren't only helpful for CI but also for the local creation and inspection of the test result, this could be handy for debugging the nito async issue in the stacktrace. Of course, they can be readded to a future branch that's specif to this nito issue.
Let me know if I should remove it.

nirinchev: I don't think we need it - it's fairly easy to get to the xml by finding the app install directory.

LaPeste: I'm a bit confused. That code isn't just to copy the file. Since the transformation replaces the original file, the code makes a copy of the original xml, so that we have both: the converted and the original.
But I'll remove it, since it's easy to re-add.

.github/templates/main.yml
LaPeste: > 2 failures per failed test  
By this, do you mean both the test run and the generated report of the test run?

.github/templates/main.yml
```diff
@@ -367,23 +371,24 @@ jobs:
         shell: powershell
       - name: Run the tests
         run: |
-          echo "test_results=$env:LOCALAPPDATA/Packages/$(get-appxpackage -name realm.uwp.tests | select -expandproperty PackageFamilyName)/LocalState/TestResults.UWP.xml" | Out-File $env:GITHUB_ENV -Encoding utf8 -Append
+          echo "test_results=$env:LOCALAPPDATA\Packages\$(get-appxpackage -name realm.uwp.tests | select -expandproperty PackageFamilyName)\LocalState\TestResults.UWP.xml" | Out-File $env:GITHUB_ENV -Encoding utf8 -Append
```
LaPeste, line 374: Backslashes in place of forwardslashes is the outcome of testing with the new test reporting GH action. It's simply done to have more consistently looking paths. I did not want the action to look at paths like `C:\bd\s\e/g/u/k`; since I set the option of `path-replace-backslashes: true`, I didn't want to risk to confuse the GH action on what to do in mixed path. And while I could have gone to look at the source code to see how the forwardslashes are replaced when that option is on, I don't see the point in having mixed path anyway.

TL;DR: Backslashes instead of forwardslashes is mainly done for consistent style given the fact that we're in powershell. This avoids potential issues with the new GH action for reporting test result.

nirinchev (approved):

PR #2600, opened 2021-08-31
Quick fixes for issues reported by CodeQL

nirinchev:
## Description

1. Bumps the version
2. Updates changelog with coveralls entry
3. Removes [Serializable] from `RealmObjectBase` inheritors
4. Fixes code style violation for `InvalidObject`


Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -534,9 +534,11 @@ internal interface IRealmCollectionBase<THandle> : IMetadataObject
     /// when deleting an element from a collection bound to UI (<see href="https://github.com/realm/realm-dotnet/issues/1903">#1903</see>).
     /// </summary>
     [SuppressMessage("StyleCop.CSharp.MaintainabilityRules", "SA1402:File may only contain a single type", Justification = "This is a special object that has a very limited meaning in the project.")]
-    internal class InvalidObject
+    internal sealed class InvalidObject
```
DominicFrei, line 537: Are those Equals / GetHashCode warnings something we should take care of? (it seems to one of those that are old anyway but I was curious)

nirinchev: Nope, we can't provide a meaningful GetHashCode() override unfortunately.

DominicFrei: Is this something we would want to ignore then (not sure if that's easily possible) to only have actionable items in those reports?

nirinchev: Yeah, I've ignored it - the warning has a `Dismiss` button that you can use to make it go away. Unfortunately, if the code surrounding the warning changes, it will pop up again, but I haven't found a way around that.

PR #2609, opened 2021-09-06
Wrap up the release workflow

nirinchev:
## Description

This should wrap up our release workflow. After publishing to Nuget/npm, it will merge the release PR, publish the Github release, tag master, then create a new PR, updating the changelog and merge that in.

##  TODO

* [x] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/.github/templates/common.lib.yml
+++ b/.github/templates/common.lib.yml
@@ -24,6 +24,8 @@
 #@ actionFindPR = "juliangruber/find-pull-request-action@f9f7484f8237cf8485e5ab826e542ba5dd9e9c6e" #! 1.5.0
 #@ actionMergePR = "juliangruber/merge-pull-request-action@8a13f2645ad8b6ada32f829b2fae9c0955a5265d" #! 1.0.6
 
+#@ mainBranch = "master"
+
 #@ def checkoutCode(submodules=False, registerProblemMatchers=True):
   - name: Checkout code
     uses: #@ actionCheckout
--- a/.github/templates/common.lib.yml
+++ b/.github/templates/common.lib.yml
@@ -20,6 +20,9 @@
 #@ actionDeleteArtifact = "geekyeggo/delete-artifact@56e063d7d8bf9972ac54aca4454d3a6675917f44" #! v1
 #@ actionGithubRelease = "ncipollo/release-action@10c84d509b28aae3903113151bfd832314964f2e" #! v1.8.8
 #@ actionDownloadAllArtifacts = "dawidd6/action-download-artifact@d0f291cf39bd21965ea9c4c6e210fc355c3844ed" #! v2.14.1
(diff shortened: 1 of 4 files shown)


.github/templates/publish-release.yml
```diff
@@ -79,11 +142,14 @@ jobs:
...
         $version = $Matches[0]
         echo "::set-output name=version::$version"
-    - #@ template.replace(publishGithubRelease())
+    - #@ template.replace(prepareGithubRelease())
     - #@ template.replace(uploadDocsToS3())
     - #@ template.replace(uploadToNuget("Realm.Fody"))
     - #@ template.replace(uploadToNuget("Realm"))
     - #@ template.replace(uploadToNpm("io.realm.unity"))
+    - #@ template.replace(mergeReleasePR())
+    - #@ template.replace(publishGithubRelease())
```
LaPeste, line 151: Is there a specific reason why we need to make a draft of the release first (line 145) and then publish after all the other actions (line 146-149) are done?

nirinchev: No reason - I guess I thought it might be a good idea to have the release prepared if any of the following steps fails, but when I think about it, I'm not sure it adds that much value. I'll rework.

.github/templates/common.lib.yml
```diff
@@ -20,6 +20,11 @@
 #@ actionDeleteArtifact = "geekyeggo/delete-artifact@56e063d7d8bf9972ac54aca4454d3a6675917f44" #! v1
 #@ actionGithubRelease = "ncipollo/release-action@10c84d509b28aae3903113151bfd832314964f2e" #! v1.8.8
 #@ actionDownloadAllArtifacts = "dawidd6/action-download-artifact@d0f291cf39bd21965ea9c4c6e210fc355c3844ed" #! v2.14.1
+#@ actionCreatePR = "peter-evans/create-pull-request@7380612b49221684fefa025244f2ef4008ae50ad" #! 3.10.1
+#@ actionFindPR = "juliangruber/find-pull-request-action@f9f7484f8237cf8485e5ab826e542ba5dd9e9c6e" #! 1.5.0
+#@ actionMergePR = "juliangruber/merge-pull-request-action@8a13f2645ad8b6ada32f829b2fae9c0955a5265d" #! 1.0.6
```
LaPeste, line 25: Given the fact that you could write a more readable reference to an action version like `juliangruber/merge-pull-request-action@1.0.6`, are you using the hash of the commit because an attacker could change the tag to point to a "reworked" version of the action?

(This comment is also for all the above lines.)

nirinchev: Yes - this is the recommendation here: https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions

Writing the more readable version means that an attacker who gains access to the action repo can push a compromised version of the action and re-tag it with an existing tag. Since commit shas are not user-definable, it would be harder to generate a sha collision.

LaPeste (approved):

PR #2612, opened 2021-09-08
Minor refactoring of uploading artifacts in main.yml template

LaPeste:
## Description
Just a minor refactoring in the template of main.yml where all the artifacts uploads now use a function instead of writing the same code over and over. I tried to make the function simple enough, so more complicated uploads can't use this function but needs to write the upload code by hand. I don't think we should over complicate the helper as what it does covers already 90% of our use cases.

Fixes: no issue was created

### DIFF
--- a/.github/templates/main.yml
+++ b/.github/templates/main.yml
@@ -28,20 +28,11 @@ with:
   key: #@ key
 #@ end
 
-#@ def storeWrapperBinaries(outputVar):
-name: Store artifacts
+#@ def uploadArtifacts(artifactName, relPath, retentionDays = "${{ github.event_name != 'pull_request' && 30 || 1 }}"):
+name: #@ "Store artifacts for " + artifactName
 uses: #@ actionUploadArtifact
 with:
-  name: #@ outputVar
-  path: wrappers/build/**
-  retention-days: 1
-#@ end
-
-#@ def uploadArtifacts(titleName, uploadName, relPath, retentionDays = "${{ github.event_name != 'pull_request' && 30 || 1 }}"):
-name: #@ "Store artifacts for " + titleName
-uses: #@ actionUploadArtifact
-with:
-  name: #@ uploadName
+  name: #@ artifactName
   path: #@ "${{ github.workspace }}/" + relPath
   retention-days: #@ retentionDays
 #@ end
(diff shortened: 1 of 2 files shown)


.github/templates/main.yml
```diff
@@ -37,6 +37,15 @@ with:
   retention-days: 1
 #@ end
 
+#@ def uploadArtifacts(titleName, uploadName, relPath, retentionDays = "${{ github.event_name != 'pull_request' && 30 || 1 }}"):
```
nirinchev, line 40: We can use `uploadName` instead of `titleName` - it's probably not going to be as visually appealing (e.g. `Store artifacts for dashboard.charts` rather than `Store artifacts for Charts Dashboard`), but that seems like a minor inconvenience.

.github/templates/main.yml
```diff
@@ -37,6 +37,15 @@ with:
   retention-days: 1
```
nirinchev, line 37: Can we remove `storeWrapperBinaries` and instead use `uploadArtifacts`?

.github/templates/main.yml
```diff
@@ -505,9 +499,4 @@ jobs:
           file: ${{ github.workspace }}/BenchmarkDotNet.Artifacts/results/${{ steps.find-results-file.outputs.benchmark-results }}
           dashboard-path: dashboard.charts
           nuget-package: ${{ github.workspace }}/Realm/packages/Realm.${{ needs.build-packages.outputs.package_version }}.nupkg
-      - name: Upload Charts Dashboard File
-        uses: #@ actionUploadArtifact
-        with:
-          name: dashboard.charts
-          path: ${{ github.workspace }}/dashboard.charts
-          retention-days: 30
+      - #@ uploadArtifacts("Charts Dashboard", "dashboard.charts", "dashboard.charts", "30")
```
nirinchev, line 502:
```suggestion
      - #@ uploadArtifacts("Charts Dashboard", "dashboard.charts", "dashboard.charts", 30)
```

nirinchev (changes requested):
Looks good. Let's also use the new function in `buildDocs()`. If you're wondering how to combine it with the `if` clause there, you can use `template.replace` to achieve this:

```yaml
  - _: #@ template.replace(uploadArtifacts("Docs", "Docs.zip", "Realm/packages/Docs.zip"))
    if: #@ docsCondition
```

.github/templates/main.yml
LaPeste: I actually went like that at the beginning but then I thought it looked a bit too long. But if you think it's fine, then I'll change it to how it was before.

nirinchev (approved):

PR #2618, opened 2021-09-10
Fix a build warning due to unreachable code

nirinchev:
Noticed that we're getting a compile time warning about unreachable code and it turned out that the inline task generator will generate a `return Success` at the end of the method.


LaPeste (approved):
Thank you for the fix.

PR #2625, opened 2021-09-14
Expose manual schema API

### LINKED ISSUE #824: Create ObjectSchema programmatically
#### Goal
I'm creating an app which the user defines database schema. With Core Data, I can create NSManagedModel programmatically. But Xamarin Realm do not have a similar solution. All ObjectSchema are created from RealmObject subclasses. And the initalize method of ObjectSchema is private method.

I found out Realm in Java can create Schema programmatically. Is there any plan these methods will port to Realm for Xamarin in future?

#### Code Sample
```csharp
> Your code here. Bigger samples should ideally be as separate Xamarin/Visual Studio project, 
> in gists/repositories or privately at help@realm.io)

```

Realm version(s): ?

Xamarin/Visual Studio version: ?

Which operating system version and device: ?

### LINKED ISSUE #1978: SIGABORT on Linux and Mac (not Windows) when schema not persisted
When opening a realm db with missing schema in one table, the exception thrown from realm reaches the catch block on Windows c# code, but on Linux and Mac, it never reaches the catch block (it crashes hard with SIGABORT)
[SigabortOnMissingSchema.zip](https://github.com/realm/realm-dotnet/files/4523095/SigabortOnMissingSchema.zip)

Attached is the project with source code, pre-built binaries and sample realm db. Go to bin\Debug\netcoreapp3.1 sub-folder and execute:

dotnet SigabortOnMissingSchema.dll <full path of v6db.realm included>

Do this on all 3 environments - Linux, Mac and Windows.

You should see SIGABORT on Linux and Mac. You will see soft landing on Windows.

## Goals

Graceful exception handling across all environments
no hard SIGABORT crash on Linux and Mac

## Expected Results

Hit my catch all exception block in C#

## Actual Results

Hard crash SIGABORT on Linux and Mac

Core dumps with following:

Unhandled exception. System.InvalidOperationException: No properties in LocationSettingsModel, has linker stripped it? See https://realm.io/docs/xamarin/latest/#linker-stripped-schema
   at Realms.Schema.ObjectSchema.Builder.Build()
   at Realms.Schema.RealmSchema.CreateFromObjectStoreSchema(Schema nativeSchema)
   at Realms.RealmConfiguration.<>c__DisplayClass24_0.<CreateRealm>b__0(Schema nativeSchema)
   at Realms.SharedRealmHandle.GetNativeSchema(Schema schema, IntPtr managedCallbackPtr)
   at Realms.SharedRealmHandle.NativeMethods.get_schema(SharedRealmHandle sharedRealm, IntPtr callback, NativeException& ex)
   at Realms.SharedRealmHandle.GetSchema(Action`1 callback)
   at Realms.RealmConfiguration.CreateRealm(RealmSchema schema)
   at Realms.Realm.GetInstance(RealmConfigurationBase config, RealmSchema schema)
   at Realms.Realm.GetInstance(RealmConfigurationBase config)
   at SigabortOnMissingSchema.Program.Main(String[] args) in /home/user/Documents/SigabortOnMissingSchema/Program.cs:line 30
Fatal error. Internal CLR error. (0x80131506)
   at Realms.SharedRealmHandle+NativeMethods.get_schema(Realms.SharedRealmHandle, IntPtr, Realms.NativeException ByRef)
   at Realms.SharedRealmHandle+NativeMethods.get_schema(Realms.SharedRealmHandle, IntPtr, Realms.NativeException ByRef)
   at Realms.SharedRealmHandle.GetSchema(System.Action`1<Realms.Native.Schema>)
   at Realms.RealmConfiguration.CreateRealm(Realms.Schema.RealmSchema)
   at Realms.Realm.GetInstance(Realms.RealmConfigurationBase, Realms.Schema.RealmSchema)
   at Realms.Realm.GetInstance(Realms.RealmConfigurationBase)
   at SigabortOnMissingSchema.Program.Main(System.String[])
Aborted (core dumped)

## Steps to Reproduce

Attached is the project with source code, pre-built binaries and sample realm db. Go to bin\Debug\netcoreapp3.1 sub-folder and execute:

dotnet SigabortOnMissingSchema.dll <full path of v6db.realm included>

Do this on all 3 environments - Linux, Mac and Windows.

You should see SIGABORT on Linux and Mac. You will see soft landing on Windows.

Latest stable release Realm 4.3.0

- Realm Object Server Version: ? (you can see this as well in Studio once connected to a Realm instance) 
- Flavor:
  - [ ] Realm Cloud
  - [ ] Self-Hosted
- Server OS & Version: ? (e.g. CentOS 6)
- Client SDK Version: ?
- Client OS & Version: ?

nirinchev:
## Description

This obsoletes `RealmConfigurationBase.ObjectClasses` and exposes `RealmConfigurationBase.Schema` instead. This allows users to specify the Realm Schema either manually or construct it from a set of c# types:

```csharp
// Construct config from c# classes as usual
var config = new RealmConfiguration
{
    Schema = new[] { typeof(Foo), typeof(Bar) }
};

// Construct config with manual schema - advanced
var config = new RealmConfiguration
{
    Schema = new RealmSchema.Builder
    {
        new ObjectSchema.Builder("Foo", isEmbedded: false)
        {
            Property.Primitive("Id", RealmValueType.Guid, isPrimaryKey: true)
            Property.Primitive("Bar", RealmValueType.String, isNullable: true, isIndexed: true)
        }
    }
}

// Mix and match - very advanced
var fooSchema = new ObjectSchema.Builder(typeof(Foo));
fooSchema.Add(Property.FromType<int>("ThisIsNotOnCSharpClass"));

// fooSchema will have all properties on the Foo class + an extra int prop - "ThisIsNotOnCSharpClass"
var config = new RealmConfiguration
{
    Schema = new[] { fooSchema.Build() }
};
```

Fixes #824
Fixes #1978

##  TODO

* [ ] Changelog entry
* [x] Tests (if applicable)
* [x] Fix warnings due to deprecation
* [ ] ~Figure out how to update the schema once the Realm is opened~ - moved to https://github.com/realm/realm-dotnet/issues/2651.
* [x] Expose an API to get a prepopulated builder from ObjectSchema
* [x] Expose an API to get a prepopulated builder from RealmSchema
* [x] Expose an API to get ObjectSchema.Builder from type
* [x] Test that mixing and matching strongly typed API and manual properties is supported
* [x] Make Property a readonly struct


nirinchev:
@LaPeste @papafe @fealebenpae I have 3 approaches I'm considering toward how we expose this API and would appreciate feedback on what you think makes most sense:

## Immutable schema

This is the approach this PR implements and it roughly means that all the components of the schema are immutable and only defined at construction time.

### Defining the schema

```csharp
var realmSchema = new RealmSchema(new[]
{
    new ObjectSchema("Person", isEmbedded: false, new[]
    {
        new Property("Id", PropertyType.Guid, isPrimaryKey: true),
        new Property("Name", PropertyType.NullableString, isIndexed: true),
        new Property("Dogs", PropertyType.Set, objectType: "Dog")
    }),
    new ObjectSchema("Dog", isEmbedded: false, new[]
    {
        new Property("Id", PropertyType.Guid, isPrimaryKey: true),
        new Property("Name", PropertyType.NullableString, isIndexed: true),
        new Property("Owners", PropertyType.LinkingObjects | PropertyType.Array, objectType: "Person", linkOriginProperty: "Dogs"),
    }),
});

var config = new RealmConfiguration
{
    Schema = realmSchema
};

var realm = Realm.GetInstance(config);
```

### Updating the schema

```csharp
realm.Write(() =>
{
    var newSchema = realm.Schema.ToList();
    newSchema.Add(new ObjectSchema(...));

    var personSchema = newSchema.Single(os => os.Name == "Person");
    newSchema.Remove(personSchema);

    var newPersonSchema = personSchema.ToList();
    newPersonSchema.Add(new Property("IsInteresting", PropertyType.Bool));

    newSchema.Add(newPersonSchema);

    realm.UpdateSchema(new RealmSchema(newSchema));
});
```

## ImmutableSchema + Builders

This is the old approach that was all internal - the schema is immutable, but we do offer builders to allow manipulation.

### Defining the schema

```csharp
var schemaBuilder = new RealmSchema.Builder();

var personSchema = new ObjectSchema.Builder("Person", isEmbedded: false);
personSchema.Add(new Property("Id", PropertyType.Guid, isPrimaryKey: true));
personSchema.Add(new Property("Name", PropertyType.NullableString, isIndexed: true));
personSchema.Add(new Property("Dogs", PropertyType.Set, objectType: "Dog"));

schemaBuilder.Add(personSchema.Build());

// Using the fluent API
schemaBuilder.Add(new ObjectSchema.Builder("Dog", isEmbedded: false)
    .Add(new Property("Id", PropertyType.Guid, isPrimaryKey: true))
    .Add(new Property("Name", PropertyType.NullableString, isIndexed: true))
    .Add(new Property("Owners", PropertyType.LinkingObjects | PropertyType.Array, objectType: "Person", linkOriginProperty: "Dogs"))
    .Build());

var config = new RealmConfiguration
{
    Schema = schemaBuilder.Build()
};

var realm = Realm.GetInstance(config);
```

### Updating the schema

```csharp
realm.Write(() =>
{
    var newSchemaBuilder = realm.Schema.GetBuilder();
    newSchemaBuilder.Add(new ObjectSchema.Builder(...).Build());

    var personSchema = newSchemaBuilder.Single(os => os.Name == "Person");
    newSchemaBuilder.Remove(personSchema);

    var newPersonSchema = personSchema.GetBuilder()
        .Add(new Property("IsInteresting", PropertyType.Bool))
        .Build();

    newSchemaBuilder.Add(newPersonSchema);

    realm.UpdateSchema(newSchemaBuilder.Build());
});
```

## Mutable Schema

This approach is the most complex in terms of implementation, but probably the most flexible in terms of API. Here, we allow schema manipulation at any time and depending on whether it has a Realm associated with it, we update it in memory or actually call into the Realm API:

### Illustrative implementation

```csharp
public struct RealmSchema
{
    private readonly Realm _realm;
    private readonly IDictionary<string, ObjectSchema> _schemas;

    public void Add(ObjectSchema schema)
    {
        if (_realm != null)
        {
            _realm.AddSchema(schema.ToNative());
        }

        _schemas.Add(schema);
    }

    public void Remove(ObjectSchema schema)
    {
        if (_realm != null)
        {
            _realm.RemoveSchema(schema.ToNative());
        }

        _schemas.Remove(schema.Name);
    }
}
```

### Defining the schema

```csharp
var schema = new RealmSchema();

var personSchema = new ObjectSchema("Person", isEmbedded: false);
personSchema.Add(new Property("Id", PropertyType.Guid, isPrimaryKey: true));
personSchema.Add(new Property("Name", PropertyType.NullableString, isIndexed: true));
personSchema.Add(new Property("Dogs", PropertyType.Set, objectType: "Dog"));

schema.Add(personSchema);

// Using the fluent API
schema.Add(new ObjectSchema("Dog", isEmbedded: false)
    .Add(new Property("Id", PropertyType.Guid, isPrimaryKey: true))
    .Add(new Property("Name", PropertyType.NullableString, isIndexed: true))
    .Add(new Property("Owners", PropertyType.LinkingObjects | PropertyType.Array, objectType: "Person", linkOriginProperty: "Dogs")));

var config = new RealmConfiguration
{
    Schema = schema
};

var realm = Realm.GetInstance(config);
```

### Updating the schema
```csharp
realm.Write(() =>
{
    realm.Schema.Add(new ObjectSchema(...));

    // Here we don't need to remove+add the person schema because it
    // is mutable. We also don't need to call realm.UpdateSchema since
    // mutations already update the Realm
    var personSchema = realm.Schema.Single(os => os.Name == "Person");
    personSchema.Add(new Property("IsInteresting", PropertyType.Bool));
});
```

papafe:
I like the `ImmutableSchema + Builders` the most. The first solution does seems the least flexible and readable. Regarding the last one, I am a not a fan because I would prefer to keep the schema immutable. Besides, in my opinion it makes a lot of sense to have a method like `realm.UpdateSchema` to be explicit about the change.

LaPeste:
From a surface perspective, hence from a developer usability stand point, in my opinion:
1. Between the first ( `Immutable schema` ) and the secon approach (`ImmutableSchema + Builders`) there's very little difference. Because of this, I find "unnecesary" and annoying having to call the builder all the time (with all its `.build()` calls). If using a builder had shorten the way a schema can be updated, my opinion would have been different, of course. So between the first 2 solutions I prefer `Immutable schema`
2. The last solution is in line with the others when it comes to defining the schema but it's so much more concise and intuitive when it comes to updating the schema, hence this last one is my favourite one.  

In summary, if we can implement `Mutable Schema` then we have I vote for that. In case we don't have enough time for that, the `Immutable schema` solution is my second favourite.

nirinchev:
After some deliberation and experimentation, I think I'll be going with a builders approach. The three main reasons are:
1. With builders we can implement a collection initializer pattern which allows for declarative schema definition:
    ```csharp
    new RealmSchema.Builder
    {
        new ObjectSchema.Builder("Person", isEmbedded: false)
        {
            new Property(...),
            new Property(...)
        },
        new ObjectSchema.Builder("Dog", isEmbedded: false)
        {
            new Property(...),
            new Property(...)
        }
    }
    ```
2. It is a bit more intuitive compared to the immutable schema approach in the sense that we expose the API to do mutate the schema rather than relying on people to figure out they can create a list from the schema, then a new schema from the list.
3. Compared to the mutable schema API - it makes it more explicit that the schema update happens atomically and is significantly easier to implement.

LaPeste:
> 
> 
> After some deliberation and experimentation, I think I'll be going with a builders approach. The three main reasons are:
> 
>     1. With builders we can implement a collection initializer pattern which allows for declarative schema definition:
>        ```cs
>        new RealmSchema.Builder
>        {
>            new ObjectSchema.Builder("Person", isEmbedded: false)
>            {
>                new Property(...),
>                new Property(...)
>            },
>            new ObjectSchema.Builder("Dog", isEmbedded: false)
>            {
>                new Property(...),
>                new Property(...)
>            }
>        }
>        ```
> 
>     2. It is a bit more intuitive compared to the immutable schema approach in the sense that we expose the API to do mutate the schema rather than relying on people to figure out they can create a list from the schema, then a new schema from the list.
> 
>     3. Compared to the mutable schema API - it makes it more explicit that the schema update happens atomically and is significantly easier to implement.

Point 2. is a really good point and I didn't think about it. About point 3., as I said already, if it's too hard/long to implement, I don't see a problem with a little longer schema update process.
So you made the `ImmutableSchema + Builders` look like the best one :)

nirinchev:
This is now ready for review. I will address the schema updating as a separate PR as this one is getting quite unmanageable. While the overall number of changes is insane, they can be roughly split into 3 categories:
1. Meaningful code changes - these are mainly concentrated in the Realm/Schema folder. I'd like a thorough review of those.
2. Mechanical changes due to deprecation. `ObjectClasses` is now deprecated and users need to use `Schema`, which resulted in a bunch of updates, mainly in test code.
3. New tests. The ObjectSchemaTests class is the primary place where I've put all the new tests - I've tried to cover all the public API as well as the relevant branches, but would appreciate feedback if you notice something is missing.

jsflax:
What is the use case for mixing and matching properties on existing Realm Objects when creating the schema? Seems like a potential foot gun.

nirinchev:
@jsflax it would enable a use case where you have a `RealmObject` inheritor with some commonly accessed information that you want to enhance with dynamic properties depending on runtime data. Let's say you have a collection of tags that you want to apply to your entities - then you can have a model like:

```csharp
public class Note : RealmObject
{
	public string Title { get; set; }
    public string Text { get; set; }
    public DateTimeOffset LastModified { get; set; }
}
```

And these are properties you'd like to access via the strongly-typed API. Then, you get a list of tags (from somewhere) and you enhance your model with those (this is not yet implemented - but something like):

```csharp
var schema = new ObjectSchema.Builder(typeof(Note));
foreach (var tag in tags)
{
    schema.Add(Property.Primitive(tag, RealmValueType.Boolean));
}
```

Then you can work with your strongly-typed properties like normal:

```csharp
var note = realm.Find<Note>(123);
EntryBox.Text = note.Text;

// Query by tag
var notesByTag = realm.All<Note>().Filter($"{tag} = TRUE");

// Set a tag
note.DynamicApi.Set(tag, true);

// Get tag value
var isNoteTagged = note.DynamicApi.Get<bool>(tag);
```

I don't expect this to come up often, but it seems like it might be useful and I don't see much downside - the only situation where it could be problematic is if you remove properties from the schema that are present in the model, but that seems like you'd be actively sabotaging yourself (though we can definitely guard against it). Do you see other situations where this can be more of an accidental footgun?

jsflax:
Mix & match / manual construction seem like they are one and the same. Any misdeclaration/misspelling would end up being a dynamic property as opposed to an exception. Probably a super rare use case/mishap anyway, but was more curious about how people might use it.

Realm/Realm/Schema/SchemaBuilderBase.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            _values.Add(key, item);
+        }
+
+        /// <summary>
+        /// Removes an element from the builder.
+        /// </summary>
+        /// <param name="item">The element to remove.</param>
+        /// <returns><c>true</c> if the element was found and removed; <c>false</c> otherwise.</returns>
```
DominicFrei, line 76: Do we need to differentiate here between `could not be found` and `could not be deleted`?

DominicFrei: same questions for the other functions below

DominicFrei (commented):
Only added some questions I was curious about.

Realm/Realm/Schema/SchemaBuilderBase.cs
nirinchev: There's no case where we wouldn't be able to delete something if we found it.

nirinchev:
I filed https://github.com/realm/realm-dotnet/issues/2651 to follow up on the runtime schema updates. I won't be tackling that as part of this PR as this feature is still useful as is.

LaPeste (approved):
I've mainly focused my attention on changes in Realm/Schema. From what I could see the overall changes and additions make sense in my mind. And the API looks to have all I can think of.

PR #2635, opened 2021-09-21
Added missing migration methods

### LINKED ISSUE #243: Cleanup repo

### LINKED ISSUE #2543: Add missing methods to migration
It seems that in the .NET API for migrations Specifically we are missing the two methods about renaming an existing property and deleting an entire class and its schema, that are presents in the Cocoa implementation, [for instance](https://github.com/realm/realm-cocoa/blob/master/Realm/RLMMigration.h).

papafe:
This PR adds two methods that were missing on `Migration`: `RemoveType` and `RenameProperty`.

## Description

Fixes #2543

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm/Migration.cs
```diff
@@ -83,11 +86,46 @@ internal bool Execute(Realm oldRealm, Realm newRealm)
...
                 NewRealm = null;
+
+                _migrationSchema = IntPtr.Zero;
             }
 
             return true;
         }
 
+        /// <summary>
+        /// Removes a type during a migration. All the data associated with the type, as well as its schema, will be removed from <see cref="Realm"/>.
```
DominicFrei, line 97: Do we need to differentiate between the (general) schema and the object schema to not confuse anyone here?

wrappers/src/shared_realm_cs.cpp
```diff
@@ -547,4 +548,38 @@ REALM_EXPORT Results* shared_realm_create_results(SharedRealm& realm, TableKey t
...
+
+REALM_EXPORT bool shared_realm_remove_type(const SharedRealm& realm, uint16_t* type_name_buf, size_t type_name_len, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        Utf16StringAccessor type_name_str(type_name_buf, type_name_len);
+
+        auto table = ObjectStore::table_for_object_type(realm->read_group(), type_name_str);
+        if (!table)
+        {
+            return false;
```
DominicFrei, line 571: Isn't this a similar problem to the one below? Should we throw here as well?

DominicFrei (commented):
Don't see anything wrong with that. :) Just added two questions for clarification.

wrappers/src/shared_realm_cs.cpp
papafe: Yeah, this is up for discussion I think 😁 
The one down throws an exception because it tries to do something that can be already done with the current API (`realm.RemoveAll`), and I suppose it's better to have developers having one way of doing certain things, as it causes less headaches

Realm/Realm/Migration.cs
papafe: I think it's clear enough with the "its", but maybe it's not?

wrappers/src/shared_realm_cs.cpp
DominicFrei: Hm, if this is a valid way, just not the preferred one I would rather add a comment to the documentation than throwing a  (probably) very unexpected and confusing error because it's supposed to work? I get your point, just trying to see it from a users point of view.

Realm/Realm/Migration.cs
DominicFrei: Clear enough for me because I know how Realm works and what the difference between a schema and an object schema is. Then again, how important is this for a user? They just want to use the framework and not understand too much about the internals I suppose. I guess it's fine both ways. ;)

DominicFrei:
The non-throwing delete would also be in line with other deletes. Take `DeleteRealm` for example. If the Realm was already deleted, it's fine, we just accept that because the goal was that that Realm does not exist anymore. Throwing an exception would be like throwing an exception just because the Realm we try to delete did not exist in the first place.

nirinchev (changes requested):
I have a couple of comments mostly around code style and docs, but for the most part it looks good. What is missing outside of the comments I added is:
1. Tests for invalid arguments (null, empty strings, a string with 1000 characters)
2. A test for renaming a property to a value that doesn't exist in the new schema
3. Address the build warnings/CodeQL errors

Realm/Realm/Migration.cs
```diff
@@ -83,11 +86,46 @@ internal bool Execute(Realm oldRealm, Realm newRealm)
...
+
+        /// <summary>
+        /// Renames a property during a migration.
+        /// </summary>
+        /// <param name="typeName">The type for which the property rename needs to be performed. </param>
+        /// <param name="oldPropertyName">The previous name of the property. </param>
+        /// <param name="newPropertyName">The new name of the property. </param>
+        /// <remarks>
+        /// It is not possible to access the renamed property in <see cref="NewRealm"/> in the migration block after this method is called.
+        /// If this is necessary, the method should be called after the property access, or the value retrieved from <see cref="OldRealm"/> can be used.
```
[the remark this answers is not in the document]
papafe: Mmh, I need to rephrase it. The idea is that once this method is called, one should not try to access the renamed property

wrappers/src/shared_realm_cs.hpp
```diff
@@ -34,13 +34,18 @@ using SharedSyncUser = std::shared_ptr<SyncUser>;
...
 
-class ManagedExceptionDuringMigration : public std::runtime_error
-{
+class ManagedExceptionDuringMigration : public std::runtime_error {
 public:
     ManagedExceptionDuringMigration() : std::runtime_error("Uncaught .NET exception during Realm migration") {
     }
 };
 
+class RemoveTypeInSchemaException : public std::runtime_error {
```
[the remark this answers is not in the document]
papafe: Makes sense

nirinchev:
Can you rebase on master? There was a Coveralls outage over the weekend which is why the base coverage report is missing.

papafe:
Final corrections done (lint correction, more tests, ...)

nirinchev (approved):
Good job!

CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+* Added two new methods on `Migration`:
```
nirinchev, line 4: Needs a link to the issue

Tests/Realm.Tests/Database/MigrationTests.cs
```diff
@@ -170,5 +172,265 @@ public void MigrationTriggersDelete()
...
+                    var newPeople = migration.NewRealm.All<Person>();
+
+                    for (var i = 0; i < newPeople.Count(); i++)
+                    {
+                        var oldPerson = oldPeople.ElementAt(i);
+                        var newPerson = newPeople.ElementAt(i);
+
+                        var oldValue = oldPerson.DynamicApi.Get<string>("TriggersSchema");
+                        var newValue = newPerson.OptionalAddress;
+                        Assert.That(newValue, Is.EqualTo(newValue));
```
nirinchev, line 198:
```suggestion
                        Assert.That(oldValue, Is.EqualTo(newValue));
```

Tests/Realm.Tests/Database/MigrationTests.cs
```diff
@@ -170,5 +172,265 @@ public void MigrationTriggersDelete()
...
+            };
+
+            using (var newRealmDynamic = GetRealm(newRealmDynamicConfig))
+            {
+                // This means that "Person" is not in the schema anymore, as we retrieve it directly from core.
+                Assert.That(newRealmDynamic.Schema.TryFindObjectSchema("Person", out _), Is.False);
+            }
+        }
+
+        public void MigrationRemoveTypeInvalidArguments()
```
nirinchev, line 410: This is missing `[Test]`

Tests/Realm.Tests/Database/MigrationTests.cs
```diff
@@ -170,5 +172,265 @@ public void MigrationTriggersDelete()
...
+                {
+                    migration.RenameProperty(nameof(Person), nameof(Person.Latitude), nameof(Person.Longitude));
+                }
+            };
+
+            ex2 = Assert.Throws<AggregateException>(() => GetRealm(configuration));
+            Assert.That(ex2.Flatten().InnerException.Message, Does.Contain("Cannot rename property 'Person.Latitude' to 'Longitude' because the source property still exists."));
+        }
+
+        public void MigrationRenamePropertyInvalidArguments()
```
nirinchev, line 299: Missing `[Test]`

Tests/Realm.Tests/Database/MigrationTests.cs
```diff
@@ -170,5 +172,265 @@ public void MigrationTriggersDelete()
...
+
+                    newValue = newPerson.OptionalAddress;
+                }
+            };
+
+            using var realm = GetRealm(configuration);
+        }
+
+        [Test]
+        public void MigrationRenamPropertyErrors()
```
nirinchev, line 232:
```suggestion
        public void MigrationRenamePropertyErrors()
```

Tests/Realm.Tests/Database/MigrationTests.cs
```diff
@@ -170,5 +172,265 @@ public void MigrationTriggersDelete()
...
+                        Assert.That(newValue, Is.EqualTo(newValue));
+                    }
+                }
+            };
+
+            using var realm = GetRealm(configuration);
+        }
+
+        [Test]
+        public void MigrationException()
```
nirinchev, line 207: Should this test be removed now?

nirinchev (changes requested):

nirinchev (approved):

PR #2654, opened 2021-09-23
Use new XCode build system for iOS wrappers

### LINKED ISSUE #2171: Investigate use of new Xcode build system for iOS wrappers
Xcode 10 introduced a [new build system](https://developer.apple.com/documentation/xcode-release-notes/build-system-release-notes-for-xcode-10) that was included as the default one in cmake, starting from version [3.19](https://cmake.org/cmake/help/latest/release/3.19.html). 

This new build system causes the following error when trying to build the wrappers for iOS:
```
error: unable to attach DB: error: accessing build database "/Users/andrea.catalini/Development/realm-dotnet/wrappers/cmake/iOS/build/XCBuildData/build.db": 
database is locked Possibly there are two concurrent builds running in the same filesystem location.
```

At the moment it is possible to force the use of the old build system by passing the following extra arguments to cmake `-T buildsystem=1`, but it will be removed in a future release, so we need to understand why it is failing with the new build system in the first place.

LaPeste:
## Description
The issue has been fixed with newer versions of CMake. So all it took was removing the cmake arg to force Cmake to use the old build system,

Fixes #2171

##  TODO

* [ ] Changelog entry

### DIFF
--- a/wrappers/build-ios.sh
+++ b/wrappers/build-ios.sh
@@ -3,7 +3,7 @@
 SCRIPT_DIRECTORY="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 REALM_CMAKE_CONFIGURATION=Debug
-EXTRA_CMAKE_ARGS="-T buildsystem=1"
+# EXTRA_CMAKE_ARGS="-T buildsystem=1"
 
 for i in "$@"
 do
--- a/wrappers/build-ios.sh
+++ b/wrappers/build-ios.sh
@@ -3,7 +3,6 @@
 SCRIPT_DIRECTORY="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 REALM_CMAKE_CONFIGURATION=Debug
-# EXTRA_CMAKE_ARGS="-T buildsystem=1"
 
 for i in "$@"
 do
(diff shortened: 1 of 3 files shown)


nirinchev:
The new build system appears to consistently increase build times - from 25-ish minutes, they go into 30-35ish - not the end of the world, but somewhat suspicious. Can you try a build with LTO enabled to see how long it will take?

nirinchev:
Would be also interesting to test build times with both build system on macOS-11 hosts.

LaPeste:
"Performance" report:
* no LTO on macOS 10.5: ~31m
* LTO on macOS 10.5: 27m
* no LTO on macOS 11: 29m 50s
* LTO on macOS 11: 27m 40s

nirinchev (approved):

wrappers/build-ios.sh
```diff
@@ -3,7 +3,6 @@
 SCRIPT_DIRECTORY="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 REALM_CMAKE_CONFIGURATION=Debug
-EXTRA_CMAKE_ARGS="-T buildsystem=1"
```
nirinchev, line 6: Can we keep this, just remove the contents: `EXTRA_CMAKE_ARGS=""`

nirinchev (approved):

PR #2700, opened 2021-10-29
Add advanced data modeling blog post

nirinchev:
Putting this up just for review. Will close once we're aligned.

### DIFF
--- a/advanced-data-types.md
+++ b/advanced-data-types.md
@@ -1,0 +1,195 @@
+# Advanced data modeling with Realm .NET
+
+Realm's intuitive data model approach means that in most cases you don't even think of Realm models as entities - you just declare your POCOs, have them inherit from `RealmObject` and you're done. Now you have persistable models, with `INotifyPropertyChanged` capabilities all wired up, that are also "live" - i.e. every time you access a property, you get the latest state and not some snapshot from who knows how long ago. This is great and most of our users absolutely love the simplicity. Still, there are some use cases where being aware that you're working with a database can really bring your data models to the next level. In this blog post we'll evaluate 3 techniques you can apply to make your models fit your needs even better.
+
+## Constructor validation
+
+One of the core requirements of Realm is that all models need to have a parameterless constructor. This is needed because Realm needs to be able to instantiate an object without figuring out what the arguments to pass to the constructor. What not many people know is that you can make this parameterless constructor private to communicate expectations  to your callers. This means that if you have an `Person` class where you absolutely expect that a `Name` is provided upon object creation, you can have a public constructor with a `name` argument and a private parameterless one for use by Realm:
+
+```csharp
+class Person : RealmObject
+{
+    public string Name { get; set; }
+
+    public Person(string name)
+    {
+        ValidateName(name);
+
+        Name = name;
+    }
+
+    // This is used by Realm, even though it's private
+    private Person()
+    {
+    }
+}
+```
+
+And I know what some of you may be thinking - "oh no 😱 does that mean Realm uses the suuuuper slow reflection to create object instances". Fortunately, the answer is "no". Instead, at compile-time, Realm injects a nested helper class in each model that has a `CreateInstance` method. Since the helper class is nested in the model classes, it has access to private members and is thus able to invoke the private constructor.
+
+## Property access modifiers
+
+Similar to the point above - another relatively unknown feature of Realm is that persisted properties don't need to be public. You can either have the entire property be private or just one of the accessors. This synergizes nicely with the private constructor technique that we mentioned above. If you expose a constructor that explicitly validates the person's name, it would be fairly annoying to do all that work and have some code accidentally set the property to `null` the very next line. So it would make sense to make the setter of the name property above private:
+
+```csharp
+class Person : RealmObject
+{
+    public string Name { get; private set; }
+
+    public Person(string name)
+    {
+        // ...
+    }
+}
+```
+
+That way you're communicating clearly to the class consumers that they need to provide the name at object creation time and that it can't be changed later. A very common use case here is to make the `Id` setter private and generate a random `Id` at object creation time:
+
+```csharp
+class Transaction : RealmObject
+{
+    public Guid Id { get; private set; } = Guid.NewGuid();
+}
+```
+
+Sometimes it makes sense to make the entire property private - typically when you want to expose a different public property that wraps it. If we go back to our `Person` and `Name` example - perhaps we want to allow changing the name, but we want to still validate the new name before we persist it. Then, we create a private autoimplemented property that Realm will use for persistence, and a public one that does the validation:
+
+```csharp
+class Person : RealmObject
+{
+    [MapTo("Name")]
+    private string _Name { get; set; }
+
+    public string Name
+    {
+        get => _Name;
+        set
+        {
+            ValidateName(value);
+            _Name = value;
+        }
+    }
+}
+```
+
+This is quite neat as it makes the public API of your model safe, while preserving their persistability. Of note is the `MapTo` attribute applied to `_Name` - it is not strictly necessary, I just added it to avoid having ugly column names in the database. You can use it or not use it - it's totally up to you. One thing to note when utilizing this technique is that Realm is completely unaware of the relationship between `Name` and `_Name`. This has two implications - 1) notifications will be emitted for `_Name` only and 2) you can't use LINQ queries to filter `Person` objects by name. Let's see how we can mitigate both:
+
+For notifications, we can override `OnPropertyChanged` and raise a notification for `Name` whenever `_Name` changes:
+
+```csharp
+class Person : RealmObject
+{
+    protected override void OnPropertyChanged(string propertyName)
+    {
+        base.OnPropertyChanged(propertyName);
+
+        if (propertyName == nameof(_Name))
+        {
+            RaisePropertyChanged(nameof(Name));
+        }
+    }
+}
+```
+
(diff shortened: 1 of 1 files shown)


papafe:
I think it's done well so far. It's good that it is short and straight to the point, without much fuss. 

It's informal at point, that is fine, but I don't know if we have any "editorial guidelines" for this aspect.

LaPeste (approved):
Overall, I like the style of the blog post for 2 reasons:
1- it's informal as it should be
1- it's dense in information and short in length, which is the sweet spot in my opinion
Additionally, I think this is going to be a really useful piece of content for new comers. A totally needed blog post.

LaPeste (commented):
One thing that you could think about adding: `enum`s in more complex data structures; like lists or sets. It's a little more convoluted but I think it could be useful to show a usage example.

advanced-data-types.md
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    // This is used by Realm, even though it's private
+    private Person()
+    {
+    }
+}
+```
+
+And I know what some of you may be thinking - "oh no 😱 does that mean Realm uses the suuuuper slow reflection to create object instances". Fortunately, the answer is "no". Instead, at compile-time, Realm injects a nested helper class in each model that has a `CreateInstance` method. Since the helper class is nested in the model classes, it has access to private members and is thus able to invoke the private constructor.
```
DominicFrei, line 28: Might some readers also think "Oh no, this is going to increase the compile time considerably. And it does so every time."?

advanced-data-types.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Property access modifiers
+
+Similar to the point above - another relatively unknown feature of Realm is that persisted properties don't need to be public. You can either have the entire property be private or just one of the accessors. This synergizes nicely with the private constructor technique that we mentioned above. If you expose a constructor that explicitly validates the person's name, it would be fairly annoying to do all that work and have some code accidentally set the property to `null` the very next line. So it would make sense to make the setter of the name property above private:
+
```
DominicFrei, line 33: Is it worth mentioning here that they could also implement a custom setter validating the name on every change?

advanced-data-types.md
nirinchev: I would be surprised if people care about it - it's just compiling like it compiles any other piece of code in their project.

advanced-data-types.md
```diff
@@ -0,0 +1,195 @@
...
+That way you're communicating clearly to the class consumers that they need to provide the name at object creation time and that it can't be changed later. A very common use case here is to make the `Id` setter private and generate a random `Id` at object creation time:
+
+```csharp
+class Transaction : RealmObject
+{
+    public Guid Id { get; private set; } = Guid.NewGuid();
+}
+```
+
+Sometimes it makes sense to make the entire property private - typically when you want to expose a different public property that wraps it. If we go back to our `Person` and `Name` example - perhaps we want to allow changing the name, but we want to still validate the new name before we persist it. Then, we create a private autoimplemented property that Realm will use for persistence, and a public one that does the validation:
```
DominicFrei, line 55: Do we actually offer a way to combine them yet?

DominicFrei: and by that I mean: do we have plans to do that?
Would also solve LINQ and notification problem I suppose.

advanced-data-types.md
nirinchev: Yes, it's below.

advanced-data-types.md
nirinchev: What do you mean?

advanced-data-types.md
```diff
@@ -0,0 +1,195 @@
...
+The situation with queries is slightly harder to work around. The main issue is that, because the property is private, you can't use it in a LINQ query - e.g. `realm.All<Person>().Where(p => p._Name == "Peter")` will result in a compile-time error. On the other hand, because Realm doesn't know that `Name` is tied to `_Name`, you can't use `p.Name == "Peter"` either. You can still use the string-based queries though, just remember to use the name that Realm knows about - i.e. the string argument of `MapTo` if you remapped the property name or the internal property (`_Name`) if you didn't:
+
+```csharp
+// _Name is mapped to 'Name' which is what we use here
+var peters = realm.All<Person>().Filter("Name == 'Peter'");
+```
+
+## Using unpersistable data types
+
+Realm has a wide variety of supported data types - most primitive types in the BCL, as  well as advanced collections, such as sets and dictionaries. But sometimes, you'll come across a data type that Realm can't store yet - the most obvious example being enums. In such cases, you can build on top of the previous technique to expose enum properties in your models and have them be persisted as one of the supported data types:
```
DominicFrei, line 105: Probably just .NET-noobie me but what is BCL? Will every reader know the abbrev.?

advanced-data-types.md
```diff
@@ -0,0 +1,195 @@
...
+enum TransactionState
+{
+    Pending,
+    Settled,
+    Error
+}
+
+class Transaction : RealmObject
+{
+    private string _State { get; set; }
```
DominicFrei, line 117: Is `string` actually our recommendation for `enum`? Wouldn't `int` be more intuitive to use?

DominicFrei: edit: I found the answer partially in the paragraph below but will leave the rest here out of interest.

advanced-data-types.md
```diff
@@ -0,0 +1,195 @@
...
+}
+
+class Powerup : RealmObject
+{
+    [MapTo("Position")]
+    private Vector3Model _Position { get; set; }
+
+    public Vector3 Position
+    {
+        get => _Position?.ToVector3() ?? Vector3.Zero;
```
DominicFrei, line 161:
```suggestion
        get => _Position?.ToVector3() ?? Vector3.zero;
```

advanced-data-types.md
LaPeste: To the first question: it's (B)ase (C)lass (L)ibrary
To the second: I don't think so. Maybe it can be rephrased as -> Base Class Library (BCL)

advanced-data-types.md
```diff
@@ -0,0 +1,195 @@
...
+        set => _Position = new Vector3Model(value);
```
DominicFrei, line 162: Positions change a lot (depending on the `GameObject`). If this `PowerUp` is floating around we will create loads of orphaned `EmbeddedObject`s here. Would it be more efficient to update the position instead of creating a new one?
I guess this is - as often - a question of `keeping the tutorial as simple as possible` vs. `our recommended way of implementing things`.

advanced-data-types.md
nirinchev: Gotcha - it's not possible for now, but we do hope to support it.

advanced-data-types.md
nirinchev: Should be fairly familiar for most .NET devs, but can expand it here - it's the Base Class Library.

advanced-data-types.md
```diff
@@ -0,0 +1,195 @@
...
+
+        if (propertyName == nameof(_Position))
+        {
+            RaisePropertyChanged(nameof(Position));
+        }
+    }
+}
+```
+
+In this example we've defined a `Vector3Model` that roughly mirrors Unity's `Vector3` - it has 3 float properties representing the 3 components of the vector. We've also utilized what we learned in the previous sections - it has a private constructor to force consumers to always construct it with a `Vector3` argument. We've also marked its properties as private as we don't want consumers directly interacting with them - we want users to always call `ToVector3` to obtain the Unity type. And for our `Powerup` model, we're doing exactly that in the publicly exposed `Position` property. Note that similarly, to our `Person` example, we're making sure to raise a notification for `Position` whenever `_Position` changes.
```
DominicFrei, line 177: This actually brought up another interesting question: A `Vector3` (https://docs.unity3d.com/ScriptReference/Vector3.html) offers more than we implemented in the other tutorial and here. Would we want to show examples on how to achieve a 'full' `Vector3` that is also persistable? Do we even have a good recommendation other than waiting for polymorphic inheritance?

advanced-data-types.md
nirinchev: Yeah, `string` is the recommended way to store enums in databases.

advanced-data-types.md
DominicFrei: Might just be a Swift thing. I've seen discussions about that a lot there around static vs. dynamic frameworks, pre-compiled, etc.

advanced-data-types.md
DominicFrei: 👍

DominicFrei (commented):
No really change request (apart from a tiny one), but some questions that I'd be interested in.

advanced-data-types.md
nirinchev: I'm not sure if there's a fundamental perf difference between the two options, but I guess people can figure it out themselves - we could just as easily have an `Update(Vector3)` method on the `Vector3Model` where the properties are updated rather than a new object being created. This blog post is intended to be more of an inspiration than an definitive guide for every single use case.

advanced-data-types.md
nirinchev: The additional properties are computed values and are derived from the x, y, z coordinates. Obviously, we can't use them in queries, but similarly to the point above - this post is intended to be merely an inspiration and not so much a definitive guide. If people need to query by magnitude, they can, of course, store that as well, but it's not required to recreate the vector. We can't support the vector algebra functions in queries because Realm 1) doesn't have vector algebra functions and 2) isn't aware that this object being stored is a vector. But in the future we may be able to offer a way for users to use their own methods in queries or even implement some common vector algebra functions in Core for use in gaming scenarios.

DominicFrei: 👍

DominicFrei (approved):

papafe (approved):

nirinchev:
Post is scheduled for publication. Closing this.

PR #2707, opened 2021-11-03
Add a dispatch workflow to upload packages to an S3 bucket

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2708

##  TODO

* [x] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/Realm/Realm.Fody/Realm.Fody.csproj
+++ b/Realm/Realm.Fody/Realm.Fody.csproj
@@ -16,7 +16,6 @@
   </PropertyGroup>
   <ItemGroup Label="References">
     <PackageReference Include="FodyHelpers" Version="6.*" PrivateAssets="All" />
-    <PackageReference Include="System.Management" Version="4.5.0" PrivateAssets="Build;Compile;Runtime" />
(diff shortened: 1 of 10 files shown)


Realm/Realm.Fody/Realm.Fody.csproj
```diff
@@ -16,7 +16,6 @@
   </PropertyGroup>
   <ItemGroup Label="References">
     <PackageReference Include="FodyHelpers" Version="6.*" PrivateAssets="All" />
-    <PackageReference Include="System.Management" Version="4.5.0" PrivateAssets="Build;Compile;Runtime" />
```
LaPeste, line 19: This I suppose is something completely unrelated or does it have to do with sleet?

nirinchev: It is indeed unrelated. It's just that when I was testing the sleet nuget feed, I noticed that we have a single dependency in the weaver package, which seemed odd and it turned out to be something we don't really need.

LaPeste (approved):

PR #2731, opened 2021-11-24
Read-only realm properly throws when refreshed

### LINKED ISSUE #2718: [Bug]: Crash when goto 10.7.0
### What happened?

I upgrade from 10.5.0  direct to 10.7.0, app crash   Message=Continuous transaction through DB object without history information

Downgrade to 10.6.0 app works fine.

### Repro steps

Upgrade from 10.5.0 to 10.7.0 

### Version

10.7.0

Local Database only

Xamarin

### Client OS and version

	Microsoft Windows 10 Pro Version	10.0.19043 Build 19043

### Code snippets

_No response_

```shell
Realms.Exceptions.RealmException
  HResult=0x80131500
  Message=Continuous transaction through DB object without history information
Exception backtrace:
<backtrace not supported on this platform>
  Source=Realm
  StackTrace:
   at Realms.NativeException.ThrowIfNecessary(Func`2 overrider)
   at Realms.SharedRealmHandle.Open(Configuration configuration, RealmSchema schema, Byte[] encryptionKey)
   at Realms.RealmConfiguration.CreateRealm()
   at LexSemanticDicPersistence.Context.DB.GetNewInstance() in C:\Users\scf\Source\Repos\Utilities\LexSemanticDicPersistence\Context\DB.cs:line 119
   at LexSemanticDicPersistence.Services.LexSemanticDicService.<SearchComposed>d__13.MoveNext() in C:\Users\scf\Source\Repos\Utilities\LexSemanticDicPersistence\Services\LexSemanticDicService.cs:line 247
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at LexSemanticDicPersistence.Services.LexSemanticDicService.<SearchVol>d__10.MoveNext() in C:\Users\scf\Source\Repos\Utilities\LexSemanticDicPersistence\Services\LexSemanticDicService.cs:line 175
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at LexSemanticClient.ViewModel.VerbeteSearchObservableCollection.<LoadSearchedCollection>d__51.MoveNext() in C:\Users\scf\Source\Repos\Utilities\LexSemanticClient\LexSemanticClient\ViewModel\VerbeteSearchViewModel.cs:line 567
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at LexSemanticClient.ViewModel.VerbeteSearchObservableCollection.<LoadSearched>d__50.MoveNext() in C:\Users\scf\Source\Repos\Utilities\LexSemanticClient\LexSemanticClient\ViewModel\VerbeteSearchViewModel.cs:line 529

  This exception was originally thrown at this call stack:
    [External Code]
    LexSemanticDicPersistence.Context.DB.GetNewInstance() in DB.cs
    LexSemanticDicPersistence.Services.LexSemanticDicService.SearchComposed(string, string) in LexSemanticDicService.cs
    [External Code]
    LexSemanticDicPersistence.Services.LexSemanticDicService.SearchVol(System.Collections.Generic.IList<string>, string) in LexSemanticDicService.cs
    [External Code]
    LexSemanticClient.ViewModel.VerbeteSearchObservableCollection.LoadSearchedCollection(string) in VerbeteSearchViewModel.cs
    [External Code]
    LexSemanticClient.ViewModel.VerbeteSearchObservableCollection.LoadSearched() in VerbeteSearchViewModel.cs
```

### Relevant log output

_No response_

LaPeste:
## Description
Because the change [#5065](https://github.com/realm/realm-core/pull/5065) in core, the wrappers needed a small change to stop a refresh call when getting a shared instance of a read-only realm.
An extra unit test was added to verify that freezing a read-only realm worked as expected.

Fixes #2718 

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Tests/Realm.Tests/Database/InstanceTests.cs
+++ b/Tests/Realm.Tests/Database/InstanceTests.cs
@@ -40,6 +40,23 @@ namespace Realms.Tests.Database
         }
 
         [Test]
+        public void ReadOnlyInstance_ThrowsOnRefresh()
+        {
+            var path = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "temp.realm");
+            var realm = Realm.GetInstance(path);
+            realm.Dispose();
+            var config = new RealmConfiguration(path)
+            {
+                IsReadOnly = true
+            };
+
+            realm = Realm.GetInstance(config);
+            Assert.Throws<RealmException>(() => realm.Refresh(), "Can't refresh a read-only Realm.");
+            realm.Dispose();
+            DeleteRealmWithRetries(realm);
+        }
+
+        [Test]
         public void InstanceIsClosedByDispose()
         {
             Realm temp;
--- a/Tests/Realm.Tests/Database/InstanceTests.cs
+++ b/Tests/Realm.Tests/Database/InstanceTests.cs
@@ -42,18 +42,16 @@ namespace Realms.Tests.Database
         [Test]
         public void ReadOnlyInstance_ThrowsOnRefresh()
         {
-            var path = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "temp.realm");
-            var realm = Realm.GetInstance(path);
+            var path = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, Guid.NewGuid().ToString());
+            var realm = GetRealm(path);
             realm.Dispose();
             var config = new RealmConfiguration(path)
             {
                 IsReadOnly = true
             };
 
-            realm = Realm.GetInstance(config);
+            realm = GetRealm(config);
             Assert.Throws<RealmException>(() => realm.Refresh(), "Can't refresh a read-only Realm.");
-            realm.Dispose();
-            DeleteRealmWithRetries(realm);
         }
 
         [Test]
--- a/Tests/Realm.Tests/Database/InstanceTests.cs
+++ b/Tests/Realm.Tests/Database/InstanceTests.cs
@@ -42,16 +42,27 @@ namespace Realms.Tests.Database
         [Test]
         public void ReadOnlyInstance_ThrowsOnRefresh()
         {
-            var path = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, Guid.NewGuid().ToString());
-            var realm = GetRealm(path);
+            var config = new RealmConfiguration(Guid.NewGuid().ToString());
+            var realm = GetRealm(config);
             realm.Dispose();
-            var config = new RealmConfiguration(path)
-            {
-                IsReadOnly = true
-            };
+
+            config.IsReadOnly = true;
 
             realm = GetRealm(config);
             Assert.Throws<RealmException>(() => realm.Refresh(), "Can't refresh a read-only Realm.");
+        }
+
+        [Test]
+        public void GetTwice_ReadOnlyInstance_DoesNotThrow()
+        {
+            var config = new RealmConfiguration(Guid.NewGuid().ToString());
+            var realm = GetRealm(config);
+            realm.Dispose();
+
+            config.IsReadOnly = true;
+
+            _ = GetRealm(config);
+            Assert.DoesNotThrow(() => GetRealm(config));
         }
 
         [Test]
--- a/Tests/Realm.Tests/Database/InstanceTests.cs
+++ b/Tests/Realm.Tests/Database/InstanceTests.cs
@@ -935,6 +935,21 @@ namespace Realms.Tests.Database
         }
 
         [Test]
+        public void Realm_Freeze_ReadOnly()
+        {
+            var config = new RealmConfiguration(Guid.NewGuid().ToString());
+            var realm = GetRealm(config);
+            realm.Dispose();
+
+            config.IsReadOnly = true;
+
+            realm = GetRealm(config);
+            Realm frozenRealm = null;
+            Assert.DoesNotThrow(() => frozenRealm = realm.Freeze());
+            frozenRealm.Dispose();
+        }
+
+        [Test]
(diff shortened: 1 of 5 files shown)


Tests/Realm.Tests/Database/InstanceTests.cs
```diff
@@ -39,6 +39,23 @@ public void GetInstanceTest()
             GetRealm().Dispose();
         }
 
+        [Test]
+        public void ReadOnlyInstance_ThrowsOnRefresh()
+        {
+            var path = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "temp.realm");
```
nirinchev, line 45: Let's not build deterministic paths and instead use `Guid.NewGuid()` here.

Tests/Realm.Tests/Database/InstanceTests.cs
```diff
@@ -39,6 +39,23 @@ public void GetInstanceTest()
...
+            var realm = Realm.GetInstance(path);
```
nirinchev, line 46: This should use `GetRealm` instead which will take care of cleaning it up and deleting it.

Tests/Realm.Tests/Database/InstanceTests.cs
```diff
@@ -39,6 +39,23 @@ public void GetInstanceTest()
...
+            realm.Dispose();
+            var config = new RealmConfiguration(path)
+            {
+                IsReadOnly = true
+            };
+
+            realm = Realm.GetInstance(config);
```
nirinchev, line 53:
```suggestion
            realm = GetRealm(config);
```

Tests/Realm.Tests/Database/InstanceTests.cs
```diff
@@ -39,6 +39,23 @@ public void GetInstanceTest()
...
+            Assert.Throws<RealmException>(() => realm.Refresh(), "Can't refresh a read-only Realm.");
+            realm.Dispose();
+            DeleteRealmWithRetries(realm);
```
nirinchev, line 56: You won't need those when you switch to using GetRealm

nirinchev (changes requested):
This also needs tests that verify that calling `GetRealm` twice for a readonly realm will work. I assume this would have failed without this change:

```csharp
var config = new RealmConfiguration(Guid.NewGuid().ToString());
var realm = GetRealm(config);
realm.Dispose();

config.IsReadOnly = true;

_ = GetRealm(config);
_ = GetRealm(config);
```

Tests/Realm.Tests/Database/InstanceTests.cs
LaPeste: Can you explain the reasoning for this?

nirinchev: There are several:
1. We have code in our tests that modifies the base path for where we store Realm files to be the temp directory.
2. If you hardcode a path, this means that running the test twice may result in a non-clean slate run. For example, if the test crashes/throws halfway through, this would prevent us from deleting the Realm file. Then the next time you run it, you'll be using the existing file rather than a brand new one.

Tests/Realm.Tests/Database/InstanceTests.cs
nirinchev: You can still dispose of the Realm even if you use `GetRealm`. Using GetRealm also ensures that the file gets deleted, even if the test fails.

Tests/Realm.Tests/Database/InstanceTests.cs
LaPeste: Thank you. That makes totally sense.

nirinchev (approved):
Minor comment about the changelog, otherwise, it's ready to go.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -128,7 +128,13 @@ Realm::Config get_shared_realm_config(Configuration configuration, SyncConfigura
 
 inline SharedRealm* new_realm(SharedRealm realm)
 {
-    if (!realm->refresh()) {
+    if (!realm->config().immutable()) {
+        if (!realm->refresh()) {
+            // this should only be needed for immutable frozen realms
```
nirinchev, line 133: [nit] This comment is not correct or it's placed in the wrong line - if we're here, then the Realm we're operating on is not immutable.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -128,7 +128,13 @@ Realm::Config get_shared_realm_config(Configuration configuration, SyncConfigura
 
 inline SharedRealm* new_realm(SharedRealm realm)
 {
-    if (!realm->refresh()) {
+    if (!realm->config().immutable()) {
```
nirinchev, line 131: This if statement can be reduced to:

```cpp
// If a Realm is immutable, it can't be refreshed
// If we can't refresh the Realm, make sure we begin a read transaction
// as a lot of functionality expects an active read transaction and OS
// doesn't always start one automatically.
if (realm->config().immutable() || !realm->refresh()) {
    realm->read_group();
}
```

Issue #1373, opened 2017-04-28
Sorting by an interface property fails with KeyNotFoundException


PedroLV:
Hello:

This code was working on previous xamarin version , but now on Realm Xamarin 1.2 fails.

## Goals
Execute a simple query like this:

`T obj = realm.All<T>().OrderByDescending(x => x.Id).FirstOrDefault();`

## Expected Results
Get an object from database. This code was working on realm 1.1.1

The query produces an exception

Error --> **System.Collections.Generic.KeyNotFoundException**: The given key was not present in the dictionary.
  at System.ThrowHelper.ThrowKeyNotFoundException () [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/10.8.0.175/src/mono/mcs/class/referencesource/mscorlib/system/throwhelper.cs:68 
  at System.Collections.Generic.Dictionary`2[TKey,TValue].get_Item (TKey key) [0x00021] in ....

## Steps to Reproduce
Create a simple class, and run the code.

## Code Sample
```
public class Dog : RealmObject, IDocument
    {
        [PrimaryKey]
        public Int32 Id { get; set; }
        public string Name { get; set; }
        public int Age { get; set; }
        
    }

public interface IDocument
    {
         Int32 Id { get; set; }
    }
//code
var realm = Realm.GetInstance();

            realm.Write(() =>
            {
                try
                {
                    Int32 newId = NewId<Dog>(realm); 
                  
                    realm.Add(new Dog { Id = newId, Name = "Rex", Age = 1 });
                }
                catch(System.Exception ex)
                {
                    Debug.WriteLine("Error --> " + ex.ToString()); 
                }
            });

private  Int32 NewId<T>( Realm realm) where T : Realms.RealmObject, IDocument
        {
            try
            {
                 //this query gets an exception
                T obj = realm.All<T>().OrderByDescending(x => x.Id).FirstOrDefault();
                if (obj == null)
                {
                    return 1;
                }
                return obj.Id + 1;
            }
            catch (InvalidOperationException)
            {
                return 1;
            }
        }


```

## Version of Realm and Tooling
Realm 1.2 Xamarin
Visual Studio 2017
Xamarin IOS

* Realm Object Server Version: none
* Flavor: Developer
* Server OS & Version: Windows 10
* Client SDK Version: Xamarin IOS SDK 10.8.0.174
* Client OS & Version: IOS 10.3

istx25:
[...]

PR #2750, opened 2021-12-13
Fix `TraverseSort` to support properties declared by interfaces

daawaan:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/1373

##  TODO

* [ ] Changelog entry
* [ ]

## Questions

I'm a first time contributor and I would like to ask what are changelog entries specifically?

### DIFF
--- a/Realm/Realm/Linq/RealmResultsVisitor.cs
+++ b/Realm/Realm/Linq/RealmResultsVisitor.cs
@@ -158,7 +158,7 @@ namespace Realms
 
             while (expression != null)
             {
-                var type = expression.Member.DeclaringType;
+                var type = expression.Expression.Type;
                 var typeName = type.GetMappedOrOriginalName();
                 if (!_realm.Metadata.TryGetValue(typeName, out var metadata))
                 {


nirinchev:
Hey @daawaan, thanks for the PR and sorry for the delay in getting this reviewed - most of the team has been off around the holidays. The change makes sense to me, but we'll need to review it more thoroughly to ensure it doesn't break existing behavior.

It appears our CI setup doesn't work well for outside contributors, so we'll need to run the tests manually. @papafe can you take point on this?

papafe:
@nirinchev I can run the tests locally, but we need to see how can make CI work for external contributors. 
Just the last couple of weeks me and @LaPeste were looking at a LINQ problem, and there was an issue where some nodes were actually of different types on different environments. So we definitely need the CI to be sure this fix works everywhere

papafe:
@daawaan To answer your question, changelog entries refer to the `CHANGELOG.md` file, where we log all the bug fixes and enhancements that are added in a certain version. We can take care of that ourselves at the end, no problem.

papafe:
The CI is green on the PR with the same code, so this could be merged. @nirinchev do we fix the changelog in another PR after this is merged?

nirinchev:
Let's merge this and fix the changelog as a follow-up

papafe (approved):

PR #2766, opened 2022-01-12
Added fallback path

### LINKED ISSUE #2751: [Bug]: Cannot open realm files on vfat / exFAT filesystems on linux
### What happened?

realm-dotnet and realm studio will not open realm files when placed on an ExFAT or similar filesystem.

Reported [here](https://github.com/ppy/osu/discussions/16063#discussioncomment-1795753).

Note that I have tested the same scenario on macOS and it works as expected.

### Repro steps

- Move an existing realm file to an ExFAT volume on linux
- Attempt to open via `Realm.GetInstance` or Realm Studio

### Version

.net 5.0

Local Database only

Other

### Client OS and version

Linux (tested multiple variants)

### Code snippets

_No response_

```shell
Unhandled exception. Realms.Exceptions.RealmFileAccessErrorException: No such file or directory
   at Realms.NativeException.ThrowIfNecessary(Func`2 overrider)
   at Realms.SharedRealmHandle.Open(Configuration configuration, RealmSchema schema, Byte[] encryptionKey)
   at Realms.RealmConfiguration.CreateRealm()
   at Realms.Realm.GetInstance(RealmConfigurationBase config)
   at osu.Game.Database.RealmContextFactory.CreateContext() in /var/lib/buildkite-agent/builds/debian-gnu-linux-vm-1/ppy/osu/osu.Game/Database/RealmContextFactory.cs:line 177
```

### Relevant log output

_No response_

papafe:
Added new config parameter for the fallback path where named pipes used by realm should be created.

Fixes #2751 

##  TODO

* [x] Changelog entry

### DIFF
--- a/Realm/Realm/Configurations/RealmConfigurationBase.cs
+++ b/Realm/Realm/Configurations/RealmConfigurationBase.cs
@@ -209,7 +209,7 @@ namespace Realms
             return new Native.Configuration
             {
                 Path = DatabasePath,
-                FallbackPipePath = FallbackPipePath ?? string.Empty,
+                FallbackPipePath = FallbackPipePath,
                 schema_version = SchemaVersion,
                 enable_cache = EnableCache,
                 max_number_of_active_versions = MaxNumberOfActiveVersions
--- a/Realm/Realm/Configurations/RealmConfigurationBase.cs
+++ b/Realm/Realm/Configurations/RealmConfigurationBase.cs
@@ -46,6 +46,17 @@ namespace Realms
         public string DatabasePath { get; protected set; }
 
         /// <summary>
+        /// Gets or sets the path where the named pipes used by Realm can be placed.
+        /// </summary>
+        /// <remarks>
+        /// In the greatest majority of cases this value can be left null.
+        /// It needs to be set if the Realm is opened on a filesystem where a named pipe cannot be created, such as external storage on Android that uses FAT32.
+        /// In this case the path should point to a location on a filesystem where the pipes can be created.
+        /// </remarks>
+        /// <value>The path where named pipes can be created.</value>
+        public string FallbackPipePath { get; protected set; }
+
+        /// <summary>
         /// Gets or sets a value indicating whether the Realm will be open in dynamic mode. If opened in dynamic mode,
         /// the schema will be read from the file on disk.
         /// </summary>
@@ -198,6 +209,7 @@ namespace Realms
             return new Native.Configuration
             {
                 Path = DatabasePath,
+                FallbackPipePath = FallbackPipePath ?? string.Empty,
                 schema_version = SchemaVersion,
                 enable_cache = EnableCache,
                 max_number_of_active_versions = MaxNumberOfActiveVersions
--- a/Realm/Realm/Configurations/RealmConfigurationBase.cs
+++ b/Realm/Realm/Configurations/RealmConfigurationBase.cs
@@ -49,12 +49,12 @@ namespace Realms
         /// Gets or sets the path where the named pipes used by Realm can be placed.
         /// </summary>
         /// <remarks>
-        /// In the greatest majority of cases this value can be left null.
+        /// In the vast majority of cases this value should be left null.
         /// It needs to be set if the Realm is opened on a filesystem where a named pipe cannot be created, such as external storage on Android that uses FAT32.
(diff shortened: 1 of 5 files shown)


Realm/Realm/Configurations/RealmConfigurationBase.cs
```diff
@@ -45,6 +45,17 @@ public abstract class RealmConfigurationBase
...
+        /// <summary>
+        /// Gets or sets the path where the named pipes used by Realm can be placed.
+        /// </summary>
+        /// <remarks>
+        /// In the greatest majority of cases this value can be left null.
+        /// It needs to be set if the Realm is opened on a filesystem where a named pipe cannot be created, such as external storage on Android that uses FAT32.
+        /// In this case the path should point to a location on a filesystem where the pipes can be created.
+        /// </remarks>
+        /// <value>The path where named pipes can be created.</value>
+        public string FallbackPipePath { get; protected set; }
```
nirinchev, line 57: Why is this protected set?

Realm/Realm/Configurations/RealmConfigurationBase.cs
```diff
@@ -198,6 +209,7 @@ internal Native.Configuration CreateNativeConfiguration()
             return new Native.Configuration
             {
                 Path = DatabasePath,
+                FallbackPipePath = FallbackPipePath ?? string.Empty,
```
nirinchev, line 212: I think we should leave that as `null` rather than `string.Empty`.

Realm/Realm/Configurations/RealmConfigurationBase.cs
```diff
@@ -45,6 +45,17 @@ public abstract class RealmConfigurationBase
         /// <value>The absolute path to the Realm.</value>
         public string DatabasePath { get; protected set; }
 
+        /// <summary>
+        /// Gets or sets the path where the named pipes used by Realm can be placed.
+        /// </summary>
+        /// <remarks>
+        /// In the greatest majority of cases this value can be left null.
```
nirinchev, line 52:
```suggestion
        /// In the vast majority of cases this value should be left null.
```

Realm/Realm/Configurations/RealmConfigurationBase.cs
papafe: And then we take care of null values in the setter? Or how does it work with passing null values to native?

nirinchev: In the setter, we can pass a null string, we just need to use the `IntPtrLength` extension method to make sure we don't cause a NRE when reading the string length. Then in native we should check for `nullptr` like:

```cpp
if (configuration.fallback_path) {
  config.fifo_files_fallback_path = Utf16StringAccessor(configuration.fallback_path, configuration.fallback_path_len);
}
```

peppy:
Since the configuration property has `Fallback` in its name, does this imply it will only be used in cases that creation beside the realm file fails?

One potential use case I have is wanting to *always* create the pipe files in a specified temporary folder to avoid compatibility issues (specifically, if a Parallels Desktop VM sees the pipe file it will seize up and require a `kill -9` of the VM to recover - can be quite detrimental when a realm file is created in my home directory for whatever reason).

papafe:
@peppy I agree the name is misleading, but actually if set it is used all the time.

peppy:
Thanks for confirmation and for the implementation. Should help us heaps.

nirinchev (approved):

smoogipoo:
Hi, I'm not sure that this is working on Linux (`5.10.79-1-MANJARO`). https://github.com/realm/realm-dotnet/pull/2766#issuecomment-1010876759 indicates that it should have an effect at all times if I understand correctly, though it still seems to create the pipe file in the original path.

We're choosing the path as follows (I've tested both as directory and as file):
```csharp
string tempPathLocation = Path.Combine(Path.GetTempPath(), @"lazer");
if (!Directory.Exists(tempPathLocation))
    Directory.CreateDirectory(tempPathLocation);
```

@peppy Claims that it works for him on macOS on the same project, so this may be a linux-specific issue?

papafe:
@smoogipoo I am really sorry, I gave you the wrong information because I misunderstood a piece of code. As you've noticed that path is being used only if it's not possible to create the pipe next to the realm file. 

As an additional note, the fallback path should just be a directory, no need to point to a file, and should have a trailing slash

PR #2768, opened 2022-01-13
Use the admin api instead of the import functionality

nirinchev:
## Description

This replace the import/export approach with using the admin API to configure BaaS apps. It's intended to prepare us for FLX where the import is not wired up yet.

##  TODO

* [x] Changelog entry
* [ ] Tests (if applicable)


papafe:
It's a little dense to follow everything, but it seems to make sense

fealebenpae (commented):
It looks okay, but I believe as it is written it'll only work when running against local (aka docker) baas, and not Atlas, because of how URLs are hardcoded in `CreateApp`.
Also I think we might want to look at `System.CommandLine` when we do introduce the ability to test against an Atlas cluster.

PR #2777, opened 2022-01-20
Implementation of the Client reset api for seamless loss

LaPeste:
## Description
Implement the Client reset api for seamless loss and integration tests for it.

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)
* [x] Write documentation on public API
* [x] Add copyrights to new files


Realm/Realm/Exceptions/ErrorCode.cs
```diff
@@ -18,6 +18,16 @@
 
 namespace Realms.Sync.Exceptions
 {
+
+    // ATTENTION: given that managed does not differentiate errors by category like core does,
```
nirinchev, line 22: Might be worth confirming this with someone in Core, but as far as I remember, client error codes are in the 100-199 range, while protocol errors are 200-299, so you can use the value to infer the category.

LaPeste: I could see that [Protocol Error](https://github.com/realm/realm-core/blob/9874d5164927ea39273b241a5af14b596a3233e9/src/realm/sync/protocol.hpp#L197) has `connection level errors` and `session level errors`. `connection level errors` use the 100-199, so that'd collide with the [ClientError](https://github.com/realm/realm-core/blob/9874d5164927ea39273b241a5af14b596a3233e9/src/realm/sync/client_base.hpp#L73) which also use the 100-199 range. Because of all of this I decided to go with the solution you see here.
In case I misunderstood something I'll ask core about it.
Or if you think we'll never want to pass `connection level errors` then, yes we can avoid all this. But "never want to" seems unlikely to me.

Realm/Realm/Exceptions/Sync/ErrorCode.cs
```diff
@@ -29,6 +38,11 @@ public enum ErrorCode
         /// </summary>
         Unknown = -1,
 
+        /// <summary>
+        /// A fatal error was encountered which prevents the completion of a client reset.
+        /// </summary>
+        AutoClientResetFailed_Cl = 132,
```
fealebenpae, line 44: I don't like adding this new value to the public enum. I suggest we add a new internal `ClientError` enum and add this value there. Then we add a new internal `SimulateError` overload that takes `ClientError`. Finally in `Session.ReportErrorForTesting` and in `realm_syncsession_report_error_for_testing` we add a new enum parameter (same enum in C# and C++) that indicates whether the error category is client or protocol.

fealebenpae (changes requested):
There are a few small things I noticed you can address on your own, but I also promised to help with the sync configuration handle cleanup, so please get in touch for a pair programming session 😄

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -225,14 +252,92 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
+            var syncConfiguration = (SyncConfigurationBase)syncConfigHandle.Target;
+
+            if (syncConfiguration.ClientResetHandler == null)
+            {
+                return true;
+            }
+
+            var discardLocalResetHandler = (DiscardLocalResetHandler)syncConfiguration.ClientResetHandler;
+            var schema = syncConfiguration.GetSchema();
+            var realmBefore = new Realm(new UnownedRealmHandle(beforeFrozen), syncConfiguration, schema);
```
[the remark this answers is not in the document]
nirinchev: I don't think we need to dispose of Realms that are unowned - we can, but it doesn't really do anything.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -225,14 +252,92 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
 
-                Session.RaiseError(session, exception);
+                if (syncConfigurationBase.SyncErrorHandler != null)
+                {
+                    syncConfigurationBase.SyncErrorHandler.OnError?.Invoke(session, exception);
+                }
+                else
+                {
+                    // after deprecation: this will need to go when Session.Error is fully deprecated
+                    Session.RaiseError(session, exception);
```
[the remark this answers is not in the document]
LaPeste: I discussed about this with @nirinchev while I was working on it. The behaviour you see is what we agreed on. @nirinchev, what do you think of @fealebenpae's point of view?

Realm/Realm/Handles/SessionHandle.cs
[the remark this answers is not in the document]
LaPeste: Then I avoid the disposal to keep it clean.

Realm/Realm/Configurations/SyncConfigurationBase.cs
```diff
@@ -40,6 +41,17 @@ public abstract class SyncConfigurationBase : RealmConfigurationBase
...
+        /// <summary>
+        /// Gets or sets a subclass of <see cref="ClientResetHandlerBase"/> to specify actions to be taken for the selected Client Reset strategy: <see cref="ManualRecoveryHandler"/> or <see cref="DiscardLocalResetHandler"/>.
+        /// If nothing is set, the strategy defaults to <see cref="DiscardLocalResetHandler"/> with no custom actions set for the before and after synchronization.
+        /// </summary>
+        public ClientResetHandlerBase ClientResetHandler { get; set; }
+
+        /// <summary>
+        /// Gets or sets a callback to handle errors that happen on a session of a syncronized Realm.
+        /// </summary>
+        public SyncErrorHandler SyncErrorHandler { get; set; }
```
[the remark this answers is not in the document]
LaPeste: I needed a way to share a delegate signature across classes that aren't related. I'll find a better way.

Realm/Realm/Sync/ErrorHandling/SyncErrorHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+using Realms.Sync.Exceptions;
+
+namespace Realms.Sync.ErrorHandling
+{
+    /// <summary>
+    /// Handler triggered whenever a sync error happens on a synchronized Realm.
+    /// To be noted that a client reset is not a sync error and in order to handle that subclasses of <see cref="ClientResetHandlerBase"/> are available.
+    /// </summary>
+    /// <seealso href="https://docs.mongodb.com/realm/sync/overview/">Sync Overview Docs</seealso>
+    public class SyncErrorHandler
```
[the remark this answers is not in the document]
LaPeste: Yes, you're totally correct.

Realm/Realm/Exceptions/Sync/ErrorCode.cs
LaPeste: Unfortunately, `SimulateError` needs to be public in order to be visible outside the assembly so that it can be used by the tests. Because of this all of the types it uses in the signature need to be of the same visibility. Do you see any problem with `ClientError` being public?

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -200,20 +212,36 @@ protected override void Unbind()
...
+                            discardLocalResetHandler.ManualResetFallback?.Invoke(session, (ClientResetException)exception);
+                        }
+                        else if (syncConfigurationBase.ClientResetHandler is ManualRecoveryHandler manualRecoveryHandler)
+                        {
+                            manualRecoveryHandler.OnClientReset?.Invoke(session, (ClientResetException)exception);
+                        }
+
+                        // after deprecation: this won't be require anymore when Session.Error will be fully deprecated
+                        return;
+                    }
```
nirinchev, line 244: This will be more readable as a swtich

Realm/Realm/Handles/SessionHandle.cs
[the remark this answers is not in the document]
nirinchev: I think it makes sense to invoke one or the other, but not both - if a user explicitly adds an `OnSessionError` handler, they probably don't want to handle the same error twice.

Realm/Realm/Handles/SessionHandle.cs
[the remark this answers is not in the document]
nirinchev: With https://github.com/realm/realm-dotnet/pull/2795 disposing these unowned Realms becomes important, so let's do it after all :D

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -225,14 +253,82 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
+            var syncConfiguration = (SyncConfigurationBase)syncConfigHandle.Target;
+
+            if (syncConfiguration.ClientResetHandler == null)
+            {
+                return true;
+            }
+
+            var discardLocalResetHandler = (DiscardLocalResetHandler)syncConfiguration.ClientResetHandler;
+            var schema = syncConfiguration.GetSchema();
+            var realmBefore = new Realm(new UnownedRealmHandle(beforeFrozen), syncConfiguration, schema);
```
nirinchev, line 286: Why do we construct the realm before we even verify that `OnBeforeReset` is not null?

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -225,14 +253,82 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
+
+            try
+            {
+                discardLocalResetHandler.OnBeforeReset?.Invoke(realmBefore);
+            }
+            catch (Exception ex)
+            {
+                Logger.Default.Log(LogLevel.Warn, $"An error has occurred while executing SyncConfigurationBase.OnBeforeReset during a client reset: {ex}");
```
nirinchev, line 294: Should this be an error instead?

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -225,14 +253,82 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
+            // no clientResetHandler means don't do anything for this callback
+            if (syncConfiguration.ClientResetHandler == null)
+            {
+                return true;
+            }
+
+            var discardLocalResetHandler = (DiscardLocalResetHandler)syncConfiguration.ClientResetHandler;
+            var schema = syncConfiguration.GetSchema();
+            var realmBefore = new Realm(new UnownedRealmHandle(beforeFrozen), syncConfiguration, schema);
+            var realmAfter = new Realm(new UnownedRealmHandle(after), syncConfiguration, schema);
```
nirinchev, line 317: Same here - don't create these if the user hasn't specified `OnAfterReset` handler.

Realm/Realm/Realm.csproj
```diff
@@ -116,6 +116,9 @@
   <ItemGroup>
     <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
   </ItemGroup>
+  <ItemGroup>
+	<Folder Include="Sync\ErrorHandling\" />
```
nirinchev, line 120: Why do we need this?

Realm/Realm/Sync/ClientResyncMode.cs
```diff
@@ -0,0 +1,29 @@
+﻿// ////////////////////////////////////////////////////////////////////////////
+// //
+// // Copyright 2021 Realm Inc.
```
nirinchev, line 3: Fix header

Realm/Realm/Sync/ClientResyncMode.cs
```diff
@@ -0,0 +1,29 @@
...
+// ////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync
+{
+        internal enum ClientResyncMode : byte
+        {
+            // Fire a client reset error
+            Manual = 0,
+
+            // Discard local changes, without disrupting accessors or closing the Realm
```
nirinchev, line 26: We don't need these comments. Also, since DiscardLocal is the default, should it be 0?

Realm/Realm/Sync/ErrorHandling/ClientResetHandlerBase.cs
```diff
@@ -0,0 +1,45 @@
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2021 Realm Inc.
```
nirinchev, line 3:
```suggestion
// Copyright 2022 Realm Inc.
```

Realm/Realm/Sync/ErrorHandling/ClientResetHandlerBase.cs
```diff
@@ -0,0 +1,45 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+using Realms.Sync.Exceptions;
+
+namespace Realms.Sync.ErrorHandling
+{
+    /// <summary>
+    /// In order to specify what strategy to use when a client reset happens, instanciate a subclass of <see cref="ClientResetHandlerBase"/>
+    /// and assign it to the chosen subclass of <see cref="SyncConfigurationBase"/> to its <see cref="SyncConfigurationBase.ClientResetHandler"/> field.
```
nirinchev, line 25:
```suggestion
    /// The base class for the different client reset handler types. The possible implementations are <see cref="DiscardLocalResetHandler"/> and <see cref="ManualResetHandler"/>.
    /// To use either of them, create a new instance and assign it to <see cref="SyncConfigurationBase.ClientResetHandler"/> on the configuration
    /// you use to open the synchronized <see cref="Realm"/> instance.
```

Realm/Realm/Sync/ErrorHandling/DiscardLocalResetHandler.cs
```diff
@@ -0,0 +1,66 @@
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2021 Realm Inc.
```
nirinchev, line 3:
```suggestion
// Copyright 2022 Realm Inc.
```

Realm/Realm/Sync/ErrorHandling/DiscardLocalResetHandler.cs
```diff
@@ -0,0 +1,66 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync.ErrorHandling
+{
+    /// <summary>
+    /// A client reset strategy where all the not yet synchronized data is automatically discarded and a fresh copy of the synchronized Realm is obtained. To be noted that the freshly downloaded
```
nirinchev, line 22: These lines are very long - I know we don't have a hard guidance on line length, but if they overflow the github diff view, it's probably too long 😄

Realm/Realm/Sync/ErrorHandling/DiscardLocalResetHandler.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <summary>
+        /// Callback that indicates a Client Reset is about to happen.
+        /// </summary>
+        /// <param name="beforeFrozen">
+        /// The frozen <see cref="Realm"/> before the reset.
+        /// </param>
+        public delegate void BeforeResetCallback(Realm beforeFrozen);
+
+        /// <summary>
+        /// Callback that indicates a Client Reset is about to happen.
```
nirinchev, line 40:
```suggestion
        /// Callback that indicates a Client Reset has just happened.
```

nirinchev (commented):
Went through most of the C# code - will review the rest later, but figured it's worth posting what I have so far.

The current approach with reset handlers is mildly problematic as it creates a GCHandle to a mutable type, accessible to the user, meaning at any point they can change the handler or the callback properties of the handler. We should probably clone the config before storing it in native, but curious to hear what others think.

Realm/Realm/Configurations/SyncConfigurationBase.cs
```diff
@@ -103,11 +118,14 @@ internal override async Task<Realm> CreateRealmAsync(CancellationToken cancellat
...
         internal virtual Native.SyncConfiguration CreateNativeSyncConfiguration()
         {
+            var syncConfHandle = GCHandle.Alloc(this);
             return new Native.SyncConfiguration
             {
                 SyncUserHandle = User.Handle,
                 session_stop_policy = SessionStopPolicy,
                 schema_mode = Schema == null ? SchemaMode.AdditiveDiscovered : SchemaMode.AdditiveExplicit,
+                client_resync_mode = ClientResetHandler is DiscardLocalResetHandler ? ClientResyncMode.DiscardLocal : ClientResyncMode.Manual,
+                managed_sync_configuration_handle = GCHandle.ToIntPtr(syncConfHandle),
```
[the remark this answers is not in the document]
LaPeste: Since we decided to move the session error handling in the `SyncConfigurationBase` now [SessionHandle](https://github.com/realm/realm-dotnet/blob/b300da911a10c3e66870f108caf895d8a7174ae7/Realm/Realm/Handles/SessionHandle.cs#L258) needs to get a hold of that error handler. In order not to send to native a handle per callback I sent the whole configuration, which when managed gets again knows what to do with it.
[Here](https://github.com/realm/realm-dotnet/blob/b300da911a10c3e66870f108caf895d8a7174ae7/wrappers/src/shared_realm_cs.cpp#L125) here the session error handler is called from native.
Shall I review how things are passed around? The only thing is that if there will be another callback that will be added in the future, one more will need to be passed through method parameters. I thought it was a little easier to manage in the long run.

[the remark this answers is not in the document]
nirinchev: I was thinking we can pass the ClientResetHandler instead of the config. Then in the callback, we cast the GCHandle to `ClientResetHandlerBase` and invoke whatever methods we want on it.

Realm/Realm/Realm.csproj
LaPeste: It's where I put the error handling for sync (all the client reset handlers). Instead of putting those in the sync folder. If you want I can put them in the parent sync folder instead.

Realm/Realm/Sync/ClientResyncMode.cs
LaPeste: These map directly to [core values](https://github.com/realm/realm-core/blob/76661758162132e9af4245bc6842980e4e0fdd1c/src/realm/sync/config.hpp#L95). Doing so avoided the need for a native mirror in our wrappers that need to be converted to the core ones.

Realm/Realm/Realm.csproj
nirinchev: I mean, MSBuild should just pick up the folder, I don't think we need to explicitly add it in the project file.

Realm/Realm/Sync/ClientResyncMode.cs
nirinchev: I see, kind of weird they chose 0 for manual, but let's keep it aligned then.

Realm/Realm/Configurations/SyncConfigurationBase.cs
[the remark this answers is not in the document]
LaPeste: Yes I understand that. But what I was trying to say is that we also need to pass to native the [SyncConfigurationBase.OnSessionError](https://github.com/realm/realm-dotnet/blob/66db6371004aced05320eed5e5f8005765f509b1/Realm/Realm/Configurations/SyncConfigurationBase.cs#L68) in order to execute it from [SessionHandle](https://github.com/realm/realm-dotnet/blob/b300da911a10c3e66870f108caf895d8a7174ae7/Realm/Realm/Handles/SessionHandle.cs#L258).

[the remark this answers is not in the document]
nirinchev: Ah, I understand, didn't see that. Let's discuss as part of the broader mutable struct discussion.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -115,9 +112,37 @@ Realm::Config get_shared_realm_config(Configuration configuration, SyncConfigura
...
+        for (const auto& p : error.user_info) {
+            user_info_pairs.push_back(std::make_pair(const_cast<char*>(p.first.c_str()), const_cast<char*>(p.second.c_str())));
+        }
+
+        s_session_error_callback(new SharedSyncSession(session), error.error_code.value(), to_capi_value(error.message), user_info_pairs.data(), user_info_pairs.size(), error.is_client_reset_requested(), sync_configuration_handle->handle());
+    };
+
     config.sync_config->client_resync_mode = ClientResyncMode::Manual;
     config.sync_config->stop_policy = sync_configuration.session_stop_policy;
+    config.sync_config->client_resync_mode = sync_configuration.client_resync_mode;
```
nirinchev, line 130: This is already set on line 128.

Realm/Realm/Sync/ErrorHandling/DiscardLocalResetHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <summary>
+        /// Callback that indicates a Client Reset has just happened.
+        /// </summary>
+        /// <param name="beforeFrozen">
+        /// The frozen <see cref="Realm"/> as it was before the reset.
+        /// </param>
+        /// <param name="after">
+        /// The <see cref="Realm"/> after the client reset.
+        /// </param>
+        public delegate void AfterResetCallback(Realm beforeFrozen, Realm after);
```
nirinchev, line 49: For these callbacks, we should clarify whether `after` is already in a write transaction or one should create their own as well as that the lifetime of these Realms and all objects within them is scoped to the callback - i.e. you can't store references to objects or the Realm itself in `Before` and expect them to be valid in `After`.

Realm/Realm/Sync/ErrorHandling/DiscardLocalResetHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        /// Gets or sets the callback that indicates a Client Reset is about to happen.
+        /// Among other things, you can use this call to temporarily store the before Realm as a backup and in the <see cref="OnAfterReset"/> callback merge the changes, if necessary.
+        /// </summary>
+        public BeforeResetCallback OnBeforeReset { get; set; }
+
+        /// <summary>
+        /// Gets or sets the callback that indicates a Client Reset just happened. Special custom actions can be taken at this point like merging local changes if the "before" realm was stored during <see cref="BeforeResetCallback"/>.
```
nirinchev, line 58: Can you split these lines to make sure they roughly fit the screen. Also, for properties, we need to supply a `value` tag otherwise the generated docs look a little ugly.

Realm/Realm/Sync/ErrorHandling/ManualRecoveryHandler.cs
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync.ErrorHandling
+{
+    /// <summary>
+    /// A client reset strategy where the user needs to fully take care of a client reset. <see cref="OnClientReset"/> is the callback where user cose is supplied.
+    /// For more information see <see cref="ClientResetHandlerBase"/>.
```
nirinchev, line 23: Again - probably not that useful to link back to the parent class; split long lines; add value tags for the property docs.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -100,6 +103,347 @@ public void Session_Error_ShouldPassCorrectSession()
...
+                        Assert.IsFalse(manualFallbackTriggered);
+                        onBeforeTriggered = true;
+                        throw new Exception("Exception thrown in OnBeforeReset");
+                    },
+                    OnAfterReset = (beforeFrozen, after) =>
+                    {
+                        onAfterResetTriggered = true;
+                        tcs.TrySetResult(true);
+                    },
+                    ManualResetFallback = (session, err) =>
```
nirinchev, line 242: Can we also check the error here?

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -100,6 +103,347 @@ public void Session_Error_ShouldPassCorrectSession()
...
+                GetSession(realm).SimulateClientReset("simulated client reset");
+
+                await tcs.Task;
+
+                Assert.IsTrue(manualOnClientResetTriggered);
+            });
+        }
+
+        [Test]
+        public void Session_ClientReset_ManualRecoveryHandler_When_Exception_OnBeforeReset()
```
nirinchev, line 218: Should this be DiscardLocalRecoveryHandler? Same for next test.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -100,6 +103,347 @@ public void Session_Error_ShouldPassCorrectSession()
...
+                    },
+                    OnAfterReset = (beforeFrozen, after) =>
+                    {
+                        Assert.IsTrue(onBeforeTriggered);
+                        Assert.IsFalse(onAfterResetTriggered);
+                        Assert.IsFalse(manualFallbackTriggered);
+                        onAfterResetTriggered = true;
+                        throw new Exception("Exception thrown in OnAfterReset");
+                    },
+                    ManualResetFallback = (session, err) =>
```
nirinchev, line 291: Check error

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -100,6 +103,347 @@ public void Session_Error_ShouldPassCorrectSession()
...
+                await tcs.Task;
+
+                Assert.IsTrue(manualFallbackTriggered);
+                Assert.IsTrue(onBeforeTriggered);
+                Assert.IsTrue(onAfterResetTriggered);
+            });
+        }
+
+        [Test]
+        public void Session_SyncError_Handler()
```
nirinchev, line 314: This name is misleading as it's testing the SyncConfiguration.OnSessionError handler, not Session.Error.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -100,6 +103,347 @@ public void Session_Error_ShouldPassCorrectSession()
...
+        {
+            SyncTestHelpers.RunBaasTestAsync(async () =>
+            {
+                var sessionErrorTriggered = false;
+                var tcs = new TaskCompletionSource<bool>();
+                var config = await GetIntegrationConfigAsync();
+                config.OnSessionError = (sender, e) =>
+                {
+                    Assert.IsInstanceOf<Session>(sender);
+                    Assert.IsInstanceOf<SessionException>(e);
```
nirinchev, line 324: Can we check the session exception details here.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -100,6 +103,347 @@ public void Session_Error_ShouldPassCorrectSession()
...
+
+                // priority is given to the newer appoach in SyncConfigurationBase, so this should never be reached
+#pragma warning disable CS0618 // Type or member is obsolete
+                Session.Error += (sender, e) =>
+                {
+                    obsoleteSessionErrorTriggered = true;
+                    tcs.TrySetResult(true);
+                };
+#pragma warning restore CS0618 // Type or member is obsolete
+
```
nirinchev, line 380: This approach - and the one in the next test are prone to race conditions - e.g. if the code triggers both the client reset handler callbacks and `Session.Error`, you would not detect this as long as the ordering is correct. It's kind of silly, but you might want to add a `await Task.Delay(1000)` or similar to ensure that `Session.Error` is indeed not triggered.

Tests/Realm.Tests/Sync/SyncConfigurationTests.cs
```diff
@@ -79,6 +80,13 @@ public void SyncConfiguration_WithAbsolutePath()
             Assert.That(config.DatabasePath, Is.EqualTo(path));
         }
 
+        [Test]
+        public void FlexibleSyncConfiguration_throws_When_Assigned_DiscardLocalResetHandler()
```
nirinchev, line 84:
```suggestion
        public void FlexibleSyncConfiguration_Throws_When_Assigned_DiscardLocalResetHandler()
```

Tests/Realm.Tests/Sync/SyncConfigurationTests.cs
```diff
@@ -79,6 +80,13 @@ public void SyncConfiguration_WithAbsolutePath()
...
+        {
+            var conf = GetFakeFLXConfig();
+            Assert.That(() => { conf.ClientResetHandler = new DiscardLocalResetHandler(); }, Throws.TypeOf<NotSupportedException>());
+        }
+
```
nirinchev, line 89: Needs a test that validates that `DiscardLocal` is the default client reset strategy.

Tests/Realm.Tests/Sync/SyncTestHelpers.cs
```diff
@@ -310,10 +313,14 @@ public static Task<T> WaitForSessionError<T>(Action<Session> sessionAssertions)
                     tcs.TrySetException(ex);
                 }
 
+#pragma warning disable CS0618 // Type or member is obsolete
```
nirinchev, line 316: Can we suppress that with a global suppressions attribute rather than pragmas? That way we won't have to hunt all the pragmas when we start fixing things.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -115,9 +112,36 @@ Realm::Config get_shared_realm_config(Configuration configuration, SyncConfigura
...
+
+        s_session_error_callback(new SharedSyncSession(session), error.error_code.value(), to_capi_value(error.message), user_info_pairs.data(), user_info_pairs.size(), error.is_client_reset_requested(), sync_configuration_handle->handle());
+    };
+
     config.sync_config->stop_policy = sync_configuration.session_stop_policy;
+    config.sync_config->client_resync_mode = sync_configuration.client_resync_mode;
+
+    if (sync_configuration.client_resync_mode == ClientResyncMode::DiscardLocal) {
+
+        config.sync_config->notify_before_client_reset = [managed_sync_configuration_base_handle = sync_configuration.managed_sync_configuration_base_handle](SharedRealm before_frozen) {
```
nirinchev, line 133: Should this capture `sync_configuration_handle` instead?

wrappers/src/sync_session_cs.hpp
```diff
@@ -23,12 +23,14 @@
 #include <realm/sync/config.hpp>
 
 using ProgressCallbackT = void(void* state, uint64_t transferred_bytes, uint64_t transferrable_bytes);
+using NotifyBeforeClientResetCallbackT = bool(SharedRealm& before_frozen, void* managed_client_reset_handler_handle);
+using NotifyAfterClientResetCallbackT = bool(SharedRealm& before_frozen, SharedRealm& after, void* managed_client_reset_handler_handle);
```
nirinchev, line 27: This has a slightly different signature from the migration callback where we use pointers instead of references. I don't know how important it is or if's a problem, but is strange. cc @fealebenpae

Realm/Realm/Configurations/RealmConfiguration.cs
```diff
@@ -84,6 +84,11 @@ public class RealmConfiguration : RealmConfigurationBase
         /// The <see cref="MigrationCallbackDelegate"/> that will be invoked if the <see cref="Realm"/> needs
         /// to be migrated.
         /// </value>
+        /* Passing a mutable reference to the MigrationCallBackDelegate to native is dangerous,
```
nirinchev, line 87: No need for this comment to be here.

wrappers/src/shared_realm_cs.hpp
```diff
@@ -76,6 +82,9 @@ struct SyncConfiguration
     SchemaMode schema_mode;
 
     bool is_flexible_sync;
+
+    ClientResyncMode client_resync_mode;
+    void* managed_sync_configuration_base_handle;
```
nirinchev, line 87: Same - don't need the entire type here.
```suggestion
    void* managed_sync_config;
```

wrappers/src/shared_realm_cs.hpp
```diff
@@ -85,29 +94,59 @@ inline const TableRef get_table(const SharedRealm& realm, TableKey table_key)
 
 namespace realm {
 namespace binding {
+
+extern std::function<void(void*)> s_release_gchandle;
+
+struct GCHandleHolder {
```
nirinchev, line 100: Fancy

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -224,14 +261,86 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
                     exception = new SessionException(messageString, errorCode);
                 }
 
-                Session.RaiseError(session, exception);
+                if (syncConfigurationBase.OnSessionError != null)
+                {
+                    syncConfigurationBase.OnSessionError.Invoke(session, exception);
```
nirinchev, line 266: This should not be invoked if the exception is a client reset exception.

nirinchev (changes requested):
I think we need to add more tests:

1. I don't see tests that try to access the before/after realms - we should ensure that they behave correctly.
2. With both handler types, we don't test what happens when a manual reset is requested and we trigger it from the callback.
3. We only have unit tests - is it possible to write an integration test for this? You should be able to trigger a client reset with code 227 by opening two Realm instances on different partitions and add an object with the same Id to each. The second should get a client reset and you should validate that the object has been removed in the `after` Realm.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -224,14 +261,86 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
+
+            if (syncConfiguration.ClientResetHandler == null)
+            {
+                return true;
+            }
+
+            var discardLocalResetHandler = (DiscardLocalResetHandler)syncConfiguration.ClientResetHandler;
+            var schema = syncConfiguration.GetSchema();
+
+            try
```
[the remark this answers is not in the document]
LaPeste: 1. A wrong cast should not happen as the types are checked way ahead of time, [here](https://github.com/realm/realm-dotnet/blob/582cd17185e7031756863d60f41508c52d93138c/Realm/Realm/Configurations/SyncConfigurationBase.cs#L139). So if we got here the type must be right.
1. I think that a wrong cast would be an error on our side, which the tests will spot right away.
1. If your concern is the fact that the user may change on the fly the type of handler, since they are mutable; that's a fair point. And I would not know what the best approach is. Likely we should not make the application go down, although shall we not tell the user that what he is doing is far from correct? This is different from a logical error that they could make in their handler. In fact the logical error may show up only in corner cases so I don't think we should make the app crash. And in fact, we rightly don't. But a change of handler type on the fly is a consistent error that should show up any time that code path is reached. Because of this, I assume that these types of mistakes are easily findable by testing that reaches that code path. Shall we assume that users don't do testing? But if that is the case, I don't know who's at fault there.
Unfortunately, this brings up another problem. How can we tell the user that changing callback, without modifying its type but only its content, after the instance of realm is created won't do what is expected?

In summary if you referred to point 3, I think we should discuss it a little further. Otherwise, I must have missed something that didn't make me understand your concern.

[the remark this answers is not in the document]
nirinchev: My concern is precisely 3. - if the user does:

```csharp
var config = new PartitionSyncConfiguration
{
    ClientResetHandler = new DiscardLocalHandler()
};

var realm = Realm.GetInstance(config);
config.ClientResetHandler = new ManualRecoveryHandler();
```

The application would crash with the code above and not necessarily in an obvious way as in native, we won't be able to catch the managed exception that we will throw here. Moving it inside the `try`-block ensures that we log the error and that the developer is notified of what they did wrong.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -224,14 +261,86 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
+            {
+                if (discardLocalResetHandler.OnBeforeReset != null)
+                {
+                    using var realmBefore = new Realm(new UnownedRealmHandle(beforeFrozen), syncConfiguration, schema);
```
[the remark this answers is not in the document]
LaPeste: The only built `RealmMetadata` that I could find is in `Realm`. Otherwise metadata is built inside the constructor of `Realm`. I don't know a more efficient way of building `RealmMetadata` then what the constructor already does. So I need a little more about what you suggest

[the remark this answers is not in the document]
nirinchev: Oh right... we don't store the metadata in the configuration. Meh... it's fine then, let's keep it as is.

Realm/Realm/Sync/ErrorHandling/DiscardLocalResetHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync.ErrorHandling
+{
+    /// <summary>
+    /// A client reset strategy where all the not yet synchronized data is automatically discarded and a fresh copy of the synchronized Realm is obtained.
+    /// To be noted that the freshly downloaded copy of the synchronized Realm triggers all change notifications as a write transaction is internally simulated.
+    /// This strategy supplies three callbacks: <see cref="OnBeforeReset"/> and <see cref="OnAfterReset"/> to give a chance to take some user actions,
```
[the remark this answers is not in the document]
LaPeste: The third is at line 26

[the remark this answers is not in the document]
nirinchev: Ah, I think the wording tricked me here - I think we should list all three callbacks in a single sentence - something like:

```
This strategy supplies three callbacks: <see cref="OnBeforeReset"/>, <see cref="OnAfterReset"/>, and <see cref="ManualResetFallback"/>. The first two give notify you just before and after the client reset has happened, while the last one will be invoked in case an error occurs during the automated process and the system needs to fallback to a manual mode.
```

Realm/Realm/Sync/ErrorHandling/DiscardLocalResetHandler.cs
LaPeste: I'm not sure I fully understand the lifetime problem.`beforeFrozen` is frozen, so refs to realm objects are not a problem as long as the memory growth, given by version pinning, is acceptable to the use case.
Saving a ref to a realm in `OnBefore` to use it in `OnAfter`, as you pointed out already, makes little sense as `OnAfter` gives you a ref to both the before and the after realm.
Are you specifically asking about a **non** realm object created in `OnBefore` to be used in `OnAfter`? I'm not really sure here.

LaPeste:
> With both handler types, we don't test what happens when a manual reset is requested and we trigger it from the callback.

@nirinchev if with "both handlers" you mean `DiscardLocalResetHandler` and `ManualRecoveryHandler`, then:
1. `DiscardLocalResetHandler.ManualResetFallback` is tested [here](https://github.com/realm/realm-dotnet/blob/582cd17185e7031756863d60f41508c52d93138c/Tests/Realm.Tests/Sync/SessionTests.cs#L218) and [here](https://github.com/realm/realm-dotnet/blob/582cd17185e7031756863d60f41508c52d93138c/Tests/Realm.Tests/Sync/SessionTests.cs#L265)
2. `ManualRecoveryHandler.OnClientReset` is tested [here](https://github.com/realm/realm-dotnet/blob/582cd17185e7031756863d60f41508c52d93138c/Tests/Realm.Tests/Sync/SessionTests.cs#L189)

If you meant something else, then please try again to explain what it is missing.

Realm/Realm/Configurations/SyncConfigurationBase.cs
```diff
@@ -131,10 +131,10 @@ internal override async Task<Realm> CreateRealmAsync(CancellationToken cancellat
         internal virtual Native.SyncConfiguration CreateNativeSyncConfiguration()
         {
             var syncConfHandle = GCHandle.Alloc(this);
-            var clientResyncMode = ClientResyncMode.DiscardLocal;
-            if (ClientResetHandler != null && ClientResetHandler is ManualRecoveryHandler)
+            var clientResyncMode = ClientResyncMode.Manual;
```
nirinchev, line 134: Why did this change?

Realm/Realm/Extensions/TestingExtensions.cs
```diff
@@ -43,7 +44,38 @@ public static void SimulateError(this Session session, ErrorCode errorCode, stri
...
+
+        /// <summary>
+        /// Simulates a client reset.
+        /// </summary>
+        /// <param name="session">The session where the simulated client reset will occur.</param>
+        /// <param name="message">Error message.</param>
+        /// <remarks>
+        /// You still need to set up a Realm integration testing environment, namely a testing app to connect to.
+        /// </remarks>
+        public static void SimulateClientReset(this Session session, string message)
```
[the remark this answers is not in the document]
LaPeste: <bump>

Tests/Realm.Tests/Sync/SyncConfigurationTests.cs
LaPeste: As a note I put the test in the `SessionTests.cs` to keep all testing regarding this new client reset API in one place.

LaPeste:
Just as a heads up, there are issues with `InitiateClientReset()` on CI. Meaning any call to this method fail on CI but do not fail in local. So I consider that the tests are correct but have other types of issues on CI, which we need to solve of course.
Finally, think if you have more tests to advice to implement.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -224,14 +249,88 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
+
+                if (discardLocalResetHandler.OnBeforeReset != null)
+                {
+                    using var realmBefore = new Realm(new UnownedRealmHandle(beforeFrozen), syncConfig, schema);
+                    discardLocalResetHandler.OnBeforeReset(realmBefore);
+                }
+            }
+            catch (Exception ex)
+            {
+                Logger.Default.Log(LogLevel.Error, $"An error has occurred while executing SyncConfigurationBase.OnBeforeReset during a client reset: {ex}");
```
nirinchev, line 295:
```suggestion
                Logger.Default.Log(LogLevel.Error, $"An error has occurred while executing DiscardLocalResetHandler.OnBeforeReset during a client reset: {ex}");
```

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -224,14 +249,88 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
+                if (discardLocalResetHandler.OnAfterReset != null)
+                {
+                    using var realmBefore = new Realm(new UnownedRealmHandle(beforeFrozen), syncConfig, schema);
+                    using var realmAfter = new Realm(new UnownedRealmHandle(after), syncConfig, schema);
+                    discardLocalResetHandler.OnAfterReset(realmBefore, realmAfter);
+                }
+            }
+            catch (Exception ex)
+            {
+                Logger.Default.Log(LogLevel.Error, $"An error has occurred while executing SyncConfigurationBase.OnAfterReset during a client reset: {ex}");
```
nirinchev, line 327:
```suggestion
                Logger.Default.Log(LogLevel.Error, $"An error has occurred while executing DiscardLocalResetHandler.OnAfterReset during a client reset: {ex}");
```

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -224,14 +249,88 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
+
+                var discardLocalResetHandler = (DiscardLocalResetHandler)syncConfig.ClientResetHandler;
+                var schema = syncConfig.GetSchema();
+
+                if (discardLocalResetHandler.OnAfterReset != null)
+                {
+                    using var realmBefore = new Realm(new UnownedRealmHandle(beforeFrozen), syncConfig, schema);
+                    using var realmAfter = new Realm(new UnownedRealmHandle(after), syncConfig, schema);
+                    discardLocalResetHandler.OnAfterReset(realmBefore, realmAfter);
+                }
```
nirinchev, line 323: We could rework this slightly like (same thing can be done for BeforeReset).

```suggestion
                if (syncConfig.ClientResetHandler is DiscardLocalResetHandler discardLocalHandler &&
                    discardLocalHandler.OnAfterReset != null)
                {
                    var schema = syncConfig.GetSchema();
                    using var realmBefore = new Realm(new UnownedRealmHandle(beforeFrozen), syncConfig, schema);
                    using var realmAfter = new Realm(new UnownedRealmHandle(after), syncConfig, schema);
                    discardLocalHandler.OnAfterReset(realmBefore, realmAfter);
                }
```

CHANGELOG.md
```diff
@@ -1,6 +1,44 @@
...
+        {
+          // executed right before a client reset is about to happen
+        },
+        OnAfterReset = (beforeFrozen, after) =>
+        {
+          // executed right after a client reset is has completed
+        },
+        ManualResetFallback = (session, err) =>
+        {
+            // user's code for manual recovery
```
nirinchev, line 28:
```suggestion
          // handle the reset manually
```

CHANGELOG.md
```diff
@@ -1,6 +1,44 @@
...
+        }
+      }
+    };
+  ```
+  If, instead, the user wants to continue using the manual solution even after the end of the deprecation period, the following should be done
```
nirinchev, line 33:
```suggestion
  If, instead, you want to continue using the manual solution even after the end of the deprecation period, the following should be done
```

CHANGELOG.md
```diff
@@ -1,6 +1,44 @@
...
 ### Enhancements
+* Added support for a new client reset strategy, called [Discard Unsynced Changes](https://docs.mongodb.com/realm/sync/error-handling/client-resets/#discard-unsynced-changes). This new stragegy greatly simplifies the handling of a client reset event on a synchronized Realm.
+This addition makes `Session.Error` **deprecated**. In order to temporarily contiue using the current `Session.Error` the following must be done:
+  ```csharp
+    var conf = new PartitionSyncConfiguration(partition, user)
+    {
+      ClientResetHandler = new ManualRecoveryHandler();
+    };
+  ```
+  In order to take advantage of the new **Discard Unsynced Changes** feature, the following should be done:
```
```suggestion
  In order to take advantage of the new **Discard Unsynced Changes** feature, the following should be done (all callbacks are optional):
```

.github/workflows/main.yml
```diff
@@ -1,8 +1,6 @@
 name: main
 "on":
   push:
-    branches:
-    - main
```
nirinchev, line 5: This should be reverted

Realm/Realm/Sync/ErrorHandling/ClientResetHandlerBase.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        /// Callback triggered when a Client Reset error happens in a synchronized Realm.
+        /// </summary>
+        /// <param name="session">
+        /// The <see cref="Session"/> where the error happened on.
+        /// </param>
+        /// <param name="clientResetException">
+        /// The specific <see cref="ClientResetException"/> that holds useful data to be used when trying to manually recover from a client reset.
+        /// </param>
+        public delegate void ClientResetCallback(Session session, ClientResetException clientResetException);
```
nirinchev, line 40: I know this comes in as a fairly late feedback, but do we need the session argument here? Previously we had it because that was the only way to figure out which Realm the reset was for, but now that this is configured on the RealmConfiguration, it seems superfluous. @fealebenpae @LaPeste do you see value in including the session in the manual reset callback?

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -1102,7 +1101,7 @@ private static async Task TryInitiateClientReset(ClientResetException ex, int ex
             Assert.That(File.Exists(config.DatabasePath), Is.True);
 
             var didReset = false;
-            for (var i = 0; i < 100 || didReset; i++)
+            for (var i = 0; i < 100 && !didReset; i++)
```
nirinchev, line 1104: 🤦‍♂️

Realm/Realm/Sync/ErrorHandling/ManualRecoveryHandler.cs
```diff
@@ -0,0 +1,59 @@
...
+            ManualClientReset = onClientReset;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ManualRecoveryHandler"/> class that uses the Session.Error as handler for client resets.
+        /// </summary>
+        /// <remarks>
+        /// This obsolete constructor should be used only when the user wants to use the obsolete Session.Error event for handling session errors and client resets.
+        /// </remarks>
+        [Obsolete("Use SyncConfigurationBase.OnSessionError in conjunction with SyncConfigurationBase.ClientResetHandler instead.")]
```
nirinchev, line 54: We should probably point people to the non-obsolete constructor in this message. @MongoCaleb hit it during docs work and it wasn't obvious to him that he should use the one that takes an argument.

Issue #1865, opened 2019-05-21
Exception 'System.Runtime.InteropServices.SEHException'


sunthx:
## Goals
I'm using RealmDb with my wpf application .  sometimes the database will throw an exception ，I don't know how to reproduce it.

## Expected Results
succeed

## Actual Results
```
UnhandledException : System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
   在 Realms.SynchronizationContextEventLoopSignal.EventLoop.<>c__DisplayClass3_0.<Post>b__0(Object _)
   在 System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   在 System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   在 System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
   在 System.Threading.ThreadPoolWorkQueue.Dispatch() 
```

## Steps to Reproduce
I don't know how to reproduce this problem.

## Version of Realm and Tooling
* Realm.Database 3.4.0
* Realm.DataBinding 1.2.0
* Win10 1809
* .Netframework 4.6.2

Jared-Aus:
I get this error when trying to use Realm with xUnit. nUnit doesn't complain however.

fealebenpae:
Hey @Jared-Aus, can you upload a small project that reproduces the crash?

Jared-Aus:
[XUnitTestProject1.zip](https://github.com/realm/realm-dotnet/files/4137170/XUnitTestProject1.zip)

Jared-Aus:
> Hey @Jared-Aus, can you upload a small project that reproduces the crash?

Done. Btw, I haven't noticed any issues in my Xamarin or nUnit projects, however it's possible that xUnit is simply less forgiving than the others.

fealebenpae:
I was able to reproduce this, thanks.
From what I can see the test passes fine, but during teardown the internal notification machinery crashes. I wasn't able to attach the exception in a debugger, unfortunately. Perhaps it's a quirk of the xUnit runner that we'll need to work around.

Jared-Aus:
> I was able to reproduce this, thanks.
[...]

PR #2787, opened 2022-01-28
Add a workaround for xunit tests

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/1865

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


fealebenpae (approved):

nirinchev:
The xUnit compatibility tests are run on release only because I don't expect we'll ever cause them to fail and it's a waste of resources to run them continuously. The line coverage drop is expected as these codepaths are not exercised by our regular tests.

PR #2803, opened 2022-02-02
Expose connection state and notifications to the sync session

### LINKED ISSUE #2801: Expose connection state and notifications to the sync session
OS has added API to access the connection state that we need to expose in the SDK: https://github.com/realm/realm-core/blob/6b05da97d0a9bd0bf61522a0342222443b35e196/src/realm/object-store/sync/sync_session.hpp#L125. 

There's also an [API to register a connection state callback](https://github.com/realm/realm-core/blob/6b05da97d0a9bd0bf61522a0342222443b35e196/src/realm/object-store/sync/sync_session.hpp#L166) which should also be added.

LaPeste:
## Description
Expose core functionalities in the title.

`Session.cs` seems to have a lot of changes, but in reality it just adds methods to respect the `INotifyPropertyChanged` and to expose the `SessionConnectionState`. The rest is reordering within the class according to "fields/properties in the private-protected-public order and methods in the public-protected-private order".

Fixes #2801

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -68,6 +74,12 @@ private static class NativeMethods
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "realm_syncsession_unregister_progress_notifier", CallingConvention = CallingConvention.Cdecl)]
             public static extern void unregister_progress_notifier(SessionHandle session, ulong token, out NativeException ex);
 
+            [DllImport(InteropConfig.DLL_NAME, EntryPoint = "realm_syncsession_register_connection_change_callback", CallingConvention = CallingConvention.Cdecl)]
+            public static extern ulong register_connection_change_callback(SessionHandle session, IntPtr state_change_callback_handle, out NativeException ex);
```
nirinchev, line 78: Instead of passing a handle to the action action here, we can pass a handle to the managed session.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -40,15 +40,21 @@ private static class NativeMethods
             [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
             public delegate void SessionWaitCallback(IntPtr task_completion_source, int error_code, PrimitiveValue message);
 
+            [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+            public delegate void SessionConnectionChangeCallback(IntPtr connectionChangeCallbackHandle);
```
nirinchev, line 44: Let's call this `SessionPropertyChangedCallback` and have it pass the property name as well. This way adding state notifications will be fairly trivial.

nirinchev (changes requested):

Realm/Realm/Handles/SessionHandle.cs
fealebenpae: That's applicable with Nikola's suggestion above.

Realm/Realm/Sync/Session.cs
```diff
@@ -185,5 +225,34 @@ internal void CloseHandle(bool waitForShutdown = false)
...
+        internal static void RaiseError(Session session, Exception error)
+        {
+            var args = new ErrorEventArgs(error);
+            Error?.Invoke(session, args);
+        }
+
+        private void SubscribeNotifications()
+        {
+            var cb = new Action(() => NotifyPropertyChanged(nameof(ConnectionState)));
+            var callbackHandle = GCHandle.Alloc(cb);
```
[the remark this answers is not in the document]
nirinchev: I don't think we need to free it from native. Instead, if we pass in a weak handle to the session, it will be freed when the session is GC-ed or when we unsubscribe.

LaPeste:
Since we are going to add more than 1 fields for notifications on the `Session`, I switched it up a little to make the solution more future proof. In this way adding a new field should be really straight forward and fast.

Realm/Realm/Sync/Session.cs
```diff
@@ -185,5 +225,34 @@ internal void CloseHandle(bool waitForShutdown = false)
...
+        {
+            var args = new ErrorEventArgs(error);
+            Error?.Invoke(session, args);
+        }
+
+        private void SubscribeNotifications()
+        {
+            var cb = new Action<string>((propertyName) => NotifyPropertyChanged(propertyName));
+            var callbackHandle = GCHandle.Alloc(cb, GCHandleType.Weak);
+            var nativeTokens = Handle.RegisterPropertyChangedCallbacks(GCHandle.ToIntPtr(callbackHandle));
```
nirinchev, line 239: Again, let's allocate a weak handle to the managed session rather than the callback.

Realm/Realm/Sync/Session.cs
```diff
@@ -185,5 +225,34 @@ internal void CloseHandle(bool waitForShutdown = false)
...
+            });
+        }
+
+        private void UnsubscribeNotifications()
+        {
+            _propertyChangedNotificationTokens?.Dispose();
+            _propertyChangedNotificationTokens = null;
+        }
+
+        private void NotifyPropertyChanged([CallerMemberName] string propertyName = "")
```
nirinchev, line 253: Let's make this internal and call it `RaisePropertyChanged` to be consistent with the naming in `RealmObjectBase`.

Realm/Realm/Sync/Session.cs
```diff
@@ -27,21 +30,76 @@ namespace Realms.Sync
...
         internal bool IsClosed => _handle.IsClosed;
 
+        /// <summary>
+        /// Occurs when a property value changes.
+        /// </summary>
+        public event PropertyChangedEventHandler PropertyChanged
+        {
+            add
+            {
+                if (_propertyChangedNotificationTokens == null)
```
nirinchev, line 71: Rather than check for `_propertyChangedNotificationTokens`, let's check for `_propertyChanged == null`, again for consistency's sake.

Realm/Realm/Sync/Session.cs
```diff
@@ -27,21 +30,76 @@ namespace Realms.Sync
...
         /// <summary>
         /// Gets the session’s current state.
         /// </summary>
         /// <value>An enum value indicating the state of the session.</value>
         public SessionState State => Handle.GetState();
 
+        /// <summary>
+        /// Gets the session’s current connection state.
+        /// </summary>
+        /// <remarks>This field notifies clients that its value has changed according to the <see href="https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.inotifypropertychanged?view=net-6.0">INotifyPropertyChanged</see> interface.</remarks>
```
nirinchev, line 99: We should remove this remark - the interface itself is implying that this is the case.

Realm/Realm/Sync/Session.cs
```diff
@@ -27,21 +30,76 @@ namespace Realms.Sync
...
         /// Triggered when an error occurs on a session. The <c>sender</c> argument will be the session which has errored.
         /// </summary>
         public static event EventHandler<ErrorEventArgs> Error;
 
+        private readonly SessionHandle _handle;
+
+        [SuppressMessage("StyleCop.CSharp.NamingRules", "SA1300:Element should begin with upper-case letter", Justification = "This is the private event - the public is uppercased.")]
+        private event PropertyChangedEventHandler _propertyChanged;
+
+        private IDisposable _propertyChangedNotificationTokens;
```
nirinchev, line 45:
```suggestion
        private IDisposable _notificationToken;
```

There's no need for `Session` to know how notifications are implemented in Core and that there are multiple tokens.

Realm/Realm/Sync/Session.cs
```diff
@@ -134,42 +192,24 @@ public class Session
...
 
-        /// <inheritdoc/>
-        public override bool Equals(object obj)
-            => obj is Session other &&
-               Handle.GetRawPointer() == other.Handle.GetRawPointer();
-
-        /// <inheritdoc/>
-        public override int GetHashCode() => Handle.GetRawPointer().GetHashCode();
-
         internal void CloseHandle(bool waitForShutdown = false)
```
nirinchev, line 213: We should unsubscribe from events in this method.

Realm/Realm/Native/PropertyChangedNotificationTokens.cs
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Native
+{
+    internal struct PropertyChangedNotificationTokens
```
nirinchev, line 21: This is a pretty generic name for something that is very specific to the Session object.
```suggestion
    internal struct SessionNotificationToken
```

Realm/Realm/Native/PropertyChangedNotificationTokens.cs
```diff
@@ -0,0 +1,25 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Native
```
nirinchev, line 19:
```suggestion
namespace Realms.Sync.Native
```

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -256,5 +290,17 @@ private static void HandleSessionWaitCallback(IntPtr taskCompletionSource, int e
...
+        private static void HandleSessionPropertyChangedCallback(IntPtr propertyChangedCallbackHandle, NotifiableProperties propertyChanged)
+        {
+            ((Action<string>)GCHandle.FromIntPtr(propertyChangedCallbackHandle).Target)(propertyChanged.ToString());
+        }
+
+        // !! The names of this enum must be kept in sync with the notifiable property names of Session !!
+        private enum NotifiableProperties : byte
+        {
+            ConnectionState = 0,
+        }
```
nirinchev, line 304: Rather than keep names in sync with properties, which is not something we can enforce, we should use a switch to map them like:

```csharp
var propertyName = notifiableProperty switch
{
    ConnectionState => nameof(Session.ConnectionState),
    _ => throw new NotSupportedException($"Unexpected notifiable property value: {notifiableProperty}")
};
```

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -256,5 +290,17 @@ private static void HandleSessionWaitCallback(IntPtr taskCompletionSource, int e
...
         }
+
+        [MonoPInvokeCallback(typeof(NativeMethods.SessionPropertyChangedCallback))]
+        private static void HandleSessionPropertyChangedCallback(IntPtr propertyChangedCallbackHandle, NotifiableProperties propertyChanged)
+        {
+            ((Action<string>)GCHandle.FromIntPtr(propertyChangedCallbackHandle).Target)(propertyChanged.ToString());
+        }
+
+        // !! The names of this enum must be kept in sync with the notifiable property names of Session !!
+        private enum NotifiableProperties : byte
```
nirinchev, line 301:
```suggestion
        private enum NotifiableProperty : byte
```

By convention, c# enums are named plural only if they are flag enums.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -256,5 +290,17 @@ private static void HandleSessionWaitCallback(IntPtr taskCompletionSource, int e
                 tcs.TrySetException(new RealmException(OuterMessage, inner));
             }
         }
+
+        [MonoPInvokeCallback(typeof(NativeMethods.SessionPropertyChangedCallback))]
+        private static void HandleSessionPropertyChangedCallback(IntPtr propertyChangedCallbackHandle, NotifiableProperties propertyChanged)
```
nirinchev, line 295:
```suggestion
        private static void HandleSessionPropertyChangedCallback(IntPtr propertyChangedCallbackHandle, NotifiableProperties property)
```

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -221,6 +223,81 @@ public void Session_Start_IsIdempotent()
...
+                    }
+                    else if (stateChanged == 2)
+                    {
+                        Assert.That(session.ConnectionState, Is.EqualTo(SessionConnectionState.Connected));
+                    }
+                    else if (stateChanged == 3)
+                    {
+                        Assert.That(session.ConnectionState, Is.EqualTo(SessionConnectionState.Disconnected));
+                        completionTCS.TrySetResult(null);
+                    }
```
nirinchev, line 258: This needs a try-catch similar to what we're doing with the client reset PR otherwise an assertion will cause the test to timeout.

wrappers/src/sync_session_cs.cpp
```diff
@@ -26,6 +26,8 @@
 #include <realm/object-store/sync/sync_session.hpp>
 #include "sync_session_cs.hpp"
 
+enum class NotifiableProperties : uint8_t;
```
nirinchev, line 29: Why is this defined here but implemented further down? Can't we just move the implementation here?

nirinchev (changes requested):
Looks good so far. I have some code style comments + a reminder to use the managed session handle rather than an action. There are also some more tests we can write:

1. Verify that sessions can be GC-ed even when someone has subscribed to PropertyChanged.
2. Verify that calling Stop/Start on one session propagates the state to other session instances for the same Realm (this will require that you open multiple Realm instances as the session is cached).

fealebenpae (changes requested):

Realm/Realm/Sync/Session.cs
LaPeste: I thought that the managed session was a recommendation mostly because there was not in place a generic mechanism like this. So it'd have helped with other fields to raise notifications for. Since this does not seem to be the case, could you explain the advantages of passing the whole `SessionHandle` rather than an action. Does it have to do with lifecycle?

Realm/Realm/Sync/Session.cs
```diff
@@ -185,5 +225,34 @@ internal void CloseHandle(bool waitForShutdown = false)
...
+
+        internal static void RaiseError(Session session, Exception error)
+        {
+            var args = new ErrorEventArgs(error);
+            Error?.Invoke(session, args);
+        }
+
+        private void SubscribeNotifications()
+        {
+            var cb = new Action<string>((propertyName) => NotifyPropertyChanged(propertyName));
```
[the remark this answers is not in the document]
LaPeste: My intention was to store it in the `_propertyChangedNotificationTokens`, which is a [NotificationToken])https://github.com/realm/realm-dotnet/blob/1152c35aabef4252232a2db49aca16d3bf6ff2f1/Realm/Realm/DatabaseTypes/INotifiable.cs#L40). Unfortunately, I sent the tokens instead of the callback. Would you agree that fixing this would be enough?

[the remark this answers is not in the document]
fealebenpae: You could pass `cb` as the first argument to `NotificationToken.Create`, yeah, and then capture `nativeTokens` in the unsubscribe lambda. It would look a little bit weird, I think, but it will work.

Personally I'm not sure I'd use the `NotificationToken` machinery here in the first place, because we're only handling subscribing and unsubscribing for the callbacks internally and there isn't an API that looks like `RealmCollectionBase<T>.SubscribeForNotifications`, but I'm not opposed to it if you prefer it for the sake of consistency.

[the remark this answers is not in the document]
LaPeste: @nirinchev What's your opinion on this?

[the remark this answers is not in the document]
nirinchev: As mentioned in my review, I think that creating a weak handle around the session will simplify the lifetime tracking and most things should work out of the box and will also be more consistent with the design of notifications in `RealmObjectBase`.

wrappers/src/sync_session_cs.cpp
LaPeste: It was just to keep it a little tidier. If stylistically it's preferred to have it in there, then I'll move the implementation there and remove this forward declaration.

Realm/Realm/Handles/SessionHandle.cs
LaPeste: Sleek. Thank you.

wrappers/src/sync_session_cs.cpp
nirinchev: I don't know - @fealebenpae you're the c++ master here. I just find it easier to have things defined once, but not going to pretend I know what I'm talking about 😅

Realm/Realm/Native/SessionNotificationToken.cs
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync.Native
+{
+    internal struct SessionNotificationToken
```
fealebenpae, line 21:
```suggestion

using System.Runtime.InteropServices;

namespace Realms.Sync.Native
{
    [StructLayout(LayoutKind.Sequential)]
    internal struct SessionNotificationToken
```
We decorate all structs that we pass to and from native with this attribute in order to enforce that the runtime layout of the struct matches the layout of the corresponding struct in native. The C# compiler generates sequential structs by default, but our convention is to be explicit about marshaling declarations.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -256,5 +290,22 @@ private static void HandleSessionWaitCallback(IntPtr taskCompletionSource, int e
...
+        [MonoPInvokeCallback(typeof(NativeMethods.SessionPropertyChangedCallback))]
+        private static void HandleSessionPropertyChangedCallback(IntPtr managedSessionHandle, NotifiableProperty property)
+        {
+            var propertyName = property switch
+            {
+                NotifiableProperty.ConnectionState => nameof(Session.ConnectionState),
+                _ => throw new NotSupportedException($"Unexpected notifiable property value: {property}")
+            };
+            var session = (Session)GCHandle.FromIntPtr(managedSessionHandle).Target;
+            session.RaisePropertyChanged(propertyName);
```
fealebenpae, line 303:
```suggestion
            System.Threading.ThreadPool.QueueUserWorkItem(_ =>
            {
                session.RaisePropertyChanged(propertyName);
            });
```
After what we learned in #2830 I think we should always invoke all user callbacks on the thread pool instead of directly on the sync client's worker thread.

fealebenpae (approved):
LGTM, with a couple of small suggestions.

Realm/Realm/Sync/SessionConnectionState.cs
```diff
@@ -0,0 +1,41 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync
+{
+    /// <summary>
+    /// The current connection state of a sync session object.
+    /// </summary>
+    public enum SessionConnectionState : byte
```
nirinchev, line 24: Should we rename this to `ConnectionState`? I don't think we need the `Session` prefix as the fact that it's an enum in the `Realms.Sync` namespace already implies it's the sync connection.

wrappers/src/sync_session_cs.cpp
```diff
@@ -34,12 +38,14 @@ using SharedSyncSession = std::shared_ptr<SyncSession>;
 using ErrorCallbackT = void(std::shared_ptr<SyncSession>* session, int32_t error_code, realm_value_t message, std::pair<char*, char*>* user_info_pairs, size_t user_info_pairs_len, bool is_client_reset);
 using ProgressCallbackT = void(void* state, uint64_t transferred_bytes, uint64_t transferrable_bytes);
 using WaitCallbackT = void(void* task_completion_source, int32_t error_code, realm_value_t message);
+using PropertyChangedCallbackT = void(void* managed_session_handle, NotifiableProperty propertyName);
```
nirinchev, line 41:
```suggestion
using PropertyChangedCallbackT = void(void* managed_session_handle, NotifiableProperty property);
```

wrappers/src/sync_session_cs.cpp
```diff
@@ -133,6 +147,29 @@ REALM_EXPORT void realm_syncsession_unregister_progress_notifier(const SharedSyn
     });
 }
 
+typedef struct PropertyChangedNotificationTokens {
+    uint64_t connection_state;
+} PropertyChangedNotificationTokens;
```
nirinchev, line 152:
```suggestion
typedef struct PropertyChangedNotificationToken {
    uint64_t connection_state;
} PropertyChangedNotificationToken;
```

wrappers/src/sync_session_cs.cpp
```diff
@@ -133,6 +147,29 @@ REALM_EXPORT void realm_syncsession_unregister_progress_notifier(const SharedSyn
...
+
+REALM_EXPORT PropertyChangedNotificationTokens realm_syncsession_register_property_changed_callbacks(const SharedSyncSession& session, void* managed_session_handle, NativeException::Marshallable& ex)
```
nirinchev, line 154:
```suggestion
REALM_EXPORT PropertyChangedNotificationTokens realm_syncsession_register_property_changed_callback(const SharedSyncSession& session, void* managed_session_handle, NativeException::Marshallable& ex)
```

wrappers/src/sync_session_cs.cpp
```diff
@@ -133,6 +147,29 @@ REALM_EXPORT void realm_syncsession_unregister_progress_notifier(const SharedSyn
...
+        auto connection_state_token = session->register_connection_change_callback([managed_session_handle](realm::SyncSession::ConnectionState old_state, realm::SyncSession::ConnectionState new_state) {
+            s_property_changed_callback(managed_session_handle, NotifiableProperty::ConnectionState);
+        });
+
+        PropertyChangedNotificationTokens tokens { connection_state_token };
+        return tokens;
+    });
+}
+
+REALM_EXPORT void realm_syncsession_unregister_property_changed_callbacks(const SharedSyncSession& session, PropertyChangedNotificationTokens tokens, NativeException::Marshallable& ex)
```
nirinchev, line 166:
```suggestion
REALM_EXPORT void realm_syncsession_unregister_property_changed_callback(const SharedSyncSession& session, PropertyChangedNotificationTokens tokens, NativeException::Marshallable& ex)
```

wrappers/src/sync_session_cs.cpp
```diff
@@ -133,6 +147,29 @@ REALM_EXPORT void realm_syncsession_unregister_progress_notifier(const SharedSyn
...
+
+REALM_EXPORT PropertyChangedNotificationTokens realm_syncsession_register_property_changed_callbacks(const SharedSyncSession& session, void* managed_session_handle, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&] {
+        auto connection_state_token = session->register_connection_change_callback([managed_session_handle](realm::SyncSession::ConnectionState old_state, realm::SyncSession::ConnectionState new_state) {
+            s_property_changed_callback(managed_session_handle, NotifiableProperty::ConnectionState);
+        });
+
+        PropertyChangedNotificationTokens tokens { connection_state_token };
+        return tokens;
```
nirinchev, line 162:
```suggestion
        PropertyChangedNotificationToken notification_token { connection_state_token };
        return notification_token;
```

nirinchev (commented):
I mostly have naming nits

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -256,5 +290,22 @@ private static void HandleSessionWaitCallback(IntPtr taskCompletionSource, int e
...
+
+        [MonoPInvokeCallback(typeof(NativeMethods.SessionPropertyChangedCallback))]
+        private static void HandleSessionPropertyChangedCallback(IntPtr managedSessionHandle, NotifiableProperty property)
+        {
+            var propertyName = property switch
+            {
+                NotifiableProperty.ConnectionState => nameof(Session.ConnectionState),
+                _ => throw new NotSupportedException($"Unexpected notifiable property value: {property}")
+            };
+            var session = (Session)GCHandle.FromIntPtr(managedSessionHandle).Target;
```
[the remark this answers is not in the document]
LaPeste: Thanks for the catch. Can I also log an error about the exception? So the user knows that something wrong happened there.

Realm/Realm/Sync/SessionConnectionState.cs
LaPeste: It makes sense as the summary on the enum should clear any possible confusion.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -96,12 +110,19 @@ public static void Initialize()
...
             GCHandle.Alloc(error);
             GCHandle.Alloc(progress);
             GCHandle.Alloc(wait);
+            GCHandle.Alloc(propertyChanged);
 
-            NativeMethods.install_syncsession_callbacks(error, progress, wait);
+            NativeMethods.install_syncsession_callbacks(error, progress, wait, propertyChanged);
+        }
+
+        ~SessionHandle()
```
nirinchev, line 123: This is not necessary - the SafeHandle finalizer should always call `ReleaseHandle`.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -322,5 +513,11 @@ private Session OpenRealmAndStopSession()
 
             return session;
         }
+
+        private static IDisposable GetNotificationToken(Session session)
+        {
+            var sessionHandle = (IDisposable)typeof(Session).GetProperty("Handle", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(session);
```
nirinchev, line 519: This will throw if the handle has been closed.

nirinchev (changes requested):

Realm/Realm/Handles/SessionHandle.cs
LaPeste: I still see the need to dispose of the notification token. Are you suggesting to put that call (`UnsubscribeNotifications`) in `ReleaseHandle`?

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -147,6 +171,26 @@ public void UnregisterProgressNotifier(ulong token)
             ex.ThrowIfNecessary();
         }
 
+        public void SubscribeNotifications(Session session)
+        {
+            Debug.Assert(!_notificationToken.HasValue, $"{nameof(_notificationToken)} must be null before subscribing.");
+
+            var managedSessionHandle = GCHandle.Alloc(session, GCHandleType.Weak);
```
nirinchev, line 178: Technically, this is leaking the handle, but it seems there are a fair amount of other places where we do it. I filed https://github.com/realm/realm-dotnet/issues/2867 to fix it and we can probably leave this for whenever we fix them all.

nirinchev (approved):
I think this should be ready now. I left a comment about the gchandle, but we can fix it as part of the broader GCHandle audit.

Realm/Realm/Handles/SessionHandle.cs
LaPeste: You're right. There were a few changes around this and apparently I overlooked free it.
Since you filed an issue for this I guess we can postpone it as you said.
In this specific case we should free the handle during `UnsubscribeNotifications`, which is also called during unbinding. I think that should be it, right?

PR #2813, opened 2022-02-13
Refactor async tests

fealebenpae:
This is a (noisy) refactor of our unit tests that extracts the logic behind `TestHelpers.RunAsyncTest` and `SyncTestHelpers.RunBaasTestAsync` into two custom NUnit attributes - `[RequiresSynchronizationContext]` and `[RequiresBaas]`. This allows us to switch to using awaitable test methods instead of wrapping async code in `TestHelpers.RunAsyncTest()` calls.

I originally intended to just replace `SyncTestHelpers.RunBaasTestAsync` calls with the `[RequiresBaas]` attribute and remove `TestHelpers.RunAsyncTest` in favor of making the test methods async themselves, but it turned out that NUnit doesn't automatically install a SynchronizationContext and the only way to tell it to do so is to mark your tests as needing a [single-threaded apartment](https://docs.microsoft.com/en-us/dotnet/api/system.threading.apartmentstate?view=net-6.0#system-threading-apartmentstate-sta), which also causes NUnit to try to run tests inside a STA thread which .NET only supports on Windows. As a workaround I ended up creating `[RequiresSynchronizationContext]`, made it inheritable, and slapped it on our base fixture class `RealmTest`, so while it exists, there shouldn't be a need to use it in test code if the test class inherits from `RealmTest`.

For all intents and purposes now it's enough for a test method to return `Task` (or another awaitable type) to be considered async, and to decorate any baas integration tests with `[RequiresBaas]`.


Tests/Realm.Tests/Helpers/NUnitAsyncTests.cs
```diff
@@ -0,0 +1,51 @@
...
+            await Task.Delay(50);
+            Assert.That(threadId, Is.EqualTo(Environment.CurrentManagedThreadId));
+        }
+
+        [Test, RequiresSynchronizationContext]
+        public async Task Post_WhenYielding_ShouldExecuteRightAway()
+        {
+            Assert.That(SynchronizationContext.Current, Is.Not.Null);
+
+            int i = 0;
```
nirinchev, line 44:
```suggestion
            var i = 0;
```

Tests/Realm.Tests/Helpers/RequiresSynchronizationContextAttribute.cs
```diff
@@ -0,0 +1,64 @@
...
+            private static readonly Type NUnitSynchronizationContext =
+                typeof(NUnitAttribute).Assembly.GetType("NUnit.Framework.Internal.SingleThreadedTestSynchronizationContext");
+
+            public ExecuteInSynchronizationContextCommand(TestCommand innerCommand) : base(innerCommand)
+            {
+            }
+
+            public override TestResult Execute(TestExecutionContext context)
+            {
+                var syncContext = (SynchronizationContext)Activator.CreateInstance(NUnitSynchronizationContext, TimeSpan.FromSeconds(10));
```
nirinchev, line 49: Should the timespan we pass to the ctor be inferred from the test timeout attribute?

Tests/Realm.Tests/Helpers/RequiresSynchronizationContextAttribute.cs
```diff
@@ -0,0 +1,64 @@
...
+                var current = SynchronizationContext.Current;
+
+                SynchronizationContext.SetSynchronizationContext(syncContext);
+                try
+                {
+                    return context.CurrentResult = innerCommand.Execute(context);
```
nirinchev, line 55: That's a little tough to read - can we split it into two lines?

Tests/Realm.Tests/Sync/FlexibleSyncTests.cs
```diff
@@ -1100,620 +1091,584 @@ public void SubscriptionSet_Enumerator_Reset()
...
+                ato.Int32Property = 15;
             });
+
+            var sessionError = await sessionErrorTask;
+            Assert.That(sessionError.ErrorCode, Is.EqualTo(ErrorCode.WriteNotAllowed));
         }
 
-        [Test]
-        public void Integration_AfterAnError_CanRecover()
+        [Test, RequiresSynchronizationContext, RequiresBaas]
```
nirinchev, line 1568: Would it make sense to make these attributes hierarchical? I.e. rename `[RequiresSynchronizationContext]` to `[SynchronizationContextTest]` and have it inherit from `[TestAttribute]` and have `[BaasTest]` inherit from `[SynchronizationContextTest]`? It'll reduce the amount of attributes we have to apply on each test.

Tests/Realm.Tests/Sync/SyncTestBase.cs
```diff
@@ -20,6 +20,7 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using MongoDB.Bson;
+using NUnit.Framework;
```
nirinchev, line 23: Is this needed?

nirinchev (commented):
Good stuff - I like this a lot better than the current approach.

Tests/Realm.Tests/Sync/FlexibleSyncTests.cs
fealebenpae: I like the composability, because it allows us to mix `RequiresSynchronizationContext` and `RequiresBaas` both with each other and with `Test`, `TestCase`, and `Theory`.
Plus with my latest change `RequiresSynchronizationContext` becomes inherited and we only need have it explicitly once on the base `RealmTest` so it's implicitly picked up everywhere.
As for `RequiresBaas` I think it's better to keep it as a separate action attribute because this allows us to apply it to both fixtures and test cases.

Tests/Realm.Tests/Helpers/RequiresSynchronizationContextAttribute.cs
fealebenpae: This timespan is not configurable - it's [hardcoded in NUnit internally](https://github.com/nunit/nunit/blob/b34eba3ac1aa6957157857bddd116256c634afab/src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs#L50) and I'm not sure what other values we should be applying.

Tests/Realm.Tests/Helpers/RequiresSynchronizationContextAttribute.cs
fealebenpae: Latest rewrite of the attribute solves this.

Tests/Realm.Tests/Sync/SyncTestBase.cs
fealebenpae: No - thanks for catching this.

Tests/Realm.Tests/Helpers/RequiresSynchronizationContextAttribute.cs
```diff
@@ -0,0 +1,74 @@
...
+                 * "SingleThreadedTestSynchronizationContext". It's enough to install their SynchronizationContext
+                 * before the test runs for the test executor to pick it up and await the returned Task in it.
+                 * This is different from Nito's AsyncContext where we explicitly had to "enter" the context,
+                 * not just installing it.
+                 * Because of this, posting work on the SynchronizationContext only executes it in async tests
+                 * where NUnit awaits the Task returned by the test method. Synchronous "void" tests won't await
+                 * and run the posted work so we don't install a SynchronizationContext for them.
+                */
+
+                if (SynchronizationContext.Current is SynchronizationContext current)
```
papafe, line 53: Why don't use the synchronization context that's already there if it exists?

papafe (approved):
Looks quite cool!

DominicFrei (approved):
Nice one. Big fan of simplifications! :+1:

Tests/Realm.Tests/Helpers/NUnitAsyncTests.cs
```diff
@@ -0,0 +1,52 @@
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2019 Realm Inc.
```
LaPeste, line 3:
```suggestion
// Copyright 2022 Realm Inc.
```

Tests/Realm.Tests/Helpers/RequiresBaasAttribute.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            if (EnsureNoSessionErrors)
+            {
+                void ErrorHandler(object _, ErrorEventArgs e)
+                {
+                    test.Properties.Set(ExceptionKey, e.Exception);
+                }
+
+                var @delegate = new EventHandler<ErrorEventArgs>(ErrorHandler);
+                Realms.Sync.Session.Error += @delegate;
```
LaPeste, line 71: As we discussed already, we'll need to re-adjust this once the client reset API PR gets merged as the `Session.Error` is deprecated from there on

LaPeste (approved):
Really good work. It's gonna make our tests so much pretties in style and faster to read and write.

PR #2830, opened 2022-03-03
Avoid possible loop / error with client reset

papafe:
This PR fixes a possible client reset error / loop. This would happen in a situation like:

```csharp
public MainPage()
{
    InitializeRealm();
}

private async void InitializeRealm()
{
    Session.Error += SessionOnError;
    await LoadDataAsync();
}

public Task LoadDataAsync()
{
	_realm = Realm.GetInstance(new SyncConfiguration(partitionKey, _user, realmName));
	await _realm.SyncSession.WaitForDownloadAsync();
    //.... load stuff in the UI
}

private void SessionOnError(object sender, ErrorEventArgs e)
{
    if (e.Exception is ClientResetException clientResetException)
    {
        _realm?.Dispose();
        var didSucceed = clientResetException.InitiateClientReset();

        if (didSucceed)
        {
            LoadDataAsync();
        }
    }
}

```

In this case if there is a client reset then the `SessionOnError` handler will be called. Then the client reset will be initiated and if `didSucceed` will be true, then `LoadDataAsync` is called again, and then we get yet again a new client reset when we reach `WaitForDownloadAsync`. This creates a loop, and is what happens on iOS. 
On Windows, at the moment there is no loop, because correctly `didSucceed` is false as the realm file is in use and can't be removed. This will make it impossible to make a client reset. 

Internally, the reason why this is happening in both cases is because the `SessionOnError` handler is being called in the same thread where the core sync client operates, and thus there are several shared pointers still referencing the sync session here. When we call `InitiateClientReset`, because the sync session is kept alive by those pointers, then there is a problem in correctly executing the client reset itself. 
In order to solve this, we are calling the error event handler on another thread in the thread loop, in order for the shared pointer in the original thread to be freed. The only thing left to do is to free the managed `SessionHandle` after the event handler has been executed. Because we need to do so, we need to instruct the developers that `InitiateClientReset` need to happen on a different thread than the one where the session error is being run. For instance:

```csharp
private void SessionOnError(object sender, ErrorEventArgs e)
{
    if (e.Exception is ClientResetException clientResetException)
    {
        _realm?.Dispose();
            MainThread.BeginInvokeOnMainThread(async () =>
            {
		            var didSucceed = clientResetException.InitiateClientReset();
		    
		            if (didSucceed)
		            {
		                LoadDataAsync();
		            }
			});
    }
}
```
In this case it's being run on the main thread, but it's just a possibility, and it just needs to be run on a different thread. The reason why we need to do so, is to have the possibility of disposing of the last handle, the managed one. (This means that in the thread where the `InitiateClientReset` is called the session is disposed, and can't be used anymore).

All of this needs to be in the docs.

Additionally, and that's something to be addressed in other PRs:
- The progress observables need to be taken care of, as they have a strong reference to the session handle
- At the moment, if a developer-driven action causes a client reset exception (like calling `WaitForDownloadAsync` above), then the exception is not a `ClientResetException`, but a generic error. We should probably change that. 

##  TODO
* [ ] Changelog entry
* [ ] Docs

### DIFF
--- a/Realm/Realm/Handles/SessionHandle.cs
+++ b/Realm/Realm/Handles/SessionHandle.cs
@@ -203,7 +203,7 @@ namespace Realms.Sync
         {
             try
             {
-                using var handle = new SessionHandle(null, sessionHandlePtr);
+                var handle = new SessionHandle(null, sessionHandlePtr);
                 var session = new Session(handle);
                 var messageString = message.AsString();
 
@@ -224,7 +224,18 @@ namespace Realms.Sync
                     exception = new SessionException(messageString, errorCode);
                 }
 
-                Session.RaiseError(session, exception);
+                System.Threading.ThreadPool.QueueUserWorkItem(_ =>
+                {
+                    try
+                    {
+                        Session.RaiseError(session, exception);
+                    }
+                    catch { }
+                    finally
+                    {
+                        handle.Dispose();
+                    }
+                });
             }
             catch (Exception ex)
             {


Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -224,7 +224,18 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
+                {
+                    try
+                    {
+                        Session.RaiseError(session, exception);
+                    }
+                    catch { }
+                    finally
+                    {
+                        handle.Dispose();
+                    }
```
fealebenpae, line 237: You should move the outer try/catch here. We don't need the outer one because we're no longer invoking user code, and the one inside the thread pool should log the exception instead of just swallow it.

fealebenpae (changes requested):

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -201,35 +201,38 @@ public void ShutdownAndWait()
...
-                    var userInfo = StringStringPair.UnmarshalDictionary(userInfoPairs, userInfoPairsLength.ToInt32());
-                    exception = new PermissionDeniedException(session.User.App, messageString, userInfo);
+                    Session.RaiseError(session, exception);
                 }
-                else
+                catch (Exception ex)
                 {
-                    exception = new SessionException(messageString, errorCode);
+                    Logger.Default.Log(LogLevel.Warn, $"An error has occurred while handling a session error: {ex}");
                 }
```
fealebenpae, line 234: We still need a `finally` block here that will dispose the session handle we created earlier.

papafe: The handle is being released in `ClientResetException` when `InitiateClientReset` is invoked

fealebenpae: What if the user doesn't call `InitiateClientReset` or the exception isn't `ClientResetException` to begin with?

papafe: We have a problem 😁

fealebenpae (approved):

nirinchev (approved):
LGTM, but let's merge this after the client reset work as it'll be a lot easier to resolve conflicts here than there.

LaPeste (approved):
Thank you for fixing this. If there is need of help with merging this fix with the client reset, don't hesitate to ask.

nirinchev:
@papafe We merged the client reset PR so this can go in now.

papafe:
@nirinchev @fealebenpae I've merged from master, so maybe it's better to give another look before merging

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 * None
 
 ### Fixed
-* None
+* Fixed an issue that could cause errors or loops with client reset on certain platforms. (PR [2830](https://github.com/realm/realm-dotnet/pull/2830))
```
LaPeste, line 7: Shouldn't it mention what platforms?

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -267,57 +267,64 @@ public override void Unbind()
         [MonoPInvokeCallback(typeof(NativeMethods.SessionErrorCallback))]
         private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorCode, PrimitiveValue message, IntPtr userInfoPairs, IntPtr userInfoPairsLength, bool isClientReset, IntPtr managedSyncConfigurationBaseHandle)
         {
-            try
+            var handle = new SessionHandle(null, sessionHandlePtr);
```
LaPeste, line 270: Why did you remove the try/catch? User code could throw at line 284 or 288. Same as you've done at line 307, where you want to catch user exceptions that could come from line 311 or 316. And `handle` should be disposed also in the `finally` block of this new try/catch.

LaPeste (changes requested):
~Beside the try/catch change, could you also fix `SessionHandle` line 411? The code analyzer is right. There `null` should be replaced by `IntPtr` as `managedSessionHandle == null` is always false.~ Never mind, Nikola fixed it in a [commit](https://github.com/realm/realm-dotnet/pull/2899/commits/a3300eb1d3475cc1077af511bab0ea34c3eaabc5) in my PR.
So just the 2 things I commented in the code need attention.

CHANGELOG.md
papafe: I did not write it because there was a different behaviour on iOS/Android/Windows and didn't want to get too much into details.

Realm/Realm/Handles/SessionHandle.cs
papafe: I agree, I did it too hastily. I've tried to refactor it, so that all the user callbacks are executed at the end so we don't need to have two try/catch.

CHANGELOG.md
LaPeste: I understand that the bug created different behaviours (errors) on different platforms. In fact, not being specific like e.g. "*error Y* on iOS and *error X* on Windows" is very fine in my opinion.
But I still see value for our users to be able to quickly read something like "fixed issue A and B that we sometimes saw happening on platform X and Y".
Just my opinion.

nirinchev: I think it's fine to just leave out the `certain platforms` altogether. It was a bug and even if it worked by accident on some platforms, that's probably not worth calling out.

papafe, line 7:
```suggestion
* Fixed an issue that could cause errors or loops with client reset. (PR [2830](https://github.com/realm/realm-dotnet/pull/2830))
```

papafe:
@nirinchev This has been open for quite some time and a lot of changes have been done since then. Should we just close this?

PR #2831, opened 2022-03-03
AWS Device Farm

fealebenpae:
## Description
Run Android tests on AWS Device Farm physical devices.


.github/templates/main.yml
```diff
@@ -213,52 +213,48 @@ steps:
...
-    if: steps.avd-cache.outputs.cache-hit != 'true'
-    uses: reactivecircus/android-emulator-runner@v2
-    with:
-      api-level: 29
-      arch: x86_64
-      force-avd-creation: false
-      emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-      disable-animations: false
-      script: echo "Generated AVD snapshot for caching."
+  - #@ msbuildOnWin("Tests/Tests.Android", target="SignAndroidPackage", AndroidSupportedAbis="armeabi-v7a", AndroidUseSharedRuntime="False", EmbedAssembliesIntoApk="True")
```
nirinchev, line 216:
```suggestion
  - #@ template.replace(msbuildOnWin("Tests/Tests.Android", target="SignAndroidPackage", AndroidSupportedAbis="armeabi-v7a", AndroidUseSharedRuntime="False", EmbedAssembliesIntoApk="True"))
```

.github/workflows/test-xamarinandroid.yml
```diff
@@ -17,14 +17,22 @@ name: test-xamarinandroid
...
+        required: true
+      DEVICEFARM_ANDROID_POOL_ARN:
+        required: true
 env:
   REALM_DISABLE_ANALYTICS: true
   DOTNET_NOLOGO: true
 jobs:
   run-tests:
-    runs-on: macos-latest
+    runs-on: windows-latest
```
nirinchev, line 33: Can this run on Linux instead?

.github/workflows/test-xamarinandroid.yml
```diff
@@ -53,57 +61,71 @@ jobs:
...
-      with:
-        path: |
-          ~/.android/avd/*
-          ~/.android/adb*
-        key: android-avd
-    - name: Create AVD for caching
-      if: steps.avd-cache.outputs.cache-hit != 'true'
-      uses: reactivecircus/android-emulator-runner@v2
+      run: msbuild Tests/Tests.Android -t:SignAndroidPackage -p:Configuration=Release -restore -p:RestoreConfigFile=Tests/Test.NuGet.Config -p:UseRealmNupkgsWithVersion=${{ inputs.version }} -p:AndroidSupportedAbis=armeabi-v7a -p:AndroidUseSharedRuntime=False -p:EmbedAssembliesIntoApk=True
+    - name: Configure AWS Credentials
```
nirinchev, line 65: Why do we need to configure AWS credentials here? Shouldn't this be done by the device farm action?

.github/workflows/test-xamarinandroid.yml
```diff
@@ -53,57 +61,71 @@ jobs:
...
+        project_arn: ${{ secrets.DEVICEFARM_PROJECT_ARN }}
+        device_pool_arn: ${{ secrets.DEVICEFARM_ANDROID_POOL_ARN }}
+        app_file: ${{ github.workspace }}/Tests/Tests.Android/bin/Release/io.realm.xamarintests-Signed.apk
+        app_type: ANDROID_APP
+        test_type: APPIUM_PYTHON
+        test_package_file: https://github.com/realm/aws-devicefarm-sample-data/releases/download/0.3/sample_env_python3.zip
+        test_package_type: APPIUM_PYTHON_TEST_PACKAGE
+        test_spec_file: test_spec.yaml
+        test_spec_type: APPIUM_PYTHON_TEST_SPEC
+        remote_src: true
```
nirinchev, line 84: These don't seem like something that should be configurable. It's fine to leave them like this for now, but I'd rather set them as defaults in the action and not have us specify them every time.

.github/workflows/test-xamarinandroid.yml
```diff
@@ -53,57 +61,71 @@ jobs:
...
+        $suites = Get-DFSuiteList -Arn $jobs[0].Arn
+        $artifacts = Get-DFArtifactList -Arn $suites[1].Arn -Type File | Where-Object { $_.Name -EQ "Logcat" }
+        echo "::group::Logcat"
+        Invoke-WebRequest -Uri $artifacts[0].Url | Select-Object -Expand RawContent
+        echo "::endgroup::"
+      name: Fetch test artifacts
+    - run: |
+        echo "::group::Data"
+        echo (ConvertFrom-Json '${{ steps.run_tests.outputs.data }}' | ConvertTo-Json)
+        echo "::endgroup::"
```
nirinchev, line 120: This too seems like something that should be done internally in the device farm action rather than here.

nirinchev (commented):
Generally, it seems fine to me, though I'd prefer to move a lot of the complexity to the device farm action as it seems counterproductive to maintain it across multiple repositories .

.github/workflows/test-xamarinandroid.yml
fealebenpae: We can't build Xamarin.Android projects on Linux.

.github/workflows/test-xamarinandroid.yml
fealebenpae: I filed https://github.com/realm/aws-devicefarm/issues/6 to track this. It's something @bwachter and I already discussed.

.github/workflows/test-xamarinandroid.yml
fealebenpae: I'll wait until I have a working version for iOS as well, or until we need to integrate this into other SDKs. It's not unlikely that other SDKs will use a different test package or spec. But I agree, this can use a refactoring on a second pass.

.github/workflows/test-xamarinandroid.yml
fealebenpae: That's a convention with actions using the AWS SDK. Besides, in this specific case I need to use some AWS PowerShell cmdlets and they also need the same credentials.

nirinchev (approved):

Issue #2850, opened 2022-03-15
Skip PR builds for docs updates


nirinchev:
We should avoid running the full build if no code has changed.

PR #2851, opened 2022-03-15
Don't run workflows when code hasn't changed

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2850

### DIFF
--- a/.github/templates/main.yml
+++ b/.github/templates/main.yml
@@ -413,6 +413,11 @@ name: main
     branches:
       - main
   pull_request:
+    paths:
+      - '**.cs'
+      - '**.cpp'
+      - '**.hpp'
+      - '**.csproj'
 env:
   REALM_DISABLE_ANALYTICS: true
   DOTNET_NOLOGO: true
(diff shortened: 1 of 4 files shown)


fealebenpae (commented):
Could we use `Realm/**` and `wrappers/**` for paths instead? Also, do you know if this will be triggered for PRs with just the realm-core submodule changed?

nirinchev:
@fealebenpae now it should run if we update realm-core. I wanted to be more explicit about the monitored types because we do have some metadata/docs files in `Realm`, `Tests`, and so on and I'd like to avoid running the PR flow for those. I went through ~10-15 merged PRs and this filter would have done the right thing for them.

.github/templates/main.yml
```diff
@@ -413,6 +413,15 @@ name: main
     branches:
       - main
   pull_request:
+    paths:
+      - '**.cs'
+      - '**.cpp'
+      - '**.hpp'
+      - '**.csproj'
```
fealebenpae, line 420:
```suggestion
      - '**.csproj'
      - '*CMakeLists.txt'
```

fealebenpae (approved):

.github/templates/main.yml
fealebenpae: Looks good, but we should also take CMakeLists.txt files into account.

PR #2861, opened 2022-03-17
Disable artifacts cleanup

fealebenpae:
[GitHub Actions can now repeat just the failed jobs in a workflow run](https://github.blog/2022-03-16-save-time-partial-re-runs-github-actions/). Our workflow doesn't support that because we always clear the wrappers artifacts.


nirinchev (changes requested):
I'll integrate that as part of https://github.com/realm/realm-dotnet/pull/2860. Let's wait until that is mergable as I don't feel like dealing with merge conflicts :D

Issue #2865, opened 2022-03-21
[Bug]: CertPathValidatorException on Android 6.


luccasclezar:
### What happened?

MongoDB Realm does not connect on Android 6, any Android version after that is working correctly.

The line `Realms.Sync.User user = await App.MongoApp.LogInAsync(credentials);` throws a `java.security.cert.CertPathValidatorException`.

I don't know if this issue is occurring on real devices or only on emulators, I don't have any Android 6 device to test.

### Repro steps

Just try to use MongoDB Realm on Android 6.

[androidmtest.zip](https://github.com/realm/realm-dotnet/files/8319752/androidmtest.zip)

### Version

10.10.0

### What SDK flavour are you using?

MongoDB Realm (i.e. Sync, auth, functions)

### What type of application is this?

Xamarin

### Client OS and version

Android 6

### Stacktrace of the exception/crash you're getting

```shell
code 999: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.
  at Realms.Sync.AppHandle.LogInAsync (Realms.Sync.Native.Credentials credentials) [0x00055] in D:\a\realm-dotnet\realm-dotnet\Realm\Realm\Handles\AppHandle.cs:282 
  at Realms.Sync.App.LogInAsync (Realms.Sync.Credentials credentials) [0x0003d] in D:\a\realm-dotnet\realm-dotnet\Realm\Realm\Sync\App.cs:208
```

LaPeste:
In the meanwhile I can reproduce the issue on for API 23 but not on another APIs (I tried API 30). As originally reported by the user.

nirinchev:
This is an issue due to older versions of Android not trusting the new root certificate of Let's Encrypt. There are no clear workaround currently.

We have two options for solving this:

### Expose a way to provide custom HttpClientHandler

Currently we [construct an http client](https://github.com/realm/realm-dotnet/blob/b425ca2dd0f242961a10781dcd7fa354ee0ec45c/Realm/Realm/Native/HttpClientTransport.cs#L107) with the default handler. We can expose a configuration mechanism for users to provide their custom http client handlers.

### Multitargeting

When we start multitargeting the Realm assembly, we can provide an android implementation that trusts the ISRG root with the following code:

```csharp
private const string ISRGRoot =
@"MIIFazCCA1OgAwIBAgIRAIIQz7DSQONZRGPgu2OCiwAwDQYJKoZIhvcNAQELBQAw
[...]

PR #2883, opened 2022-03-25
Expose a HttpClientHandler property on app config

nirinchev:
## Description

This pulls in the GCHandleHolder mechanics from https://github.com/realm/realm-dotnet/pull/2777 as I didn't want to block this on the client reset work.

Fixes https://github.com/realm/realm-dotnet/issues/2865

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm/Sync/App.cs
```diff
@@ -150,6 +153,8 @@ public static App Create(AppConfiguration config)
                 }
             }
 
+            var httpClient = new HttpClient(config.HttpClientHandler ?? new HttpClientHandler());
```
fealebenpae, line 156: I think we should initialize the `AppConfiguration.HttpClientHandler` property with the default value, instead of leaving it null and instantiating the default handler here.

fealebenpae (approved):
Looks good, but I have a small default value suggestion.

PR #2899, opened 2022-04-01
Write/Commit Async API

LaPeste:
## Description
This PR adds Write/Commit Async API. 

The only thing that testing in this PR doesn't do is checking if cancelling while waiting for an async transaction does indeed cancel the queued callback in CORE. I don't see an easy way of testing this with the current implementation. My expectation is that this aspect is going to be easier to test when we introduce `CancellationToken`s in async calls. There will be a PR for this coming soon ™️.

##  TODO

* [x] Changelog entry
* [x] Write API documentation
* [x] Tests (if applicable)
* [ ] Manually close the JIRA ticket once it's merged as there was not a github issue to link this PR to.


wrappers/src/shared_realm_cs.cpp
```diff
@@ -63,6 +65,7 @@ namespace binding {
     std::function<LogMessageT> s_log_message;
     std::function<MigrationCallbackT> s_on_migration;
     std::function<ShouldCompactCallbackT> s_should_compact;
+    std::function<HandleTaskCompletionCallbackT> s_handle_write_commit_async;
```
nirinchev, line 68: wrong var name

wrappers/src/shared_realm_cs.cpp
```diff
@@ -160,7 +163,8 @@ REALM_EXPORT void shared_realm_install_callbacks(
     ObjectNotificationCallbackT* notify_object,
     DictionaryNotificationCallbackT* notify_dictionary,
     MigrationCallbackT* on_migration,
-    ShouldCompactCallbackT* should_compact)
+    ShouldCompactCallbackT* should_compact,
+    HandleTaskCompletionCallbackT* handle_write_commit_async)
```
nirinchev, line 167: arg name

wrappers/src/shared_realm_cs.cpp
```diff
@@ -369,6 +374,41 @@ REALM_EXPORT uint64_t shared_realm_get_schema_version(SharedRealm& realm, Native
     });
 }
 
+REALM_EXPORT uint32_t shared_realm_begin_transaction_async(SharedRealm& realm, void* tcs_ptr, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        return realm->async_begin_transaction([tcs_ptr, ex]()
+        {
+            s_handle_write_commit_async(tcs_ptr, ex);
+        }, true);
```
nirinchev, line 383: Can we add a comment why `notify_only` is always set to `true`?

wrappers/src/shared_realm_cs.cpp
```diff
@@ -369,6 +374,41 @@ REALM_EXPORT uint64_t shared_realm_get_schema_version(SharedRealm& realm, Native
     });
 }
 
+REALM_EXPORT uint32_t shared_realm_begin_transaction_async(SharedRealm& realm, void* tcs_ptr, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        return realm->async_begin_transaction([tcs_ptr, ex]()
+        {
+            s_handle_write_commit_async(tcs_ptr, ex);
```
nirinchev, line 382: We should not be reusing `ex` here.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -369,6 +374,41 @@ REALM_EXPORT uint64_t shared_realm_get_schema_version(SharedRealm& realm, Native
...
+        }, true);
+    });
+}
+
+REALM_EXPORT uint32_t shared_realm_commit_transaction_async(SharedRealm& realm, void* tcs_ptr, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        return realm->async_commit_transaction([tcs_ptr](std::exception_ptr err) {
+            NativeException::Marshallable nativeEx { RealmErrorType::NoError };
```
nirinchev, line 391: Isn't `NoError` the default here?

wrappers/src/shared_realm_cs.cpp
```diff
@@ -369,6 +374,41 @@ REALM_EXPORT uint64_t shared_realm_get_schema_version(SharedRealm& realm, Native
...
+            if (err) {
+                nativeEx = convert_exception(err).for_marshalling();
+            }
+            s_handle_write_commit_async(tcs_ptr, nativeEx);
+        }, false);
```
nirinchev, line 396: We don't need `false` here

nirinchev (changes requested):
Reviewed the implementation, but haven't looked at the tests or docs. @fealebenpae can you double check the native code to make sure we haven't missed something.

wrappers/src/shared_realm_cs.cpp
fealebenpae: We can't pass `ex` in because it's a reference to a stack-allocated variable in C#, and the callback to `async_begin_transaction` can be called some time in the future (like if the write lock is held by another thread or process) when the variable no longer exists in C# and it'll crash if we try to access it. If `async_begin_transaction` doesn't pass in an `std::exception_ptr` to its callback like `async_commit_transaction` then we simply don't need the `NativeException` argument to `s_handle_write_commit_async`. If you need to reuse the callback type then just initialize a dummy `NativeException` with `NoError`.

fealebenpae (changes requested):

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -238,6 +255,7 @@ public static void Initialize()
             DictionaryHandle.KeyNotificationCallback notifyDictionary = DictionaryHandle.NotifyDictionaryChanged;
             NativeMethods.MigrationCallback onMigration = OnMigration;
             NativeMethods.ShouldCompactCallback shouldCompact = ShouldCompactOnLaunchCallback;
+            NativeMethods.HandleTaskCompletionCallback handleWriteCommitAsync = HandleWriteCommitAsyncCallback;
```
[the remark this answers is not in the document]
LaPeste: What is that you'd like to see? Something like `handleTaskCompletion`?

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -454,22 +473,67 @@ public void BeginTransaction()
...
             nativeException.ThrowIfNecessary();
+        }
+
+        public bool IsInTransaction()
+        {
+            var result = NativeMethods.is_in_transaction(this);
+            return result;
+        }
+
+        public bool IsInAsyncTransaction()
```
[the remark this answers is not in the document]
LaPeste: Are you suggesting that it is not that important to internally, at the SDK level, know if the transaction is async or not?
I'm already merging the 2 checks in the public API => property `Realm.IsInTransaction`.

Realm/Realm/Transaction.cs
```diff
@@ -55,30 +70,80 @@ public void Dispose()
...
+        public async Task CommitAsync()
+        {
+            EnsureActionFeasibility();
+            _asyncTransactionHandle = await _realm.SharedRealmHandle.CommitTransactionAsync();
+            FinishTransaction();
+        }
+
+
+        // TODO andrea: check usefulness of this method
+        private bool IsAsync()
```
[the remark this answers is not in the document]
LaPeste: Do you see any value in the `Debug.Assert` in this method? I thought it'd be a good idea to have it around. If you do see value, where would you recommend to have it? At the beginning of the `Transaction.Rollback`, perhaps?

Realm/Realm/Handles/SharedRealmHandle.cs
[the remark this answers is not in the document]
nirinchev: Yeah - the var name should be closer to the type name.

Realm/Realm/Handles/SharedRealmHandle.cs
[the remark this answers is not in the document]
nirinchev: We don't seem to use it for anything at the SDK layer, so I don't see a reason to differentiate. We can always expose it later if it turns out we care about it.

Realm/Realm/Transaction.cs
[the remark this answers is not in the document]
nirinchev: No, I don't see much value in debug asserting here - this will get optimized away when we compile, so it won't reach user code. I expect Core already validates this, but if it doesn't, we can throw an exception in the wrappers for commit/rollback

wrappers/src/shared_realm_cs.cpp
LaPeste: If you're referring to the fact that `handle_error` sets the type to `NoError`. That's correct. If you are just asking in general, then no because `NoError` is -1. (You can see it [here](https://github.com/realm/realm-dotnet/blob/67e05668a650b2aa7e5e262fc652d9e29aac478c/wrappers/src/realm_error_type.hpp#L27))
Either way, are you suggesting that being as implicit as possible is what we should strive for? That doesn't sound as a good practice as it can be one more source of bugs, plus while reading it'd require browsing around (as I had to do) to see the default value.
Have I misunderstood something?

wrappers/src/shared_realm_cs.cpp
LaPeste: This must be on a similar note as the comment above. Is this right?

Realm/Realm/Handles/SharedRealmHandle.cs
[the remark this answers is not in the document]
LaPeste: Then I prefer to change the type name to be more explicit about what it's supposed to handle. Is that ok with you?

wrappers/src/shared_realm_cs.cpp
nirinchev: I would have preferred that a default-constructed `NativeException::Marshallable` sets the exception type to no error, but since that's not the case, then the code as it is right now is just fine.

Realm/Realm/Handles/SharedRealmHandle.cs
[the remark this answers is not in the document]
nirinchev: Isn't the same callback used for both begin and commit write?

Realm/Realm/Transaction.cs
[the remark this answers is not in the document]
LaPeste: I only meant that `Debug.Assert` for our development only. Mostly to ensure correctness in the way managed keeps track of `_asyncTransactionHandle`.
I was more worried about our side than core :D. I will just take that part out.
Thanks.

wrappers/src/shared_realm_cs.cpp
nirinchev: Yes - if we're providing explicit argument values, we should name the arguments. As a reader, I don't know what `false` means in this case, so it's more harmful than helpful.

Realm/Realm/Handles/SharedRealmHandle.cs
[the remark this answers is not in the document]
LaPeste: Yes, it is. That's why I'd rename the type to be `HandleWriteCommitAsync`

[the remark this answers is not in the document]
nirinchev: I think `HandleTaskCompletion` is better in this case because `HandleWriteCommit` sounds like it will only handle the commit part, not the begin write part.

CHANGELOG.md
```diff
@@ -1,7 +1,51 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+* Added a stroger write and commit async API. This deprecates `Realm.WriteAsync(Action<Realm> action)`, `Real.WriteAsync<T>(Func<Realm, IQueryable<T>> function)`, `Realm.WriteAsync<T>(Func<Realm, IList<T>> function)` and `Realm.WriteAsync<T>(Func<Realm, T> function)`. The new API is made of `Realm.WriteAsync<T>(Func<T> function)`, `Realm.WriteAsync(Action action)`, `Realm.BeginWriteAsync()` and `Transaction.CommitAsync()`. While the `Transaction.Rollback()` stays unchanged as it does not need an async version. Importantly, the async calls are executed in the FIFO order. And additionally you can mix and match async and sync calls, albeit only the async calls are guaranteed to be executed in the FIFO order. As an example, the following is fully legal code:
```
peppy, line 4: "stroger" seems like a typo here

LaPeste: Thank you for spotting that.

CHANGELOG.md
```diff
@@ -1,7 +1,22 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+* Added a stronger write and commit async API. This deprecates `Realm.WriteAsync(Action<Realm> action)`, `Real.WriteAsync<T>(Func<Realm, IQueryable<T>> function)`, `Realm.WriteAsync<T>(Func<Realm, IList<T>> function)` and `Realm.WriteAsync<T>(Func<Realm, T> function)`. The new API is made of `Realm.WriteAsync<T>(Func<T> function)`, `Realm.WriteAsync(Action action)`, `Realm.BeginWriteAsync()` and `Transaction.CommitAsync()`. While the `Transaction.Rollback()` stays unchanged as it does not need an async version. Importantly, you can mix and match async and sync calls. And the new WriteAsync API executes on the same thread where it's called, so you can capture Realms and RealmObjects/EmbeddedObjects in the delegate like so:
```
nirinchev, line 4: I think we can reword this a little to be more explicit - e.g. "stronger" is a bit vague and the key difference of the API is left to the end sentence, while we should probably explain the distinction between the old and the new earlier in the text. I'm thinking something like:

```
* Added a more efficient replacement for `Realm.WriteAsync`. The previous API would start a background thread, open the Realm there and run a synchronous write transaction on the background thread. The new API will asynchronously acquire the write lock (begin transaction) and asynchronously commit the transaction, but the actual write block will execute on the original thread. This means that objects/queries captured before the block can be used inside the block without relying on threadsafe references. 
  * Enumerate all the new API
  * Enumerate the deprecated API
  * Explain the behavior when started on a bg thread (i.e. it will run synchronously)
  * example
```

Realm/Realm/Realm.cs
```diff
@@ -785,6 +785,38 @@ public T Write<T>(Func<T> function)
             return result;
         }
 
+        /// <summary>
+        /// Factory for a write <see cref="Transaction"/>. Essential object to create scope for updates.
+        /// </summary>
```
nirinchev, line 790: The original docs are kind of vague, so rather than copying them, I suggest we rework them. You may want to update the `BeginWrite` docs to be more helpful as well.

```suggestion
        /// <summary>
        /// Asynchronously begin a write transaction for this Realm.
        /// </summary>
        /// <remarks>
        /// This method asynchronously acquires the write lock and then dispatches the continuation on the original
        /// thread the Realm was opened on. The transaction can then be committed either asynchronously or
        /// synchronously.
        /// <br/>
        /// When invoked on a thread without SynchronizationContext (i.e. typically background threads), this method
        /// calls <see cref="BeginWrite"/> and executes synchronously.
        /// </remarks>
```

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
                 }
             });
         }
+
+        [Test]
+        public void AsyncWrite_Action()
```
nirinchev, line 163: We're super inconsistent about test case naming, but it's a good idea to at least be consistent within the same PR 😄 My suggestion would be to use the `Action_OptionalCondition_Expectation` pattern. For this test, this should roughly translate to `AsyncWrite_WithActionCallback_PersistsChanges`.

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+                Assert.That(_realm.All<Person>().Count(), Is.EqualTo(0));
+            });
+        }
+
+        [Test]
+        public void AsyncBeginWriteAndClose_DoesNot_PersistData()
+        {
+            TestHelpers.RunAsyncTest(async () =>
+            {
+                IQueryable<Person> peopleQuery = null;
```
nirinchev, line 223: Why do we need this declared here?

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+
+                using (var innerRealm = GetRealm(_realm.Config))
```
nirinchev, line 225: Why do we need an extra Realm instance for this test?

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+                    await writeTask;
+
+                    var peopleQuery = _realm.All<Person>();
+                    Assert.That(peopleQuery.Count, Is.EqualTo(2));
+                    Assert.That(peopleQuery.ElementAt(0).FirstName, Is.EqualTo("Marco"));
+                    Assert.That(peopleQuery.ElementAt(1).FirstName, Is.EqualTo("Giovanni"));
+                }
+                catch
+                {
+                    transaction.Dispose();
```
nirinchev, line 277: Why do we need the try-catch in this method?

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+                }
+            });
+        }
+
+        [Test]
+        public void WriteAsync_CaptureObjectsInDelegate_DoesNotThrow()
```
nirinchev, line 283: Not sure I understand the purpose of this test - what could happen for this to ever throw?

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+                {
+                    var person = new Person { FullName = personName };
+                    _realm.Add(person);
+                    return person;
+                });
+
+                Assert.That(peopleQuery.Count(), Is.EqualTo(1));
+                Assert.That(peopleQuery.First().FullName, Is.EqualTo(personName));
+                Assert.That(person, Is.Not.Null);
+                Assert.That(person.FullName, Is.EqualTo(personName));
```
nirinchev, line 318: These conditions could be simplified to

```suggestion
                Assert.That(peopleQuery.Single(), Is.EqualTo(person));
```

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+            });
+        }
+
+        [Test]
+        public void AsyncWrite_Func_List()
```
nirinchev, line 323: I don't understand this test - it seems to be testing that you can return a container object, containing a managed `RealmObject`, but that's implicitly the case if the test above it passes. Did you mean for this to test returning a `RealmList` property?

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+                Assert.That(peopleQuery.Count(), Is.EqualTo(1));
+                Assert.That(peopleQuery.First().FullName, Is.EqualTo(personName));
+                Assert.That(list, Is.Not.Null);
+                Assert.That(list.Count, Is.EqualTo(1));
+                Assert.That(list.First().FullName, Is.EqualTo(personName));
+            });
+        }
+
+        [Test]
+        public void AsyncWrite_Func_Queryable()
```
nirinchev, line 347: Same here - why are we testing this?

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+                Assert.That(peopleQuery.Count(), Is.EqualTo(1));
+                Assert.That(peopleQuery.First().FullName, Is.EqualTo(personName));
+                Assert.That(queryable, Is.Not.Null);
+                Assert.That(queryable.Count, Is.EqualTo(1));
+                Assert.That(queryable.First().FullName, Is.EqualTo(personName));
+            });
+        }
+
+        [Test]
+        public void AsyncBeginWrite_StackReferencedNativeException_Does_Not_Throw()
```
nirinchev, line 371: Not sure I understand this test - can you add some comments around it?

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+                }, ct);
+
+                // give time for the lock to be acquired from the other task
+                await Task.Delay(2000);
+
+                Exception ex = null;
+
+                try
+                {
+                    await _realm.BeginWriteAsync();
```
nirinchev, line 394: How does this complete? As far as I can tell, the background Realm acquires a write lock first and never releases it - how can we acquire the write lock on the `_realm` instance?

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+                {
+                    new Person { FirstName = "Anderson" },
+                    new Person { FirstName = "Banderson" },
+                    new Person { FirstName = "Canderson" },
+                    new Person { FirstName = "Danderson" }
+                };
+
+                ConcurrentQueue<Person> markers = new();
+                Queue<Person> actualWriters = new();
+                Task[] tasks = new Task[people.Length];
```
nirinchev, line 443: Can we replace the explicit type declarations with `var`s?

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+                _realm.Write(() =>
+                {
+                    _realm.RemoveAll<Person>();
+                });
+
+                var thread = new AsyncContextThread();
+                Parallel.For(0, people.Length, index =>
+                {
+                    tasks[index] = thread.Factory.Run(async () =>
```
nirinchev, line 452: You're scheduling work on the same thread, so I don't think this test is testing our internal FIFO guarantees.

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+        public void Async_And_Sync_Write_Mixed_No_Deadlock()
+        {
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+            // RunAsyncTest needs a Func<Task> and Task.WaitAll isn't awaitable as it's synchronous.
+            // But we still need to run with an SyncronizationContext and a timer on the tests,
+            // which is exactly what RunAsyncTest supplies.
+            TestHelpers.RunAsyncTest(async () =>
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
+            {
+                Person[] people = new[]
```
nirinchev, line 484:
```suggestion
                var people = new[]
```

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+                    new Person { FirstName = "Danderson" }
+                };
+
+                var tasks = new Task[people.Length];
+
+                // opening realm here avoids that in the following ror loop multiple threads
+                // try to add Person to the schema at the same time
+                var realm = GetRealm(_realm.Config);
+
+                for (int i = 0; i < people.Length; i++)
```
nirinchev, line 498:
```suggestion
                for (var i = 0; i < people.Length; i++)
```

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -155,5 +158,405 @@ static async Task<long> MeasureTiming(Func<Task> func)
...
+                            using var realm = GetRealm(_realm.Config);
+                            realm.Write(() =>
+                            {
+                                realm.Add(people[index]);
+                            });
+                        }));
+                    }
+                }
+
+                Task.WaitAll(tasks);
```
nirinchev, line 527: You can use `await Task.WhenAll` here instead.

nirinchev (changes requested):
Looks pretty close to being done - I have some style comments and a bunch of questions around the tests, but the core feature seems sound to me.

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -454,25 +474,75 @@ public void BeginTransaction()
...
         }
 
-        public bool IsInTransaction()
+        public void CancelTransaction()
         {
-            var result = NativeMethods.is_in_transaction(this, out var nativeException);
+            var nativeException = new NativeException
+            {
+                type = RealmExceptionCodes.NoError
+            };
```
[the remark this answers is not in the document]
LaPeste: `RealmExceptionCodes::NoError = -1`, because of this I thought that's just safer to be explicit about it. This avoids a possible gotcha if in the future `CancelTransaction` is changed and the `nativeException` won't only be an `out` var.

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -728,6 +785,35 @@ private static bool ShouldCompactOnLaunchCallback(IntPtr delegatePtr, ulong tota
             return compactDelegate(totalSize, dataSize);
         }
 
+        [MonoPInvokeCallback(typeof(NativeMethods.HandleTaskCompletionCallback))]
+        private static void HandleTaskCompletionCallback(IntPtr tcs_ptr, NativeException ex)
+        {
+            // The task awaiting on this tcs should only be unblocked once the native call has finished.
+            // If not scheduled for later execution, Realm::run_writed, the caller of this callback, losing control
+            // of the flow which is given back to the managed task that was awaiting.
```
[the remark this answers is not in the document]
LaPeste: We can get in a call and I can try to express it in a less concise manner. There is also a typo, as the method is `Realm::run_writes`. Sorry about that.

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -48,6 +48,8 @@ internal class SharedRealmHandle : StandaloneHandle
         // it is important that children always have a reference path to their root for this to work
         private bool _noMoreUserThread;
 
+        private uint? _asyncTransactionHandle;
```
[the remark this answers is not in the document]
nirinchev: I think we should remove this field because even with cancellations implemented, I don't suppose we'll ever store the handle as a field-level variable. For now, we can disregard the returned values from begin/commit async and start using them in your next PR.

Realm/Realm/Handles/SharedRealmHandle.cs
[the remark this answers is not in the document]
nirinchev: I think if we change the signature of `CancelTransaction`, we should update the code then.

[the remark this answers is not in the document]
nirinchev: We never initialize the native exception variable because it always gets overwritten by native. E.g. if you have

```csharp
var somethingFromNative = NativeMethods.invoke_some_method(out var ex);
```

we allocate the memory in managed code, but the fields get set by native. Similarly, if you have an if clause, you can do:

```csharp
NativeException ex;
if (condition)
{
    NativeMethods.call_method_one(out ex);
}
else
{
    NativeMethods.call_method_two(out ex);
}
```

Tests/Realm.Tests/Database/AsyncTests.cs
LaPeste: It was mostly to make sure that the new API reflects the capabilities of the old. But you're right, from a technical standpoint this and the next test don't test anything. I'll remove them both.

Tests/Realm.Tests/Database/AsyncTests.cs
nirinchev: But in the test you have the transaction inside a `using` statement - so before the Realm is closed, the transaction is going to get disposed and thus rolled back.

Tests/Realm.Tests/Database/AsyncTests.cs
LaPeste: 1. The cancellation token is useless, so I removed it.
1. The task was running without a sync context. It does now.
1. It completes because they are running on different sync contexts on different Realms. So now that I look at it I'm not even sure that it tests what we thought it does. Again @fealebenpae should help us here.

Tests/Realm.Tests/Database/AsyncTests.cs
nirinchev: This would have worked previously as well because your person instance is unmanaged and can travel between threads freely.

Tests/Realm.Tests/Database/AsyncTests.cs
LaPeste: The scope doc says
> Ensure FIFO ordering of async write transactions originating from the same Scheduler

And talking with Jørgen, he also confirmed that the async write/commit is supposed to all happen within the same thread, which for us translates in "the same synchronization context". So I think that the test is correct.
If you think I'm wrong, you could explain in a call, as by writing I think it'd take too long.

nirinchev: Hm... in this case the test can be a lot simpler then - I got thrown off by the `Parallel.For` here. A simpler test would be:

```csharp
var acquisitionOrder = new List<int>();
var tasks = new List<Task>();

for (var i = 0; i < 5; i++)
{
    var index = i;
    tasks.Add(realm.WriteAsync(() =>
    {
        acquisitionOrder.Add(index);
    });
}

await Task.WhenAll(tasks);

Assert.That(acquisitionOrder, Is.EqualTo(new[] { 0, 1, 2, 3, 4 }));
```

Tests/Realm.Tests/Database/AsyncTests.cs
LaPeste: I see the mistake. I'll make that instance managed, so the test becomes meaningful. Thanks.

Tests/Realm.Tests/Database/AsyncTests.cs
LaPeste: Marked `Explicit` for the future

Realm/Realm/Transaction.cs
```diff
@@ -55,30 +55,46 @@ public void Dispose()
...
+
+        /// <summary>
+        /// Use to save the changes to the realm. If <see cref="Transaction"/> is declared in a <c>using</c> block,
+        /// it must be used before the end of that block.
+        /// </summary>
+        /// <returns>An awaitable <see cref="Task"/>.</returns>
+        public async Task CommitAsync()
+        {
+            EnsureActionFeasibility("commit");
+            await _realm.SharedRealmHandle.CommitTransactionAsync();
```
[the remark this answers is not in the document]
LaPeste: Good catch, thanks.

nirinchev (approved):

PR #2900, opened 2022-04-01
Add functionality to convert a non-sync to a sync Realm.

### LINKED ISSUE #2746: Implement Convert Non-Sync to Sync Realm

DominicFrei:
## Description

This PR adds the functionality to convert non-sync to sync Realms and back so that we can now offer all four combinations of conversations.

All four operations are combined in the overloaded `Realm.WriteCopy` which in return uses Core's `SharedRealm.WriteCopy`, `SharedRealm.ExportTo` and also `Group.Write` to accomplish the copy/conversation.

The corresponding tests in `InstanceTests.cs` and `SynchronizedInstanceTests.cs` have not been touched and keep testing the local->local and sync->sync part, the two additional tests test the other two parts (local->sync and sync->local).

Fixes #2746 

##  TODO

* [X] Changelog entry
* [X] Tests (if applicable)


wrappers/src/shared_realm_cs.cpp
```diff
@@ -460,13 +460,44 @@ REALM_EXPORT void thread_safe_reference_destroy(ThreadSafeReference* reference)
...
+        if (encryption_key) {
+            auto& key = *reinterpret_cast<std::array<char, 64>*>(encryption_key);
+            config.encryption_key = std::vector<char>(key.begin(), key.end());
+        }
+
+        realm->get()->export_to(std::move(config));
+    });
+}
+
+REALM_EXPORT void shared_realm_export_to_sync(SharedRealm* realm, Configuration configuration, SyncConfiguration sync_configuration, SchemaObject* objects, int objects_length, SchemaProperty* properties, uint8_t* encryption_key, NativeException::Marshallable& ex)
```
nirinchev, line 477: We don't need to pass the schema to this method as it's unlikely to be used in the copy. We can unify the `export_to_local` and `export_to_sync` functions by treating the sync configuration as optional (since it's a struct, it can't be really optional, but we can check for presence of `User` or something).

Realm/Realm/Realm.cs
```diff
@@ -1414,7 +1414,22 @@ public void WriteCopy(RealmConfigurationBase config)
...
+                else
+                {
+                    SharedRealmHandle.ExortTo(config.CreateNativeConfiguration(), config.EncryptionKey, null, null);
+                }
+            }
+            else
+            {
+                var copyIsSync = config is SyncConfigurationBase;
+                SharedRealmHandle.WriteCopy(config.DatabasePath, config.EncryptionKey, copyIsSync);
+            }
```
nirinchev, line 1432: We don't need all this separate handling. We should be able to delete the `shared_realm_write_copy` method and instead only use the `ExportTo` functionality.

nirinchev (changes requested):
I think it should be possible to unify all of this in a single method call with some optional parameters. The reason this is not the case for js is because they accept an overload that takes a string rather than a config. We don't so we don't need all the special casing they're doing.

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -523,9 +533,24 @@ public IntPtr ResolveReference(ThreadSafeReference reference)
             return result;
         }
 
-        public void WriteCopy(string path, byte[] encryptionKey)
+        public void ExortTo(Configuration configuration, byte[] encryptionKey, Sync.Native.SyncConfiguration? syncConfiguration = null, RealmSchema schema = null)
```
fealebenpae, line 536: Typo: should be `ExportTo`.

fealebenpae (changes requested):

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -396,6 +396,89 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+
+                using var originalRealm = GetRealm(originalConfig);
+
+                AddDummyData(originalRealm, true);
+
+                originalRealm.WriteCopy(copyConfig);
+
+                using var copiedRealm = GetRealm(copyConfig);
+
+                Assert.AreEqual(copiedRealm.All<ObjectIdPrimaryKeyWithValueObject>().Count(), DummyDataSize / 2);
```
LaPeste, line 430: We should use the constrain model for assertions as NUnit recommends in [their official guideline](https://docs.nunit.org/articles/nunit/writing-tests/assertions/assertions.html):
```suggestion
                Assert.That(copiedRealm.All<ObjectIdPrimaryKeyWithValueObject>().Count(), Is.EqualTo(DummyDataSize / 2));
```

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -396,6 +396,89 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+                AddDummyData(originalRealm, true);
+
+                await WaitForUploadAsync(originalRealm);
+                await WaitForDownloadAsync(originalRealm);
+
+                originalRealm.WriteCopy(copyConfig);
+
+                using var copiedRealm = GetRealm(copyConfig);
+
+                Assert.AreEqual(copiedRealm.All<ObjectIdPrimaryKeyWithValueObject>().Count(), DummyDataSize / 2);
```
LaPeste, line 478: Same as above comment.

LaPeste (changes requested):

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -396,6 +396,89 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+                    return copiedRealm.Add(new ObjectIdPrimaryKeyWithValueObject
+                    {
+                        StringValue = "Added from copy"
+                    });
+                });
+
+                // We cannot re-sync like we do it in `WriteCopy_CanSynchronizeData` since the `originalRealm` is not synced but instead
+                // at least try to upload and download the data in the `copiedRealm`.
+                await WaitForUploadAsync(copiedRealm);
+                await WaitForDownloadAsync(copiedRealm);
```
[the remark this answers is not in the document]
DominicFrei: That was my idea, yeah. Since only one of the two Realms is synced I can't test a full sync cycle like I did in the other test where we copy a synced Realm to a synced Realm but I wanted to test as much as possible and figured I could at least try to call upload and download to get soooome kind of indication if something completely went wrong.

Do you have any other ideas what would be good to check here?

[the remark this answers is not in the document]
papafe: I think that we already test that sync works in other contexts, so probably we don't need to do this here too in my opinion. Checking that the synced realm has the same kind of objects as the original one should be enough

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -396,6 +396,89 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+
+                AddDummyData(originalRealm, true);
+
+                originalRealm.WriteCopy(copyConfig);
+
+                using var copiedRealm = GetRealm(copyConfig);
+
+                Assert.AreEqual(copiedRealm.All<ObjectIdPrimaryKeyWithValueObject>().Count(), DummyDataSize / 2);
+
+                var fromCopy = copiedRealm.Write(() =>
```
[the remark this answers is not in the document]
DominicFrei: [af29ecc](https://github.com/realm/realm-dotnet/pull/2900/commits/af29eccefda22e78cae23bac6b3381643d068094)

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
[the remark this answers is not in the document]
DominicFrei: [af29ecc](https://github.com/realm/realm-dotnet/pull/2900/commits/af29eccefda22e78cae23bac6b3381643d068094)

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -396,6 +396,89 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+                {
+                    return copiedRealm.Add(new ObjectIdPrimaryKeyWithValueObject
+                    {
+                        StringValue = "Added from copy"
```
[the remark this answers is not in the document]
DominicFrei: The `fromCopy` added to the `copiedRealm` is synced and then checked in

```
var itemInOriginal = originalRealm.Find<ObjectIdPrimaryKeyWithValueObject>(fromCopy.Id);
Assert.That(itemInOriginal, Is.Not.Null);
Assert.That(itemInOriginal.StringValue, Is.EqualTo(fromCopy.StringValue));
```

if it could be synchronized back to the `originalRealm`.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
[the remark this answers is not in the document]
DominicFrei: [af29ecc](https://github.com/realm/realm-dotnet/pull/2900/commits/af29eccefda22e78cae23bac6b3381643d068094)

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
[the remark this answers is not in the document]
DominicFrei: [af29ecc](https://github.com/realm/realm-dotnet/pull/2900/commits/af29eccefda22e78cae23bac6b3381643d068094)

Realm/Realm/Handles/SharedRealmHandle.cs
DominicFrei: Fixed. 👍

Realm/Realm/Realm.cs
DominicFrei: I ended up keeping the name `write_copy` to keep it consistent with the SDK API. 👍

wrappers/src/shared_realm_cs.cpp
DominicFrei: Done.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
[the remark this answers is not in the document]
DominicFrei: [c8663c3](https://github.com/realm/realm-dotnet/pull/2900/commits/c8663c3eac1e5507a4c5c87bb7b14453b29208d9)

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
DominicFrei: [c8663c3](https://github.com/realm/realm-dotnet/pull/2900/commits/c8663c3eac1e5507a4c5c87bb7b14453b29208d9)

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
DominicFrei: [c8663c3](https://github.com/realm/realm-dotnet/pull/2900/commits/c8663c3eac1e5507a4c5c87bb7b14453b29208d9)

LaPeste (approved):

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -156,7 +156,7 @@ private static class NativeMethods
             public static extern IntPtr resolve_realm_reference(ThreadSafeReferenceHandle referenceHandle, out NativeException ex);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_write_copy", CallingConvention = CallingConvention.Cdecl)]
-            public static extern void write_copy(SharedRealmHandle sharedRealm, [MarshalAs(UnmanagedType.LPWStr)] string path, IntPtr path_len, byte[] encryptionKey, out NativeException ex);
+            public static extern void write_copy(SharedRealmHandle sharedRealm, Configuration configuration, IntPtr sync_configuration, byte[] encryptionKey, out NativeException ex);
```
nirinchev, line 159: Why is the sync config IntPtr?

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -523,9 +523,20 @@ public IntPtr ResolveReference(ThreadSafeReference reference)
...
 
-        public void WriteCopy(string path, byte[] encryptionKey)
+        public void WriteCopy(RealmConfigurationBase config)
         {
-            NativeMethods.write_copy(this, path, (IntPtr)path.Length, encryptionKey, out var nativeException);
+            IntPtr syncConfiguration;
+            if (config is SyncConfigurationBase syncConfig)
+            {
+                var configHandle = GCHandle.Alloc(syncConfig.CreateNativeSyncConfiguration());
+                syncConfiguration = GCHandle.ToIntPtr(configHandle);
```
nirinchev, line 532: This seems wrong - we should just pass the sync config as a value similarly to what we're doing in `OpenWithSync`.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -396,6 +396,76 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+                    originalConfig.EncryptionKey = TestHelpers.GetEncryptionKey(42);
+                }
+
+                var copyConfig = await GetIntegrationConfigAsync(Guid.NewGuid().ToString());
+                if (copyEncrypted)
+                {
+                    copyConfig.EncryptionKey = TestHelpers.GetEncryptionKey(23);
+                }
+
+                File.Delete(copyConfig.DatabasePath);
```
nirinchev, line 420: That should not be necessary - if it is, we have a bug somewhere.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -396,6 +396,76 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+
+                using var originalRealm = GetRealm(originalConfig);
+
+                AddDummyData(originalRealm, true);
+
+                originalRealm.WriteCopy(copyConfig);
+
+                using var copiedRealm = GetRealm(copyConfig);
+
+                Assert.That(copiedRealm.All<ObjectIdPrimaryKeyWithValueObject>().Count(), Is.EqualTo(originalRealm.All<ObjectIdPrimaryKeyWithValueObject>().Count()));
```
nirinchev, line 430: We're missing verification that the copied Realm can actually sync. There are two tests here - first is that we can call `WaitForUploadAsync` and that will complete successfully and that another user opening a Realm with the same partition can download the changes uploaded by the copied Realm.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -396,6 +396,76 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+                    originalConfig.EncryptionKey = TestHelpers.GetEncryptionKey(42);
+                }
+
+                var copyConfig = new RealmConfiguration(Guid.NewGuid().ToString());
+                if (copyEncrypted)
+                {
+                    copyConfig.EncryptionKey = TestHelpers.GetEncryptionKey(23);
+                }
+
+                File.Delete(copyConfig.DatabasePath);
```
nirinchev, line 452: Not needed.

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -396,6 +396,76 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
...
+                };
+                if (originalEncrypted)
+                {
+                    originalConfig.EncryptionKey = TestHelpers.GetEncryptionKey(42);
+                }
+
+                var copyConfig = await GetIntegrationConfigAsync(Guid.NewGuid().ToString());
+                if (copyEncrypted)
+                {
+                    copyConfig.EncryptionKey = TestHelpers.GetEncryptionKey(23);
```
nirinchev, line 417: When the copy is encrypted, we should test that we can't open it with a different encryption key.

nirinchev (commented):
Minor comments from me, otherwise looks much nicer with all the various cases moved into Core.

nirinchev (approved):

nirinchev:
Baas job failed due to a 503 and the C# CodeQL job is unrelated to these changes.

PR #2906, opened 2022-04-13
Target .NET 6

fealebenpae:
Update our NuGet package with support for `net6-android` and `net6-ios` runtimes:

- Split the iOS wrappers into a framework for iOS and a framework for iOS Simulator. The binaries go to `runtimes/ios/native` and `runtimes/iossimulator/native`, respectively.
- Move the Android wrappers to `runtimes/android-${ABI}/native`.
- Build the Realm assembly for .NET Standard 2.0 and .NET 6. That's because the .NET 6 runtime doesn't support the dll-mapping feature of Mono we used to redirect P/Invoke to the right `realm-wrappers.framework` binary at runtime. .NET 6 introduces the `NativeLibrary` API that allows native library resolution to be controlled at runtime.
- Split the old wrappers props file into TFM-specific props files for `monoandroid`, `xamarinios`, `xamarinmac`, and `net461`. Everything else relies on NuGet to resolve runtime-specific binaries from the right `runtimes` folder.


.github/workflows/wrappers.yml
```diff
@@ -310,63 +326,18 @@ jobs:
...
       if: steps.check-cache.outputs.cache-hit != 'true'
     - name: Store artifacts for wrappers-ios-${{ matrix.arch }}
       uses: actions/upload-artifact@v2
       with:
         name: wrappers-ios-${{ matrix.arch }}
-        path: ${{ github.workspace }}/wrappers/cmake/**/realm-wrappers.framework
-        retention-days: 1
-  ios:
-    runs-on: macos-latest
-    name: iOS XCframework
```
nirinchev, line 326: Are we no longer producing an xcframework?

Realm/Realm/Realm.csproj
```diff
@@ -33,19 +33,28 @@
...
       <LastGenOutput>Operator.cs</LastGenOutput>
     </None>
     <!-- iOS -->
-    <None Update="ios-dllmap.config" Pack="true" PackagePath="native\ios\$(AssemblyName).dll.config" />
-    <None Include="..\..\wrappers\build\iOS\$(Configuration)\realm-wrappers.xcframework\**">
+    <Content Include="ios-dllmap.config" Pack="true">
+      <PackagePath>contentFiles\any\xamarinios\$(AssemblyName).dll.config</PackagePath>
+      <PackageCopyToOutput>true</PackageCopyToOutput>
+    </Content>
+    <Content Include="wrappers-props\xamarinios\wrappers.props" Pack="true" PackagePath="build\xamarinios\Realm.props" />
```
nirinchev, line 45: Is `build/xamarinios` the folder being used both for Xamarin.iOS and net6-ios TFMs?

nirinchev (approved):
Looks like we have a failing test for string.Contains - likely due to now having that in the BCL. You'll probably need to #ifdef the `StringExtenstions.Contains` method as well as have the expression visitor use the system type instead.

Realm/Realm/Realm.csproj
fealebenpae: No, it's only the Xamarin.iOS TFM that uses it.

.github/workflows/wrappers.yml
fealebenpae: No, we ship the device framework under `runtimes/ios` and the simulator framework under `runtimes/iossimulator`.

PR #2922, opened 2022-05-06
Add support for InitialSubscriptions

nirinchev:
## Description

This is very much WIP - putting it up as a PR because I want to run tests on CI to see how much I broke things.

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


papafe:
Looks fine to me for what I can see. Did we change the way we wrap notifications because we expect to have multiple notifications coming from the callbacks from core?

nirinchev:
The reason why I changed how we do callbacks is because we didn't have a way to propagate errors from those. For example, if you had:

```csharp
var realm = Realm.GetInstance(new RealmConfiguration
{
    ShouldCompact = (_, __) =>
    {
        throw new Exception("I want to know what this is");
    }
});
```

We would catch the exception and opening the Realm would fail, but we would not propagate the correct exception to users. An alternative approach would have been to allocate a GCHandle for the exception, pass it down to native as `void*` and then somehow propagate it all the way back to the caller, but that seemed more involved in terms of changes.

fealebenpae (approved):

PR #2931, opened 2022-05-17
Modified type delegator and added test

### LINKED ISSUE #2930: Generalize implementation of IReflectableType
`TypeInfo` returned by the implementation of `IReflectableType` overrides only `GetDeclaredProperty`, but not `GetProperty`. The second one is more general and we should override that instead

papafe:
Fixes #2930

### DIFF
--- a/Realm/Realm/DataBinding/RealmObjectTypeDelegator.cs
+++ b/Realm/Realm/DataBinding/RealmObjectTypeDelegator.cs
@@ -34,7 +34,7 @@ namespace Realms.DataBinding
         protected override PropertyInfo GetPropertyImpl(string name, BindingFlags bindingAttr, Binder binder, Type returnType, Type[] types, ParameterModifier[] modifiers)
         {
             var result = base.GetPropertyImpl(name, bindingAttr, binder, returnType, types, modifiers);
-            var wovenAttribute = result.GetCustomAttribute<WovenPropertyAttribute>();
+            var wovenAttribute = result?.GetCustomAttribute<WovenPropertyAttribute>();
             if (wovenAttribute != null)
             {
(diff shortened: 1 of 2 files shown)


Realm/Realm/DataBinding/RealmObjectTypeDelegator.cs
```diff
@@ -31,19 +31,16 @@ internal RealmObjectTypeDelegator(Type type) : base(type)
         {
         }
 
-        public override PropertyInfo GetDeclaredProperty(string name)
+        protected override PropertyInfo GetPropertyImpl(string name, BindingFlags bindingAttr, Binder binder, Type returnType, Type[] types, ParameterModifier[] modifiers)
         {
-            return _propertyCache.GetOrAdd(name, n =>
+            var result = base.GetPropertyImpl(name, bindingAttr, binder, returnType, types, modifiers);
+            var wovenAttribute = result.GetCustomAttribute<WovenPropertyAttribute>();
```
fealebenpae, line 37:
```suggestion
            var wovenAttribute = result?.GetCustomAttribute<WovenPropertyAttribute>();
```
This way if the binding expression has a typo (e.g. `{Binding FulName}`) we won't crash because we tried to look up a non-existing property. I think there was an issue about that somewhere

fealebenpae (approved):
Looks good, and makes total sense. Just a small suggestion, though.

LaPeste (approved):

Realm/Realm/DataBinding/RealmObjectTypeDelegator.cs
papafe: Makes sense! Corrected

papafe:
I've also tested this with the changes made by Avalonia, and it works :)

PR #2932, opened 2022-05-18
Adds user deletion from the server

### LINKED ISSUE #2675: User account Deletion API
Add API to provide deletion of the user account by the app user.

LaPeste:
## Description
Adds user deletion from the server

Fixes #2675 

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


wrappers/src/app_cs.cpp
```diff
@@ -263,6 +274,15 @@ extern "C" {
         });
     }
 
+    REALM_EXPORT void shared_app_delete_user(SharedApp& app, SharedSyncUser& user, void* tcs_ptr, NativeException::Marshallable& ex)
+    {
+        return handle_errors(ex, [&]() {
+            app->delete_user(user, [tcs_ptr](util::Optional<AppError> err) {
```
nirinchev, line 280: Can this use `get_callback_handler`?

nirinchev (commented):
A few nits, otherwise looks solid

wrappers/src/app_cs.cpp
LaPeste: I read `get_callback_handler` too quickly 🤦🏻 .

Tests/Realm.Tests/Sync/UserManagementTests.cs
```diff
@@ -147,6 +147,35 @@ public void AppRemoveUser_RemovesUser()
...
         }
 
+        [Test]
+        public void AppDeleteUserFromServer_RemovesUser()
+        {
+            SyncTestHelpers.RunBaasTestAsync(async () =>
+            {
+                var username = SyncTestHelpers.GetVerifiedUsername();
+                var password = SyncTestHelpers.DefaultPassword;
+                var first = await GetUserAsync(app: null, username, password);
```
nirinchev, line 157: We don't have `second`, so this can be just `user`.

nirinchev (approved):
Looks great - a final nit from me

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+* Added the method `App.DeleteUserFromServerAsync` to delete a user from the server. The biproduct is that the user is also removed from the device together with their data. (Issue [#2675](https://github.com/realm/realm-dotnet/issues/2675))
```
papafe, line 4: It's "byproduct". 
Otherwise I think we could just say "Additionally the user is also removed...."

Tests/Realm.Tests/Sync/SynchronizedInstanceTests.cs
```diff
@@ -487,6 +487,11 @@ public void EmbeddedObject_WhenAdditiveExplicit_ShouldThrow()
 
                 originalRealm.WriteCopy(copyConfig);
 
+                // In the schema server-side for each model there's the field to hold what partition each object belongs to, namely realm_id.
```
papafe, line 490: In the server-side schema

Tests/Realm.Tests/Sync/UserManagementTests.cs
```diff
@@ -147,6 +147,35 @@ public void AppRemoveUser_RemovesUser()
...
+                {
+                    await DefaultApp.LogInAsync(Credentials.EmailPassword(username, password));
+                }
+                catch (Exception e)
+                {
+                    ex = e;
+                }
+
+                Assert.That(ex, Is.TypeOf<AppException>());
+                Assert.That(ex.Message, Is.EqualTo("InvalidPassword: invalid username/password"));
```
papafe, line 175: If you want to make it slightly shorter you can do something like:
```csharp
var ex = Assert.Throws<AppException>(() => await DefaultApp.Login....)
Assert.That(ex.Message, Is.EqualTo("InvalidPassword: invalid username/password"));

```
This way you don't need to write the try catch. Unfortunately the new style (Assert.That) does not allow to have the exception returned.

nirinchev: I'm not sure this will work with await 😕

CHANGELOG.md
nirinchev: A general good practice is to avoid passive voice in documentation. This is because we want to convey that an actor does something that leads to a result. With passive voice, things happen, making the source of the occurrence somewhat ambiguous. So in this instance, a more direct statement would be:

```
Added the method `App.DeleteUserFromServerAsync` to delete a user from the server. It will also invalidate the user locally as well as remove all their local data. It will not remove any data the user has uploaded from the server.
```

Tests/Realm.Tests/Sync/UserManagementTests.cs
papafe: It seems that there is the async version: https://docs.nunit.org/articles/nunit/writing-tests/assertions/classic-assertions/Assert.ThrowsAsync.html

LaPeste: I tried the async version already. It never returns. That's why I went with that solution. Also I understood that we wanted to only use the constraint model, which I think it's a good decision. So that was one more reason not to dig deeper.

PR #2936, opened 2022-05-19
Adds cancellation token to async write/commit API

### LINKED ISSUE #2925: Implement cancellation in async API
With the addition of the async write/commit API, it seems clear that we should allow the user to stop waiting on write lock acquisition and commit. So we should extend the API to accept cancellation tokens.

LaPeste:
## Description
Adds cancellation token to async write/commit API.

I've been debating whether or not to add the cancellation token to `Transaction.CommitAsync`. I was mostly concerned about users possibly thinking that they could cancel a scheduled commit. When in reality cancelling a `CommitAsync` simply avoids the triggering of the callback that indicates that the commit has actually just finished.
To add to this, testing seems quite difficult. I'd simply write a test where a `CommitAsync` is awaited and control is given back to the thread, which would just cancel the `CancellationTokenSource`. However, since a slow commit is generally a result of a disk being already quite busy with other I/O operations, I don't know of an easy way to simulate this to slow down the commit operation itself.

<br/>

Fixes #2925 

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -767,6 +774,13 @@ private static void HandleTaskCompletionCallback(IntPtr tcs_ptr, NativeException
 
         private static void HandleTaskCompletion<T>(IntPtr tcsPtr, Func<T> resultBuilder, NativeException ex)
         {
+            // if the cancellation of the task is executed just an instant before core calls this callback,
+            // the handle is already freed by the finally block of the managed caller.
+            if (tcsPtr == IntPtr.Zero)
+            {
+                return;
+            }
```
nirinchev, line 782: I don't think `tcsPtr` can ever be zero here. Even if the handle is freed by the finally block, we would have already passed a non-zero value to Core, which we'd get here.

That being said, we're in a bit of a trouble here, because Mono has a bug where calling `GCHandle.FromIntPtr` on a freed gchandle will crash the process. So if there's a chance that this callback gets invoked after we free the handle, we'll need to rework the code to make sure it never happens.

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -782,6 +796,17 @@ private static void HandleTaskCompletion<T>(IntPtr tcsPtr, Func<T> resultBuilder
             }
         }
 
+        private void OnTaskCancellation(uint? asyncTransactionHandle, TaskCompletionSource<object> tcs)
+        {
+            if (asyncTransactionHandle.HasValue)
+            {
+                NativeMethods.cancel_async_transaction(this, asyncTransactionHandle.Value, out var innerNativeException);
+                innerNativeException.ThrowIfNecessary();
```
nirinchev, line 804: If a native exception occurs we should set it on the `tcs` rather than throw it here.

Realm/Realm/Realm.cs
```diff
@@ -1002,9 +1011,9 @@ public async Task<T> WriteAsync<T>(Func<T> function)
                 return Write(function);
             }
 
-            using var transaction = await BeginWriteAsync();
+            using var transaction = await BeginWriteAsync(cancellationToken);
             var result = function();
-            await transaction.CommitAsync();
+            await transaction.CommitAsync(cancellationToken);
```
nirinchev, line 1016: We should not be passing the cancellation token to `CommitAsync` here - the reason is that it implies that the transaction will be cancelled, when this is not really the case. Consider the following code:

```csharp
try
{
    var cts = ...;
    await realm.WriteAsync(() => { ... }, cts);
}
catch (TaskCancelledException)
{
    // Here the user expects the transaction has been cancelled.
    // If we only canceled `CommitAsync`, the write will have
    // succeeded and will eventually be committed.
}
```

Realm/Realm/Transaction.cs
```diff
@@ -74,10 +75,23 @@ public void Commit()
 
         /// <summary>
         /// Use to save the changes to the realm. If <see cref="Transaction"/> is declared in a <c>using</c> block,
-        /// it must be used before the end of that block.
+        /// it must be used before the end of that block. It terminates when the changes are effectively written to disk.
```
nirinchev, line 78:
```suggestion
        /// it must be used before the end of that block. It completes when the changes are effectively written to disk.
```

Realm/Realm/Transaction.cs
```diff
@@ -74,10 +75,23 @@ public void Commit()
...
         /// </summary>
-        /// <returns>An awaitable <see cref="Task"/>.</returns>
-        public async Task CommitAsync()
+        /// <remarks>
+        /// Cancelling the returned <see cref="Task"/> simply makes the <see cref="Task"/>'s execution continue;
+        /// it does not cancel the commit action.
```
nirinchev, line 82: This language construct is a bit awkward. I'd suggest saying something like "Cancelling the returned Task will not prevent the write to disk but will immediately resolve the task with a `TaskCanceledException`".

Realm/Realm/Transaction.cs
```diff
@@ -74,10 +75,23 @@ public void Commit()
...
+        /// In fact, the commit action can't be stopped and continues running to completion in the background.<br/>
+        /// A use case for cancelling this action could be that you want to show users a pop-up indicating that the
+        /// data is being saved. But, you want to automcatically close such pop-up after a certain amount of time.
```
nirinchev, line 85:
```suggestion
        /// data is being saved. But, you want to automatically close such pop-up after a certain amount of time.
```

Realm/Realm/Transaction.cs
```diff
@@ -74,10 +75,23 @@ public void Commit()
...
+        /// Or, you may want to allow users to manually dismiss that pop-up.
+        /// </remarks>
+        /// <param name="cancellationToken">
+        /// Optional cancellation token to stop waiting on the returned <see cref="Task"/>.
+        /// </param>
+        /// <returns>
+        /// An awaitable <see cref="Task"/> that terminates when the committed changes are effectively written to disk.
```
nirinchev, line 92:
```suggestion
        /// An awaitable <see cref="Task"/> that completes when the committed changes are effectively written to disk.
```

nirinchev (commented):
Will need to review tests, but leaving some comments for the core implementation.

Realm/Realm/Realm.cs
LaPeste: I've had an exchange with @papafe exactly on the usefulness of the CancellationToken on the `CommitAsync` and exactly the unknown that could come from not knowing what the `WriteAsync` actually didn't do. I also wrote my initial thoughts on it in the description of this PR.
@papafe mentioned that's quite common for async calls with CancellationToken (e.g. http request) not to know what's actually been done. It made sense to me there that the user would check after the operation to know what was written and what not.
I suppose that your point is to keep things simple, which is a pretty good point. Let's see if @papafe has anything to add. Otherwise I'll go ahead with removing the ct from the `CommitAsync` call.

Realm/Realm/Handles/SharedRealmHandle.cs
LaPeste: > I don't think tcsPtr can ever be zero here. Even if the handle is freed by the finally block, we would have already passed a non-zero value to Core, which we'd get here.

You're right. I misjudged that. I was trying to avoid managed to resolve the handle when the pointer points to garbage.

> we're in a bit of a trouble here, because Mono has a bug where calling GCHandle.FromIntPtr on a freed gchandle will crash the process

Not good

> if there's a chance that this callback gets invoked after we free the handle

I have not seen this happening while testing, but I thought that being in a multi-threaded scenario we can't know what timing does, so I wanted to be on the safer side.

Realm/Realm/Realm.cs
nirinchev: I think it's fine to have a cancellation token for `CommitAsync` (i.e. if the user wants to manually pass a token there, that's fine). My argument is only for `WriteAsync` - i.e. there we only pass the cancellation token to `BeginWriteAsync` and not to `CommitAsync`. I'm thinking something like:

```csharp
using var transaction = await BeginWriteAsync(cancellationToken);
var result = function();

// Last chance to rollback the transaction - we manually check for cancellation
if (cancellationToken.IsCancellationRequested)
{
    throw new TaskCanceledException();
}

// No cancellation token here because we can't really cancel the commit
await transaction.CommitAsync();
```

Realm/Realm/Handles/SharedRealmHandle.cs
LaPeste: So I can report that with some "good timing" core ends up calling our managed callback when we have already freed the `tcs`. A typical scenario for this would be the following:
1. `BeginWriteAsync` is called and the await starts
2. The managed cb is queued by core for later execution
3. The `CancellationToken` is called right when core is about to dequeue
4. At this point the finally block of `BeginTransactionAsync` and the `tcs` handle is freed
5. Core calls the managed cb (`SharedRealmHandle.HandleTaskCompletion`) and `GCHandle.FromIntPtr(tcsPtr);` returns a null reference.

LaPeste: Then I'll move ownership of the `handle` to `HandleTaskCompletion` and add a bunch of comments around who owns what.

Tests/Realm.Tests/Database/AsyncTests.cs
```diff
@@ -426,5 +426,85 @@ public void SyncWrite_AsyncCommit_Mixed()
...
+                var syncTask = Task.Run(() => AsyncContext.Run(() =>
+                {
+                    using var realm = GetRealm(_realm.Config);
+                    var transaction = realm.BeginWrite();
+
+                    tcs.TrySetResult(null);
+
+                    cts.Cancel();
+                    transaction.Rollback();
+                }));
```
nirinchev, line 449: Can you try to see if this fixes the windows issue:

```suggestion
                var syncTask = Task.Run(() => AsyncContext.Run(async () =>
                {
                    using var realm = GetRealm(_realm.Config);
                    using var transaction = realm.BeginWrite();

                    tcs.TrySetResult(null);

                    cts.Cancel();
                    await Task.Delay(1000);
                }));
```

nirinchev (approved):

PR #2951, opened 2022-06-07
Update GHA workflow to use a single cluster

nirinchev:
## Description

This is a fairly massive change, but the end goal is to allow running many parallel tests against the same cluster. This is beneficial from a resource utilization perspective because our tests are fairly lightweight when it comes to cluster load, but spinning up multiple clusters has a non-negligible cost, even when using a shared tier. Below is a more detailed summary of changes.

* Moved all logic for setting up the test apps to `BaasClient`. Now the majority of the methods are private and all the app configuration logic is encapsulated inside `BaasClient`. This is motivated by a desire to avoid spreading app configuration across multiple files and by a desire to reuse `BaasClient` in multiple projects.
* Added a command-line tool that can run the `BaasClient` app import. This is intended to be used in jobs that run in parallel with the build and are expected to complete before the tests start. This shaves off 30-60 seconds of app importing that would otherwise happen in the test job.
* Restructured `BaasClient` to accept a differentiator as opposed to the setup cluster job accepting it. This differentiator is used to configure all the databases used by tests to have a unique name inside the cluster, allowing multiple apps to work with the cluster in parallel.

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


nirinchev:
Test failures are unrelated and sporadic.

Tools/DeployApps/DeployApps.csproj
```diff
@@ -0,0 +1,41 @@
+<Project Sdk="MSBuild.Sdk.Extras">
```
fealebenpae, line 1: The project file looks wrong - I'll try to fix it.

fealebenpae (approved):

LaPeste (approved):
I'm not super familiar with all the details about setting up the cluster and apps. However, by following the PR it generally made sense.
Overall this concentration of logic for the whole setup makes it easier to follow the subject when one is new to this part of the codebase.

PR #2956, opened 2022-06-10
Implement client reset with automatic recovery

### LINKED ISSUE #2745: Implement Client Reset w/ Recovery

LaPeste:
## Description

Fixes #2745 

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm/Sync/ErrorHandling/AutomaticRecoveryOrDiscardLocalHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// while the last one will be invoked in case an error occurs during the automated process and the system needs to fallback to a manual mode.
+    /// The overall recommendation for using this strategy is that using the three available callbacks should only be considered when:
+    /// 1. the user needs to be notified (in <see cref="OnBeforeReset"/>) of an incoming potential data loss of unsynced data as a result of a merge
+    ///    or a complete discard of local changes
+    /// 2. the user needs to be notified (in <see cref="OnAfterReset"/>) that the reset process has completed
+    /// 3. advanced use cases for data-sensitive applications where the developer wants to recover in the most appropriate way the unsynced data
+    /// 4. backup the whole realm before the client reset happens (in <see cref="OnBeforeReset"/>). Such backup could, for example, be used to restore the unsynced data (see 3.)
+    /// </remarks>
+    /// <seealso href="https://docs.mongodb.com/realm/sdk/dotnet/advanced-guides/client-reset/">Client Resets - .NET SDK</seealso>
+    public sealed class AutomaticRecoveryOrDiscardLocalHandler : ClientResetHandlerBase
```
nirinchev, line 41: Let's not have two of those (especially with such an obnoxious name). Instead we can add an optional ctor argument that controls the fallback:

```csharp
public AutomaticReccoveryHandler(Fallback fallback = Fallback.DiscardLocal)
```

Realm/Realm/Configurations/SyncConfigurationBase.cs
```diff
@@ -133,11 +134,18 @@ internal override async Task<SharedRealmHandle> CreateHandleAsync(Schema.RealmSc
...
         internal virtual Native.SyncConfiguration CreateNativeSyncConfiguration()
         {
-            var clientResyncMode = ClientResyncMode.DiscardLocal;
+            var clientResyncMode = ClientResyncMode.AutomaticRecoveryOrDiscardLocal;
 
-            if (ClientResetHandler != null && ClientResetHandler is ManualRecoveryHandler)
+            if (ClientResetHandler != null)
             {
-                clientResyncMode = ClientResyncMode.Manual;
+                clientResyncMode = ClientResetHandler switch
```
nirinchev, line 141: Instead of switching on type, can we add an abstract internal property to the ClientResetHandlerBase class and have inheritors be explicit about the resync mode they support.

nirinchev (changes requested):
Haven't reviewed the tests, but have some comments about the implementation.

Realm/Realm/Configurations/SyncConfigurationBase.cs
LaPeste: Beside allowing a param to the constructor of `AutomaticReccoveryHandler`, as you suggested [here](https://github.com/realm/realm-dotnet/pull/2956#discussion_r899977885). Is there any other advantage to this approach? E.g. do you see handlers ever changing internal behaviour after being released? Also reading a property is surely faster than switching on the type, but were you concerned with performance?

LaPeste: All I'm trying to say is, the type of the recovery strategy already implies what the recovery behaviour is gonna be. Why do we need to write it "again" in the class?

Realm/Realm/Sync/ErrorHandling/AutomaticRecoveryHandler.cs
```diff
@@ -0,0 +1,96 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync.ErrorHandling
+{
+    // TODO andrea: ask how this strategy may fail if it isn't for the server not allowing automatic merge
```
LaPeste, line 21: I'm waiting on James on this one.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -39,6 +42,13 @@ public class SessionTests : SyncTestBase
     {
         private readonly ConcurrentQueue<EventHandler<ErrorEventArgs>> _sessionErrorHandlers = new();
 
+        private static IEnumerable<IClientResetHandler> _allClientResetHandlers()
+        {
+            yield return new ClientResetHandler<DiscardLocalResetHandler>();
+            yield return new ClientResetHandler<AutomaticRecoveryHandler>(AutomaticRecoveryHandler.Fallback.DiscardLocal);
+            yield return new ClientResetHandler<AutomaticRecoveryHandler>(AutomaticRecoveryHandler.Fallback.None);
+        }
```
nirinchev, line 50: This seems a bit overengineered and will not work for Unity. When passing arguments to `TestCaseSource`, you need to have them be `object[]`. My suggestion would be to do something like:

```csharp
public static readonly object[] AllClientResetHandlers = new[]
{
    new object[] { typeof(DiscardLocalHandler), (AutomaticRecoveryHandler.Fallback?)null },
    new object[] { typeof(AutomaticRecoveryHandler), AutomaticRecoveryHandler.Fallback.DiscardLocal },
    new object[] { typeof(AutomaticRecoveryHandler), AutomaticRecoveryHandler.Fallback.None },
};

[TestCaseSource(nameof(AllClientResetHandlers))]
public void MyTest(Type handlerType, AutomaticRecoveryHandler.Fallback? fallback)
{

}

private static ClientResetHandlerBase GetClientResetHandler(Type type, 
    AutomaticRecoveryHandler.Fallback? fallback,
    BeforeResetCallback beforeCb = null,
    AfterResetCallback afterCb = null,
    ClientResetCallback manualCb = null)
{
    var handler = fallback.HasValue ? Activator.CreateInstance(type, new[] { fallback.Value }) : Activator.CreateInstance(type);

    if (beforeCb != null)
    {
        type.GetProperty(nameof(DiscardLocalHandler. OnBeforeReset)).SetValue(handler, beforeCb);
    }

    // rest of callbacks
}
```

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -270,31 +237,388 @@ public void Session_ClientReset_DiscardLocal_ManualResetFallback_InitiateClientR
...
+        public void SessionIntegrationTest_ClientResetHandlers_OutOfBoundArrayMove_Discarded()
+        {
+            SyncTestHelpers.RunBaasTestAsync(async () =>
+            {
+                var user = await GetUserAsync();
+                var partition = Guid.NewGuid().ToString();
+
+                // ===== clientA =====
+                var tscAfterClientResetMergeA = new TaskCompletionSource<object>();
+                var configA = await GetIntegrationConfigAsync(partition, null, "realm123.realm", user);
```
nirinchev, line 297: It's a bad practice to pass `null` as a positional argument as it's not clear what it means. Since that is an optional argument, it's a better idea to name the remaining arguments. Also, let's not specify the path of the realm explicitly.

```suggestion
                var configA = await GetIntegrationConfigAsync(partition, user: user);
```

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -270,31 +237,388 @@ public void Session_ClientReset_DiscardLocal_ManualResetFallback_InitiateClientR
...
+                realmA.Write(() =>
+                {
+                    originalObj.Strings.RemoveAt(2);
+                });
 
-                        onBeforeTriggered = true;
-                        tcs.TrySetResult(null);
-                    })
+                // ===== clientB =====
+                var configB = await GetIntegrationConfigAsync(partition, null, "realm456.realm", user);
```
nirinchev, line 335:
```suggestion
                var configB = await GetIntegrationConfigAsync(partition, optionalPath: configA.DatabasePath + "_B", user: user);
```

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -270,31 +237,388 @@ public void Session_ClientReset_DiscardLocal_ManualResetFallback_InitiateClientR
...
+                var afterCbB = GetOnAfterHandler(tcsAfterClientResetB, (before, after) =>
+                {
+                    var listB = after.All<SyncObjectWithRequiredStringList>().Single().Strings.ToList();
+                    var list = after.All<SyncObjectWithRequiredStringList>().Single().Strings;
+                    Assert.That(list.Count, Is.EqualTo(3));
+                    Assert.That(list[0], Is.EqualTo("2"));
+                    Assert.That(list[1], Is.EqualTo("0"));
+                    Assert.That(list[2], Is.EqualTo("1"));
+                });
+                configB.ClientResetHandler = new AutomaticRecoveryHandler(AutomaticRecoveryHandler.Fallback.DiscardLocal)
```
nirinchev, line 347:
```suggestion
                configB.ClientResetHandler = new AutomaticRecoveryHandler
```

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -270,31 +237,388 @@ public void Session_ClientReset_DiscardLocal_ManualResetFallback_InitiateClientR
...
+                var realmB = await GetRealmAsync(configB);
+
+                await WaitForDownloadAsync(realmB);
+
+                Assert.That(realmB.All<SyncObjectWithRequiredStringList>().Count, Is.EqualTo(1));
+                var originalObjStr = realmB.All<SyncObjectWithRequiredStringList>().Single().Strings;
+                Assert.That(originalObjStr.Count, Is.EqualTo(3));
+                Assert.That(originalObjStr.First(), Is.EqualTo("0"));
+                Assert.That(originalObjStr.ElementAt(1), Is.EqualTo("1"));
+                Assert.That(originalObjStr.ElementAt(2), Is.EqualTo("2"));
```
nirinchev, line 360:
```suggestion
                Assert.That(originalObjStr, Is.EqualTo(new[] { "0", "1", "2"}));
```

nirinchev (changes requested):

Tests/Tests.iOS/Tests.iOS.csproj
```diff
@@ -30,6 +30,8 @@
     <MtouchLink>SdkOnly</MtouchLink>
     <MtouchDebug>true</MtouchDebug>
     <MtouchExtraArgs>--nosymbolstrip --linkskip=nunit.framework --linkskip=Realm.Tests --nolinkaway --linkskip=Newtonsoft.Json</MtouchExtraArgs>
+    <CodesignProvision></CodesignProvision>
+    <CodesignKey>iPhone Developer</CodesignKey>
```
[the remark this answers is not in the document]
LaPeste: I don't really know how I didn't see this. Thanks.

Tests/Realm.Tests/Sync/SessionTests.cs
LaPeste: Is there a specific reason for `AllClientResetHandlers` to be public?

Tests/Tests.iOS/Tests.iOS.csproj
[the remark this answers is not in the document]
nirinchev: VS will sometimes add these automatically and it's kind of annoying.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -270,31 +237,388 @@ public void Session_ClientReset_DiscardLocal_ManualResetFallback_InitiateClientR
...
+                    if (eventArgs.PropertyName == nameof(SyncObjectWithRequiredStringList.Strings))
+                    {
+                        var list = realmA.All<SyncObjectWithRequiredStringList>().Single().Strings;
+                        Assert.That(list.Count, Is.EqualTo(3));
+                        Assert.That(list[0], Is.EqualTo("0"));
+                        Assert.That(list[1], Is.EqualTo("1"));
+                        Assert.That(list[2], Is.EqualTo("3"));
+                        tscAfterRemoteUpdateA.TrySetResult(null);
+                    }
+                };
```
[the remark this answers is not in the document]
LaPeste: Then, I think we have a problem because all works fine.
You're right, I should have subscribed to `SubscribeForNotifications`, not to `PropertyChanged`.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -281,6 +286,7 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
                     var clientResetEx = new ClientResetException(session.User.App, messageString, errorCode, userInfo);
 
                     if (syncConfig.ClientResetHandler is DiscardLocalResetHandler ||
+                        syncConfig.ClientResetHandler is AutomaticRecoveryHandler ||
                         syncConfig.ClientResetHandler?.ManualClientReset != null)
```
nirinchev, line 290: [nit] I know that's code you didn't change, but we can remove the null propagation here
```suggestion
                        syncConfig.ClientResetHandler.ManualClientReset != null)
```

Realm/Realm/Sync/ErrorHandling/AutomaticRecoveryHandler.cs
```diff
@@ -0,0 +1,96 @@
...
+            /// </summary>
+            Manual = 0,
+
+            /// <summary>
+            /// If the automatic merge fails, automatically discards all local changes and uses the latest realm that is available on the remote.
+            /// </summary>
+            DiscardLocal = 1,
+        }
+
+        internal Fallback FallbackStrategy { get; private set; }
```
nirinchev, line 61: We can probably expose this publicly. Not super useful, but doesn't hurt either.

Tests/Tests.iOS/PreserveFromStripping.xml
```diff
@@ -0,0 +1,6 @@
+<linker>
```
nirinchev, line 1: We should not have this. If the managed linker strips those, we should annotate them with `[Preserve(AllMembers = true)]`.

Tools/DeployApps/BaasClient.cs
```diff
@@ -41,8 +41,19 @@ public static class AppConfigType
...
     }
 
-    public class BaasClient : IDisposable
+    public class BaasClient
     {
+        public class FunctionReturn
+        {
+            public enum Result
+            {
+                success = 0,
```
nirinchev, line 50: These should be documented or the warning should be suppressed.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -240,23 +210,22 @@ public void Session_ClientReset_ManualRecovery_InitiateClientReset()
...
-                    ManualResetFallback = (err) =>
-                    {
-                        manualResetFallbackHandled = true;
-                        errorTcs.TrySetResult(err);
-                    }
+                    manualResetFallbackHandled = true;
+                    errorTcs.TrySetResult(err);
                 };
 
+                config.ClientResetHandler = GetClientResetHandler(handlerType, fallback, beforeCb: null, afterCb: null, manualCb);
```
nirinchev, line 227: `beforeCb` and `afterCb` are optional, so you can omit them.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -876,23 +1058,23 @@ public void Session_Error_OnSessionError_Coexistence()
             });
         }
 
-        [TestCase(ProgressMode.ForCurrentlyOutstandingWork)]
-        [TestCase(ProgressMode.ReportIndefinitely)]
-        public void Session_ProgressObservable_IntegrationTests(ProgressMode mode)
+        [TestCaseSource(nameof(DoubleModeClientResetHandlers))]
+        public void SessionIntegrationTest_ProgressObservable((ProgressMode Mode, ClientResetHandlerBase Handler) tuple)
```
nirinchev, line 1062: Why change this test? Seems unrelated to the client reset work.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -1330,6 +1524,40 @@ protected override void CustomTearDown()
 
         }
 
+        public interface IClientResetHandler
+        {
+            ClientResetHandlerBase BuildHandler(BeforeResetCallback beforeCb = null,
+                                                AfterResetCallback afterCb = null,
+                                                ClientResetCallback manualCb = null);
+        }
```
nirinchev, line 1532: Should this be deleted

nirinchev (commented):
A couple of final touches from me, but feels like it's almost ready to go.

Tools/DeployApps/BaasClient.cs
LaPeste: I missed it. I'm gonna suppress it. It's only for our internal functions results.

Tests/Realm.Tests/Sync/SessionTests.cs
LaPeste: Just to keep it consistent with the other naming convention I picked for the other integration tests for the session. Is this wrong?

Tests/Tests.iOS/PreserveFromStripping.xml
LaPeste: @fealebenpae suggested that it's better to avoid adding that attribute when in the main assembly unless strictly necessary. In our case I get the setter of all callbacks for `AutomaticRecoveryHandler` and `DiscardLocalResetHandler` stripped. I just supposed that users should not run in the same issue. I just spotted this stripping problem because I used reflections in tests, it would otherwise have gone unnoticed.
So I'm not sure what the best course of action is here. When dealing with user facing API, should we preserve more often then not or keep it at a minimum?

nirinchev: Hm... good point. I didn't realize it's due to reflection that those were stripped. In that case, let's preserve them by using them in the tests rather than with a xml file. Those can easily go out of sync with the code they're preserving and are a bit annoying to maintain (e.g. you need to also add this for MAUI and Unity). I think we can preserve them like:

```csharp
[Preserve]
static SessionTests()
{
    var preserveHandler = new AutomaticRecoveryHandler
    {
        OnBeforeReset = (...) => { },
	...
    }
}
```

Tools/DeployApps/BaasClient.cs
nirinchev: You can see the warning when you browse the code:

Screen Shot 2022-07-01 at 18 23 51

To see that in VS, you'll probably need to rebuild the project, otherwise msbuild will not show warnings for code it doesn't compile.

Tests/Realm.Tests/Sync/SessionTests.cs
nirinchev: But it's changing the arguments of the test for no obvious reason. I don't see why we'd want to test different types of client reset handlers in this progress test.

LaPeste: I understand what you mean now. You're right, client resets have nothing to do here. I went over it too fast. I'll revert it to what it was before.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -280,8 +285,10 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
                     var userInfo = StringStringPair.UnmarshalDictionary(userInfoPairs, userInfoPairsLength.ToInt32());
                     var clientResetEx = new ClientResetException(session.User.App, messageString, errorCode, userInfo);
 
-                    if (syncConfig.ClientResetHandler is DiscardLocalResetHandler ||
-                        syncConfig.ClientResetHandler?.ManualClientReset != null)
+                    if (syncConfig.ClientResetHandler is DiscardUnsyncedChangesHandler ||
+                        syncConfig.ClientResetHandler is RecoverUnsyncedChangesHandler ||
+                        syncConfig.ClientResetHandler is RecoverOrDiscardUnsyncedChangesHandler ||
+                        syncConfig.ClientResetHandler.ManualClientReset != null)
```
nirinchev, line 291: Rather than check the types here, can we check the resync mode?

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -323,48 +330,67 @@
...
                     var schema = syncConfig.GetSchema();
                     using var realmBefore = new Realm(new UnownedRealmHandle(beforeFrozen), syncConfig, schema);
-                    discardLocalHandler.OnBeforeReset(realmBefore);
+                    cb.Invoke(realmBefore);
                 }
             }
             catch (Exception ex)
             {
-                Logger.Default.Log(LogLevel.Error, $"An error has occurred while executing DiscardLocalResetHandler.OnBeforeReset during a client reset: {ex}");
+                Logger.Default.Log(LogLevel.Error, $"An error has occurred while executing {syncConfig.ClientResetHandler.GetType().Name}.{cb.GetType().Name} during a client reset: {ex}");
```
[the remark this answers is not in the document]
nirinchev: This warning is correct - we should check for null on the sync config and the callback.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -323,48 +330,67 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
```
nirinchev, line 357: Let's call this `OnBeforeReset` - there's no need to check the type of the callback.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -323,48 +330,67 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
                     using var realmBefore = new Realm(new UnownedRealmHandle(beforeFrozen), syncConfig, schema);
                     using var realmAfter = new Realm(new UnownedRealmHandle(after), syncConfig, schema);
-                    discardLocalHandler.OnAfterReset(realmBefore, realmAfter);
+                    cb.Invoke(realmBefore, realmAfter);
                 }
             }
             catch (Exception ex)
             {
-                Logger.Default.Log(LogLevel.Error, $"An error has occurred while executing DiscardLocalResetHandler.OnAfterReset during a client reset: {ex}");
+                Logger.Default.Log(LogLevel.Error, $"An error has occurred while executing {syncConfig.ClientResetHandler.GetType().Name}.{cb.GetType().Name} during a client reset: {ex}");
```
nirinchev, line 393: `cb.GetType().Name` is unlikely to give us anything meaningful. We should just use `OnAfterReset`.

Realm/Realm/Configurations/FlexibleSyncConfiguration.cs
```diff
@@ -182,20 +182,22 @@ internal override Native.SyncConfiguration CreateNativeSyncConfiguration()
...
         /// <seealso href="https://docs.mongodb.com/realm/sdk/dotnet/advanced-guides/client-reset/">Client reset docs</seealso>
         public override ClientResetHandlerBase ClientResetHandler
         {
             get => base.ClientResetHandler;
             set
             {
-                if (value is DiscardLocalResetHandler)
+                if (value is DiscardUnsyncedChangesHandler ||
+                    value is RecoverUnsyncedChangesHandler ||
+                    value is RecoverOrDiscardUnsyncedChangesHandler)
```
nirinchev, line 198: It's kind of a moot point considering we're going to support all reset handler types soon, but this is not a great way to check. Apart from checking by type, it's also missing a check for `DiscardLocalResetHandler`.

Realm/Realm/Sync/ErrorHandling/DiscardUnsyncedChangesHandler.cs
```diff
@@ -33,37 +35,52 @@ namespace Realms.Sync.ErrorHandling
...
+        /// </summary>
+        /// <value>Callback invoked if automatic Client Reset handling fails.</value>
+        public ClientResetCallback ManualResetFallback
+        {
+            get => ManualClientReset;
+            set => ManualClientReset = value;
+        }
+    }
 
+#pragma warning disable SA1402 // This type is now obsolete and will be deleted in the next major version
```
nirinchev, line 65:
```suggestion
#pragma warning disable SA1402 // This type is now obsolete and will be deleted in the next major version

```

Realm/Realm/Sync/ErrorHandling/DiscardUnsyncedChangesHandler.cs
```diff
@@ -86,4 +105,5 @@ public ClientResetCallback ManualResetFallback
             set => ManualClientReset = value;
         }
     }
+#pragma warning restore SA1402 // This type is now obsolete and will be deleted in the next major version
```
nirinchev, line 108:
```suggestion

#pragma warning restore SA1402 // This type is now obsolete and will be deleted in the next major version

```

Realm/Realm/Sync/ErrorHandling/RecoverOrDiscardUnsyncedChangesHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <summary>
+        /// Gets or sets the callback that indicates a Client Reset is about to happen.
+        /// </summary>
+        /// <value>Callback invoked right before a Client Reset.</value>
+        public BeforeResetCallback OnBeforeReset { get; set; }
+
+        /// <summary>
+        /// Gets or sets the callback that indicates that an automatic Client Reset just happened.
+        /// </summary>
+        /// <value>Callback invoked right after a Client Reset.</value>
```
nirinchev, line 63: In the remarks we should indicate that only one of `OnAfterRecovery` or `OnAfterDiscard` but not both.

Tests/Realm.Tests/Sync/SyncConfigurationTests.cs
```diff
@@ -109,10 +109,12 @@ public void Test_SyncConfigRelease()
         }
 
         [Test]
-        public void FlexibleSyncConfiguration_Throws_When_Assigned_DiscardLocalResetHandler()
+        public void FlexibleSyncConfiguration_WhenAssignedNewRecoveryHandlers_Throws()
         {
             var conf = GetFakeFLXConfig();
-            Assert.That(() => conf.ClientResetHandler = new DiscardLocalResetHandler(), Throws.TypeOf<NotSupportedException>());
```
nirinchev, line 115: This should still test the obsolete case.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -323,48 +330,67 @@ private static void HandleSessionError(IntPtr sessionHandlePtr, ErrorCode errorC
...
+            SyncConfigurationBase syncConfig = null;
+
             try
             {
                 var syncConfigHandle = GCHandle.FromIntPtr(managedSyncConfigurationHandle);
-                var syncConfig = (SyncConfigurationBase)syncConfigHandle.Target;
+                syncConfig = (SyncConfigurationBase)syncConfigHandle.Target;
+                cb = syncConfig.ClientResetHandler switch
+                {
+                    DiscardUnsyncedChangesHandler handler => handler.OnBeforeReset,
```
nirinchev, line 342: What if it's `DiscardLocalResetHandler`? Even though the type is obsolete, we should not break users who were using it.

nirinchev (changes requested):
Looks mostly fine. I've commented on several places, but be sure to audit all usages where you're explicitly checking the type of a reset handler (like `xxx is SomeResetHandler`) to make sure it also checks for `DiscardLocalResetHandler`, otherwise we're silently breaking users.

Realm/Realm/Handles/SessionHandle.cs
LaPeste: All this was done to give you e.g. `DiscardUnsyncedChangesHandler.OnBeforeReset`, which I checked and it works as expected.

nirinchev: I'm not sure how it's possible to give you `DiscardUnsyncedChangesHandler.OnBeforeReset`. I haven't tried it, but I expect `cb` here to be of type `BeforeResetCallback` or `Action<Realm>`. There's no obvious reason/way for it to ever be of type `OnBeforeReset`, considering that's the name of a property and not an actual type.

LaPeste: What I said makes no sense. Without reflection this is impossible possible. And yet my memory told me "yes yes it's fine". Eh...🤦🏻

nirinchev (approved):
Looks good, just make sure to suppress the build warnings for the obsolete reset handler

PR #2958, opened 2022-06-13
[SG] Add support for Source Generator

papafe:


papafe:
@nirinchev @LaPeste @fealebenpae 
I would like your opinion on `ManuallyGeneratedClass.cs`. It shows how the source generation would look like for an "average" class. It would be good to get some feedback on that before getting deeper into the source generation itself. 
I've created this PR so it's easier to give comments.

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
```diff
@@ -0,0 +1,270 @@
...
+
+        public bool IsManaged => _accessor.IsManaged;
+
+        public bool IsValid => _accessor.IsValid;
+
+        public bool IsFrozen => _accessor.IsFrozen;
+
+        public Realm Realm => _accessor.Realm;
+
+        public ObjectSchema ObjectSchema => _accessor.ObjectSchema;
```
nirinchev, line 101: I wonder if exposing this on the user class is particularly useful. We may want to remove it from the interface and instruct people to use the accessor if they need it.

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
```diff
@@ -0,0 +1,270 @@
...
+
+                if (IsManaged &&
+                    _propertyChanged == null)
+                {
+                    UnsubscribeFromNotifications();
+                }
+            }
+        }
+
+        protected virtual void OnPropertyChanged(string propertyName)
```
nirinchev, line 164: Should this be partial instead? We don't expect people to be inheriting from this class, so they can't really override it.

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
```diff
@@ -0,0 +1,270 @@
...
+        {
+        }
+
+        protected void RaisePropertyChanged([CallerMemberName] string propertyName = null)
+        {
+            _propertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+            OnPropertyChanged(propertyName);
+        }
+
+        protected internal virtual void OnManaged()
```
nirinchev, line 174: This should probably not be virtual either.

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
```diff
@@ -0,0 +1,270 @@
...
+                    UnsubscribeFromNotifications();
+                }
+            }
+        }
+
+        protected virtual void OnPropertyChanged(string propertyName)
+        {
+        }
+
+        protected void RaisePropertyChanged([CallerMemberName] string propertyName = null)
```
nirinchev, line 168: Why is this protected? Should it be private?

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
```diff
@@ -0,0 +1,270 @@
...
+                SubscribeForNotifications();
+            }
+
+            OnManaged();
+        }
+        #endregion
+
+        private event PropertyChangedEventHandler _propertyChanged;
+
+        public event PropertyChangedEventHandler PropertyChanged
```
nirinchev, line 140: I'm not sure I see how unmanaged object notifications are wired up.

nirinchev: Actually for managed as well - should we be raising propertychanged every time a property is set?

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
```diff
@@ -0,0 +1,270 @@
...
+        public ManagedAccessor CreateAccessor() => new ManuallyGeneratedClassManagedAccessor();
+
+        public IRealmObjectBase CreateInstance()
+        {
+            return new ManualllyGeneratedClass();
+        }
+
+        public bool TryGetPrimaryKeyValue(IRealmObjectBase instance, out object value)
+        {
+            value = (instance.Accessor as IManuallyGeneratedClassAccessor).PrimaryKeyValue;
```
nirinchev, line 209: I don't like using `as` casts like that. Can we do a regular one so that we throw properly?

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
```diff
@@ -0,0 +1,270 @@
...
+
+    internal interface IManuallyGeneratedClassAccessor : IRealmAccessor
+    {
+        int IntValue { get; set; }
+
+        IList<int> ListValue { get; }
+
+        public string StringValue { get; set; }
+
+        public int PrimaryKeyValue { get; set; }
```
nirinchev, line 222: Why are these public?

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
```diff
@@ -0,0 +1,270 @@
...
+         */
+        public int IntValue
+        {
+            get => (int)GetValue("IntValue");
+            set => SetValue("IntValue", value);
+        }
+
+        public IList<int> ListValue
+        {
+            get => GetListValue<int>("ListValue");
```
nirinchev, line 239: Previously, we would only create the collection once rather than for every access. I suggest we employ a similar strategy here as well. Something like:

```csharp
private Lazy<IList<int>> _listValue = new(() => GetListValue<int>("ListValue"));
public IList<int> ListValue => _listValue.Value;
```

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
```diff
@@ -0,0 +1,270 @@
...
+        {
+            _accessor.UnsubscribeFromNotifications();
+        }
+    }
+}
+
+// Having a separate namespace allows to hide the implementation details better.
+namespace Realm.Generated
+{
+    internal class ManuallyGeneratedClassObjectHelper : IRealmObjectHelper
```
nirinchev, line 193: Can we add `[EditorBrowsable(Never)]` to these generated accessors?

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
papafe: I really thought notifications for unmanaged objects didn't work... my bad. I'm adding more tests to the manual generation test fixture. 

For managed that's done in the same way as it was done before. After `SubscribeToNotifications` is called, then `RaisePropertyChanged` it's called from native every time it's needed.

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
nirinchev: Both are using a backing field :P Whether the field is lazy or not, doesn't matter all that much, so whatever you prefer. The reason we previously used a backing field instead of a lazy was that we already had the backing field for the auto property on the object and constructing a lazy in IL is... fun 😫

LaPeste (commented):
`ManuallyGeneratedClass.cs` seems fine to me. I just added a super minor style comment.

Realm/Realm.SourceGenerator/Realm.SourceGenerator.csproj
```diff
@@ -0,0 +1,43 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+	<PropertyGroup>
```
nirinchev, line 2: Spaces 😛

Realm/Realm.SourceGenerator/Realm.SourceGenerator.csproj
```diff
@@ -0,0 +1,43 @@
...
+		<IsRoslynComponent>true</IsRoslynComponent>
+	</PropertyGroup>
+	<ItemGroup>
+		<!-- Unity does not support the latest version of CodeAnalysis.CSharp. According to the docs it should support version 3.8.0, 
+		but it works also with 3.9.0. We need to verify manually if it works also with some later versions. -->
+		<PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.9.0" PrivateAssets="all" />
+		<PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.3">
+			<PrivateAssets>all</PrivateAssets>
+			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+		</PackageReference>
```
nirinchev, line 21: Can we add `Microsoft.CodeAnalysis.FxCopAnalyzers` and `StyleCop.Analyzers` so that we get warnings about style violations?

nirinchev (commented):
I went through 1/3 of the files and left some preliminary comments. Will continue next week.

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -205,11 +215,20 @@ Analytics payload
...
                 {
-                    return WeaveType(type);
+                    if (isGenerated)
+                    {
+                        return WeaveGeneratedType(type);
+                    }
+                    else
+                    {
+                        return WeaveType(type);
+                    }
```
nirinchev, line 231:
```suggestion
                    return isGenerated ? WeaveGeneratedType(type) : WeaveType(type);
```

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -234,6 +253,118 @@ Analytics payload
...
+                        sequencePoint);
+
+                    return WeaveTypeResult.Error(type.Name, isGenerated: true);
+                }
+            }
+
+            return WeaveTypeResult.Success(type.Name, persistedProperties, isGenerated: true);
+        }
+
+        private WeavePropertyResult WeaveGeneratedClassProperty(PropertyDefinition prop, TypeDefinition interfaceType)
```
nirinchev, line 289: I believe this can also remove the compiler generated backing fields?

Realm/Realm/Attributes/GeneratedAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+// //
+// ////////////////////////////////////////////////////////////////////////////
+
+using System;
+using System.ComponentModel;
+
+namespace Realms
+{
+
+    //TODO Add documentation
```
nirinchev, line 25: I don't think we should document this - instead, let's hide it from the docs, as users are not supposed to interact with it.

nirinchev (commented):
Some more comments. I'll put this on pause so that you can clean it up and address existing comments.

Realm/Realm/DatabaseTypes/RealmValue.cs
```diff
@@ -144,7 +144,7 @@ private RealmValue(IRealmObjectBase obj) : this()
 
         private static RealmValue String(string value) => new RealmValue(value);
 
-        internal static RealmValue Object(IRealmObjectBase value) => new RealmValue(value);
+        public static RealmValue Object(IRealmObjectBase value) => new RealmValue(value);
```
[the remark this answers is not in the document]
papafe: This is necessary to have the implicit conversion operator like:
```public static implicit operator RealmValue(AllTypesClass val) => RealmValue.Object(val);```

File: Realm/Realm/Schema/Property.cs [hunk 16]
```diff
@@ -18,6 +18,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Data.SqlTypes;
```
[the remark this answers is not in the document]
papafe: Really not sure where it came from.... I'll delete it

Tests/Realm.Tests.SourceGeneratorPlayground/Realm.Tests.SourceGeneratorPlayground.csproj
```diff
@@ -0,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
papafe: Agreed. Before it was just to try the SG, now it also contains the classes that are used for testing. I'll give it another name

Tests/Weaver/Realm.FakeForWeaverTests/Realm.FakeForWeaverTests.csproj
```diff
@@ -1,38 +1,39 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <AssemblyName>Realm</AssemblyName>
-  </PropertyGroup>
+	<PropertyGroup>
```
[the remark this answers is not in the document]
papafe: Who is Miguel??? XD

I realised that after I had to repair Visual Studio my settings for tabs/spaces got reset to default... That's why it's everywhere 😢

Realm/Realm.SourceGenerator/Diagnostics.cs
```diff
@@ -0,0 +1,447 @@
...
+                $"{className} derives from another class and this is not yet supported",
+                "RealmClassGeneration",
+                DiagnosticSeverity.Error,
+                true
+                );
+
+            return Diagnostic.Create(descriptor, location);
+        }
+
+        public static Diagnostic ObjectWithNoProperties(string className, Location location)
```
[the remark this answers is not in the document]
papafe: At the moment this is an error in the weaver. Also, is there any use to have a table without columns?

Realm/Realm.SourceGenerator/Diagnostics.cs
```diff
@@ -0,0 +1,447 @@
...
+                DiagnosticSeverity.Error,
+                true);
+
+            return Diagnostic.Create(descriptor, location);
+        }
+
+        public static Diagnostic EmbeddedObjectWithPrimaryKey(string className, string propertyName, Location location)
+        {
+            DiagnosticDescriptor descriptor = new
+                ("REALM001",
```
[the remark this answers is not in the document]
papafe: We need to change all of them. We need to decide if we want to have them grouped in some way or we just give them sequential values as we add them. 
This was just a test to see if the values get correctly shown.

Realm/Realm.SourceGenerator/Diagnostics.cs
```diff
@@ -0,0 +1,447 @@
...
+                true);
+            return Diagnostic.Create(descriptor, location);
+        }
+
+        public static Diagnostic RequiredWrongType(string className, string propertyName, string propertyType, Location location)
+        {
+            DiagnosticDescriptor descriptor = new
+                ("REALM001",
+                "[Required] is only allowed on specific types",
+                $"{className}.{propertyName} is marked as [Required] which is only allowed on strings or nullable scalar types, not on {propertyType}.",
```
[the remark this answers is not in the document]
papafe: Agreed, I'll correct it also in the weaver

Realm/Realm.Weaver/RealmWeaver.cs
papafe: Most probably yes, I remember we talked about it. I'll add another ticket for this so I don't forget

Tests/Weaver/Realm.FakeForWeaverTests/Realm.FakeForWeaverTests.csproj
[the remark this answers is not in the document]
nirinchev: Miguel de Icaza - he's been vocally on the tabs side of the tabs vs spaces debate

Realm/Realm.SourceGenerator/Diagnostics.cs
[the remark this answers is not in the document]
nirinchev: I don't think it's super useful, but I can come up with artificial use cases - e.g. a table where the user dynamically edits the schema. But even then, you can probably come up with 1-2 properties that are static.

Realm/Realm.SourceGenerator/Generator.cs
```diff
@@ -0,0 +1,722 @@
...
+                {{
+{skipDefaultsContent}
+                }}";
+
+            }
+
+            var isEmbedded = _classInfo.IsEmbedded ? "true" : "false";
+            var schema = @$"        public static ObjectSchema RealmSchema = new ObjectSchema.Builder(""{_classInfo.Name}"", isEmbedded: {isEmbedded})
+        {{
+{schemaProperties}
```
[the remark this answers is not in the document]
papafe: There are several places where this happens at the moment. The "NormalizeWhiteSpace" method takes care of this if we decide to use it.
I would say to think about it together with the indent method though

Realm/Realm.SourceGenerator/Parser.cs
```diff
@@ -0,0 +1,387 @@
...
+                    return PropertyTypeInfo.Unsupported;
+                }
+
+                propertyType.InternalType = GetSingleLevelPropertyTypeInfo(argument);
+            }
+            else if (propertyType.IsIQueryable)
+            {
+                // TODO: these checks are a bit too restrictive as they operate on IQueryable that may not have
+                // been annotated with [Backlink] - e.g. IQueryable<string> Foo { get; set; }
+                // We should make sure we only return unsupported if the property is actually a backlink.
```
nirinchev, line 250: Ping @papafe

papafe: just fixed

nirinchev (commented):
Some minor issues as I'm going making my way through this

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -0,0 +1,305 @@
...
+            SimpleTypeEnum.Guid,
+        };
+
+        private static HashSet<SimpleTypeEnum?> _requiredTypes = new()
+        {
+            SimpleTypeEnum.String,
+            SimpleTypeEnum.Data,
+        };
+
+        public virtual SimpleTypeEnum? SimpleType { get; set; }
```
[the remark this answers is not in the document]
papafe: Yes, the naming can be better. It's trying to convey that it's a scalar value. Maybe we could call it ScalarType (also the enum) ?
It can be null because it could be a collection for example

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -0,0 +1,305 @@
...
+        public PropertyTypeInfo TypeInfo { get; set; }
+
+        public Accessibility Accessibility { get; set; }
+
+        public string Initializer { get; set; }
+    }
+
+    internal abstract record PropertyTypeInfo
+    {
+        private static HashSet<SimpleTypeEnum?> _indexableTypes = new()
```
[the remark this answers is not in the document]
papafe: If we keep enum properties like "SimpleType" as nullables, then this need to be a nullable too, otherwise can't use "Contains" directly

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -0,0 +1,305 @@
...
+
+        private static HashSet<SimpleTypeEnum?> _requiredTypes = new()
+        {
+            SimpleTypeEnum.String,
+            SimpleTypeEnum.Data,
+        };
+
+        public virtual SimpleTypeEnum? SimpleType { get; set; }
+
+        public virtual CollectionTypeEnum? CollectionType { get; set; }
```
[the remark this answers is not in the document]
papafe: Yes it doesn't need a setter. Regarding the `None`, I don't have very strong opinions on both ways to be honest, I suppose they are more or less the same.

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -0,0 +1,305 @@
...
+
+        public virtual bool IsRealmInteger { get; set; }
+
+        public virtual bool IsIQueryable { get; set; }
+
+        public bool IsNullable => NullableAnnotation == NullableAnnotation.None || NullableAnnotation == NullableAnnotation.Annotated;
+
+        // NullabilityAnnotation != None for all value types and for ref types with nullability annotations enabled
+        public virtual NullableAnnotation NullableAnnotation { get; set; } = NullableAnnotation.None;
```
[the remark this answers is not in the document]
papafe: Got carried away with virtual properties 😁

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -0,0 +1,305 @@
...
+
+        public static PropertyTypeInfo RealmInteger => new RealmIntegerTypeInfo();
+
+        public static PropertyTypeInfo IQueryable => new IQueryableTypeInfo();
+
+        public bool IsSupportedIndexType()
+        {
+            if (IsRealmInteger)
+            {
+                if (HasNullabilityAnnotation)
```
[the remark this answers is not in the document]
papafe: Yes, that is the case. 
We can't use `IsNullable` because of nullability annotation. For example if we have a string:
- Without nullability annotations enabled, `string` will have ` NullableAnnotation == NullableAnnotation.None`
- With nullability annotations enabled, `string` will have ` NullableAnnotation == NullableAnnotation.NotAnnotated`

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -0,0 +1,305 @@
...
+
+        public bool HasCorrectNullabilityAnnotation()
+        {
+            if (NullableAnnotation == NullableAnnotation.Annotated &&
+                (IsCollection || IsIQueryable || SimpleType == SimpleTypeEnum.RealmValue))
+            {
+                return false;
+            }
+
+            if (NullableAnnotation == NullableAnnotation.NotAnnotated &&
```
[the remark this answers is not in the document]
papafe: Then it's fine. `None` is applicable only for reference values without nullability annotations enabled (that we support, and means that the value is nullable by default). 
Value types always have either a specific annotation of `Annotated` or `NotAnnotated`. Reference values have a specific type only if nullable annotations are enabled.

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -0,0 +1,305 @@
...
+
+    internal record SimpleTypeInfo : PropertyTypeInfo
+    {
+        public SimpleTypeInfo(SimpleTypeEnum type)
+        {
+            SimpleType = type;
+        }
+    }
+
+    internal enum CollectionTypeEnum
```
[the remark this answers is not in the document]
papafe: Agreed. It started to create less confusion but then I never changed it

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        public bool IsSupportedPrimaryKeyType()
+        {
+            return _primaryKeyTypes.Contains(SimpleType);
+        }
+
+        public bool IsSupportedRequiredType()
+        {
+            if (IsListOrSet)
```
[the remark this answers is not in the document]
papafe: An error

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -0,0 +1,305 @@
...
+            SimpleTypeEnum.Data,
+        };
+
+        public virtual SimpleTypeEnum? SimpleType { get; set; }
+
+        public virtual CollectionTypeEnum? CollectionType { get; set; }
+
+        public virtual bool IsRealmInteger { get; set; }
+
+        public virtual bool IsIQueryable { get; set; }
```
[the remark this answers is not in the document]
papafe: I can understand the feeling, but I think most of the time IQueryable needs to be treated quite differently than the other two, so this could make the code even more difficult to read.

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -0,0 +1,326 @@
...
+
+namespace Realms.SourceGenerator
+{
+    internal record ClassInfo
+    {
+        public string Name { get; set; }
+
+        public ObjectType ObjectType { get; set; }
+
+        public bool IsEmbedded => ObjectType == ObjectType.EmbeddedObject;
```
nirinchev, line 31: Let's get rid of `IsEmbedded` and instead use the `ObjectType` directly.

Realm/Realm.SourceGenerator/RealmGenerator.cs
```diff
@@ -0,0 +1,60 @@
...
+        /* Not explicitly supported:
+         * - Inheritance of any kind (classes cannot derive from anything)
+         * - Full nullability support
+         * - Nested classes
+         */
+
+        /* Notes for developers:
+         * - "OnManaged" is a partial method now
+         * - "OnPropertyChanged" is a partial method now
+         *
```
nirinchev, line 35:
```suggestion
```

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -234,6 +247,118 @@ Analytics payload
...
+
+            if (prop.SetMethod != null)
+            {
+                ReplaceGeneratedClassSetter(prop, interfaceType, accessorGetter);
+            }
+
+            return WeavePropertyResult.Success(prop);
+        }
+
+        private void ReplaceGeneratedClassGetter(PropertyDefinition prop, TypeDefinition interfaceType, MethodReference accessorGetter)
```
nirinchev, line 297:
```suggestion
        private static void ReplaceGeneratedClassGetter(PropertyDefinition prop, TypeDefinition interfaceType, MethodReference accessorGetter)
```

Realm/Realm/DataBinding/RealmObjectTypeDelegator.cs
```diff
@@ -27,15 +29,23 @@ internal class RealmObjectTypeDelegator : TypeDelegator
         // Holds property name -> PropertyInfo map to avoid creating a new WovenPropertyInfo for each GetDeclaredProperty call.
         private readonly ConcurrentDictionary<string, PropertyInfo> _propertyCache = new ConcurrentDictionary<string, PropertyInfo>();
 
-        internal RealmObjectTypeDelegator(Type type) : base(type)
+        private ObjectSchema _schema;
```
nirinchev, line 32:
```suggestion
        private readonly ObjectSchema _schema;
```

nirinchev (changes requested):
I went through a portion of this and have mostly stylistic comments. A few general comments which didn't have a concrete line number:
1. All the changes to `Realm-Windows.sln` need to also be applied to `Realm.sln`.
2. There seem to still be some places where we can end up with awkward spacing - while not critical, it does trigger my OCD and I'd like to do our best to generate code that is nice to look at.

Realm/Realm.Weaver/RealmWeaver.cs
papafe: Already done in the other PR (I will merge back the changes in Realm and Realm.SourceGenerator afterI go through the comments)

Realm/Realm.SourceGenerator/ClassCodeBuilder.cs
```diff
@@ -0,0 +1,811 @@
...
+                            var internalTypeNullable = property.IsRequired ? "false" : internalType.IsNullable.ToCodeString();
+
+                            schemaProperties.AppendLine(@$"Property.{builderMethodName}(""{property.GetMappedOrOriginalName()}"", {internalTypeString}, areElementsNullable: {internalTypeNullable}, managedName: ""{property.Name}""),");
+                        }
+
+                        skipDefaultsContent.AppendLine($"newAccessor.{property.Name}.Clear();");
+
+                        copyToRealm.AppendLine($@"
+CollectionExtensions.PopulateCollection(oldAccessor.{property.Name}, newAccessor.{property.Name}, update, skipDefaults);
+");
```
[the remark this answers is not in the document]
papafe: Yes

Realm/Realm.SourceGenerator/ClassCodeBuilder.cs
```diff
@@ -0,0 +1,811 @@
...
+
+#region {baseInterface} implementation
+
+private {_accessorInterfaceName} _accessor;
+
+IRealmAccessor IRealmObjectBase.Accessor => Accessor;
+
+internal {_accessorInterfaceName} Accessor => _accessor = _accessor ?? new {_unmanagedAccessorClassName}(typeof({_classInfo.Name}));
+
+[IgnoreDataMember, XmlIgnore]
```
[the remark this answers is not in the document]
papafe: I thought that the reason why we added `XmlIgnore` only to some data members was that in some cases `IgnoreDataMember` was not enough. Given that adding it shouldn't be a problem, I've added it in all places in which we also have `IgnoreDataMember`

Realm/Realm.SourceGenerator/ClassCodeBuilder.cs
```diff
@@ -0,0 +1,811 @@
...
+
+    OnManaged();
+}}
+
+#endregion
+
+partial void OnManaged();
+
+{(_classInfo.HasPropertyChangedEvent ? string.Empty :
+$@"private event PropertyChangedEventHandler _propertyChanged;
```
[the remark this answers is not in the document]
papafe: Yes, if the original class already has it implemented we don't add our implementation.

Realm/Realm.SourceGenerator/ClassCodeBuilder.cs
```diff
@@ -0,0 +1,811 @@
...
+[IgnoreDataMember, XmlIgnore]
+public ObjectSchema ObjectSchema => Accessor.ObjectSchema;
+
+[IgnoreDataMember, XmlIgnore]
+public DynamicObjectApi DynamicApi => Accessor.DynamicApi;
+
+[IgnoreDataMember, XmlIgnore]
+public int BacklinksCount => Accessor.BacklinksCount;
+
+{parameterlessConstructorString}
```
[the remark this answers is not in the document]
papafe: In `CodeEmitter` there is a regex that is taking care of similar cases.

Realm/Realm.SourceGenerator/CodeEmitter.cs
```diff
@@ -0,0 +1,73 @@
...
+                if (!ShouldEmit(classInfo))
+                {
+                    continue;
+                }
+
+                try
+                {
+                    var generatedSource = new ClassCodeBuilder(classInfo).GenerateSource();
+
+                    // Replace all occurrences of at least 3 newlines with only 2
```
nirinchev, line 50: Not sure I understand this - why would we ever have 2 newlines?

Realm/Realm/DatabaseTypes/Accessors/ManagedAccessor.cs
```diff
@@ -199,16 +221,31 @@ private void RaisePropertyChanged(string propertyName = null)
...
         }
 
+        /// <summary>
+        /// Returns all the objects that link to this object in the specified relationship.
+        /// </summary>
+        /// <param name="objectType">The type of the object that is on the other end of the relationship.</param>
+        /// <param name="property">The property that is on the other end of the relationship.</param>
+        /// <returns>A queryable collection containing all objects of <c>objectType</c> that link to the current object via <c>property</c>.</returns>
+        [Obsolete("Use realmObject.DynamicApi.GetBacklinksFromType() instead.")]
         public IQueryable<dynamic> GetBacklinks(string objectType, string property) => DynamicApi.GetBacklinksFromType(objectType, property);
```
nirinchev, line 231: I'm not sure how I feel about this - we're obsoleting functionality that nobody should be using, can't we just delete it?

Realm/Realm/DatabaseTypes/Accessors/UnmanagedAccessor.cs
```diff
@@ -18,25 +18,24 @@
...
 using System.Linq;
 using Realms.Schema;
 
 namespace Realms
 {
-    internal class UnmanagedAccessor
-        : IRealmAccessor
+    // Should be used by generator and undocumented
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public abstract class UnmanagedAccessor : IRealmAccessor
```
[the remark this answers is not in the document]
nirinchev: Should we suppress this?

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -541,17 +541,19 @@
     /// when deleting an element from a collection bound to UI (<see href="https://github.com/realm/realm-dotnet/issues/1903">#1903</see>).
     /// </summary>
     [SuppressMessage("StyleCop.CSharp.MaintainabilityRules", "SA1402:File may only contain a single type", Justification = "This is a special object that has a very limited meaning in the project.")]
-    internal sealed class InvalidObject
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public sealed class InvalidObject
```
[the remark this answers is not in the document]
nirinchev: Should we suppress this one as well?

Realm/Realm/Dynamic/DynamicObjectApi.cs
```diff
@@ -0,0 +1,264 @@
+﻿// ////////////////////////////////////////////////////////////////////////////
```
nirinchev, line 1: Double comments

Tests/Realm.Tests/Realm.Tests.csproj
```diff
@@ -140,6 +140,7 @@
     </EmbeddedResource>
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\Realm\Realm.SourceGenerator\Realm.SourceGenerator.csproj" />
```
nirinchev, line 143: Should this be referencing the project like this? What about on CI - I imagine we want to do something similar to what we do with `Realm.csproj` where we pick it up from the nuget package.

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/ErrorClasses/ClassWithBaseType.cs
```diff
@@ -0,0 +1,28 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+using Realms;
+
+namespace SourceGeneratorPlayground
+{
+    public partial class ClassWithBaseType : RealmObj, IRealmObject
+    {
+    }
+
```
nirinchev, line 26: Empty line

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/ErrorClasses/CollectionErrors.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Threading.Tasks;
+using Realms;
+
+namespace SourceGeneratorPlayground
+{
+    public partial class CollectionErrors : IRealmObject
+    {
+        public IDictionary<int, string> UnsupportetDictionaryKeyProp { get; }
+        
+        public ISet<EmbeddedObj> SetofEmbeddedObj { get; }
```
nirinchev, line 32:
```suggestion
        public ISet<EmbeddedObj> SetOfEmbeddedObj { get; }
```

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/ErrorClasses/CollectionErrors.cs
```diff
@@ -0,0 +1,42 @@
...
+        
+        public IList<int> CollectionWithSetter { get; set; }
+
+        public IList<RealmInteger<int>> CollectionOfRealmInteger { get; }
+
+        public IList<DateTime> CollectionOfUnsupportedType { get; }
+        
+        public List<int> ListInsteadOfIlist { get; }
```
nirinchev, line 40:
```suggestion
        public List<int> ListInsteadOfIList { get; }
```

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/ErrorClasses/RealmObjectAndEmbeddedObjectClass.cs
```diff
@@ -0,0 +1,28 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+using Realms;
+
+namespace SourceGeneratorPlayground
+{
+    public partial class RealmObjectAndEmbeddedObjectClass : IRealmObject, IEmbeddedObject
+    {
+    }
+
```
nirinchev, line 26:
```suggestion
```

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/ErrorClasses/UnsupportedBacklink.cs
```diff
@@ -0,0 +1,42 @@
...
+        public IQueryable<BacklinkObj> CorrectBacklinkProp { get; }
+
+        public IQueryable<string> IgnoredBacklink { get; }
+    }
+
+    public partial class BacklinkObj : IRealmObject
+    {
+        public int Id { get; set; }
+
+        [PrimaryKey]
```
nirinchev, line 39: Why is this applied on the link property?

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/AllTypesClass_generated.cs
```diff
@@ -0,0 +1,259 @@
...
+
+        public void SetManagedAccessor(IRealmAccessor managedAccessor, IRealmObjectHelper helper = null, bool update = false, bool skipDefaults = false)
+        {
+            var newAccessor = (IAllTypesClassAccessor)managedAccessor;
+            var oldAccessor = _accessor as IAllTypesClassAccessor;
+            _accessor = newAccessor;
+
+            if (helper != null)
+            {
+
```
nirinchev, line 64: We should clean that up

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/AllTypesClass_generated.cs
```diff
@@ -0,0 +1,259 @@
...
+    {
+        char CharProperty { get; set; }
+    }
+
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    internal class AllTypesClassManagedAccessor : ManagedAccessor, IAllTypesClassAccessor
+    {
+        public char CharProperty
+        {
+            get => (char)GetValue("CharProperty");
```
nirinchev, line 196: Should we use `nameof` here? I know we're generating it and is probably irrelevant, but might look nicer?

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/BacklinkObj.diagnostics.cs
```diff
@@ -0,0 +1,14 @@
+﻿[
```
nirinchev, line 1: Why are these `.cs` files? They look like json

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/Dog_generated.cs
```diff
@@ -0,0 +1,288 @@
...
+            _accessor = newAccessor;
+
+            if (helper != null)
+            {
+
+                if(!skipDefaults || oldAccessor.Name != default(string))
+                {
+                    newAccessor.Name = oldAccessor.Name;
+                }
+                if(oldAccessor.Owner != null)
```
nirinchev, line 70: Missing newline

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/Dog_generated.cs
```diff
@@ -0,0 +1,288 @@
...
+        [IgnoreDataMember, XmlIgnore]
+        public DynamicObjectApi DynamicApi => Accessor.DynamicApi;
+
+        [IgnoreDataMember, XmlIgnore]
+        public int BacklinksCount => Accessor.BacklinksCount;
+
+        public void SetManagedAccessor(IRealmAccessor managedAccessor, IRealmObjectHelper helper = null, bool update = false, bool skipDefaults = false)
+        {
+            var newAccessor = (IDogAccessor)managedAccessor;
+            var oldAccessor = _accessor as IDogAccessor;
```
nirinchev, line 60: Why are we casting the accessors in a different way?

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/Dog_generated.cs
```diff
@@ -0,0 +1,288 @@
...
+            Property.Object("Owner", "Person", managedName: "Owner"),
+        }.Build();
+
+        #region IRealmObject implementation
+
+        private IDogAccessor _accessor;
+
+        IRealmAccessor IRealmObjectBase.Accessor => Accessor;
+
+        internal IDogAccessor Accessor => _accessor = _accessor ?? new DogUnmanagedAccessor(typeof(Dog));
```
nirinchev, line 34: Shouldn't this be:
```suggestion
        internal IDogAccessor Accessor => _accessor ?? (_accessor = new DogUnmanagedAccessor(typeof(Dog)));
```

It's almost the same, but I find it a little clearer.

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/InnerClass_generated.cs
```diff
@@ -0,0 +1,230 @@
...
+namespace SourceGeneratorAssemblyToProcess.TestClasses
+{
+    [Generated]
+    [Woven(typeof(InnerClassObjectHelper))]
+    private partial class InnerClass : IRealmObject, INotifyPropertyChanged, IReflectableType
+    {
+        public static ObjectSchema RealmSchema = new ObjectSchema.Builder("InnerClass", ObjectSchema.ObjectType.RealmObject)
+        {
+
+        }.Build();
```
nirinchev, line 25: Should this be empty? And if it is, should we avoid generating the `{ }`?

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/InnerClass_generated.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        public void SetManagedAccessor(IRealmAccessor managedAccessor, IRealmObjectHelper helper = null, bool update = false, bool skipDefaults = false)
+        {
+            var newAccessor = (IInnerClassAccessor)managedAccessor;
+            var oldAccessor = _accessor as IInnerClassAccessor;
+            _accessor = newAccessor;
+
+            if (helper != null)
+            {
+
```
nirinchev, line 64: Same here - probably should skip generating it if it doesn't have properties

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/InnerClass_generated.cs
```diff
@@ -0,0 +1,230 @@
...
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    internal class InnerClassManagedAccessor : ManagedAccessor, IInnerClassAccessor
+    {
+
+    }
+
+    internal class InnerClassUnmanagedAccessor : UnmanagedAccessor, IInnerClassAccessor
+    {
+
```
nirinchev, line 195: Empty line

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/NamespaceObj_generated.cs
```diff
@@ -0,0 +1,289 @@
...
+        public void SetManagedAccessor(IRealmAccessor managedAccessor, IRealmObjectHelper helper = null, bool update = false, bool skipDefaults = false)
+        {
+            var newAccessor = (INamespaceObjAccessor)managedAccessor;
+            var oldAccessor = _accessor as INamespaceObjAccessor;
+            _accessor = newAccessor;
+
+            if (helper != null)
+            {
+
+                if(!skipDefaults || oldAccessor.Id != default(int))
```
nirinchev, line 67: Can we instead generate ` ... !=  default`?

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/Person_generated.cs
```diff
@@ -0,0 +1,308 @@
...
+
+namespace SourceGeneratorPlayground
+{
+    [Generated]
+    [Woven(typeof(PersonObjectHelper))]
+    public partial class Person : IRealmObject, INotifyPropertyChanged, IReflectableType
+    {
+        public static ObjectSchema RealmSchema = new ObjectSchema.Builder("Person", ObjectSchema.ObjectType.RealmObject)
+        {
+            Property.Primitive("Id", RealmValueType.Guid, isPrimaryKey: true, isIndexed: false, isNullable: false, managedName: "Id"),
```
nirinchev, line 24: Do we need to specify `managedName` when it's the same as the Core name?

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/SupportClasses/EmbeddedObj.cs
```diff
@@ -0,0 +1,28 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+using Realms;
+
+namespace SourceGeneratorPlayground
+{
+    public partial class EmbeddedObj : IEmbeddedObject
+    {
+        public int Id { get; set; }
+
```
nirinchev, line 26: Empty line

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/TestClasses/AllTypesClass.cs
```diff
@@ -0,0 +1,95 @@
...
+using MongoDB.Bson;
+using Realms;
+
+namespace SourceGeneratorPlayground
+{
+    public partial class AllTypesClass : IRealmObject
+    {
+        public char CharProperty { get; set; }
+
+        //public byte ByteProperty { get; set; }
```
nirinchev, line 28: Should these be uncommented.

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/TestClasses/ClassWithoutParameterlessConstructor.cs
```diff
@@ -0,0 +1,37 @@
+﻿// ////////////////////////////////////////////////////////////////////////////
```
nirinchev, line 1: Double commenting

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/TestClasses/DifferentNamespaces.cs
```diff
@@ -0,0 +1,44 @@
+﻿// ////////////////////////////////////////////////////////////////////////////
```
nirinchev, line 1: Double comments

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/TestClasses/DifferentNamespaces.cs
```diff
@@ -0,0 +1,44 @@
...
+// //
+// // Unless required by applicable law or agreed to in writing, software
+// // distributed under the License is distributed on an "AS IS" BASIS,
+// // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// // See the License for the specific language governing permissions and
+// // limitations under the License.
+// //
+// ////////////////////////////////////////////////////////////////////////////
+
+using System;
```
nirinchev, line 19: Do we need all those usings?

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/TestClasses/InnerClass.cs
```diff
@@ -0,0 +1,37 @@
+﻿// ////////////////////////////////////////////////////////////////////////////
```
nirinchev, line 1: Double comments

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/TestClasses/PartialClass.cs
```diff
@@ -0,0 +1,32 @@
+﻿// ////////////////////////////////////////////////////////////////////////////
+// //
```
nirinchev, line 2: Comments

Tests/Weaver/Realm.FakeForWeaverTests/Realm.FakeForWeaverTests.csproj
```diff
@@ -1,38 +1,39 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <AssemblyName>Realm</AssemblyName>
-  </PropertyGroup>
+    <PropertyGroup>
```
nirinchev, line 3: Why do we need these changes?

nirinchev (commented):
I think this is almost ready to go. I have some code cleanup comments and a bunch of places where I've commented on the generated class where we need to more carefully handle corner cases to ensure we generate beautiful code.

Realm/Realm.SourceGenerator/CodeEmitter.cs
papafe: Two newlines means that there is one empty line between code, that is what we want at most.

Realm/Realm/DatabaseTypes/Accessors/ManagedAccessor.cs
papafe: We definitely can, I was trying to be too conservative probably

papafe: If we delete it then it'll be breaking change. because this is used by the weaved classes too.

Tests/Realm.Tests/Realm.Tests.csproj
papafe: This file is heavily changed in the other PR, so that it uses the local reference locally and the nuget package on CI

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/BacklinkObj.diagnostics.cs
papafe: This is a limitation of source generators (there is a comment about it in `DiagnosticsEmitter`), because at the moment they can only emit code files, so .cs is added on all generated files. There's a discussion about supporting this https://github.com/dotnet/roslyn/issues/57608. 
Given that this is just for internal use I didn't worried about it too much.

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/Dog_generated.cs
papafe: My bad

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/Person_generated.cs
papafe: It's not necessary, but I think it doesn't hurt, and we avoid complicating the generator code.

Tests/Weaver/Realm.FakeForWeaverTests/Realm.FakeForWeaverTests.csproj
papafe: The only difference is that I've added the "Include" to the Generated attribute. It looks like this because I run the autoformat on the file

Realm/Realm.SourceGenerator/ClassCodeBuilder.cs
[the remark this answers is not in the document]
papafe: What do you think, is it fine to keep them?

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/AllTypesClass_generated.cs
papafe: This is a more general problem of using string interpolation with substrings that could be empty or not. 
If we want to avoid all string we could build the whole method using string builder with append/appendLine, but the code (our code) will become less readable in my opinion. Do you have any alternative suggestion?

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/NamespaceObj_generated.cs
papafe: `default` literal has been introduced in C# 7.1.  Is it old enough that we expect all the users of realm to use an older version?

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/ErrorClasses/UnsupportedBacklink.cs
papafe: The expectation here is that all the classes in this folder are generating only diagnostics and no actual code. This was to be sure that this class would not be generated.

papafe:
As a note, don't look at the CI. I've merged back into this branch only the changes I've done in the Realm folder, so they were easier to read here, so there's a lot of things missing that are in the other PR

Realm/Realm.SourceGenerator/CodeEmitter.cs
nirinchev: Ah 🤦‍♂️ makes sense

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/AllTypesClass_generated.cs
nirinchev: Good point.

Issue #2406, opened 2021-06-01
[Unity] Extend a github action for building Unity projects


geragray:
We're thinking about using [Unity builder](https://github.com/game-ci/unity-builder) to build our Unity project. It doesn't support floating build licensing so we should extend the github action/fork it and add support: https://github.com/game-ci/unity-builder/issues/185.

nirinchev:
Point of contact from Realm: @nirinchev and @LaPeste

bwachter:
TL;DR for mac: I have a VM with OS-X/intel capable of doing Unity builds with a fixed seat license. Can be used to get mac tests running while I work on getting it to cooperate with the license server as soon as I get a github token to register the runner there.

Unity on OS-X has quite a few issues, and mostly does error handling by not logging anything and not terminating - so a lot of hung processes without obvious reason. Adding to that is that they're wildly jumping between putting data in the users home and system wide location for the licensing client - not all of them are writeable on a default installation, and not in all instances do they handle write errors gracefully. On top of that parts of the license validation seem to call APIs which only work if the user running the license service also has an active desktop session - so currently unity on mac is running as a user logged into the desktop, with the ability to gain elevated privileges, which is not quite ideal. 

I've also observed some freezes which seemed to have been due to powershell - so to rule that out as problem currently the unity-editor wrapper on mac is just exec /Applications/Unity/Unity.app/Contents/MacOS/Unity $@

I've cloned the VM with the working builds to one of our servers, so it can be registered as github worker, and am now working on a copy of it to get the interaction with the license server working.

bmunkholm:
Nikola will look into this issue next week.

bmunkholm:
No progress so far. Andrea will write wiki on how to simulate CI run locally.

PR #2963, opened 2022-06-20
Start running Unity tests on CI

nirinchev:
## Description

This adds Windows Mono tests for Unity. This is the only platform we can test for right now because of issues still being sorted out with our self-hosted runners. Other platforms/backends will be added in a follow-up PR.

A few notes:
1. We have 3 failing tests. Fixing them is out of scope for this PR and I'll address them in a follow-up one.
1. PR builds will fail if more than one build tries to build for Unity at the same time. This is being addressed at the runner level and should be fixed at a later point.

Part of https://github.com/realm/realm-dotnet/issues/2406.


nirinchev:
@LaPeste since @fealebenpae went on a tactical vacation prior to reviewing this PR, you've been volunteered to take his spot 😝

.github/templates/test-unity.yml
```diff
@@ -0,0 +1,97 @@
...
+    if: ${{ contains(inputs.platform, 'Linux') }}
+    steps:
+      - #@ template.replace(checkoutCode())
+      - name: Download Test Runner
+        uses: #@ actionDownloadArtifact
+        with:
+          name: UnityTestsRunner.linux
+          path: TestRunner
+      - name: Install xvfb
+        run: sudo apt-get install xvfb
```
LaPeste, line 70:
```suggestion
        run: sudo apt install -y xvfb
```

LaPeste (approved):
Looks fine. I just suggested a minuscule change.

nirinchev:
@cla-bot check

PR #2980, opened 2022-07-27
Fix CI Android

### LINKED ISSUE #2979: CI failures because of changes in Github Actions virtual environments
CI is failing because of https://github.com/actions/virtual-environments/issues/5879

papafe:
Fixes #2979


papafe:
@LaPeste Let's see if this fixes the issue for now. 
There is a still an ongoing discussion about this in the issue https://github.com/actions/virtual-environments/issues/5879#issuecomment-1195156618. Maybe they'll fix those variables, so we need to keep an eye on it

nirinchev:
@papafe should we merge this? The sync test failures seem to be unrelated and something @LaPeste is following up on.

PR #2993, opened 2022-08-12
Implement asymmetric sync for the .NET SDK

### LINKED ISSUE #2908: Implement Asymmetric Sync

LaPeste:
## Description
Implements asymmetric sync for the .NET SDK
Fixes #2908 

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -46,29 +46,32 @@ SequencePoint GetPropSequencePoint()
         }
     }
 
-    public static bool IsRealmObjectDescendant(this TypeReference @this, ImportedReferences references)
+    private static bool IsTObjectDescendant(TypeReference target, params TypeReference[] instanceOf)
```
nirinchev, line 49:
```suggestion
    private static bool IsDescendantOf(TypeReference @this, params TypeReference[] targetTypes)
```

I got a little confused as to what `target` and `instanceOf` mean. I'm not sure my suggestion is much better, but I think it's a step in the right direction.

Realm/Realm/DatabaseTypes/AsymmetricObject.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// The benefit of using <see cref="AsymmetricObject"/> is that the performance of each sync operation is much higher.
+    /// The drawback is that an <see cref="AsymmetricObject"/> is synced unidirectionally, so it cannot be queried.
+    /// You should use this base when you have a write-heavy use case.
+    /// If, instead you want to persist an object that you can also query against, use <see cref="RealmObject"/> instead.
+    /// RealmObjects and EmbeddedObjects can't link (or backlink) to AsymmetricObjects. AsymmetricObjects can only link to EmbeddedObjects.
+    /// </remarks>
+    /// <seealso cref="TODO andrea: add asymmetric sync link page here"/>
+    public class AsymmetricObject : RealmObjectBase, IAsymmetricObject
+    {
+        // TODO andrea: should this be only instantiated in a write transaction? Since we want to enforce them to be short-lived
```
nirinchev, line 34: No - I think it's fine to instantiate them outside. An example is a sensor which records readings, but doesn't want to sync them one by one. Instead, the developer can accumulate readings from 10-30s in memory, then sync everything in a single transaction.

Realm/Realm/Dynamic/DynamicAsymmetricObject.cs
```diff
@@ -0,0 +1,34 @@
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
```
nirinchev, line 3:
```suggestion
// Copyright 2022 Realm Inc.
```

Realm/Realm/Dynamic/MetaRealmObject.cs
```diff
@@ -29,6 +29,7 @@
 
 namespace Realms.Dynamic
 {
+    // TODO andrea: this class may need adjustment as there are no tests written yet
```
nirinchev, line 32: This should be tested in `Dynamic***Tests`.

Realm/Realm/Extensions/ReflectionExtensions.cs
```diff
@@ -47,8 +48,26 @@ public static bool HasCustomAttribute<T>(this MemberInfo member)
...
+
         public static bool IsRealmObject(this Type type) => typeof(IRealmObject).IsAssignableFrom(type);
 
+        public static ObjectSchema.ObjectSchemaType GetRealmSchemaType(this Type type)
+        {
+            if (type.IsAsymmetricObject())
+            {
+                return ObjectSchema.ObjectSchemaType.TopLevelAsymmetric;
+            }
+            else if (type.IsEmbeddedObject())
```
nirinchev, line 61: You don't need the `else`-s here.

Realm/Realm/Native/SchemaObject.cs
```diff
@@ -32,6 +32,6 @@ internal struct SchemaObject
         internal int properties_end;
 
         [MarshalAs(UnmanagedType.U1)]
```
nirinchev, line 34: That should not be necessary anymore

Realm/Realm/Native/SchemaObject.cs
```diff
@@ -32,6 +32,6 @@ internal struct SchemaObject
...
-        internal bool is_embedded;
+        internal byte table_type;
```
nirinchev, line 35: You should be able to use the enum here, not the underlying type.

Realm/Realm/Realm.cs
```diff
@@ -608,9 +621,11 @@ internal void ManageEmbedded(IEmbeddedObject obj, ObjectHandle handle)
             obj.SetManagedAccessor(new ManagedAccessor(this, handle, metadata), metadata.Helper, false, true);
         }
 
-        private void AddInternal(IRealmObject obj, Type objectType, bool update)
+        // TODO andrea: see if revert this to IRealmObject and have another AddInternal with different rules
+        private void AddInternal<T>(T obj, Type objectType, bool update)
+            where T : IRealmObjectBase
         {
-            if (!ShouldAddNewObject(obj))
+            if (!ShouldAddNewObject(obj) || obj.GetType().IsEmbeddedObject())
```
nirinchev, line 628: If we call this with `EmbeddedObject`, this should be an error, shouldn't it?

Realm/Realm/Schema/ObjectSchema.cs
```diff
@@ -35,6 +35,16 @@ namespace Realms.Schema
...
     public class ObjectSchema : IReadOnlyCollection<Property>
     {
+        // TODO andrea:
+        // 1 - move this to the right place
+        // 2- add documentation
+        public enum ObjectSchemaType : byte
+        {
+            TopLevel = 0,
+            Embedded = 1,
+            TopLevelAsymmetric = 2,
```
nirinchev, line 45: If those are public, we should match the terminology we use for objects - i.e. `RealmObject`, `EmbeddedObject`, `AsymmetricObject`.

Realm/Realm/Schema/ObjectSchema.cs
```diff
@@ -59,12 +69,22 @@ public class ObjectSchema : IReadOnlyCollection<Property>
...
         public bool IsEmbedded { get; }
 
-        private ObjectSchema(string name, bool isEmbedded, IDictionary<string, Property> properties)
+        // TODO andrea: review this documentation
+        /// <summary>
+        /// Gets a <see cref="ObjectSchemaType"/> indicating whether this <see cref="ObjectSchema"/> describes
+        /// a top level object, an embedded object or an asymmetric object.
+        /// </summary>
+        /// <value>The type of ObjectSchema.</value>
+        public ObjectSchemaType RealmSchemaType { get; }
```
nirinchev, line 81: The name of this property is kind of weird. I'd suggest just `Type` or `ObjectType`.

Realm/Realm/Schema/ObjectSchema.cs
```diff
@@ -35,6 +35,16 @@ namespace Realms.Schema
     [DebuggerDisplay("Name = {Name}, Properties = {Count}")]
     public class ObjectSchema : IReadOnlyCollection<Property>
     {
+        // TODO andrea:
+        // 1 - move this to the right place
+        // 2- add documentation
+        public enum ObjectSchemaType : byte
```
nirinchev, line 41: The name of the enum is a little misleading as this is not the type of the objectschema but rather the type of the object the objectschema represents.

Realm/Realm/Schema/ObjectSchema.cs
```diff
@@ -145,31 +165,54 @@ public class Builder : SchemaBuilderBase<Property>
...
+            /// </summary>
+            /// <param name="name">The name of the <see cref="ObjectSchema"/> this builder describes.</param>
+            /// <param name="isEmbedded">A flag indicating whether the object is embedded or not.</param>
+            /// <seealso cref="EmbeddedObject"/>
+            /// <exception cref="ArgumentNullException">Thrown if <paramref name="name"/> is <c>null</c>.</exception>
+            /// <exception cref="ArgumentException">Thrown if <paramref name="name"/> is the empty string.</exception>
+            [Obsolete("Use Builder(string name, ObjectSchemaType schemaType) instead.")]
+            public Builder(string name, bool isEmbedded = false) : this(
+                name,
+                isEmbedded ? ObjectSchemaType.Embedded : ObjectSchemaType.TopLevel)
```
nirinchev, line 204:
```suggestion
            public Builder(string name, bool isEmbedded = false)
                : this(name, isEmbedded ? ObjectSchemaType.Embedded : ObjectSchemaType.TopLevel)
```

Realm/Realm/Schema/ObjectSchema.cs
```diff
@@ -145,31 +165,54 @@ public class Builder : SchemaBuilderBase<Property>
...
+            // TODO andrea: review the documentation
             /// <summary>
             /// Initializes a new instance of the <see cref="Builder"/> class populated with properties from the
             /// provided <paramref name="type"/>.
             /// </summary>
             /// <param name="type">
-            /// The <see cref="System.Type"/> that will be used to populate the builder. It must be a <see cref="RealmObject"/>
-            /// or <see cref="EmbeddedObject"/> inheritor.
+            /// The <see cref="System.Type"/> that will be used to populate the builder. It must be a <see cref="RealmObject"/>,
+            /// an <see cref="EmbeddedObject"/>  or a <see cref="AsymmetricObject"/>inheritor.
```
nirinchev, line 215:
```suggestion
            /// an <see cref="EmbeddedObject"/> , or a <see cref="AsymmetricObject"/>inheritor.
```

Realm/Realm/Schema/ObjectSchema.cs
```diff
@@ -212,10 +255,13 @@ public Builder(string name, bool isEmbedded = false)
             public Builder(Type type)
             {
                 Argument.NotNull(type, nameof(type));
-                Argument.Ensure(type.IsRealmObject() || type.IsEmbeddedObject(), $"The class {type.FullName} must descend directly from RealmObject or EmbeddedObject", nameof(type));
+                Argument.Ensure(type.IsRealmObject() || type.IsEmbeddedObject() || type.IsAsymmetricObject(),
+                    $"The class {type.FullName} must descend directly from either RealmObject, EmbeddedObject or AsymmetricObject", nameof(type));
```
nirinchev, line 259:
```suggestion
                    $"The class {type.FullName} must descend directly from either RealmObject, EmbeddedObject, or AsymmetricObject", nameof(type));
```

Realm/Realm/Schema/ObjectSchema.cs
```diff
@@ -145,31 +165,54 @@ public class Builder : SchemaBuilderBase<Property>
             /// Gets or sets a value indicating whether this <see cref="Builder"/> describes an embedded object.
             /// </summary>
             /// <value><c>true</c> if the schema pertains to an <see cref="EmbeddedObject"/> instance; <c>false</c> otherwise.</value>
+            [Obsolete("Check against RealmSchemaType instead.")]
             public bool IsEmbedded { get; set; }
```
nirinchev, line 169: This should be implemented in terms of `RealmSchemaType`.

nirinchev (commented):
Haven't reviewed the tests yet, but so far looks pretty good. I've mostly have nitty comments.

Realm/Realm/Schema/ObjectSchema.cs
LaPeste: Agree. How about one of the following, in order of my personal preference?
1. `RealmObjectType`
1. `BaseObjectType`
1. `ObjectType`

LaPeste:
@nirinchev you can start reviewing the PR. The only thing missing is that I want to see if I can include more test values in `AsymmetricObjectTests.SetAndRemotelyReadValue`. This test is already working, beside the min and max `Decimal128`, which is should be related to the [bug](https://github.com/realm/realm-core/pull/5740) in core you mentioned.

Realm/Realm/Realm.cs
```diff
@@ -595,7 +595,81 @@ public void Add<T>(IEnumerable<T> objs, bool update = false)
...
+        /// <exception cref="ObjectDisposedException">
+        /// If you invoke this method on a closed <see cref="Realm"/>.
+        /// </exception>
+        /// <exception cref="ArgumentNullException">
+        /// If you pass a null argument to this method.
+        /// </exception>
+        /// <remarks>
+        /// If the object is already managed by this <see cref="Realm"/>, this method does nothing.
+        /// This method modifies the object in-place, meaning that after it has run, <see cref="AsymmetricObject"/> will be managed.
+        /// Once an <see cref="AsymmetricObject"/> becomes managed dereferencing the original <see cref="AsymmetricObject"/>
```
nirinchev, line 625: I don't understand this part.

Realm/Realm/Realm.cs
```diff
@@ -608,8 +682,14 @@ internal void ManageEmbedded(IEmbeddedObject obj, ObjectHandle handle)
             obj.SetManagedAccessor(new ManagedAccessor(this, handle, metadata), metadata.Helper, false, true);
         }
 
-        private void AddInternal(IRealmObject obj, Type objectType, bool update)
+        private void AddInternal<T>(T obj, Type objectType, bool update)
+            where T : IRealmObjectBase
         {
+            if (obj.GetType().IsEmbeddedObject())
```
nirinchev, line 688:
```suggestion
            if (objectType.IsEmbeddedObject())
```

Realm/Realm/Schema/ObjectSchema.cs
nirinchev: Let's go with `ObjectType`.

Tests/Realm.Tests/Database/DynamicAsymmetricObjectTests.cs
```diff
@@ -0,0 +1,87 @@
+﻿// ////////////////////////////////////////////////////////////////////////////
```
nirinchev, line 1: Double comments

Tests/Realm.Tests/Database/DynamicAsymmetricObjectTests.cs
```diff
@@ -0,0 +1,87 @@
...
+using Realms.Tests.Sync;
+
+namespace Realms.Tests.Database
+{
+    [TestFixture, Preserve(AllMembers = true)]
+    public class DynamicAsymmetricObjectTests : SyncTestBase
+    {
+        [TestCase(true)]
+        [TestCase(false)]
+        public void ReadWriteVarsInWriteTransaction(bool isDynamic)
```
nirinchev, line 31:
```suggestion
        [Test]
        public void ReadWriteVarsInWriteTransaction([Values(true, false)] bool isDynamic)
```

Tests/Realm.Tests/Sync/SyncTestBase.cs
```diff
@@ -197,10 +197,10 @@ protected async Task<FlexibleSyncConfiguration> GetFLXIntegrationConfigAsync(App
             return UpdateConfig(new FlexibleSyncConfiguration(user, optionalPath));
         }
 
-        protected async Task<Realm> GetFLXIntegrationRealmAsync(App app = null)
+        protected async Task<Realm> GetFLXIntegrationRealmAsync(App app = null, FlexibleSyncConfiguration flxConfig = null)
         {
-            var config = await GetFLXIntegrationConfigAsync(app);
-            return await GetRealmAsync(config);
+            flxConfig ??= await GetFLXIntegrationConfigAsync(app);
+            return await GetRealmAsync(flxConfig);
```
nirinchev, line 203: Why this change? If you pass in the config, it's better to just use `GetRealmAsync` directly.

Tests/Realm.Tests/Database/DynamicAsymmetricObjectTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            SyncTestHelpers.RunBaasTestAsync(async () =>
+            {
+                var flxConfig = await GetFLXIntegrationConfigAsync();
+                flxConfig.IsDynamic = isDynamic;
+                flxConfig.Schema = new[] { typeof(AsymmetricObjectWithAllTypes) };
+                using var realm = await GetFLXIntegrationRealmAsync(flxConfig: flxConfig);
```
nirinchev, line 38:
```suggestion
                using var realm = await GetRealmAsync(flxConfig);
```

Tests/Realm.Tests/Database/ObjectSchemaTests.cs
```diff
@@ -884,7 +891,8 @@ public void ObjectSchema_GetBuilder_ContainsAllProperties()
 
             Assert.That(originalSchema, Is.EquivalentTo(builder));
             Assert.That(originalSchema.Name, Is.EqualTo(builder.Name));
-            Assert.That(originalSchema.IsEmbedded, Is.EqualTo(builder.IsEmbedded));
+            Assert.That(originalSchema.ObjectType == ObjectSchema.ObjectSchemaType.EmbeddedObject,
+                Is.EqualTo(builder.RealmSchemaType == ObjectSchema.ObjectSchemaType.EmbeddedObject));
```
nirinchev, line 895: This seems like a weird check - why not just assert that `originalSchema.ObjectType, Is.EqualTo(builder.RealmSchemaType)`?

Tests/Weaver/Realm.Fody.Tests/WeaverTests.cs
```diff
@@ -521,7 +522,16 @@ public void MatchErrorsAndWarnings()
                 "RealmSetWithSetter.People has a setter but its type is a ISet which only supports getters.",
                 "RealmDictionaryWithSetter.People has a setter but its type is a IDictionary which only supports getters.",
                 "RealmDictionaryWithNonStringKey.People is a Dictionary<Int32, Person> but only string keys are currently supported by Realm.",
-                "MixOfCollectionsObject.EmbeddedSet is a Set<EmbeddedObject> which is not supported. Embedded objects are always unique which is why List<EmbeddedObject> already has Set semantics."
+                "MixOfCollectionsObject.EmbeddedSet is a Set<EmbeddedObject> which is not supported. Embedded objects are always unique which is why List<EmbeddedObject> already has Set semantics.",
+                "Measurement.Sensor is of type AsymmetricObject, but AsymmetricObject-s aren't allowed to be the receiving end of any relationship.",
```
nirinchev, line 526: Why is there a dash here?

Tools/DeployApps/BaasClient.cs
```diff
@@ -699,6 +717,17 @@ private static class Schemas
...
+            private static object GenericFlxBaasRule(string differentiator, string collectionName) => new
+            {
+                collection = collectionName,
+                database = $"FLX_{differentiator}",
+                roles = new[] { _flxDefaultRoles }
+            };
+
+            public static object AsymmetricObjectWithAllTypesBaasRule(string differentiator) => GenericFlxBaasRule(differentiator, "AsymmetricObjectWithAllTypes");
+
+            public static object HugeSyncAsymmetricObjectBaasRule(string differentiator) => GenericFlxBaasRule(differentiator, "HugeSyncAsymmetricObject");
```
nirinchev, line 729: We probably don't need these explicit functions, we can use the common one directly. Also, we should use `nameof` here instead of hardcoded strings.

Tests/Realm.Tests/Database/ObjectSchemaTests.cs
```diff
@@ -25,6 +25,8 @@
 
 namespace Realms.Tests.Database
 {
+#pragma warning disable CS0618 // ObjectSchema.Builder(string, bool) is obsolete, Builder(string, ObjectSchemaType) should be used instead
```
nirinchev, line 28: Do we need to disable this for the entire file? Can we migrate all the tests to use the new constructor and if you want to test the old one, then create a new test and mark it `[Obsolete]`?

Tests/Realm.Tests/Database/AsymmetricObjectTests.cs
```diff
@@ -0,0 +1,382 @@
...
+                    realm.Write(() =>
+                    {
+                        realm.Add(asymmetricObj);
+                    });
+
+                    realm.Write(() =>
+                    {
+                        realm.Add(asymmetricObj);
+                    });
+                });
```
nirinchev, line 216: Why doesn't this throw? After the first write, we detach the object, then try to add it again - this should be an error as we're trying to add a deleted object to the Realm.

Tests/Realm.Tests/Database/AsymmetricObjectTests.cs
```diff
@@ -0,0 +1,382 @@
...
+            });
+        }
+
+        [TestCaseSource(nameof(SetAndGetValueCases))]
+        [TestCaseSource(nameof(SetAndReplaceWithNullCases))]
+        public void SetAndRemotelyReadValue(string propertyName, object propertyValue)
+        {
+            SyncTestHelpers.RunBaasTestAsync(async () =>
+            {
+                Guid objId = Guid.Empty;
```
nirinchev, line 233:
```suggestion
                var objId = Guid.Empty;
```

Tests/Realm.Tests/Database/AsymmetricObjectTests.cs
```diff
@@ -0,0 +1,382 @@
...
+                });
+
+                await realm.SyncSession.WaitForUploadAsync();
+                var documents = await GetObjFromRemoteThroughMongoClient<AsymmetricObjectWithAllTypes>(
+                    flxConfig, propertyName, BsonValue.Create(propertyValue));
+
+                foreach (var doc in documents)
+                {
+                    Assert.That(TestHelpers.GetPropertyValue(doc, propertyName), Is.EqualTo(propertyValue));
+                }
```
nirinchev, line 254: This assertion is weird. Since you're querying for objects with that property value, it's normal that all the returned documents will have the correct property. Instead, you should probably query by the objId and then assert on the property value.

Tests/Realm.Tests/Database/AsymmetricObjectTests.cs
```diff
@@ -0,0 +1,382 @@
...
+                        realm.Add(new PrimaryKeyInt32Object
+                        {
+                            Id = id
+                        });
+                    });
+                });
+            });
+        }
+
+        [Test, NUnit.Framework.Explicit("Once Daniel Tabacaru's work is done on \"error actions\" this will be an \"application bug\" action: https://github.com/10gen/baas/blob/9f32d54aa79aff6dfb36a6c07742594e38b07441/realm/sync/protocol/protocol_errors.go#L439")]
```
nirinchev, line 294: This should be in `main` now.

Tests/Realm.Tests/Database/AsymmetricObjectTests.cs
```diff
@@ -0,0 +1,382 @@
...
+        {
+            TestHelpers.RunAsyncTest(async () =>
+            {
+                var config = (RealmConfiguration)RealmConfiguration.DefaultConfiguration;
+                config.Schema = new[] { typeof(BasicAsymmetricObject) };
+
+                Exception ex = null;
+                try
+                {
+                    using var realm = await GetRealmAsync(config);
```
nirinchev, line 327: Why do you need the async version here - wouldn't you get the same error with `GetRealm`?

Tests/Realm.Tests/Database/AsymmetricObjectTests.cs
```diff
@@ -0,0 +1,382 @@
...
+                {
+                    ex = e;
+                }
+
+                Assert.That(ex, Is.InstanceOf<RealmSchemaValidationException>());
+                Assert.That(ex.Message.Contains($"Asymmetric table \'{nameof(BasicAsymmetricObject)}\'"), Is.True);
+            });
+        }
+
+        private static Task<T[]> GetObjFromRemoteThroughMongoClient<T>(FlexibleSyncConfiguration config, string remoteFieldName, BsonValue fieldValue, string mongoClientCondition = MongoClientCondition.Equality)
```
nirinchev, line 339:
```suggestion
        private static Task<T[]> GetObjFromRemoteThroughMongoClient<T>(User user, string remoteFieldName, BsonValue fieldValue, string mongoClientCondition = MongoClientCondition.Equality)
```

I thought it's quite confusing we're passing a config here.

Realm/Realm/Realm.cs
LaPeste: Could you be more specific?

Realm/Realm/Schema/ObjectSchema.cs
LaPeste: Then how would you like the field called? It's currently called exactly `ObjectType`. We could call it `SchemaType`? Would that be ok?

Realm/Realm/Realm.cs
nirinchev: "dereferencing the original reference throws an exception" - I don't believe this is the case, why would it throw?

LaPeste: Once the `AsymmetricObject` becomes managed, you get a runtime error if trying to use that reference.
We all agreed that this is how it should be. Why do you disagree?
The test for this is `AccessAsymmetricObjAfterAddedToRealm_Throws`

Tests/Realm.Tests/Sync/SyncTestBase.cs
LaPeste: I don't understand this comment. If the config is passed, it already directly uses `GetRealmAsync`.

Tests/Weaver/Realm.Fody.Tests/WeaverTests.cs
LaPeste: We usually write with the back-ticks, where it's easy to understand how the object is called -> `AsymmetricObject`s.
 I wanted to keep it simple to read when not having the back-ticks in place. Nothing else.
Should I just have AsymmetricObjects?

Realm/Realm/Realm.cs
LaPeste: Conversation taken offline. Resolved.

Tests/Realm.Tests/Database/AsymmetricObjectTests.cs
LaPeste: Wouldn't this be true for almost any test? Is there a technical reason you're asking this question on this specific test?
And in general, unless one is testing the async API, what would be the reason to want to specifically use one (the async one) over the other (the sync one)?
(I assume you're asking here because there's no need to use the async method to grab the configuration)

CHANGELOG.md
```diff
@@ -1,6 +1,46 @@
 ## vNext (TBD)
 
 ### Enhancements
+* Introduced `AsymmetricObject` intended for write-heavy workloads, where high performance is generally important. This new object:
+  1. syncs data unidirectionaly, from the clients to the server
+  1. can't be queried, deleted or modified once added to the Realm
+  1. and as a result the previous point, at the end of a write transaction dereferencing the added `AsymmetricObject` throws an exception
+  1. is only usable with flexible sync
+  1. can't be the receiveing end of any type of relationship
+  1. can only have a one-to-many relationship with `EmbeddedObject` and collections of the latter
```
nirinchev, line 10: It can also a have -to-one relationships.

CHANGELOG.md
```diff
@@ -1,6 +1,46 @@
...
+
+  In the same write transaction, it is perfetcly legal to add `AsymmetricObject`-s and `RealmObject`-s
```
```suggestion
  In the same write transaction, it is perfetcly legal to add `AsymmetricObject`s and `RealmObject`s
```

CHANGELOG.md
```diff
@@ -1,6 +1,46 @@
...
+
+  //.....
+
+  var asymmetricObject = new BasicAsymmetricObject();
+
+  realm.Write(() =>
+  {
+      realm.Add(asymmetricObject);
+
+      realm.Add(new PrimaryKeyInt32Object
```
nirinchev, line 35: Should this be `Person`?

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -381,13 +381,21 @@ private WeavePropertyResult WeaveProperty(PropertyDefinition prop, TypeDefinitio
...
 
             var backlinkAttribute = prop.CustomAttributes.FirstOrDefault(a => a.AttributeType.Name == "BacklinkAttribute");
-            if (backlinkAttribute != null && !prop.IsIQueryable())
+            if (backlinkAttribute != null)
             {
-                return WeavePropertyResult.Error($"{type.Name}.{prop.Name} has [Backlink] applied, but is not IQueryable.");
+                if (!prop.IsIQueryable())
+                {
+                    return WeavePropertyResult.Error($"{type.Name}.{prop.Name} has [Backlink] applied, but is not IQueryable.");
+                }
```
nirinchev, line 398: Why was that split?

Realm/Realm/Realm.cs
```diff
@@ -595,7 +595,85 @@ public void Add<T>(IEnumerable<T> objs, bool update = false)
...
+        /// dereferencing any property of the original <see cref="AsymmetricObject"/> reference throw an exeception.
+        /// Hence, none of the properties of the elements in the collection can be deferenced anymore after the transaction.
+        /// </remarks>
+        public void Add<T>(IEnumerable<T> objs)
+            where T : IAsymmetricObject
+        {
+            ThrowIfDisposed();
+            Argument.NotNull(objs, nameof(objs));
+            Argument.Ensure(objs.All(o => o != null), $"{nameof(objs)} must not contain null values.", nameof(objs));
+            Argument.Ensure(objs.All(o => !o.IsManaged), $"{nameof(objs)} must not contain already managed objects by a Realm.", nameof(objs));
```
nirinchev, line 672: This is inefficient because we're iterating the collection twice. We could instead have a foreach loop that checks for both conditions.

Realm/Realm/Schema/ObjectSchema.cs
```diff
@@ -145,31 +174,64 @@ public class Builder : SchemaBuilderBase<Property>
...
+            {
+                get
+                {
+                    return RealmSchemaType == ObjectType.EmbeddedObject;
+                }
+
+                set
+                {
+                    // being an obsolete method, it's assumed that it's old code that doesn't use AsymmetricObjects
+                    _ = value ? RealmSchemaType = ObjectType.EmbeddedObject : RealmSchemaType = ObjectType.RealmObject;
```
nirinchev, line 188: That is super awkward - let's simplify to

```suggestion
                    RealmSchemaType = value ? ObjectType.EmbeddedObject : ObjectType.RealmObject;
```

nirinchev (commented):
Final nits from me

Realm/Realm.Weaver/RealmWeaver.cs
LaPeste: A forgotten leftover of a check that later was moved down. Thank you for noticing it.

Realm/Realm/Schema/ObjectSchema.cs
LaPeste: Emh.... 🤦🏻

PR #3002, opened 2022-08-23
Try to fix some flx reset tests

nirinchev:
## Description

Fixes #

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


LaPeste (approved):
Cleaner solution. I didn't know you could do that with parameters of tests.
I see what you tried to do with the failing test. I guess that'll need to be removed when core addresses the issue as it wouldn't be necessary anymore

PR #3005, opened 2022-08-24
[SG] Diagnostics tweaks

nirinchev:


Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/CollectionErrors.diagnostics.cs
```diff
@@ -1,6 +1,6 @@
 ﻿[
   {
-    "Id": "REALM001",
+    "Id": "???",
```
nirinchev, line 3: This file doesn't appear to be exercised in tests which is why I put ??? here - when we write tests for it, they should fail.

Realm/Realm.SourceGenerator/Parser.cs
```diff
@@ -374,7 +374,6 @@ private static PropertyTypeInfo GetSingleLevelPropertyTypeInfo(ITypeSymbol typeS
             propInfo.TypeSymbol = typeSymbol;
             propInfo.CompleteTypeSymbol = completeTypeSymbol;
             propInfo.NullableAnnotation = nullableAnnotation;
-            propInfo.Namespace = typeSymbol.ContainingNamespace.ToString();
```
papafe, line 377: This we actually need

papafe (approved):

PR #3006, opened 2022-08-25
Android MAUI testing

fealebenpae:
Add CI testing for .NET 6 on Android.

The new tooling is still finicky, and builds for `net6.0-android` choked because `Realm.Tests` had the Xamarin frameworks in its target frameworks list. So I ripped them out. Now `Realm.Tests` builds for `netstandard2.0`, `netcoreapp3.1` (except on ARM64 macOS where there is no native runtime host), `net6.0` (if the tooling supports it - the Xamarin Mono toolchain does not), and `net461` (if running on Windows). I got rid of `net5.0` altogether, because it's no longer supported by Microsoft and Visual Studio has a nag prompt to migrate away from it. This makes the whole `AdditionalFrameworks` song and dance unnecessary, but it did cause some cascading changes in `TestHelpers.IsAOTPlatform` and `TestHelpers.IsUnity`, which produced more noise than I'd have liked.


nirinchev (approved):

nirinchev:
You need to update `TestHttpClientHandler` in `AppTests.cs` to inherit from `AndroidClientHandler` for MAUI (similarly to what we've done for Xamarin) to get the failing test to pass.

PR #3015, opened 2022-08-30
Fixes some Realm.Test annoyances in VS

LaPeste:
## Description
* Fixes the way wrappers are selected when they are copied over during the compilation of `Realm.Test`
* Adds `Microsoft.NET.Test.Sdk` to satisfy `.netcore3.1` which before was not finding the `testhost`
* Changes the order in which `TargetFrameworks` are inserted in `Realm.Test` since Visual Studio Run Tests context menu just picks up the first in line


nirinchev:
That seems to be based off the asymmetric sync branch. If you want to merge it independently, then can you cherry pick the last commit on top of main instead?

PR #3025, opened 2022-09-16
[SG] Conversion of tests to use source generators

papafe:


Tests/Realm.Tests/Sync/AsymmetricObjectTests.cs
```diff
@@ -318,27 +327,9 @@ public void AsymmetricObjectInPbs_Throws()
...
-                        tcs.TrySetException(e);
-                    }
-
-                    tcs.TrySetResult(null);
-                };
 
-                using var realm = await GetRealmAsync(config);
-
-                await tcs.Task;
+                var ex = Assert.Throws<RealmSchemaValidationException>(() => GetRealm(config));
```
LaPeste, line 331: .....🤔
I hope this is a change in the way the exception is thrown and not something that never worked before.

LaPeste (commented):
It was a long one. Overall it seems to makes sense to me, I just had a few comments and questions.

Tests/Realm.Tests/Sync/AsymmetricObjectTests.cs
papafe: I think it's something due to the place where the exception is thrown from, it's a change in core I suppose

Tests/Realm.Tests/Database/RealmValueTests.cs
```diff
@@ -402,9 +411,9 @@ private void RunNumericTests(RealmValue rv, long value, bool isManaged)
 
             Assert.That(rv.Type, Is.EqualTo(RealmValueType.Object));
 
-            Assert.That((RealmObjectBase)rv, Is.EqualTo(value));
-            Assert.That(rv.As<RealmObjectBase>(), Is.EqualTo(value));
-            Assert.That(rv.AsRealmObject(), Is.EqualTo(value));
+            Assert.That(rv.AsIRealmObject(), Is.EqualTo(value));
```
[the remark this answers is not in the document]
papafe: It should, but unfortunately we can't do castings anymore, because we can't have implicit/explicit conversion to/from interfaces, so it can be done. I'll remove this line, given somethign similar is happening some line down

Tests/Realm.Tests/Realm.Tests.csproj
```diff
@@ -1,149 +1,159 @@
...
+    <ItemGroup Condition="'$(UnityBuild)' != 'true'">
+        <PackageReference Include="NUnit" Version="3.13.2" />
+        <PackageReference Include="NUnitLite" Version="3.13.2" />
+        <PackageReference Include="NUnit3TestAdapter" Version="4.2.1"
+          Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' OR '$(TargetFrameworkIdentifier)' == '.NETCoreApp'" />
+        <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.1" />
+        <!-- Removing the generated files from compilation, adding them back as non-content files. -->
+        <Compile Remove="$(CompilerGeneratedFilesOutputPath)/**/*.cs" />
+        <None Include="$(CompilerGeneratedFilesOutputPath)/**/*.cs" />
```
nirinchev: This can probably be done unconditionally, not only when `UnityBuild` is false.

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
```diff
@@ -28,6 +28,7 @@ namespace Realms.Tests.SourceGeneration.TestClasses
 {
 #pragma warning disable CA1507 // Use nameof to express symbol names
 
+    [Ignored]
     public partial class ManualllyGeneratedClass : IRealmObject
```
nirinchev, line 32: Do we need this class anymore?

Tests/Realm.Tests/Sync/MergeByPKTests.cs
```diff
@@ -0,0 +1,93 @@
+﻿////////////////////////////////////////////////////////////////////////////
```
nirinchev, line 1: What happened to this file?

Tests/SourceGenerators/Realm.SourceGenerator.Tests/Realm.SourceGenerator.Tests.csproj
```diff
@@ -5,7 +5,9 @@
 		<ImplicitUsings>true</ImplicitUsings>
 		<Nullable>disable</Nullable>
 		<IsPackable>false</IsPackable>
```
nirinchev, line 7: Mixed spacing - let's convert to spaces only.

Tests/Tests.iOS/Tests.iOS.csproj
```diff
@@ -127,6 +127,7 @@
     <Reference Include="System" />
     <Reference Include="System.Xml" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Runtime.Serialization" />
```
nirinchev, line 130: I'm a bit worried about this - if our generated code doesn't compile on a default project, it's going to be annoying.

nirinchev (changes requested):

Tests/Realm.Tests/SourceGeneration/TestClasses/ManuallyGeneratedClass.cs
papafe: It was useful to test the proof of concept. It has not been updated so probably we can remove it

Tests/Realm.Tests/Realm.Tests.csproj
```diff
@@ -1,149 +1,159 @@
...
+    </ItemGroup>
 
-    <!--
+    <ItemGroup Condition="'$(UnityBuild)' == 'true'">
+        <Reference Include="nunit.framework">
+            <HintPath>..\..\dependencies\com.unity.ext.nunit@1.0.6\nunit.framework.dll</HintPath>
+        </Reference>
+
+        <!--
     In the Unity Tests project, we're referincing Realm.dll that has bundled dependencies in it. We should reference
```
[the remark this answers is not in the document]
papafe: It has been formatted like that by Visual Studio formatter, so maybe it's a standard practice? Same for the other comments

.github/workflows/test-xamarinmacos.yml
```diff
@@ -0,0 +1,74 @@
+name: test-xamarinmacos
```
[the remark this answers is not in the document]
papafe: Bad merge

Tests/Tests.iOS/Tests.iOS.csproj
nirinchev: What was the resolution here - did it turn out we don't need it or?

papafe: It seems this is not an issue anymore. I suppose that I was doing something wrong with the packaging before, as the project is building without issues now

PR #3043, opened 2022-10-04
Validate that deleting an in-memory realm does not throw if properly disposed.

### LINKED ISSUE #2808: Disposing a frozen Realm doesn't release resources
A user on the forums [has reported](https://www.mongodb.com/community/forums/t/frozen-objects-unit-testing/146442/3) that disposing a frozen realm doesn't appear to be properly closing the Realm, resulting in an exception when deleting the file. We should repro this and investigate what is holding the file.

elle-j:
## Description

Fixes #2808 but is likely not a bug; but rather, incorrect usage by the user. The user (see [conversation](https://www.mongodb.com/community/forums/t/frozen-objects-unit-testing/146442)) could not delete a realm containing frozen objects because the realm was still opened.

It is unclear how the user was using the realm but it seems like the user was either (a) disposing of the object's frozen realm or (b) disposing of the live realm before deleting the realm. The user should instead dispose of **both** prior to deleting.

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,7 @@
 * If you set a subscription on a link in flexible sync, the server would not know how to handle it ([#5409](https://github.com/realm/realm-core/issues/5409), since v11.6.1)
 * If a case insensitive query searched for a string including an 4-byte UTF8 character, the program would crash. (Core upgrade)
 * Added validation to prevent adding a removed object using Realm.Add. (Issue [#3020](https://github.com/realm/realm-dotnet/issues/3020))
+* Added test to validate that deleting a realm containing frozen objects does not throw if the frozen object's realm and the live realm have been disposed. (Issue [#2808](https://github.com/realm/realm-dotnet/issues/2808))
 
 ### Compatibility
 * Realm Studio: 12.0.0 or later.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -90,6 +90,7 @@
 * If you set a subscription on a link in flexible sync, the server would not know how to handle it ([#5409](https://github.com/realm/realm-core/issues/5409), since v11.6.1)
 * If a case insensitive query searched for a string including an 4-byte UTF8 character, the program would crash. (Core upgrade)
 * Added validation to prevent adding a removed object using Realm.Add. (Issue [#3020](https://github.com/realm/realm-dotnet/issues/3020))
+* Added test to validate that deleting a realm containing frozen objects does not throw if the frozen object's realm and the live realm have been disposed. (Issue [#2808](https://github.com/realm/realm-dotnet/issues/2808))
```
papafe, line 93: We don't add info about new tests to the changelog. We use it for user-facing changes, so I would remove this

LaPeste (commented):
The test looks fine, but what's the reason for the test not to be done for each type of configuration?

nirinchev (approved):

PR #3046, opened 2022-10-06
Fix crash in RealmObject finalizer due to null _accessor.

### LINKED ISSUE #3045: [Bug]: Crash in finaliser if `RealmObject` implementation throws in object initialisation
### What happened?

Regressed in https://github.com/realm/realm-dotnet/commit/2d2f5957e9b96f431b039c56688fbd530003752c#diff-ffe0558165702693f2ba96b95df2b7f6bf559ca7a2ef344e9cab6af24f75f6ae (due to missing null check).

This is a bit of a caveat with NRT and pre-constructor time crashes. We get around similar issues by using a helper method to check for null (https://github.com/ppy/osu-framework/blob/7fea93e6f5a30d5a47e2bd609fe1cf7ef45c507d/osu.Framework/Extensions/ObjectExtensions/ObjectExtensions.cs#L30-L33, usage: https://github.com/ppy/osu-framework/blob/3a9acd01eac691fcbea0163038743a05bb0ba41d/osu.Framework/IO/Stores/RawCachingGlyphStore.cs#L178). Another way would be to make the `_accessor` nullable

With the sample code included below, a crash is observed which can halt an application that is not manually handling all unobserved exceptions:

https://user-images.githubusercontent.com/191335/194010081-dab08fab-ca10-4720-9a29-ecc703237204.mp4

### Repro steps

- Have a class that throws an exception before reaching the constructor (ie. object initialisers).
- Observe crash on first GC/finalize operation.

### Version

10.16.0

Local Database only

Console/Server

### Client OS and version

n/a

### Code snippets

```csharp
// See https://aka.ms/new-console-template for more information

using System.Diagnostics;
using Realms;

while (true)
{
    try
    {
        var model = new TestModel();
    }
    catch (Exception e)
    {
        Console.WriteLine("Caught exception");
    }
}

class TestModel : RealmObject
{
    public string Test { get; set; }

    public object ThisWillThrowBeforeConstruction = new ThrowingTest();

    internal class ThrowingTest
    {
        public ThrowingTest()
        {
            throw new Exception("die in ctor");
        }
    }
}
```

```shell
System.NullReferenceException: Object reference not set to an instance of an object.
   at Realms.RealmObjectBase.UnsubscribeFromNotifications()
   at Realms.RealmObjectBase.Finalize()
```

### Relevant log output

_No response_

elle-j:
## Description

Fixes #3045.

When creating a `RealmObject` that throws before being initialized, `_accessor` will be `null`. When the object was garbage collected, `RealmObject`'s destructor would call `_accessor.UnsubscribeFromNotifications()` causing a crash due to `NullReferenceException`.

Solution:
`RealmObject`'s destructor was removed and instead added the call to `UnsubscribeFromNotifications()` in `ManagedAccessor`'s destructor. (For an `UnmanagedAccessor`, a call to `UnsubscribeFromNotifications()` does nothing.)

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,7 @@
 * If you set a subscription on a link in flexible sync, the server would not know how to handle it ([#5409](https://github.com/realm/realm-core/issues/5409), since v11.6.1)
 * If a case insensitive query searched for a string including an 4-byte UTF8 character, the program would crash. (Core upgrade)
 * Added validation to prevent adding a removed object using Realm.Add. (Issue [#3020](https://github.com/realm/realm-dotnet/issues/3020))
+* Fixed a NullReferenceException occurring in RealmObject's finalizer whenever an exception is thrown before the object gets initialized. (Issue [#3045](https://github.com/realm/realm-dotnet/issues/3045))
 
 ### Compatibility
 * Realm Studio: 12.0.0 or later.
--- a/Realm/Realm/DatabaseTypes/Accessors/ManagedAccessor.cs
+++ b/Realm/Realm/DatabaseTypes/Accessors/ManagedAccessor.cs
@@ -71,6 +71,11 @@ namespace Realms
             _hashCode = new Lazy<int>(() => _objectHandle.GetObjHash());
         }
 
+        ~ManagedAccessor()
+        {
+            UnsubscribeFromNotifications();
+        }
+
         public RealmValue GetValue(string propertyName)
         {
             return _objectHandle.GetValue(propertyName, _metadata, _realm);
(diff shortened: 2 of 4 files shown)


peppy:
Thanks for the fix, it looks solid.

As an aside, this is also really cool to see – finalizers are not zero overhead and we've seen performance degradation from having thousands-millions of objects with finalizers. This was unavoidable with `RealmObject` derived models, but looks like it may become a reality with this change (for cases of `UnmanagedAccessor`, which is beneficial to us as we use the same model in non-managed batch processing).

nirinchev (approved):

elle-j:
Thank you @peppy for providing the necessary details in your ticket so that we could provide a fix quicker!

PR #3062, opened 2022-10-20
Cleanup trigger client reset function

nirinchev:
## Description

Just a little bit of tweaking - I noticed there was a typo - `useId` vs `userId` which is why we weren't logging the user id (even though the actual deletion was succeeding). I also reshuffled the code a little bit to make it shorter.

### DIFF
--- a/Tools/DeployApps/BaasClient.cs
+++ b/Tools/DeployApps/BaasClient.cs
@@ -76,25 +76,22 @@ namespace Baas
                 };";
 
         private const string TriggerClientResetOnSyncServerFuncSource =
-            @"exports = async function(useId, appId = '') {
+            @"exports = async function(userId, appId = '') {
                 const mongodb = context.services.get('BackingDB');
                 console.log('user.id: ' + context.user.id);
-                let deletionResult;
                 try {
                   let dbName = '__realm_sync';
                   if (appId !== '')
                   {
                     dbName = [dbName, '_', appId].join('');
                   }
-                  deletionResult = await mongodb.db(dbName).collection('clientfiles').deleteMany({ ownerId: useId });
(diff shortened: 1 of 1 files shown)


Tools/DeployApps/BaasClient.cs
```diff
@@ -76,25 +76,22 @@ public enum Result
                 };";
 
         private const string TriggerClientResetOnSyncServerFuncSource =
-            @"exports = async function(useId, appId = '') {
```
LaPeste, line 79: 🤦🏻

Tools/DeployApps/BaasClient.cs
```diff
@@ -76,25 +76,22 @@ public enum Result
...
                   }
-                  deletionResult = await mongodb.db(dbName).collection('clientfiles').deleteMany({ ownerId: useId });
+                  const deletionResult = await mongodb.db(dbName).collection('clientfiles').deleteMany({ ownerId: userId });
                   console.log('Deleted documents: ' + deletionResult.deletedCount);
+
+                  return { status: deletionResult.deletedCount > 0 ? 'success' : 'failure' };
                 } catch(err) {
                   throw 'Deletion failed: ' + err;
                 }
-                if (deletionResult.deletedCount > 0) {
```
LaPeste, line 94: How do we know if anything was deleted? If we don't care, why?

nirinchev: I just moved the return inside the `try` block:

```
return { status: deletionResult.deletedCount > 0 ? 'success' : 'failure' };
```

LaPeste: eh... completely blind... Sorry

LaPeste (approved):

Issue #2815, opened 2022-02-15
Don't snapshot frozen results


nirinchev:
We are not gaining anything, so we should avoid snapshotting frozen collections.

PR #3067, opened 2022-10-25
Stop snapshotting during iteration on frozen collections

gagik:
## Description
Removes the snapshotting step if the collection is frozen and adds a basic test for that.
Fixes https://github.com/realm/realm-dotnet/issues/2815

##  TODO

* [X] Changelog entry
* [X] Tests (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 ### Fixed
 * Fixed a NullReferenceException being thrown when subscribing to `PropertyChanged` notifications on a `Session` instance that is then garbage collected prior to unsubscribing. (PR [#3061](https://github.com/realm/realm-dotnet/pull/3061))
 * Removed bitcode support from the iOS binary as it's no longer accepted for App Store submissions. (Issue [#3059](https://github.com/realm/realm-dotnet/issues/3059))
-* Slightly increased performance and reduced allocations when creating an enumerator for frozen collections (PR [#3067](https://github.com/realm/realm-dotnet/pull/3067)).
+* Slightly increased performance and reduced allocations when creating an enumerator for frozen collections (Issue [#2815](https://github.com/realm/realm-dotnet/issues/2815)).
 
 ### Compatibility
 * Realm Studio: 11.0.0 or later.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Fixed
 * Fixed a NullReferenceException being thrown when subscribing to `PropertyChanged` notifications on a `Session` instance that is then garbage collected prior to unsubscribing. (PR [#3061](https://github.com/realm/realm-dotnet/pull/3061))
 * Removed bitcode support from the iOS binary as it's no longer accepted for App Store submissions. (Issue [#3059](https://github.com/realm/realm-dotnet/issues/3059))
+* Slightly increased performance and reduced allocations when creating an enumerator for frozen collections (PR [#3067](https://github.com/realm/realm-dotnet/pull/3067)).
 
 ### Compatibility
 * Realm Studio: 11.0.0 or later.
--- a/Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
+++ b/Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
@@ -487,17 +487,10 @@ namespace Realms
                 _index = -1;
 
                 // If we didn't snapshot the parent, we should not dispose the results handle, otherwise we'll invalidate the
-                // parent collection after iterating it. Only collections of objects support snapshotting.
+                // parent collection after iterating it. Only collections of objects support snapshotting and we do not need to
+                // snapshot if the collection is frozen.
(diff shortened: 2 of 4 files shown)


Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -488,8 +488,16 @@ internal Enumerator(RealmCollectionBase<T> parent)
 
                 // If we didn't snapshot the parent, we should not dispose the results handle, otherwise we'll invalidate the
                 // parent collection after iterating it. Only collections of objects support snapshotting.
-                _shouldDisposeHandle = parent.IsValid && parent.Handle.Value.CanSnapshot && parent.Metadata != null;
-                _enumerating = _shouldDisposeHandle ? new RealmResults<T>(parent.Realm, parent.Handle.Value.Snapshot(), parent.Metadata) : parent;
+                _shouldDisposeHandle = parent.IsValid && !parent.IsFrozen && parent.Handle.Value.CanSnapshot && parent.Metadata != null;
```
gagik, line 491: I am unsure whether adding the `&& !parent.IsFrozen` condition is correct or necessary here

Tests/Realm.Tests/Database/CollectionTests.cs
```diff
@@ -247,6 +247,37 @@ public void ObjectList_WhenEnumeratingAndRemovingFromList_ShouldBeStable()
...
+
+            TestStableIteration(
+                i => container.Items.Add(new IntPropertyObject { Int = i }),
+                () => frozenList = container.Items.Freeze(),
+                item =>
+                {
+                    // Just remove last from container since item from
+                    // frozen version is not equivalent.
+                    j -= 1;
+                    container.Items.RemoveAt(j);
```
gagik, line 270: I had trouble coming up with a meaningful test for this since I don't think you are meant to be removing things from frozen collections... so this is a somewhat weird experiment that mixes frozen collection with original collection (not that the two are affecting each other anyways).

Tests/Realm.Tests/Sync/AsymmetricObjectTests.cs
```diff
@@ -341,7 +341,8 @@ public void EmbeddedObject_WhenParentAccessed_ReturnsParent()
             SyncTestHelpers.RunBaasTestAsync(async () =>
             {
                 var flxConfig = await GetFLXIntegrationConfigAsync();
-                flxConfig.Schema = new[] {
+                flxConfig.Schema = new[]
+                {
```
gagik, line 345: I also figured I should fix this since the warning annoyed me

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -488,8 +488,16 @@ internal Enumerator(RealmCollectionBase<T> parent)
...
+
+                if (parent.IsFrozen)
```
nirinchev, line 493: Do we need this check here? Since `_shouldDisposeHandle = ... && !parent.IsFrozen`, I would expect `_shouldDisposeHandle` to be `false` for frozen collections, which means that the ternary on line 499 would have been sufficient.

nirinchev (commented):
Good stuff - a minor cleanup suggestion from me and we need a changelog entry. Since snapshotting is not something we expose to users directly, it should be something vague like "Slightly increased performance and reduced allocations when creating an enumerator for frozen collections".

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
gagik: that's a great point, I will remove this and perhaps add a comment specifying the isFrozen bit

Tests/Realm.Tests/Sync/AsymmetricObjectTests.cs
LaPeste: This was also fixed by @elle-j in [#3064](https://github.com/realm/realm-dotnet/pull/3064/files)

gagik: thanks for the heads up, I can remove this from my change to avoid unnecessary merge conflict

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 ### Fixed
 * Fixed a NullReferenceException being thrown when subscribing to `PropertyChanged` notifications on a `Session` instance that is then garbage collected prior to unsubscribing. (PR [#3061](https://github.com/realm/realm-dotnet/pull/3061))
 * Removed bitcode support from the iOS binary as it's no longer accepted for App Store submissions. (Issue [#3059](https://github.com/realm/realm-dotnet/issues/3059))
+* Slightly increased performance and reduced allocations when creating an enumerator for frozen collections (PR [#3067](https://github.com/realm/realm-dotnet/pull/3067)).
```
nirinchev, line 9: We prefer to list issues in the changelog rather than PRs.

```suggestion
* Slightly increased performance and reduced allocations when creating an enumerator for frozen collections (Issue [#2815](https://github.com/realm/realm-dotnet/issues/2815)).
```

nirinchev (approved):

Issue #2555, opened 2021-08-13
Make RealmSet queryable


nirinchev:
### Description

Similar to https://github.com/realm/realm-dotnet/issues/1499 but for `ISet`.
See https://github.com/realm/realm-dotnet/issues/2647 for dictionary.

### How important is this improvement for you?
Fairly niche, but nice to have anyway

carbonete:
It's useful and create facility. At #1499 example this have [IGNORED] and will calculate at runtime, but it's possible result write and  indexed to global query of all objects in this collection ?

I have a new doubt after read this. It's possible to have like a static class properties in Realm Object  ? 
I have a case that i need  create many to many relation , where the second class is only a string indexed. To determine if exist in a list of first class without need to read all objects and interact through list . They can be exist in more than one object of first class.

Bsr
Carbonete

nirinchev:
I'm sorry, it's really difficult to understand what you mean. Can you try rewording it, possibly with code snippets so that we can see what you're trying to achieve.

DominicFrei:
Related issues / PRs:
- https://github.com/realm/realm-dotnet/pull/2554
- https://github.com/realm/realm-dotnet/issues/2555

carbonete:
> I'm sorry, it's really difficult to understand what you mean. Can you try rewording it, possibly with code snippets so that we can see what you're trying to achieve.

Sorry for my bad communication.
Maybe this code clarifying.

    public class FavoriteList : RealmObject
    {
       public static IDictionay<FavoritedVerbete, List<FavoriteList.Name>> AllFavoritedVerbetes; /* only illustratoin this will valid for all objects _partitionKey to sync maybe a challenger*/

        [PrimaryKey]
        public string _id { get; set; } = Guid.NewGuid().ToString("N");

        [Required]
        public string _partitionKey { get; set; } /* user Id */

        [Required]
        [Indexed]
        public string Name { get; set; }

        [Required]
        public string Description { get; set; }

        public IList<string> FavoritedVerbetes { get; }

        #region Constructor
        public FavoriteList( string partition, string name)
        {
            this._partitionKey = partition;
            this.Name = name;
        }
      
public Favorite(string) 
{
add to FavoritedVerbetes and to AllFavoriteVerbetes
}
        public FavoriteList() { }
        #endregion

    }

In substituition of 

    public class FavoriteList : RealmObject
    {
        [PrimaryKey]
        public string _id { get; set; } = Guid.NewGuid().ToString("N");

[...]

Issue #1499, opened 2017-08-03
Make RealmList queryable


tom-pratt:
## Goals
To be able to do this and ActiveSelections returns a live collection

```
public class Market : RealmObject
{
    public IList<Selection> Selections { get; }

    [Ignored]
    public IRealmCollection<Selection> ActiveSelections => Selections.Where(s => s.Active);
}
```

nirinchev:
Note: to achieve it, we'll likely need to create a custom `IQueryableList` or similar interface.

nirinchev:
This was done in #2554 and will be released with the next Realm .NET version. The syntax is slightly modified compared to the original proposal in order to avoid introducing a new collection type:


```csharp
public class Market : RealmObject
{
    public IList<Selection> Selections { get; }

    [Ignored]
    public IRealmCollection<Selection> ActiveSelections => Selections.AsRealmQueryable()
                                                                     .Where(s => s.Active)
                                                                     .AsRealmCollection();
}
```

PR #3068, opened 2022-10-26
Make dictionaries actively queryable

gagik:
## Description
Adds the ability to use `asRealmQueryable` to get a live `IQueryable` object from a Realm dictionary.
Related to https://github.com/realm/realm-dotnet/issues/2555 for sets and https://github.com/realm/realm-dotnet/issues/1499 for lists.
Fixes https://github.com/realm/realm-dotnet/issues/2647

##  TODO

* [ ] Changelog entry
* [X] Tests (if applicable)


Realm/Realm/DatabaseTypes/RealmDictionary.cs
```diff
@@ -98,6 +98,12 @@ public ICollection<TValue> Values
             }
         }
 
+        internal RealmResults<TValue> ToResults()
+        {
+            var resultsHandle = _dictionaryHandle.GetValues();
+            return new RealmResults<TValue>(Realm, resultsHandle, Metadata);
```
gagik, line 104: it could have been nice to just do i.e. `this.Values` but that would involve double-casting to go back to `RealmResults` from `Collection`, I am not sure if that's a good idea

Realm/Realm/Extensions/CollectionExtensions.cs
```diff
@@ -267,6 +267,42 @@ public static void Move<T>(this IList<T> list, int from, int to)
             return realmDictionary;
         }
 
+        /// <summary>
+        /// Converts a Realm-backed <see cref="IDictionary{String, T}"/> to a Realm-backed <see cref="IQueryable{T}"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of the values contained in the dictionary.</typeparam>
+        /// <param name="dictionary">The dictionary of objects as obtained from a to-many relationship property.</param>
```
gagik, line 274: I don't quite understand what "obtained from a to-many relationship property" means but I kept it here, not sure if that is correct

papafe: Originally this was written for properties like `IList<Dogs`, where this actually represent a to-many relationship between a person and his dogs. I would say that this is still a to-many relationship. so it could make sense to keep it.

Tests/Realm.Tests/Database/RealmDictionaryTests.cs
```diff
@@ -1073,6 +1073,56 @@ public void ObjectFromAnotherRealm_ThrowsRealmException()
...
 
+        [Test]
+        public void AsRealmQueryable_RaisesNotifications()
+        {
+            var obj = _realm.Write(() =>
+            {
+                return _realm.Add(new DictionariesObject());
+            });
+
+            var dogToAge = obj.ObjectDictionary;
```
papafe, line 1084: It's a very small thing, and I understand why you did it, but it looks a little strange to see variables called `dogToAge` and `oldDogs` when there's not property called like that 😁 .
I think it would make sense to have more appropriate names

Tests/Realm.Tests/Database/RealmDictionaryTests.cs
```diff
@@ -1073,6 +1073,56 @@ public void ObjectFromAnotherRealm_ThrowsRealmException()
...
+                if (i >= 5)
+                {
+                    Assert.That(changeSets.Count, Is.EqualTo(i - 4));
+
+                    var changeSet = changeSets.Last();
+                    Assert.That(changeSet.InsertedIndices.Length, Is.EqualTo(1));
+                    Assert.That(changeSet.DeletedIndices, Is.Empty);
+                    Assert.That(changeSet.ModifiedIndices, Is.Empty);
+
+                    Assert.That(oldDogs.ElementAt(changeSet.InsertedIndices[0]).Int, Is.EqualTo(i));
```
papafe, line 1111: Probably we should add some tests also for all the `changeSet` property (deleted, modified, newModified,...). It doesn't need to be in the same loop though

gagik: could you elaborate what you mean by tests for `changeSet`?

gagik: got it, this is basically line-for-line copy of the tests for sets and lists, so I imagine it'd be useful for this to be there too then.

papafe: I suppose the reason why it's not heavily tested is the fact that we have tests for "RealmResults" in other places too, and we already expect the notifications to work properly. So this is just to be sure that the new method works property. On a second though, we could probably do the same here, and just keep it as it was

Tests/Realm.Tests/Database/RealmDictionaryTests.cs
```diff
@@ -1073,6 +1073,260 @@ public void ObjectFromAnotherRealm_ThrowsRealmException()
...
+            {
+                return _realm.Add(new DogDictionaryObject());
+            });
+
+            var dogDict = obj.Dict;
+            var fiDogs = dogDict.Filter("Name BEGINSWITH[c] $0", "Fi");
+
+            _realm.Write(() =>
+            {
+                dogDict.Add("Ris", new Dog { Name = "Rick" });
```
gagik: I am using different keys just to make sure there's no overlap between keys and values in functionality

wrappers/src/dictionary_cs.cpp
```diff
@@ -226,4 +227,11 @@ extern "C" {
         });
     }
 
+    REALM_EXPORT Results* realm_dictionary_get_filtered_results(const object_store::Dictionary& dictionary, uint16_t* query_buf, size_t query_len, realm_value_t* arguments, size_t args_count, NativeException::Marshallable& ex)
+    {
+        return handle_errors(ex, [&]() {
+            realm::Results values = dictionary.get_values();
```
gagik, line 233: it's safe to refer to `get_values()` directly right? No need to create a new Ream Results from this?

gagik (commented):
Some things I am not sure about

Tests/Realm.Tests/Database/RealmDictionaryTests.cs
```diff
@@ -1073,6 +1073,235 @@ public void ObjectFromAnotherRealm_ThrowsRealmException()
...
+        }
+
+        [Test]
+        public void DictionaryAsRealmQueryable_WhenNotRealmDictionary_Throws()
+        {
+            var dict = new DictionariesObject().ObjectDictionary;
+
+            Assert.That(
+                () => dict.AsRealmQueryable(),
+                Throws.TypeOf<ArgumentException>().And.Message.Contains("dictionary must be an instance of RealmDictionary<IntPropertyObject>"));
```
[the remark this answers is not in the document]
papafe: Yes, the "nameof" is getting the name of the variable that is used inside the function. The advantage here over using just a string is that if you do a rename this will be changed too

Tests/Realm.Tests/Database/RealmDictionaryTests.cs
```diff
@@ -1073,6 +1073,235 @@ public void ObjectFromAnotherRealm_ThrowsRealmException()
             Assert.That(() => realm2.Write(() => obj2.EmbeddedObjectDictionary["foo"] = embeddedItem), Throws.TypeOf<RealmException>().And.Message.Contains("embedded object that is already managed"));
         }
 
+        public class DogDictionaryObject : RealmObject
+        {
+            public IDictionary<string, Dog> Dict { get; }
+        }
+
+        [Test]
+        public void DictionaryAsRealmQueryable_RaisesNotifications()
```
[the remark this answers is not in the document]
papafe: I am not sure what you mean by stuttering here

CHANGELOG.md
```diff
@@ -634,6 +634,11 @@ If for some reason, you want to opt out of the fixed behavior, you can temporari
...
 * Query parser now accepts `BETWEEN` operator. Can be used like `realm.All<Person>().Filter("Age BETWEEN {20, 60}")` which means "'Age' must be in the open interval ]20;60[". (Core upgrade)
+* Added two extension methods on `IDictionary` to get an `IQueryable` collection wrapping the dictionary:
+  * `dictionary.AsRealmQueryable()` allows you to get a `IQueryable<T>` from `IDictionary<string, T>` that can be then treated as a regular queryable collection and filtered/ordered with LINQ or `Filter(string)`.
+  * `dictionary.Filter(query, arguments)` will filter the list and return the filtered collection. It is roughly equivalent to `dictionary.AsRealmQueryable().Filter(query, arguments)`.
+
+  Note that the resulting queryable collection is tied to the dictionary's **values** not key-value pairs. For those values, it will behave similarly to a live Realm Results object i.e. it will emit notifications when it changes and automatically update itself. (Issue [#2647](https://github.com/realm/realm-dotnet/issues/2647))
```
gagik, line 641: Not sure if there is a better way to put this in words.

Tests/Realm.Tests/Database/RealmDictionaryTests.cs
```diff
@@ -1073,6 +1073,235 @@ public void ObjectFromAnotherRealm_ThrowsRealmException()
             Assert.That(() => realm2.Write(() => obj2.EmbeddedObjectDictionary["foo"] = embeddedItem), Throws.TypeOf<RealmException>().And.Message.Contains("embedded object that is already managed"));
         }
 
+        public class DogDictionaryObject : RealmObject
```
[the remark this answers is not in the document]
papafe: I think that you could maybe add the dictionary to the `Owner` class in `TestObjects.cs`, so it's consistent

wrappers/src/dictionary_cs.cpp
papafe: @nirinchev what do you think about this?

CHANGELOG.md
papafe: I think it mostly makes sense. I would remove the line about the fact that this is about values, and include that info already in the "main text" (for example: "collection wrapping the dictionary values" and later too). 
Also, I think talking "Realm Results" is a little confusing, because that's an internal name. You can see how is the changelog for the sets equivalent., where it says that it acts in a similar way to `realm.All`

Realm/Realm/Extensions/CollectionExtensions.cs
```diff
@@ -267,6 +267,41 @@ public static void Move<T>(this IList<T> list, int from, int to)
...
+        /// <returns>A queryable collection that represents the values contained in the dictionary.</returns>
+        /// <remarks>
+        /// This method works differently from <see cref="Queryable.AsQueryable"/> in that it only returns a collection of values,
+        /// not a collection of `KeyPair` and it actually creates an underlying Realm query that represents the dictionary's values.
```
papafe, line 278: You probably need to use a cref on KeyPair

papafe (changes requested):

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -164,6 +164,12 @@ internal RealmResults<T> GetFilteredResults(string query, RealmValue[] arguments
             return new RealmResults<T>(Realm, resultsHandle, Metadata);
         }
 
+        internal RealmResults<TValue> GetFilteredResults<TValue>(string query, RealmValue[] arguments)
```
papafe, line 167: I don't think it's a good idea to have this here in the base class, because it's not applicable to the other collections (list, sets)

Realm/Realm/Extensions/CollectionExtensions.cs
```diff
@@ -267,6 +267,41 @@ public static void Move<T>(this IList<T> list, int from, int to)
             return realmDictionary;
         }
 
+        /// <summary>
+        /// Converts a Realm-backed <see cref="IDictionary{String, T}"/> to a Realm-backed <see cref="IQueryable{T}"/>.
```
papafe, line 271: I would put it in the summary too that IQueryable is of the values

Tests/Realm.Tests/Database/RealmDictionaryTests.cs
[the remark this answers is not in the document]
papafe: I think the naming is ok, I wouldn't worry too much

papafe (changes requested):

Realm/Realm/DatabaseTypes/RealmDictionary.cs
```diff
@@ -98,6 +98,19 @@ public ICollection<TValue> Values
...
+        }
```
nirinchev, line 105: This is the same as `Values` - we could just use that and cast to `RealmResults<T>`.

gagik: I did have a comment mentioning this, is this weird casting to Collection and then casting to Results again a good idea?

nirinchev: Yeah, I think it should be fine - as long as we have tests for it, any refactoring that breaks assumptions will immediately show up as a failed test.

Realm/Realm/DatabaseTypes/RealmDictionary.cs
```diff
@@ -98,6 +98,18 @@
             }
         }
 
+        internal RealmResults<TValue> ToResults()
+        {
+            return (RealmResults<TValue>)this.Values;
```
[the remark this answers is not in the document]
gagik: not sure if this is an issue @nirinchev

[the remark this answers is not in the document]
nirinchev: No, we can dismiss it - we know better than the compiler here :)

wrappers/src/dictionary_cs.cpp
nirinchev: Yeah, don't see a reason to allocate it on the heap.

Realm/Realm/DatabaseTypes/RealmDictionary.cs
```diff
@@ -98,6 +98,18 @@ public ICollection<TValue> Values
             }
         }
 
+        internal RealmResults<TValue> ToResults()
+        {
+            return (RealmResults<TValue>)Values;
+        }
+
+        // Get filtered results from dictionary's values
+        internal RealmResults<TValue> GetFilteredValueResults(string query, RealmValue[] arguments)
```
gagik, line 107: I think moving it here and renaming it slightly was most straightforward way to make this dictionary specific and not confusing in terms of overriding `GetFilteredResults` from base class but perhaps there's better naming/ approach.

papafe (approved):
Looks good for me! You should merge from main and fix the changelog though

CHANGELOG.md
```diff
@@ -634,6 +634,11 @@ If for some reason, you want to opt out of the fixed behavior, you can temporari
 * Improve performance of creating collection notifiers for Realms with a complex schema. In the SDKs this means that the first run of a synchronous query, first call to subscribe for notifications will do significantly less work on the calling thread.
 * Improve performance of calculating changesets for notifications, particularly for deeply nested object graphs and objects which have List or Set properties with small numbers of objects in the collection.
 * Query parser now accepts `BETWEEN` operator. Can be used like `realm.All<Person>().Filter("Age BETWEEN {20, 60}")` which means "'Age' must be in the open interval ]20;60[". (Core upgrade)
+* Added two extension methods on `IDictionary` to get an `IQueryable` collection wrapping the dictionary's values:
```
papafe, line 637: All of this makes sense, but you need to move this to the top of the changelog.

Issue #2233, opened 2021-02-08
Refactor RemoveAll method


papafe:
The `RemoveAll` method in `Realm.cs` looks like this at the moment:
```
public void RemoveAll()
{
    ThrowIfDisposed();

    foreach (var metadata in Metadata.Values)
    {
        using var resultsHandle = SharedRealmHandle.CreateResults(metadata.TableKey);
        resultsHandle.Clear(SharedRealmHandle);
    }
}
```
It would be best to implement the method natively. The current implementation could cause some surprise if a user opens a `realm` with only a subset of the schema, and then `RemoveAll` doesn't leave the database file empty. As a nice secondary effect this should also speed up this method

PR #3072, opened 2022-10-31
Add native removeAll implementation

gagik:
## Description
Uses a single native call to remove all objects from Realm through `Realm.RemoveAll()`
Fixes https://github.com/realm/realm-dotnet/issues/2233

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,7 +80,6 @@ In order to use the source generation the model classes need to be declared impl
 * Seamlessly handle migrating an App Services application deployment model. (Core upgrade)
 * Slightly improve performance of `Realm.RemoveAll()` which removes all objects from an open Realm database. (Issue [#2233](https://github.com/realm/realm-dotnet/issues/2194))
 
-
 ### Fixed
 * Fix a use-after-free when a sync session is closed and the app is destroyed at the same time. (Core upgrade)
 * Fixed a `NullReferenceException` occurring in `RealmObjectBase`'s finalizer whenever an exception is thrown before the object gets initialized. (Issue [#3045](https://github.com/realm/realm-dotnet/issues/3045))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## vNext (TBD)
 
 ### Enhancements
-* Slightly improve performance of `Realm.RemoveAll()` which removes all objects from an open Realm database. (Issue [#2233](https://github.com/realm/realm-dotnet/issues/2194))
 * Added two extension methods on `IDictionary` to get an `IQueryable` collection wrapping the dictionary's values:
   * `dictionary.AsRealmQueryable()` allows you to get a `IQueryable<T>` from `IDictionary<string, T>` that can be then treated as a regular queryable collection and filtered/ordered with LINQ or `Filter(string)`.
   * `dictionary.Filter(query, arguments)` will filter the list and return a filtered collection of dictionary's values. It is roughly equivalent to `dictionary.AsRealmQueryable().Filter(query, arguments)`.
@@ -9,6 +8,7 @@
   The resulting queryable collection will behave identically to the results obtained by calling `realm.All<T>()`, i.e. it will emit notifications when it changes and automatically update itself. (Issue [#2647](https://github.com/realm/realm-dotnet/issues/2647))
 * Improve performance of client reset with automatic recovery and converting top-level tables into embedded tables. (Core upgrade)
 * Flexible sync will now wait for the server to have sent all pending history after a bootstrap before marking a subscription as Complete. (Core upgrade)
+* Slightly improve performance of `Realm.RemoveAll()` which removes all objects from an open Realm database. (Issue [#2233](https://github.com/realm/realm-dotnet/issues/2194))
 
 ### Fixed
 * Prevented `IEmbeddedObject`s and `IAsymmetricObject`s from being used as `RealmValue`s when added to a realm, and displaying more meaningful error messages.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## vNext (TBD)
 
 ### Enhancements
+* Slightly improve performance of `Realm.RemoveAll()` which removes all objects from an open Realm database. (Issue [#2233](https://github.com/realm/realm-dotnet/issues/2194))
 * Added two extension methods on `IDictionary` to get an `IQueryable` collection wrapping the dictionary's values:
   * `dictionary.AsRealmQueryable()` allows you to get a `IQueryable<T>` from `IDictionary<string, T>` that can be then treated as a regular queryable collection and filtered/ordered with LINQ or `Filter(string)`.
   * `dictionary.Filter(query, arguments)` will filter the list and return a filtered collection of dictionary's values. It is roughly equivalent to `dictionary.AsRealmQueryable().Filter(query, arguments)`.
@@ -78,7 +79,6 @@ In order to use the source generation the model classes need to be declared impl
 * Prioritize integration of local changes over remote changes - shorten the time users may have to wait when committing local changes. Stop storing downloaded changesets in history. (Core upgrade)
 * Greatly improve the performance of sorting or distincting a Dictionary's keys or values. The most expensive operation is now performed O(log N) rather than O(N log N) times, and large Dictionaries can see upwards of 99% reduction in time to sort. (Core upgrade)
 * Seamlessly handle migrating an App Services application deployment model. (Core upgrade)
-* Slightly improve performance of `Realm.RemoveAll()` which removes all objects from an open Realm database. (Issue [#2233](https://github.com/realm/realm-dotnet/issues/2194))
 
 ### Fixed
 * Fix a use-after-free when a sync session is closed and the app is destroyed at the same time. (Core upgrade)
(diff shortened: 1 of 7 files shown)


wrappers/src/shared_realm_cs.cpp
```diff
@@ -754,6 +754,19 @@ REALM_EXPORT bool shared_realm_remove_type(const SharedRealm& realm, uint16_t* t
     });
 }
 
+REALM_EXPORT bool shared_realm_remove_all(const SharedRealm& realm, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        realm->verify_in_write();
```
gagik, line 760: not sure if I should add more tests, there's `RemoveAllObjectsShouldClearTheDatabase` in theory the only thing that's not being tackled here is making sure it errors when not in write... but I do not see this tested in other functions that are expected to do.

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@
 * Prioritize integration of local changes over remote changes - shorten the time users may have to wait when committing local changes. Stop storing downloaded changesets in history. (Core upgrade)
 * Greatly improve the performance of sorting or distincting a Dictionary's keys or values. The most expensive operation is now performed O(log N) rather than O(N log N) times, and large Dictionaries can see upwards of 99% reduction in time to sort. (Core upgrade)
 * Seamlessly handle migrating an App Services application deployment model. (Core upgrade)
+* Slightly improve performance of `Realm.RemoveAll()` which removes all objects from an open Realm database. (Issue [#2233](https://github.com/realm/realm-dotnet/issues/2194))
```
gagik, line 23: let me know if I should word this differently

wrappers/src/shared_realm_cs.cpp
gagik: thank you, I was unsure how to check on this. How does one get schema from core rather than OS? Core API does not seem clear in this differentiation.

gagik: I found `get_full_schema` in core.... but it is a private method

wrappers/src/shared_realm_cs.cpp
```diff
@@ -754,6 +754,20 @@ REALM_EXPORT bool shared_realm_remove_type(const SharedRealm& realm, uint16_t* t
...
+
+        auto full_schema = ObjectStore::schema_from_group(realm->read_group());
```
gagik, line 762: This is related to [get_full_schema from core](https://github.com/realm/realm-core/blob/b86d2107fec4246b55af24a3b11e00275575f808/src/realm/object-store/shared_realm.cpp#L329) which is unfortunately a private field right now... I am not sure how many of those checks the function does is crucial beyond caching for sake of performance... Immutability check seems relevant here, but I am unsure; current implementation like this passes both basic and partial schema tests that used to not pass. @nirinchev

CHANGELOG.md
papafe: I think we should specify that we also changed the behaviour (and eventually change the text for the docs too)

Tests/Realm.Tests/Database/RemoveTests.cs
```diff
@@ -153,6 +153,38 @@ public void RemoveAllObjectsShouldClearTheDatabase()
             Assert.That(_realm.All<AllTypesObject>().Count(), Is.EqualTo(0));
         }
 
+        [Test]
+        public void RemoveAllObjectsShouldClearEntireDatabaseWithPartialSchema()
+        {
+            var c1 = new RealmConfiguration
+            {
+                Schema = new[] { typeof(Person), typeof(IntPrimaryKeyWithValueObject) }
```
gagik, line 161: Maybe not best classes to choose but unfortunately many other classes in schema reference other classes which would make it more confusing as it'd require those referenced classes to be included in schema too

wrappers/src/shared_realm_cs.cpp
papafe: I forgot that actually we have a long-standing ticket somewhere where we request that the schema at object store level should be the same that core sees. 
I am also not sure of how to do it to be honest 😁 I would need to take a look to do it. 
Another thing, we need to be sure that this works fine with Flexible Sync. So, if for example we call this method, this should only delete the objects that have been synced, and doesn't delete more on the server

nirinchev: `ObjectStore::schema_from_group(realm->read_group())` is probably an overkill for this as you don't really care about the entire schema - you only want the tables (hint: you can get those from the group).

Tests/Realm.Tests/Database/RemoveTests.cs
```diff
@@ -153,6 +153,38 @@ public void RemoveAllObjectsShouldClearTheDatabase()
...
+            };
+            var r1 = Realm.GetInstance(c1);
```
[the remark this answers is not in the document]
papafe: I think it's fine that we keep it r1 and r2. You should use `GetRealm` or similar from the base class though, as it takes care of cleaning up

Tests/Realm.Tests/Database/RemoveTests.cs
```diff
@@ -153,6 +153,36 @@ public void RemoveAllObjectsShouldClearTheDatabase()
...
+                Schema = new[] { typeof(IntPrimaryKeyWithValueObject) }
+            };
+            var r2 = GetRealm(c2);
+            r2.Write(() =>
+            {
+                r2.RemoveAll();
+            });
+
+            // Reopen with the complete schema
+            r1 = Realm.GetInstance(c1);
```
gagik, line 182: So, interestingly enough. My attempt at just using `read_group()` without getting a full schema as @nirinchev suggested and it did allow me to pass the first test but throws the following here:
```
Realms.Tests.Database.RemoveTests.RemoveAllObjectsShouldClearEntireDatabaseWithPartialSchema:
    Outcome: Failed
    Error Message:
    Realms.Exceptions.RealmException : Object was deleted
    Stack Trace:
       at Realms.NativeException.ThrowIfNecessary(Func`2 overrider) in /Users/gagik.amaryan/Documents/realm-dotnet/Realm/Realm/Native/NativeException.cs:line 88
   at Realms.SharedRealmHandle.Open(Configuration configuration, RealmSchema schema, Byte[] encryptionKey) in /Users/gagik.amaryan/Documents/realm-dotnet/Realm/Realm/Handles/SharedRealmHandle.cs:line 398
   at Realms.RealmConfiguration.CreateHandle(RealmSchema schema) in /Users/gagik.amaryan/Documents/realm-dotnet/Realm/Realm/Configurations/RealmConfiguration.cs:line 142
   at Realms.RealmConfigurationBase.CreateRealm() in /Users/gagik.amaryan/Documents/realm-dotnet/Realm/Realm/Configurations/RealmConfigurationBase.cs:line 211
   at Realms.Realm.GetInstance(RealmConfigurationBase config) in /Users/gagik.amaryan/Documents/realm-dotnet/Realm/Realm/Realm.cs:line 104
   at Realms.Tests.Database.RemoveTests.RemoveAllObjectsShouldClearEntireDatabaseWithPartialSchema() in /Users/gagik.amaryan/Documents/realm-dotnet/Tests/Realm.Tests/Database/RemoveTests.cs:line 183
```
I am assuming it deletes more than objects with this call. What kind of "tables" are there?

papafe: So... There should be also some tables that are used by realm for bookkeeping, so it would be good to ask someone from core how to recognise those

nirinchev: You could also try debugging this and seeing which tables get cleared.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -754,6 +754,22 @@ REALM_EXPORT bool shared_realm_remove_type(const SharedRealm& realm, uint16_t* t
...
 
+REALM_EXPORT bool shared_realm_remove_all(const SharedRealm& realm, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        realm->verify_in_write();
+
+        realm::Group& all_tables = realm->read_group();
+        for(realm::TableKey table_key : all_tables.get_table_keys()) {
+            auto table = all_tables.get_table(table_key);
+            if (table->get_name().begins_with("class_")) {
```
gagik, line 765: Thanks to conversation with Jørgen, this way only the classes get cleared and passes both tests 🎉

wrappers/src/shared_realm_cs.cpp
```diff
@@ -754,6 +754,22 @@ REALM_EXPORT bool shared_realm_remove_type(const SharedRealm& realm, uint16_t* t
     });
 }
 
+REALM_EXPORT bool shared_realm_remove_all(const SharedRealm& realm, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        realm->verify_in_write();
+
+        realm::Group& all_tables = realm->read_group();
+        for(realm::TableKey table_key : all_tables.get_table_keys()) {
```
nirinchev, line 763:
```suggestion
        for (auto table_key : all_tables.get_table_keys()) {
```

wrappers/src/shared_realm_cs.cpp
```diff
@@ -754,6 +754,22 @@ REALM_EXPORT bool shared_realm_remove_type(const SharedRealm& realm, uint16_t* t
     });
 }
 
+REALM_EXPORT bool shared_realm_remove_all(const SharedRealm& realm, NativeException::Marshallable& ex)
+{
+    return handle_errors(ex, [&]() {
+        realm->verify_in_write();
+
+        realm::Group& all_tables = realm->read_group();
```
nirinchev, line 762:
```suggestion
        auto& group = realm->read_group();
```

`all_tables` is a bit of a weird name for the group - even though we're using it to get the tables, that is not its sole purpose, so let's go with the more boring, but accurate `group`.

nirinchev (approved):
Good stuff. Two very minor nits from me.

papafe:
Looks good. I think we need two things:
- Make the CI green. It seems that there are errors while building the wrappers
- It would be good to add some tests for sync too. We need to see what's the behaviour with clearing the table. With flexible sync, for instance, would it remove all objects, including the ones that have not been synced because they are not part of the queries?

PR #3080, opened 2022-11-03
Overhaul metrics in Realm

LaPeste:
## Description
This PR ports the current metrics collection that's in Realm to the source generator. Additionally, it extends the details of the collected metrics by taking advantage of the data surfaced by the source generator.

The impact this has on performance is not tested yet, as the work is not finished.

##  TODO
* [x] Changelog entry
* [x] Tests
* [x] Verify performance impact
* [ ] ~Get diagnostics to properly work~
* [ ] ~Verify if the analytics thread is allowed to finish even if roslyn has finished its duty~
* [x] Add Unity compatibility


Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+        }
+
+        public void AnalyzeSyntaxNodeForApiUsage(GeneratorSyntaxContext context)
+        {
+            if (context.Node is not IdentifierNameSyntax identifierNameSyntax)
+            {
+                return;
+            }
+
+            switch (identifierNameSyntax.ToString())
```
papafe, line 253: I discussed this already with Andrea, but maybe we can find a solution in which if we already know that a certain feature has already been found we don't need to do a comparison with that string value anymore. 

(We definitely need to have those benchmark running to evaluate the impact of that though)

Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+                            break;
+                        case "ISet":
+                            _realmFeaturesToAnalyse[SdkFeature.SetSubscribeForNotifications] = 1;
+                            break;
+                        case "IDictionary":
+                            _realmFeaturesToAnalyse[SdkFeature.DictionarySubscribeForNotifications] = 1;
+                            break;
+                        default:
+                            Diagnostics.Add(
+                                SourceGenerator.Diagnostics.AnalyticsDebugInfo($"{referenceType.Name} is not a collection type that is supported for notifications"));
```
papafe, line 319: I suppose that this is only for debug, right?

Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+        private bool IsCredentials(IdentifierNameSyntax identifierNameSyntax, GeneratorSyntaxContext context)
+        {
+            try
+            {
+                var parentType = context.SemanticModel.GetTypeInfo(identifierNameSyntax.Parent.ChildNodes().First()).Type.ToDisplayString();
+                return parentType == "Realms.Sync.Credentials" || parentType == "Realms.Sync.Credentials.AuthProvider";
+            }
+            catch (Exception ex)
+            {
+                Diagnostics.Add(SourceGenerator.Diagnostics.AnalyticsDebugInfo(
```
papafe, line 435: Same question as before

Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+                            $"{identifierNameSyntax} is likely some user defined syntax token.{Environment.NewLine}{ex.Message}"));
+                return false;
+            }
+        }
+
+        private string GetJsonPayload(bool pretty)
```
papafe, line 441: We can benchmark if this is more or less efficient than calling string.format on a premade string (that we can generate manually)

Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+                    return "UNKNOWN";
+                }
+            }
+        }
+
+        public void SubmitAnalytics(GeneratorExecutionContext context)
+        {
+            var payload = string.Empty;
+
+            if (Environment.GetEnvironmentVariable("REALM_DISABLE_ANALYTICS") != null ||
```
papafe, line 118: We should probably check this also in the analytics collection phase

Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+                Environment.GetEnvironmentVariable("CI") != null)
+            {
+                return;
+            }
+
+            _realmEnvMetrics[SdkFeature.UserId] = AnonymizedUserID;
+
+            ComputeHostOSNameAndVersion(out var osName, out var osVersion);
```
papafe, line 126: Maybe this method could return a tuple?

Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+                    _realmEnvMetrics[SdkFeature.HostOsType] = Metric.OperatingSystem.Windows;
+                    break;
+                case string name when name.Contains(Metric.OperatingSystem.MacOS):
+                    _realmEnvMetrics[SdkFeature.HostOsType] = Metric.OperatingSystem.MacOS;
+                    break;
+                case string name when name.Contains(Metric.OperatingSystem.Linux):
+                    _realmEnvMetrics[SdkFeature.HostOsType] = Metric.OperatingSystem.Linux;
+                    break;
+                default:
+                    Diagnostics.Add(
```
papafe, line 141: Same comment for the other diagnostics

Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+            _realmEnvMetrics[SdkFeature.HostOsVersion] = osVersion;
+
+            bool prettyJson = false;
+#if DEBUG
+            prettyJson = true;
+#endif
+            payload = GetJsonPayload(prettyJson);
+
+#if !DEBUG
+            var base64Payload = Convert.ToBase64String(Encoding.UTF8.GetBytes(payload));
```
papafe, line 155: This should be probably be part of the initial check of this method. If we're not going to send the analytics we shouldn't even collect them

Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+{payload}
+----------------------------------"));
+            }
+
+            // TODO andrea: we may actually look for something like INTERNAL_DEBUG, so that the user never sees this
+#if DEBUG
+            foreach (var diagnostic in Diagnostics)
+            {
+                // TODO andrea: this does not up show anywhere. Investigate why.
+                context.ReportDiagnostic(diagnostic);
```
papafe, line 180: We talked about this... maybe it's because you're not on the main source generator thread? Just an idea

Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+{
+    // TODO andrea: add Unity support, for the example how one can
+    // enable or disable analytics from the menu of Unity
+
+    internal class Analytics
+    {
+        public List<Diagnostic> Diagnostics { get; } = new();
+
+        // TODO andrea: adding C#10's type aliases to give meaning to each string seems nice. Opinions?
+        private Dictionary<string, byte> _realmFeaturesToAnalyse = new()
```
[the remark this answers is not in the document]
LaPeste: Good point. Thanks.

Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+                    jsonPayload.AppendLine();
+                }
+            }
+        }
+
+        private static void ComputeHostOSNameAndVersion(out string name, out string version)
+        {
+            var platformRegex = new Regex("^(?<platform>[^0-9]*) (?<version>[^ ]*)", RegexOptions.Compiled);
+            var osDescription = platformRegex.Match(RuntimeInformation.OSDescription);
+            if (osDescription.Success)
```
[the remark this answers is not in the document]
LaPeste: Same as above: This is taken straight from the weaver's analytics. I haven't gotten to test anything on another platform than Windows. I'll be able to answer later on.

Realm/Realm.SourceGenerator/Analytics.cs
LaPeste: Yes yes, it's a good point. Performance testing needs to be done first and foremost. But improving on this isn't gonna be hard, so I'll likely do it regardless of the performance result.

Realm/Realm.SourceGenerator/Analytics.cs
LaPeste: Correct. I was struggling with Diagnostics not showing so I filled the code with it. In the next iteration, that's only for `Debug.WriteLine`.

Realm/Realm.SourceGenerator/Analytics.cs
LaPeste: Correct. I was struggling with Diagnostics not showing so I filled the code with it. In the next iteration, that's only for `Debug.WriteLine`.

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,846 @@
...
+            },
+            {
+                nameof(RealmChanged),  (instruction, featureDict, references) =>
+                {
+                    featureDict[RealmChanged] = 1;
+                    return true;
+                }
+            },
+            {
+                "SubscribeForNotifications", (instruction, featureDict, references) =>
```
papafe, line 228: I suppose it's unlikely, but it seems like this could succeed even if it's called on non-Realm classes, am I right? 
I imagine that could be difficult to check

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,846 @@
...
+                _classAnalysisSetters.Remove(nameof(IEmbeddedObject));
+            }
+
+            foreach (var property in type.Properties)
+            {
+                var key = property.PropertyType.Name;
+                if (!_classAnalysisSetters.ContainsKey(key))
+                {
+                    // when looking for "Class" type
+                    key = property.PropertyType.MetadataType.ToString();
```
papafe, line 599: If this line is used only for Class type, can't we make a special case just for that?

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,846 @@
...
+                _classAnalysisSetters.Remove(nameof(IAsymmetricObject));
+            }
+            else if (_classAnalysisSetters.TryGetValue(nameof(IEmbeddedObject), out featureFunc) &&
+                (type.IsIEmbeddedObjectImplementor(_references) || type.IsEmbeddedObjectDescendant(_references)) &&
+                featureFunc.Invoke(type, _realmFeaturesToAnalyse, _references))
+            {
+                _classAnalysisSetters.Remove(nameof(IEmbeddedObject));
+            }
+
+            foreach (var property in type.Properties)
```
papafe, line 593: Are we getting all properties here, including the ones that will not be part of the schema?

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,846 @@
...
+                    jsonPayload.AppendLine();
+                }
+            }
+        }
+
+        private static async Task SendRequest(string prefixAddr, string payload, string suffixAddr)
+        {
+            using (var httpClient = new HttpClient())
+            {
+                httpClient.Timeout = TimeSpan.FromSeconds(4);
```
papafe, line 741: Isn't 4 seconds too much?

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,846 @@
...
+                await httpClient.GetAsync(new Uri(prefixAddr + payload + suffixAddr));
+            }
+        }
+
+        private void AnalyzeTypeMethods(TypeDefinition type)
```
papafe, line 746: A small thing, but I think it would make sense to have close related methods. So maybe first all the analysis methods, and then all the sending methods (including the json building)

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,846 @@
...
+                if (_classAnalysisSetters.TryGetValue(key, out featureFunc) &&
+                    featureFunc.Invoke(property, _realmFeaturesToAnalyse, _references))
+                {
+                    // if the byte is set, remove the entry from the dict to avoid future unnecessary work
+                    _classAnalysisSetters.Remove(key);
+                }
+            }
+        }
+
+        internal void AnalyzeUserAssembly(ModuleDefinition module)
```
papafe, line 611: Small thing, but sometimes you use analyse, some other analyze. I noticed because I couldn't find some methods with search 😁

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,846 @@
...
+                    if (_apiAnalysisSetters.TryGetValue(key, out var featureFunc) &&
+                        featureFunc.Invoke(cil, _realmFeaturesToAnalyse, _references))
+                    {
+                        _apiAnalysisSetters.Remove(key);
+                    }
+                }
+            }
+        }
+
+        private static bool IsValidRealmType(TypeDefinition type, ImportedReferences references) =>
```
papafe, line 802: Would it make sense to move it together with the other extension methods?

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -0,0 +1,164 @@
...
+        private static string ConvertOsNameToMetricsVersion(string osName) =>
+            WrapInTryCatch(() =>
+            {
+                switch (osName)
+                {
+                    case string s when s.ContainsIgnoreCase(Metric.OperatingSystem.Windows):
+                        return Metric.OperatingSystem.Windows;
+                    case string s when s.ContainsIgnoreCase(Metric.OperatingSystem.MacOS):
+                        return Metric.OperatingSystem.MacOS;
+                    case string s when s.ContainsIgnoreCase(Metric.OperatingSystem.Linux):
```
papafe, line 118: Does this work for various Linux distro? (Just curious)

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -0,0 +1,164 @@
...
+using Mono.Cecil;
+using static RealmWeaver.Metric;
+
+namespace RealmWeaver
+{
+    internal static class AnalyticsUtils
+    {
+        // TODO andrea: https://github.com/realm/realm-dotnet/issues/2706#event-7877818404
+        // add a stable ID not reliant on MACAddress
+        public static string AnonymizedUserID
```
papafe, line 34: Why is this a property while the others are methods?

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            get
+            {
+                try
+                {
+                    var id = GenerateComputerIdentifier;
+                    return id != null ? SHA256Hash(id) : "UNKNOWN";
+                }
+                catch
```
papafe, line 43: What kind of exceptions can you get here?

Realm/Realm.SourceGenerator/Realm.SourceGenerator.csproj
```diff
@@ -13,6 +13,8 @@
         <LangVersion>9.0</LangVersion>
         <IsRoslynComponent>true</IsRoslynComponent>
         <IsPackable>true</IsPackable>
+        <ReportAnalyzer Condition="'$(Configuration)' == 'Debug'">true</ReportAnalyzer>
+        <Verbosity Condition="'$(Configuration)' == 'Debug'">detailed</Verbosity>
```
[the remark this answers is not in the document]
LaPeste: Yes, it's for extracting performance from the SG. I thought to leave them given that they are only used when in debug. Do you think it's a bad idea?

Tests/Weaver/Realm.FakeForWeaverTests/Sync/ErrorHandling/ManualRecoveryHandler.cs
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms.Sync.ErrorHandling
+{
+    public sealed class ManualRecoveryHandler : ClientResetHandlerBase
```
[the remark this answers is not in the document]
LaPeste: It's part of testing

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,846 @@
...
+    // - What version of Realm is being used
+    // - What OS you are running on
+    // - What OS you are building for
+    // - An anonymized MAC address and assembly name ID to aggregate the other information on.
+    internal class Analytics
+    {
+        private readonly ImportedReferences _references;
+        private readonly ILogger _logger;
+
+        #region FeatureDiciontaries
```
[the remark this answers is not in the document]
LaPeste: Thanks 😅

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,846 @@
...
+using System.Threading.Tasks;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+
+using static RealmWeaver.AnalyticsUtils;
+using static RealmWeaver.Metric.SdkFeature;
+
+namespace RealmWeaver
+{
+    // TODO andrea: review and update this comment
```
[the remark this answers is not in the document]
LaPeste: That's always been the case. What are you asking? Do you want it moved?

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,846 @@
...
+    // - What OS you are building for
+    // - An anonymized MAC address and assembly name ID to aggregate the other information on.
+    internal class Analytics
+    {
+        private readonly ImportedReferences _references;
+        private readonly ILogger _logger;
+
+        #region FeatureDiciontaries
+
+        private readonly Dictionary<string, byte> _realmFeaturesToAnalyse = new Dictionary<string, byte>()
```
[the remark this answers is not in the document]
LaPeste: Yes, I just needed 0 and 1.

Realm/Realm.Weaver/Analytics/Analytics.cs
LaPeste: If you mean avoiding to call into the `Func`s of the dictionary, I did it to keep consistency with the rest.

Realm/Realm.Weaver/Analytics/Analytics.cs
LaPeste: I used what it's [currently](https://github.com/realm/realm-dotnet/blob/f7a10a89d0e8ea3cc8d91056e7f63c572f983d6c/Realm/Realm.Weaver/Analytics.cs#L162) being done. We can change that if we want.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: This is never been tested outside of my windows machine. It could be a good idea to add unit tests to test these details. I'll think about what can be done.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: I just got what we were doing in the previous work. Knowing that I have to work on the github issue, I did not put effort on changing it for now.

Realm/Realm.SourceGenerator/Realm.SourceGenerator.csproj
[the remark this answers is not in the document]
papafe: Mmh, not sure myself. It depends how much info is actually being printed in the output

Realm/Realm.Weaver/Analytics/Analytics.cs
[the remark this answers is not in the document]
papafe: It's fine, I was just wondering if we want to put it somewhere else where it's more visible, but maybe not ;)

Realm/Realm.Weaver/Analytics/Analytics.cs
papafe: Makes sense

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
papafe: Reasonable!

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -236,7 +228,11 @@ Analytics payload
             var wovenAssemblyAttribute = new CustomAttribute(_references.WovenAssemblyAttribute_Constructor);
             _moduleDefinition.Assembly.CustomAttributes.Add(wovenAssemblyAttribute);
 
-            submitAnalytics.Wait();
+            analyzeAPITask?.Wait();
+            if (!string.IsNullOrEmpty(analyticsConfig.AnalyticsLogPath))
+            {
+                File.WriteAllText(analyticsConfig.AnalyticsLogPath, metricsResult);
```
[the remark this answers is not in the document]
LaPeste: No. Whenever a path is given to the config, we always write a file, regardless of if the analytics was turned on or not.
We specifically use this to [test](https://github.com/realm/realm-dotnet/blob/fcb05bbbf55f59d82419899a1f1ef027ced77701/Tests/Weaver/Realm.Fody.Tests/AnalyticsTests.cs#L124) that disabling analytics through the settings effectively disable analytics (by checking the content of the file).

Realm/Realm.Weaver/Analytics/Analytics.cs
LaPeste: Good point. I'll look into what can be done, but I'm not sure how easy it is going to be.

Realm/Realm.Weaver/Analytics/Analytics.cs
LaPeste: Yes. It looks at the assembly whle the weaving task is carried out on that same assembly.
I guess you're correct in suggesting of not analyzing the objects that don't make it into the schema. I'll look at how to piggyback what we already have in `WeaveSchema`.

papafe: I'm mostly worried about properties that are not in the schema because developers could have other properties that are not persisted. The objects that are not in the schema...maybe it's slightly less important

Realm/Realm.Weaver/Analytics/Analytics.cs
LaPeste: I looked at this and I can't resolve the generic type of the collection. Not being able to do so makes checking for type impossible.
@nirinchev  just to have your eye on this, for example given ```System.Collections.Generic.ISet`1<!!0>``` I'd like to have a `Mono.Cecil.TypeDefinition` in order to check if it's a collection of `RealmObject`s I tried
```cs
((GenericInstanceType)(TypeSpecification)methodSpecification.Parameters[0].ParameterType).GenericArguments[0].Resolve();
```
but the resolution is null. I don't really understand why as I expected all the signatures to be compiled at this point, but mono.cecil's [test](https://github.com/jbevain/cecil/blob/master/Test/Mono.Cecil.Tests/ResolveTests.cs#L212-L221) expects the same.
@nirinchev is there something I'm missing or it's just not possible? If the latter is the case, do you know why so?

Realm/Realm.SourceGenerator/Analytics.cs
```diff
@@ -0,0 +1,524 @@
...
+namespace Realms.SourceGenerator
+{
+    // TODO andrea: add Unity support, for the example how one can
+    // enable or disable analytics from the menu of Unity
+
+    internal class Analytics
+    {
+        public List<Diagnostic> Diagnostics { get; } = new();
+
+        // TODO andrea: adding C#10's type aliases to give meaning to each string seems nice. Opinions?
```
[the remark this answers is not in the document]
LaPeste: I can't remember what this was

Realm/Realm.SourceGenerator/Analytics.cs
LaPeste: Do you still believe we have to look into this?

Realm/Realm.SourceGenerator/Analytics.cs
LaPeste: This is about SGs. We've now moved this to the weaver.

Realm/Realm.SourceGenerator/Analytics.cs
LaPeste: This is about SGs. We've now moved this to the weaver.

Realm/Realm.SourceGenerator/Analytics.cs
LaPeste: This is about SGs. We've now moved this to the weaver.

Realm/Realm.SourceGenerator/Analytics.cs
LaPeste: This is about SGs. We've now moved this to the weaver.

Realm/Realm.SourceGenerator/Realm.SourceGenerator.csproj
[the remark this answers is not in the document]
LaPeste: To my knowledge nothing is printed out. It just creates a `.binlog` file with performance information. Such file can be opened with [MSBuild Log Viewer](https://msbuildlog.com/) or in VS throught [Project System Tools](https://marketplace.visualstudio.com/items?itemName=VisualStudioProductTeam.ProjectSystemTools). However, in my experience the viewer in VS isn't as detailed as the standalone log viewer.
If you are interested in more, you can find more info at https://jaylee.org/archive/2020/10/10/profiling-csharp-9-source-generators.html .

Realm/Realm.Weaver/Analytics/Analytics.cs
LaPeste: Should this be resolved?

Realm/Realm.Weaver/Analytics/Analytics.cs
LaPeste: I worked this out and now only properties that are part of the schema are checked. You can resolve this if you are satisfied, let me know otherwise.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: Testing for this is going to be part of Phase 2 of this endevour.

LaPeste:
@nirinchev @papafe there are a bunch of `TODO andrea` left behind, 4 to be precise. I left those there either as a reminder to have a conversation about what they refer or as a future reminder for what's mostly going to be phase 2 of 
overhauling metrics in Realm.
Advice of what do write instead of `TODO andrea` as this "tag" is generally flagged by our *Verify TODOs* job.

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,885 @@
...
+                Func<IMemberDefinition, Dictionary<string, byte>, ImportedReferences, bool> featureFunc = null;
+
+                foreach (var propertyResult in type.Properties)
+                {
+                    var property = propertyResult.Property;
+
+                    var key = property.PropertyType.Name;
+                    if (!_classAnalysisSetters.ContainsKey(key))
+                    {
+                        // when looking for "Class" type
```
papafe, line 654: What is class type?

Realm/Realm.Weaver/Analytics/Analytics.cs
nirinchev: Can we check the assembly from which `SubscribeForNotifications` comes?

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -46,6 +49,76 @@ SequencePoint GetPropSequencePoint()
...
+            case MetadataScopeType.AssemblyNameReference:
+                return ((AssemblyNameReference)@this.Scope).FullName;
+            case MetadataScopeType.ModuleReference:
+                return ((ModuleReference)@this.Scope).Name;
+            default:
+                return ((ModuleDefinition)@this.Scope).Assembly.FullName;
+        }
+    }
+
+    public static string ToFriendlyString(this TypeReference type)
```
nirinchev, line 84: Not sure I'm super happy about all these reorderings - perhaps they make sense, but I would much rather we do them in a separate PR as it inflates the number of lines everyone needs to go through and there's no clear indication what changes are real and what are just moving code around.

Tests/Weaver/AnalyticsAssembly/Program.cs
```diff
@@ -0,0 +1,360 @@
...
+#endif
+
+#if FIND
+    public static void FindByPk()
+    {
+        _ = Realm.GetInstance().Find<RootRealmClass>("aKey");
+    }
+#endif
+
+
```
nirinchev, line 66:
```suggestion
```

Tests/Weaver/AnalyticsAssembly/Program.cs
```diff
@@ -0,0 +1,360 @@
...
+#endif
+
+#if BACKLINK
+        [Backlink(nameof(JustForObjectReference.UseAsBacklink))]
+        IQueryable<JustForObjectReference> JustBackLink { get; }
+#endif
+}
+
+#if ASYMMETRIC_OBJECT
+public partial class AsymmetricTestClass : TestAsymmetricObject
```
nirinchev, line 356:
```suggestion
    public partial class AsymmetricTestClass : TestAsymmetricObject
```

Tests/Weaver/AnalyticsAssembly/Program.cs
```diff
@@ -0,0 +1,360 @@
...
+
+#if REALM_VALUE
+        RealmValue RealmValue { get; set; }
+#endif
+
+#if BACKLINK
+        [Backlink(nameof(JustForObjectReference.UseAsBacklink))]
+        IQueryable<JustForObjectReference> JustBackLink { get; }
+#endif
+}
```
nirinchev, line 353:
```suggestion
    }
```

nirinchev (changes requested):
There's a lot of changes here, so I've done only a superficial review of some of the files. I hope to continue with a more thorough review tomorrow.

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,884 @@
...
+    // - What APIs of the Realm SDK you're using
+    // - An anonymized MAC address and assembly name ID to aggregate the other information on.
+    internal class Analytics
+    {
+        private readonly ImportedReferences _references;
+        private readonly ILogger _logger;
+
+        #region FeatureDictionaries
+
+        private readonly Dictionary<string, byte> _realmFeaturesToAnalyze = new Dictionary<string, byte>()
```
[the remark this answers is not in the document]
LaPeste: Target-typed obeject creation is not available in C# 8.0. It's only from C# 9.0 onwards.
It seems that Unity [supports](https://docs.unity3d.com/Manual/CSharpCompiler.html) most of C# 9.0. If you want, later I can look into bumping up the version of the language in `Realm.Fody` and all the projects that involve unity.
Let me know, so that I know if to create an issue for that.

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,884 @@
...
+            [JWT] = 0,
+            [ApiKey] = 0,
+            [ServerApiKey] = 0,
+            [Function] = 0,
+            [CallAsync] = 0,
+            [GetMongoClient] = 0,
+            [DynamicApi] = 0,
+        };
+
+        private readonly Dictionary<string, string> _realmEnvMetrics = new Dictionary<string, string>()
```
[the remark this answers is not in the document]
LaPeste: Ditto

Realm/Realm.Weaver/Analytics/Analytics.cs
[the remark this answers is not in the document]
nirinchev: We are compiling those assemblies using msbuild, so Unity support should not affect us.

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -0,0 +1,884 @@
...
+    // - What framework and what framework version Realm is being used with (e.g. Xamarin, MAUI, etc.)
+    // - How the Realm SDK was installed (e.g. Nuget, manual, etc.)
+    // - What APIs of the Realm SDK you're using
+    // - An anonymized MAC address and assembly name ID to aggregate the other information on.
+    internal class Analytics
+    {
+        private readonly ImportedReferences _references;
+        private readonly ILogger _logger;
+
+        #region FeatureDictionaries
```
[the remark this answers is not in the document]
LaPeste: Everything is fine with the refactoring but mutating `_realmFeaturesToAnalyze` in the `Func`s or to more generally say: having custom code in the `Func`s.
There are special cases that still need to be handled manually as the value from mono.cecil is not gonna be enough. Examples are Collections. I need to analyze if it's a collection of reference type of primitive types. The most obvious way to do the analysis is in the `Func`. Now if that returns a bool and a string indicating the right key or just a bool and internally setting the right value in the dictionary is a matter of taste. But what I'm after is: the most elegant way to leave methods like `AnalyzeRealmClassProperties` elegant and small is to take the handling in the functions.
Other nuances are with generics, where fields like `RealmInteger` come out from mono.cecil like **RealmInteger`1**. To avoid custom code I need to have a key in  `SdkFeatures` with the arity in the key. As personal preference, I don't see that as a good thing, since we're using zero help from the compiler given that we're using strings for keys. It's annoying to have to type arity everywhere and while it's easy to add 2 lines once in the `Func` to handle fields with arity.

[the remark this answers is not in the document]
LaPeste: Of course, the drawback of this is that even simple fields now need more lines in the func, not just `return true/false`.
So alternatively I need to put all that analysis in `AnalyzeRealmClassProperties`, but I don't personally like it as it makes it longer to read.

Realm/Realm.Weaver/Extensions/PropertyDefinitionExtensions.cs
```diff
@@ -30,8 +29,6 @@
 [EditorBrowsable(EditorBrowsableState.Never)]
 internal static class PropertyDefinitionExtensions
 {
-    private static readonly Regex NullableRegex = new Regex("^System.Nullable`1<(?<typeName>.*)>$");
-
```
LaPeste, line 34: Moved to `TypeReferenceExtension` line 30

Realm/Realm.Weaver/Extensions/PropertyDefinitionExtensions.cs
```diff
@@ -115,11 +112,6 @@ internal static bool IsNullable(this PropertyDefinition property)
         return property.PropertyType.IsNullable();
     }
 
-    internal static bool IsNullable(this TypeReference reference)
-    {
-        return NullableRegex.IsMatch(reference.FullName);
-    }
-
```
LaPeste, line 122: Moved to `TypeReferenceExtension` line 116

Realm/Realm.Weaver/Extensions/PropertyDefinitionExtensions.cs
```diff
@@ -203,20 +195,6 @@ internal static bool IsIndexable(this PropertyDefinition property, ImportedRefer
...
-        type.BaseType.IsSameAs(references.RealmObject);
-
-    public static bool IsAsymmetricObjectInheritor(this TypeDefinition type, ImportedReferences references) =>
-        type.BaseType.IsSameAs(references.AsymmetricObject);
-
-    public static bool IsValidRealmObjectBaseInheritor(this TypeDefinition type, ImportedReferences references) =>
-        type.IsEmbeddedObjectInheritor(references) ||
-        type.IsRealmObjectInheritor(references) ||
-        type.IsAsymmetricObjectInheritor(references);
-
```
LaPeste, line 219: Moved to `TypeReferenceExtension` line 94

Realm/Realm.Weaver/Extensions/TypeDefinitionExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+    public static bool IsRealmObjectInheritor(this TypeDefinition type, ImportedReferences references) =>
+        type.BaseType.IsSameAs(references.RealmObject);
+
+    public static bool IsAsymmetricObjectInheritor(this TypeDefinition type, ImportedReferences references) =>
+        type.BaseType.IsSameAs(references.AsymmetricObject);
+
+    public static bool IsValidRealmObjectBaseInheritor(this TypeDefinition type, ImportedReferences references) =>
+        type.IsRealmObjectInheritor(references) ||
+        type.IsEmbeddedObjectInheritor(references) ||
+        type.IsAsymmetricObjectInheritor(references);
```
LaPeste, line 42: Moved from `PropertyDefinitionExtension`

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -26,6 +27,8 @@
 [EditorBrowsable(EditorBrowsableState.Never)]
 internal static class TypeReferenceExtensions
 {
+    private static readonly Regex NullableRegex = new Regex("^System.Nullable`1<(?<typeName>.*)>$");
```
LaPeste, line 30: Moved from `PropertyDefinitionExtensions` line 33

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -46,6 +49,76 @@ SequencePoint GetPropSequencePoint()
...
+            genericArgumentType = genericType.GenericArguments.Single();
+            return true;
+        }
+
+        genericArgumentType = null;
+        return false;
+    }
+
+    public static bool IsNullable(this TypeReference reference) =>
+        NullableRegex.IsMatch(reference.FullName);
```
LaPeste, line 117: Moved from `PropertyDefinitionExtensions` line 118

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -46,6 +49,76 @@ SequencePoint GetPropSequencePoint()
...
+        if (result)
+        {
+            var genericType = (GenericInstanceType)type;
+            genericArgumentType = genericType.GenericArguments.Single();
+            return true;
+        }
+
+        genericArgumentType = null;
+        return false;
+    }
```
LaPeste, line 114: Moved from `PropertyDefinitionExtensions` line 229

Realm/Realm.Weaver/Extensions/PropertyDefinitionExtensions.cs
```diff
@@ -226,28 +204,6 @@ internal static bool IsIndexable(this PropertyDefinition property, ImportedRefer
...
-        {
-            var genericType = (GenericInstanceType)type;
-            genericArgumentType = genericType.GenericArguments.Single();
-            return true;
-        }
-
-        genericArgumentType = null;
-        return false;
-    }
-
```
LaPeste, line 250: Moved to `TypeReferenceExtensions` line 94

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -83,46 +156,29 @@ private static bool IsDescendantOf(TypeReference @this, params TypeReference[] t
...
-
-    public static string ToFriendlyString(this TypeReference type)
-    {
-        if (!string.IsNullOrEmpty(type.Namespace))
+        catch
         {
-            return type.ToString().Replace(type.Namespace, string.Empty).TrimStart('.');
+            // Unity may fail to resolve some of its assemblies, but that's okay
+            // they don't contain RealmObject classes.
         }
```
LaPeste, line 180: Moved to this same file line 84

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
LaPeste, line 84: Moved from this same file from line 119

LaPeste: I'll avoid this in the future. I've added comments everywhere it's not new code but just shuffled around.

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -83,46 +156,29 @@ private static bool IsDescendantOf(TypeReference @this, params TypeReference[] t
...
-        switch (@this.Scope.MetadataScopeType)
-        {
-            case MetadataScopeType.AssemblyNameReference:
-                return ((AssemblyNameReference)@this.Scope).FullName;
-            case MetadataScopeType.ModuleReference:
-                return ((ModuleReference)@this.Scope).Name;
-            default:
-                return ((ModuleDefinition)@this.Scope).Assembly.FullName;
+            return TypeDefinitionExtensions.IsImplementorOf(definition, targetInterfaces);
         }
```
LaPeste, line 175: Moved in the same file to line 71

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -46,6 +49,76 @@ SequencePoint GetPropSequencePoint()
...
+        switch (@this.Scope.MetadataScopeType)
+        {
+            case MetadataScopeType.AssemblyNameReference:
+                return ((AssemblyNameReference)@this.Scope).FullName;
+            case MetadataScopeType.ModuleReference:
+                return ((ModuleReference)@this.Scope).Name;
+            default:
+                return ((ModuleDefinition)@this.Scope).Assembly.FullName;
+        }
+    }
```
LaPeste, line 82: Moved from the same file from line 106

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -83,46 +156,29 @@ private static bool IsDescendantOf(TypeReference @this, params TypeReference[] t
         return false;
     }
 
-    public static bool IsRealmObjectDescendant(this TypeReference @this, ImportedReferences references)
-    {
-        return IsDescendantOf(@this, references.RealmObject, references.EmbeddedObject, references.AsymmetricObject);
-    }
```
LaPeste, line 89: Moved in same file to line 52

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -46,6 +49,76 @@ SequencePoint GetPropSequencePoint()
         }
     }
 
+    public static bool IsRealmObjectDescendant(this TypeReference @this, ImportedReferences references) =>
+        IsDescendantOf(@this, references.RealmObject, references.EmbeddedObject, references.AsymmetricObject);
```
LaPeste, line 53: Moved from the same file from line 86

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -83,46 +156,29 @@ private static bool IsDescendantOf(TypeReference @this, params TypeReference[] t
...
 
-    public static bool IsRealmObjectDescendant(this TypeReference @this, ImportedReferences references)
-    {
-        return IsDescendantOf(@this, references.RealmObject, references.EmbeddedObject, references.AsymmetricObject);
-    }
-
-    public static bool IsAsymmetricObjectDescendant(this TypeReference @this, ImportedReferences references)
-    {
-        return IsDescendantOf(@this, references.AsymmetricObject);
-    }
```
LaPeste, line 94: Moved in the same file to line 55

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -46,6 +49,76 @@ SequencePoint GetPropSequencePoint()
         }
     }
 
+    public static bool IsRealmObjectDescendant(this TypeReference @this, ImportedReferences references) =>
+        IsDescendantOf(@this, references.RealmObject, references.EmbeddedObject, references.AsymmetricObject);
+
+    public static bool IsAsymmetricObjectDescendant(this TypeReference @this, ImportedReferences references) =>
+        IsDescendantOf(@this, references.AsymmetricObject);
```
LaPeste, line 56: Moved from the same file from line 91

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -46,6 +49,76 @@ SequencePoint GetPropSequencePoint()
...
+
+    public static bool IsEmbeddedObjectDescendant(this TypeReference @this, ImportedReferences references) =>
+        IsDescendantOf(@this, references.EmbeddedObject);
```
LaPeste, line 59: This is newly added

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -83,46 +156,29 @@ private static bool IsDescendantOf(TypeReference @this, params TypeReference[] t
...
-    }
-
-    public static bool IsSameAs(this TypeReference @this, TypeReference other)
+    private static bool IsImplementorOf(TypeReference @this, params TypeReference[] targetInterfaces)
     {
-        if (@this is null || other is null)
+        try
         {
-            return false;
-        }
```
LaPeste, line 101: Moved in the same file to line 61

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -83,46 +156,29 @@ private static bool IsDescendantOf(TypeReference @this, params TypeReference[] t
...
+            if (@this == null || !Weaver.ShouldTraverseAssembly(@this.Module.Assembly.Name))
+            {
+                return false;
+            }
 
-        return @this.FullName == other.FullName && @this.GetAssemblyName() == other.GetAssemblyName();
-    }
```
LaPeste, line 104: Moved in the same file to line 61

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -46,6 +49,76 @@ SequencePoint GetPropSequencePoint()
...
+
+    public static bool IsSameAs(this TypeReference @this, TypeReference other)
+    {
+        if (@this is null || other is null)
+        {
+            return false;
+        }
+
+        return @this.FullName == other.FullName && @this.GetAssemblyName() == other.GetAssemblyName();
+    }
```
LaPeste, line 69: Moved from the same file from line 96

Realm/Realm.Weaver/Extensions/TypeReferenceExtensions.cs
```diff
@@ -46,6 +49,76 @@ SequencePoint GetPropSequencePoint()
...
+
+        genericArgumentType = null;
+        return false;
+    }
+
+    public static bool IsNullable(this TypeReference reference) =>
+        NullableRegex.IsMatch(reference.FullName);
+
+    public static bool IsIRealmObjectBaseImplementor(this TypeReference type, ImportedReferences references) =>
+        IsImplementorOf(type, references.IRealmObjectBase);
```
LaPeste, line 120: Newly added in this PR

Realm/Realm.Weaver/Extensions/TypeDefinitionExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                if (@interface.InterfaceType.IsSameAs(targetInterface))
+                {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
```
LaPeste, line 64: Newly added in this PR

Realm/Realm.Weaver/Extensions/PropertyDefinitionExtensions.cs
LaPeste, line 250: Moved to `TypeReferenceExtensions` line 94

LaPeste:
I littered the PR with comments on what's new and what's only been moved around when it comes to the files `PropertyDefinitionExtensions.cs`, `TypeDefinitionExtensions.cs` and `TypeReferenceExtensions.cs`.
I know it's not much of an improvement but hopefully it helps a little.
For the future, I'll make sure not to do again such large reshuffling in a PR that's meant to add a feature.

Tests/Weaver/Realm.Fody.Tests/AnalyticsTests.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.Xml.Linq;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization;
+using NUnit.Framework;
+using RealmWeaver;
+
+namespace Analytics
+{
+    extern alias realm;
+    using Metric = realm::RealmWeaver.Metric;
```
papafe, line 34: Why do we need this?

Tests/Weaver/AnalyticsAssembly/Program.cs
```diff
@@ -0,0 +1,359 @@
...
+using TestAsymmetricObject = Realms.IAsymmetricObject;
+using TestEmbeddedObject = Realms.IEmbeddedObject;
+using TestRealmObject = Realms.IRealmObject;
+#endif
+
+public class Program
+{
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+
+    public static async Task Main(string[] args)
```
papafe, line 39: Why do you need this async method?

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -0,0 +1,263 @@
...
+                    _ => $"{platformID} is an unsupported operating system."
+                });
+
+        public static (string Name, string Version) GetFrameworkAndVersion(ModuleDefinition module, Config config)
+        {
+            if (config.TargetFramework.ContainsIgnoreCase("unity"))
+            {
+                return config.TargetFramework.ContainsIgnoreCase("editor") ?
+                    (Framework.UnityEditor, config.TargetFrameworkVersion) :
+                    (Framework.Unity, config.TargetFrameworkVersion);
```
papafe, line 127: What's the difference between unity and unityEditor?

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -0,0 +1,263 @@
...
+            }
+            else
+            {
+                // TODO andrea: the correctness of these names need to be verified in projects that use each of the packages
```
papafe, line 131: Did you manage to find a xamarin project to test this on?

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -0,0 +1,263 @@
...
+                // I didn't have any handy one.
+                var possibleFrameworks = new string[] { "Xamarin.Forms", "Mono.Android", "Xamarin.iOS", "Microsoft.Maui.Sdk" };
```
papafe, line 133: Not Xamarin.Android?

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: The editor is when you're running the project in the editor and the player is when you're running the binary of your game as standalone.
So we need to know if it's weaving to target the editor or to target the standalone binary.

Tests/Weaver/AnalyticsAssembly/Program.cs
LaPeste: No need. I'll remove it, thanks.

PR #3083, opened 2022-11-07
Update README.md

papafe:
Fixes bad link and some wording

Fixes #3082

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,18 +36,18 @@ Refer to [this guide](https://www.visualstudio.com/en-us/docs/package/nuget/cons
 
 ## Building Realm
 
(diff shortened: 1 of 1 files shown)


LaPeste:
What bad link is this fixing? It just removes the link to source, which was correctly pointing to the Realm .NET repo.

papafe:
> What bad link is this fixing? It just removes the link to source, which was correctly pointing to the Realm .NET repo.
> 
> EDIT: I can see that this issue comes from #3082 . I guess that what we're tying to say is that we don't need to link to the repo since we're already in the repo?

Yes, it doesn't make sense to link to the same repo in which we are already in.

LaPeste (approved):

papafe:
@nirinchev anything against merging this?

nirinchev:
No, go for it

PR #3088, opened 2022-11-11
Remove backing fields when weaving generated classes

gagik:
## Description
We are weaving the properties of the generated classes to go directly to the accessors, so the backing fields are unused. This PR attempts to remove them for generated classes.

Fixes https://github.com/realm/realm-dotnet/issues/2994

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -310,15 +313,16 @@ private static void ReplaceGeneratedClassGetter(PropertyDefinition prop, TypeDef
             //// This is equivalent to:
             ////   get => Accessor.Property;
 
-            var start = prop.GetMethod.Body.Instructions.First();
             var il = prop.GetMethod.Body.GetILProcessor();
+            prop.GetMethod.Body.Instructions.Clear();
```
gagik, line 317: I believe the previous implementation of replacing the getter was causing some unintended behavior, it was actually just adding instructions to the getter rather than removing the old instructions---the reason why this was not noticeable was because the old instruction because unreachable because of the `ret`/`return` statement. My ILSpy was actually giving me an error about this prior. One issue this was causing was it kept instructions that have a reference to the backing field, making it error-prone when removing them. This change makes it identical to the functionality in the `ReplaceGeneratedClassSetter`.

Tests/SourceGenerators/SourceGeneratorAssemblyToProcess/ErrorClasses/UnsupportedRequiredTypes.cs
```diff
@@ -51,11 +51,11 @@ public partial class UnsupportedRequiredTypes : IRealmObject
         [Required]
         public decimal DecimalProp { get; set; }
 
-        [Required]
-        public ObjectId ObjectIdProp { get; set; }
+        // [Required]
+        // public ObjectId ObjectIdProp { get; set; }
```
gagik, line 55: Here is where the main issue lays right now and I am not sure how to tackle it. If you have a property whose type is imported from another module i.e. `ObjectId` is imported from `MongoDB.BSON` then removing the backing fields to the following error:
```
  Fody: An unhandled exception occurred:
Exception:
Member 'MongoDB.Bson.ObjectId <Id>k__BackingField' is declared in another module and needs to be imported
Type:
System.ArgumentException
StackTrace:
   at Mono.Cecil.MetadataBuilder.LookupToken(IMetadataTokenProvider provider) in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 2264
   at Mono.Cecil.Cil.CodeWriter.WriteOperand(Instruction instruction) in C:\projects\fody\cecil\Mono.Cecil.Cil\CodeWriter.cs:line 270
   at Mono.Cecil.Cil.CodeWriter.WriteInstructions() in C:\projects\fody\cecil\Mono.Cecil.Cil\CodeWriter.cs:line 180
   at Mono.Cecil.Cil.CodeWriter.WriteResolvedMethodBody(MethodDefinition method) in C:\projects\fody\cecil\Mono.Cecil.Cil\CodeWriter.cs:line 118
   at Mono.Cecil.Cil.CodeWriter.WriteMethodBody(MethodDefinition method) in C:\projects\fody\cecil\Mono.Cecil.Cil\CodeWriter.cs:line 54
   at Mono.Cecil.MetadataBuilder.AddMethod(MethodDefinition method) in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 1675
   at Mono.Cecil.MetadataBuilder.AddMethods(TypeDefinition type) in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 1668
   at Mono.Cecil.MetadataBuilder.AddType(TypeDefinition type) in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 1468
   at Mono.Cecil.MetadataBuilder.AddTypes() in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 1440
   at Mono.Cecil.MetadataBuilder.BuildTypes() in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 1289
   at Mono.Cecil.MetadataBuilder.BuildModule() in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 1059
   at Mono.Cecil.MetadataBuilder.BuildMetadata() in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 1029
   at Mono.Cecil.ModuleWriter.<>c.<BuildMetadata>b__2_0(MetadataBuilder builder, MetadataReader _) in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 147
   at Mono.Cecil.ModuleDefinition.Read[TItem,TRet](TItem item, Func`3 read) in C:\projects\fody\cecil\Mono.Cecil\ModuleDefinition.cs:line 971
   at Mono.Cecil.ModuleWriter.BuildMetadata(ModuleDefinition module, MetadataBuilder metadata) in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 146
   at Mono.Cecil.ModuleWriter.Write(ModuleDefinition module, Disposable`1 stream, WriterParameters parameters) in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 119
   at Mono.Cecil.ModuleWriter.WriteModule(ModuleDefinition module, Disposable`1 stream, WriterParameters parameters) in C:\projects\fody\cecil\Mono.Cecil\AssemblyWriter.cs:line 78
   at Mono.Cecil.ModuleDefinition.Write(String fileName, WriterParameters parameters) in C:\projects\fody\cecil\Mono.Cecil\ModuleDefinition.cs:line 1160
   at InnerWeaver.WriteModule() in C:\projects\fody\FodyIsolated\ModuleWriter.cs:line 18
   at InnerWeaver.Execute() in C:\projects\fody\FodyIsolated\InnerWeaver.cs:line 115
Source:
Mono.Cecil
TargetSite:
Mono.Cecil.MetadataToken LookupToken(Mono.Cecil.IMetadataTokenProvider)
```
I am assuming there is something in this imported type reference that refers to the backing field for some reason but after a lot of looking I have not been able to find it. I also tried adding the type of the backing field to `type.NestedTypes` in case it does not actually import the type unless you do this, alas without success.

I am going to try some more tricks but for the sake of testing on this branch right now I commented out the imported fields since doing so does lead to a full re-generation of other classes that do have their backing fields removed! (and seem to work ok!).

nirinchev: One thing you could try and do is:

```csharp
var backingFieldName = prop.GetBackingField().FullName;
var backingField = type.Fields.Single(f => f.FullName == backingFieldName);
type.Fields.Remove(backingField);
```

Not sure if that'll work, but might be worth a try.

gagik: thank for the tip, sadly this does not work 😔 I think the issue here is that it does successfully remove the backing field, it's just that doing so causes some kind of reference to that field which is related to the backing field to fail.
I found some questions loosely related to this: [here](https://stackoverflow.com/questions/38003104/member-httpclient-is-declared-in-another-module-and-needs-to-be-imported-in), [here](https://www.reddit.com/r/csharp/comments/evo5a0/helpmonocecil_how_to_delete_a_method_from_a_type/) and [here](https://stackoverflow.com/questions/32666075/why-mono-cecil-argues-for-method-import-while-i-have-done-it-already). Sadly none of them have direct solutions, I am going to try to manually run through all methods similar to the third link's answer to figure out where the backing field is referred.

nirinchev: Okay, I think I can see what the problem is. When you have a class definition like:

```cs
partial class Foo : IRealmObject
{
    public ObjectId { get; set; } = ObjectId.GenerateNewId();
}
```

The field is getting assigned in the object constructor, even if that's not obvious from the code. The generated ctor looks kind of like this:

```
	// Id = ObjectId.GenerateNewId();
	IL_0000: ldarg.0
	IL_0001: call valuetype [MongoDB.Bson]MongoDB.Bson.ObjectId [MongoDB.Bson]MongoDB.Bson.ObjectId::GenerateNewId()
	IL_0006: stfld valuetype [MongoDB.Bson]MongoDB.Bson.ObjectId Realms.Tests.ObjectIdPrimaryKeyWithValueObject::'<Id>k__BackingField'
```

I think the issue manifests with ObjectId/Guid because those are the types we most often assign a default value to, but it'll likely happen for other types as long as you have an assignment in the model. Unfortunately, it's probably going to be a bit tricky to clean those up - we can probably bounce some ideas next week about how it can be achieved.

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -247,6 +247,36 @@ Analytics payload
...
+                        if(backingFields.Contains(field.MetadataToken)) {
+                            backingFieldInstructionsEnd = i;
+                        }
+                    }
+
+                    // If it comes across "Ldarg 0", 
+                    // it considers this th start index of backing field initializaion instructions.
+                    // And removes all backing field instructions from end to start.
+                    else if (instruction.OpCode == OpCodes.Ldarg_0) {
+                        for(var j = backingFieldInstructionsEnd; j >= i; j--) {
```
gagik, line 272: Setting `backingFieldInstructionsEnd=-1` actually takes care of cases when we never come across any backing fields beforehand, but I could also more explicitly avoid this by doing something like `if (instruction.OpCode == OpCodes.Ldarg_0 && backingFieldInstructionsEnd != -1)`

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -247,6 +247,36 @@ Analytics payload
             return WeaveModuleResult.Success(weaveResults);
         }
 
+        private void RemoveConstructorsBackingFields(TypeDefinition type, HashSet<MetadataToken> backingFields)
```
gagik, line 250: Is this the type of function I should create in terms of what I'm passing in?

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -256,16 +286,25 @@ private WeaveTypeResult WeaveGeneratedType(TypeDefinition type)
...
 
             // We need to weave all (and only) the properties in the accessor interface
             foreach (var interfaceProperty in interfaceType.Properties)
             {
                 var prop = type.Properties.First(p => p.Name == interfaceProperty.Name);
-
                 try
                 {
+                    // Stash and remove the backing field before weaving as it depends on get method.
+                    var backingField = prop.GetBackingField();
```
gagik, line 298: So if you look at source code of `.GetBackingField` it looks at `prop.GetMethod` which does get cleared when weaving, which is why I am doing this that way, maybe there is a better way to structure it though

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -256,16 +286,25 @@ private WeaveTypeResult WeaveGeneratedType(TypeDefinition type)
...
+                    if (backingField != null) {
+                        type.Fields.Remove(backingField.Resolve());
```
gagik, line 300: I am removing it at this point as it's possible there won't be any constructors referencing an existing backing field, right? This seemed like the best place to remove it, I can also move it to a function so it's cleaner

gagik (commented):
This seems to be working! At least with the classes we have so far, will probably need some more proper testing.

nirinchev (commented):
Looks reasonable - you'll need to add some spaces between `if/for/foreach` and `(` and move all the opening braces to a newline 😄 I have a low priority background task to figure out a linter that will automatically do this for you, but for now it's a matter of looking at warnings and doing what the analyzer tells you.

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -247,6 +247,36 @@ Analytics payload
...
+            foreach(var constructor in type.GetConstructors()) {
+                // Index of the most recent strfld <backing_field> instruction
+                var backingFieldInstructionsEnd = -1;
+                for(var i = constructor.Body.Instructions.Count - 1; i >= 0; i--) {
+                    var instruction = constructor.Body.Instructions[i];
+                    
+                    // If it comes across "strfld <backing_field>", 
+                    // it considers this the end index of backing field initializaion instructions.
+                    if (instruction.OpCode == OpCodes.Stfld && instruction.Operand is FieldReference) {
+                        var field = (instruction.Operand as FieldReference);
```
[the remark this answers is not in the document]
gagik: Oh cool!

Realm/Realm.Weaver/RealmWeaver.cs
```diff
@@ -256,16 +300,25 @@ private WeaveTypeResult WeaveGeneratedType(TypeDefinition type)
...
+                    var backingField = prop.GetBackingField();
+                    if (backingField != null)
+                    {
+                        backingFields.Add(backingField.MetadataToken);
+                    }
+
                     var weaveResult = WeaveGeneratedClassProperty(type, prop, interfaceType);
                     persistedProperties.Add(weaveResult);
+
+                    RemoveBackingFields(type, backingFields);
```
papafe, line 321: Why are we doing this inside the loop?

Realm/Realm.Weaver/RealmWeaver.cs
papafe: Good eye! Need to take a look at this

Tests/Realm.Tests/Database/RealmObjectTests.cs
```diff
@@ -520,6 +520,18 @@ public void RealmObject_Equals_WhenOtherIsNull_ReturnsFalse()
...
+        public void RealmObject_InitializedFields_GetCorrectValues()
+        {
+            var obj0 = new InitializedFieldObject();
+            var obj1 = new InitializedFieldObject();
+            var obj2 = new InitializedFieldObject();
+
+            Assert.That(obj0.Id, Is.EqualTo(0));
+            Assert.That(obj1.Id, Is.EqualTo(1));
+            Assert.That(obj2.Id, Is.EqualTo(2));
+        }
```
nirinchev, line 533: I think it'll be valuable to add a short comment here that explains what exactly is being tested here.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## vNext (TBD)
 
 ### Enhancements
+* Removed backing fields of Realm Objects' properties which should provide minor improvements to memory used by Realm Objects (Issue [#2647](https://github.com/realm/realm-dotnet/issues/2994))
```
nirinchev, line 4: I think we should specify this is impacting only generated classes, not `RealmObject` inheritors.

nirinchev (approved):
LGTM, only needs a minor clarification in the changelog

papafe (approved):

Issue #3086, opened 2022-11-10
[Bug]: After the Upgrade 10.18.0 - Crash Occured using Binding RealmObject on Xaml Page


Engincode:
### What happened?

We upgraded Realm-Dotnet from 10.17.0. to 10.18.0.
After the upgrade some xaml pages doesn't work.
We found the main reason;
When we use a Binding Property in a class that inherits from RealmObject errror is occuring.

Example:
**Customer.cs** (inherits from RealmObject)
![image](https://user-images.githubusercontent.com/12773298/201109434-d5f3b63d-0b50-431f-ad9c-49f2e2758d17.png)

**Xaml Page**
![image](https://user-images.githubusercontent.com/12773298/201109894-42dbfca8-a52f-43da-8eeb-b5e972de494c.png)

When the page opening
Example Error Message (Android or iOS) doesn't matter.

	0x49 in Realms.DataBinding.RealmObjectTypeDelegator.GetPropertyImpl at D:\a\realm-dotnet\realm-dotnet\Realm\Realm\DataBinding\RealmObjectTypeDelegator.cs:48,13	C#
 ............

### Repro steps

Crash.

### Version

.NET Standart 2.1

### What SDK flavour are you using?

Local Database only

### What type of application is this?

Xamarin

### Client OS and version

10.18.0.

### Stacktrace of the exception/crash you're getting

```shell
0x49 in Realms.DataBinding.RealmObjectTypeDelegator.GetPropertyImpl at D:\a\realm-dotnet\realm-dotnet\Realm\Realm\DataBinding\RealmObjectTypeDelegator.cs:48,13	C#
```

meierhoeferjannis:
Could the Problem be that he is using Compiled Bindings for the DataTemplate and Realm is doing some Black magic in the background?. 
See also #2270

papafe:
Hi @Engincode,
Could you show how you're setting up your ViewModel and initialising the bindings?

Engincode:
@papafe First of all, we were not getting any errors when we upgraded to Realm before.
We got it in the **last version.**
When you switch to the previous library; it doesn't give an error.
I shared the code sample.
What detail do you need?

papafe:
@Engincode Version 10.18.0 introduced support for source generated models, so there has been quite a lot of changes to the Realm code internally. 
[...]

PR #3097, opened 2022-11-17
Fix unmanaged schema for bindings

### LINKED ISSUE #3094: ArgumentNullException error
### Discussed in https://github.com/realm/realm-dotnet/discussions/3093

<div type='discussions-op-text'>

<sup>Originally posted by **Unreal852** November 13, 2022</sup>
Hello, i'm making an application with AvaloniaUI and i decided to use Realm as my database but when i bind my view to fields that belongs to an IRealmObject it throws ArgumentNullException. I created a separate project to replicate the issue https://github.com/Unreal852/RealmAvalonia
I did not create an issue because i don't know if i'm doing things wrong or not.</div>

papafe:
Added schema for unmanaged generated objects and fixed `RealmObjectTypeDelegator` when used with unmanaged objects

## Description

Fixes #3094 and #3086

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Realm/Realm/DataBinding/RealmObjectTypeDelegator.cs
+++ b/Realm/Realm/DataBinding/RealmObjectTypeDelegator.cs
@@ -45,7 +45,18 @@ namespace Realms.DataBinding
                 return null;
             }
 
-            if (_schema.Any(p => p.ManagedName == name))
+            if (_schema != null)
+            {
+                if (_schema.Any(p => p.ManagedName == name))
+                {
+                    return _propertyCache.GetOrAdd(name, n => new WovenPropertyInfo(result));
+                }
+
+                return result;
+            }
+
+            // Schema is null only for woven unmanaged objects
+            if (result?.GetCustomAttribute<WovenPropertyAttribute>() != null)
             {
                 return _propertyCache.GetOrAdd(name, n => new WovenPropertyInfo(result));
             }
--- a/Realm/Realm/DataBinding/RealmObjectTypeDelegator.cs
+++ b/Realm/Realm/DataBinding/RealmObjectTypeDelegator.cs
@@ -45,18 +45,8 @@ namespace Realms.DataBinding
                 return null;
             }
 
-            if (_schema != null)
-            {
-                if (_schema.Any(p => p.ManagedName == name))
-                {
-                    return _propertyCache.GetOrAdd(name, n => new WovenPropertyInfo(result));
-                }
-
-                return result;
-            }
-
             // Schema is null only for woven unmanaged objects
-            if (result?.GetCustomAttribute<WovenPropertyAttribute>() != null)
+            if (_schema?.Any(p => p.ManagedName == name) ?? result.GetCustomAttribute<WovenPropertyAttribute>() != null)
             {
                 return _propertyCache.GetOrAdd(name, n => new WovenPropertyInfo(result));
             }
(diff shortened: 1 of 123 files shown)


Realm/Realm/DataBinding/RealmObjectTypeDelegator.cs
```diff
@@ -45,7 +45,18 @@ protected override PropertyInfo GetPropertyImpl(string name, BindingFlags bindin
...
+                }
+
+                return result;
+            }
+
+            // Schema is null only for woven unmanaged objects
+            if (result?.GetCustomAttribute<WovenPropertyAttribute>() != null)
             {
                 return _propertyCache.GetOrAdd(name, n => new WovenPropertyInfo(result));
             }
```
nirinchev, line 62: All of that is very repetitive :D

Can we rework to something like:

```suggestion
        if (_schema?.Any(p => p.ManagedName == name) == true || result.GetCustomAttribute<WovenPropertyAttribute>() != null)
        {
            return _propertyCache.GetOrAdd(name, n => new WovenPropertyInfo(result));
        }
```

papafe: It is repetitive, but it's also slightly more efficient. If `schema != null`, then we don't need to look for the `WovenPropertyAttribute` anyways, as the schema will always have all the necessary info.
What you wrote was my first version, but I thought it would make sense to be a little more efficient with this if we can.

papafe: It is repetitive, but it's also slightly more efficient. If schema != null, then we don't need to look for the WovenPropertyAttribute anyway, as the schema will always have all the necessary info.
What you wrote was my first version, but I thought it would make sense to be a little more efficient with this if we can.

gagik: I am curious, would the other "or" statements be evaluated if the first one? I would have assumed if the first condition i.e. `_schema?.Any(p => p.ManagedName == name) == true` is true then it would not bother to evaluate others, thus functionally being similar to the more explicit, longer version.

papafe: You are correct, if the first condition is true then it would not check the second one. Here the situation is a little different though.
If the schema is not null and the property is found then the second condition is not evaluated and it's all good.
If the schema is not null and the property is not found, then we don't need to check the second condition anyway because we have all the properties in the schema and it will be useless. 
Only if the schema is null we should check the second condition.

LaPeste: In C# `&&` and `||` are short-circuiting, so what Nikola suggested is as efficient as the longer version.
If you specifically want to avoid the short-circuiting you need to use `&` and `|`.

LaPeste (approved):
After being explained the problem, the solution makes sense to me.

Realm/Realm/DataBinding/RealmObjectTypeDelegator.cs
papafe: That was it! I didn't think about this 😁

gagik (approved):

nirinchev (approved):

PR #3102, opened 2022-11-18
[SG] Added fully qualified types

### LINKED ISSUE #3099: [Bug]: SourceGenerated classes cant be in 'Realm' namespace
### What happened?

If you have a namespace ending in `Realm` the SG class ends up getting confused.

```
using Realms;

namespace MyApp.Realm;

public partial class Foo : IRealmObject { }
```
Results in:
> CS0738: 'Foo' does not implement interface member 'IRealmObjectBase.Realm'. 'Foo.Realm' cannot implement 'IRealmObjectBase.Realm' because it does not have the matching return type of 'Realm'.

This is because the naming collision between `MyApp.Realm` and `Realms.Realm`, and the generated class not using the fully-qualified `Realms.Realm`. I'm not sure theres a way to fix it by overriding, I think since this collides with codegen it should be simple enough to just fully qualify names on that side?

### Repro steps

See above

### Version

6

Local Database only

Other

### Client OS and version

Maui

### Code snippets

_No response_

_No response_

### Relevant log output

_No response_

papafe:
Added fully qualified names to most types in the source generated files, to avoid naming collisions

## Description

Fixes #3099

##  TODO

* [x] Changelog entry


LaPeste (approved):

nirinchev (approved):
Kind of ugly, but I guess better to be on the safe side

nirinchev:
Need to fix code generation tests though

PR #3103, opened 2022-11-21
Update to core 13.0.0

papafe:


LaPeste (approved):

nirinchev (approved):

nirinchev (changes requested):
Actually, all the tests seem to be crashing, so need to investigate that.

papafe:
@fealebenpae In PR https://github.com/realm/realm-core/pull/5980 (that was included in the version to which we are updating to) they have updated the references of logger to use a `shared_ptr` instead of a `unique_ptr`. I don't know if this has any influence in the way that we create the logger in the wrappers here: 

https://github.com/realm/realm-dotnet/blob/cf6ea7112a4182ee606a8072fbc9269fd4d4f549/wrappers/src/app_cs.cpp#L177#L180

Do you have any suggestions?

Tests/Realm.Tests/Database/RealmDictionaryTests.cs
```diff
@@ -1133,7 +1133,7 @@ public void DictionaryAsRealmQueryable_RaisesNotifications()
             Assert.That(changeSet.InsertedIndices.Length, Is.EqualTo(2));
             Assert.That(changeSet.DeletedIndices.Length, Is.EqualTo(2));
             Assert.That(changeSet.ModifiedIndices.Length, Is.EqualTo(1));
-            Assert.That(oldDogs.ElementAt(changeSet.ModifiedIndices[0]).Name, Is.EqualTo("foo"));
+            Assert.That(oldDogs.ElementAt(changeSet.NewModifiedIndices[0]).Name, Is.EqualTo("foo"));
```
nirinchev, line 1136: Why this change?

wrappers/src/app_cs.cpp
```diff
@@ -87,7 +87,7 @@ namespace realm {
             void* managed_http_client;
         };
 
-        class SyncLogger : public util::RootLogger {
+        class SyncLogger : public util::Logger {
```
nirinchev, line 90: Do we need to change the logger factory as well?

papafe: I am not completely sure if we need to do any changes to our logger factory or not given what has been done in https://github.com/realm/realm-core/pull/5980. There is a comment a little bit up in which I invoked Yavor for suggestions 😁

nirinchev: Looking at the c api changes, it seems to be a matter of replacing `make_unique` with `make_shared`: https://github.com/realm/realm-core/pull/5980/files#diff-317d315bdd0e8e2bc5b3bc3a914dcaea2ea477eb8f96b05a9a8b7dfa2ee6e980R61-R62

Tests/Realm.Tests/Database/RealmDictionaryTests.cs
papafe: It was failing on CI. The index of the newly inserted elements now is different than before. Anyways before it was working by chance, given that we were checking the order in the previous collection instead of the new one

wrappers/src/app_cs.cpp
papafe: I've pushed a commit for this, hoping it makes sense

wrappers/src/app_cs.cpp
```diff
@@ -175,9 +175,9 @@ extern "C" {
             void* managed_logger = app_config.managed_logger;
             if (managed_logger) {
                 sync_client_config.logger_factory = [managed_logger](util::Logger::Level level) {
-                    auto logger = std::make_unique<SyncLogger>(managed_logger);
+                    auto logger = std::make_shared<SyncLogger>(managed_logger);
                     logger->set_level_threshold(level);
-                    return std::unique_ptr<util::Logger>(logger.release());
+                    return std::shared_ptr<util::Logger>(logger);
```
nirinchev, line 180: Can't we return just the logger?

nirinchev (approved):

PR #3104, opened 2022-11-21
Jg/transaction state

### LINKED ISSUE #2551: Add State property to Transaction
At the moment it's not possible to know what is the state of a certain transaction, and the developer needs to be careful not to commit or rollback the transaction twice in order to avoid exceptions

JacobOscarGunnarsson:
## Description
Allows a user to check the state of a transaction so it doesn't mistakenly commit/rollback twice

Fixes #2551


Tests/Realm.Tests/Database/InstanceTests.cs
```diff
@@ -199,6 +199,22 @@ public void InstancesHaveDifferentHashes()
             Assert.That(realm1.GetHashCode(), Is.Not.EqualTo(realm2.GetHashCode()));
         }
```
JacobOscarGunnarsson, line 201: Not entirely sure if this is the correct place to put the test

papafe (changes requested):

papafe:
You should also check the CI ;) 
If something is not clear we can talk about it! The UI is a little messy

Realm/Realm/Transaction.cs
```diff
@@ -23,6 +23,28 @@
 
 namespace Realms
 {
+        /// <summary>
+        /// Represents the state of a <see cref="Transaction"/>.
+        /// </summary>
+    [Flags]
```
nirinchev, line 29: This should probably not be flags - MS has pretty good docs about what `[Flags]` designates and when to use it [here](https://learn.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-7.0).

Realm/Realm/Transaction.cs
```diff
@@ -59,6 +88,7 @@ public void Rollback()
         {
             EnsureActionFeasibility("roll back");
             _realm.SharedRealmHandle.CancelTransaction();
+            State = TransactionState.RolledBack;
             FinishTransaction();
```
nirinchev, line 92: We can probably move setting the state to `FinishTransaction` and then pass the value as an argument. Not saving a whole lot of code, but it's a bit more expressive as "this is the final thing we need to do as we close out the transaction".

Realm/Realm/Transaction.cs
```diff
@@ -23,6 +23,27 @@
 
 namespace Realms
 {
+        /// <summary>
```
nirinchev, line 26: Spacing seems off here.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## vNext (TBD)
 
 ### Enhancements
+* Added a field `Transaction.State` which describes the current state of the transaction.
```
LaPeste, line 4:
```suggestion
* Added a field `Transaction.State` which describes the current state of the transaction. (Issue [#2551](https://github.com/realm/realm-dotnet/issues/2551)
```

LaPeste (changes requested):
Just some very minor details. Otherwise, the work loos fine.

papafe:
Looks good! I would just add one test, to verify that it works also for `CommitAsync` and then if the CI agrees, it's good to go ;)

PR #3106, opened 2022-11-22
add null check for functionArgs in CallResetPasswordFunctionAsync

### LINKED ISSUE #3011: Improve error message when null is passed as argument of params for `EmailPasswordAuth.CallResetPasswordFunctionAsync`
To avoid users receiving a weird `realm-dotnet/wrappers/realm-core/src/realm/object-store/util/bson/bson.hpp:219: [realm-core-12.4.0] Assertion failed: m_type == Bson::Type::Array` (like [so](https://github.com/realm/realm-dotnet/issues/3009)) from core when passing `null` as a third to [EmailPasswordAuth.CallResetPasswordFunctionAsync](https://github.com/realm/realm-dotnet/blob/c6c3b2da60d2a0b1fc07fceed867be6fb55e8681/Realm/Realm/Sync/App.cs#L426) we should just show a readable message that explains the issue.

JacobOscarGunnarsson:
## Description
passing null as third arg will now throw a more descriptive `ArgumentNullException` error

Fixes #3011

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## vNext (TBD)
 
 ### Enhancements
+* Improved error message when null is passed as argument to params for EmailPasswordAuth.CallResetPasswordFunctionAsync. (Issue [#3011](https://github.com/realm/realm-dotnet/issues/3011))
 * Removed backing fields of generated classes' properties which should provide minor improvements to memory used by Realm Objects (Issue [#2647](https://github.com/realm/realm-dotnet/issues/2994))
 * Added two extension methods on `IDictionary` to get an `IQueryable` collection wrapping the dictionary's values:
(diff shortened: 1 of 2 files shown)


papafe:
The code looks nice, you just need to add a Changelog

LaPeste (approved):
After adding the entry in the changelog, for me it's good to go.

nirinchev (approved):

nirinchev (commented):
Let's add a small test to ensure it works and we don't break it in the future.

fealebenpae (approved):

papafe:
@nirinchev Given that we don't have tests for any of the null checks for methods in that class, I've suggested @JacobOscarGunnarsson to avoid adding one just for this. It's better if we create a ticket to add them all together

papafe (approved):

PR #3107, opened 2022-11-22
disambiguate error message when getting a defaultStorageFolder for realm or app-config

### LINKED ISSUE #2863: Config file paths for realm and app config need to have different errors
At the moment the code path for getting the default path for `Realm` and `App` is the same, and thus the exception for both is the same when the path is not writable, with message "Couldn't determine a writable folder where to store realm file. Specify absolute path manually."
We need to have two different string messages, so that developers can understand which of the two is causing issues.

JacobOscarGunnarsson:
## Description
Added helper method to allow us to tell whether we try to get a defaultStorageFolder for a realm or app-config 

Fixes #2863

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Improve performance of client reset with automatic recovery and converting top-level tables into embedded tables. (Core upgrade)
 * Flexible sync will now wait for the server to have sent all pending history after a bootstrap before marking a subscription as Complete. (Core upgrade)
 * Slightly improve performance of `Realm.RemoveAll()` which removes all objects from an open Realm database. (Issue [#2233](https://github.com/realm/realm-dotnet/issues/2194))
-* Improve error messages when setting a BaseFilePath for realm or app configuration
+* Improve error messages when not setting a BaseFilePath for realm or app configuration (Issue [2863](https://github.com/realm/realm-dotnet/issues/2863))
 
 ### Fixed
 * Fixed issue where Realm parameters' initialization would get run twice, resulting in unexpected behavior.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Improve performance of client reset with automatic recovery and converting top-level tables into embedded tables. (Core upgrade)
 * Flexible sync will now wait for the server to have sent all pending history after a bootstrap before marking a subscription as Complete. (Core upgrade)
 * Slightly improve performance of `Realm.RemoveAll()` which removes all objects from an open Realm database. (Issue [#2233](https://github.com/realm/realm-dotnet/issues/2194))
+* Improve error messages when setting a BaseFilePath for realm or app configuration
 
 ### Fixed
 * Fixed issue where Realm parameters' initialization would get run twice, resulting in unexpected behavior.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Improve performance of client reset with automatic recovery and converting top-level tables into embedded tables. (Core upgrade)
 * Flexible sync will now wait for the server to have sent all pending history after a bootstrap before marking a subscription as Complete. (Core upgrade)
 * Slightly improve performance of `Realm.RemoveAll()` which removes all objects from an open Realm database. (Issue [#2233](https://github.com/realm/realm-dotnet/issues/2194))
-* Improve error messages when not setting a BaseFilePath for realm or app configuration (Issue [2863](https://github.com/realm/realm-dotnet/issues/2863))
+* Improve error messages when not setting a BaseFilePath for realm or app configuration. (Issue [2863](https://github.com/realm/realm-dotnet/issues/2863))
 
 ### Fixed
 * Fixed issue where Realm parameters' initialization would get run twice, resulting in unexpected behavior.
--- a/Realm/Realm/InteropConfig.cs
+++ b/Realm/Realm/InteropConfig.cs
@@ -59,27 +59,19 @@ namespace Realms
                 }
             }
 
-            throw new InvalidOperationException("Couldn't determine a writable folder where to store realm file. Specify absolute path manually.");
+            return null;
         });
 
         private static string _customStorageFolder;
 
-        public static string DefaultStorageFolder
-        {
-            get => _customStorageFolder ?? _defaultStorageFolder.Value;
-            set => _customStorageFolder = value;
-        }
+        public static string GetDefaultStorageFolder(string type) =>
+            _customStorageFolder ??
+            _defaultStorageFolder.Value ??
+            throw new InvalidOperationException($"Couldn't determine a writable folder where to store {type} file. Specify absolute path manually.");
 
-        public static string TryDefaultStorageFolder(string type)
+        public static void SetDefaultStorageFolder(string value)
         {
-            try
-            {
-                return DefaultStorageFolder;
-            }
-            catch (InvalidOperationException e)
-            {
-                throw new InvalidOperationException($"Couldn't determine a writable folder where to store {type} file. Specify absolute path manually.");
-            }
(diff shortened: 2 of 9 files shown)


JacobOscarGunnarsson:
I'm a bit uncertain whether tests should be added or not for this, and if so then where

Realm/Realm/InteropConfig.cs
```diff
@@ -70,6 +70,18 @@ public static string DefaultStorageFolder
             set => _customStorageFolder = value;
         }
 
+        public static string TryDefaultStorageFolder(string type)
+        {
+            try
+            {
+                return DefaultStorageFolder;
+            }
+            catch (InvalidOperationException e)
```
LaPeste, line 79:
```suggestion
            catch (InvalidOperationException)
```
Since you're not using the exception itself

LaPeste (commented):
I'd have just put a try-catch in the `App` at line 161. Making a whole method where you have to send the string of the caller seems over done considering that 2 out of the 3 usages of the method didn't need changes.
I'd add a method of this style later along the lines if more than 1 place requires changing error messages.

But this is just my opinion. If the others prefer it this way, I'm ok too.

LaPeste:
> I'm a bit uncertain whether tests should be added or not for this, and if so then where

We don't have any on this atm. I don't think it's straight forward to test this given that you need to remove writable rights on user's Documents folder.
For UWP also doesn't look straight forward to get a not writable path given that we're specifically asking Windows for a path in the allotted sandbox where we can write.
Ah, and there is also Unity which I don't know if has even extra special needs regarding this default path.
Maybe @nirinchev can tip in some knowledge here.

Anyway, if I'm wrong and it is easier than I think, having 3 tests on this is certainly a good idea.

papafe:
I don't like the fact that we now have two exceptions. We should probably try to change the way that the default folder is build. 
Probably we can make the body of `_defaultStorageFolder` in a method already, so we don't need to try catch this too

nirinchev (changes requested):

Realm/Realm/InteropConfig.cs
```diff
@@ -70,6 +70,18 @@ public static string DefaultStorageFolder
             set => _customStorageFolder = value;
         }
 
+        public static string TryDefaultStorageFolder(string type)
+        {
+            try
+            {
+                return DefaultStorageFolder;
+            }
```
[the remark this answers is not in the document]
nirinchev: What is the method?

Realm/Realm/Sync/App.cs
```diff
@@ -158,7 +158,7 @@ public static App Create(AppConfiguration config)
             var nativeConfig = new Native.AppConfiguration
             {
                 AppId = config.AppId,
-                BaseFilePath = config.BaseFilePath ?? InteropConfig.DefaultStorageFolder,
+                BaseFilePath = config.BaseFilePath ?? InteropConfig.GetDefaultStorageFolder("app configuration"),
```
nirinchev, line 161: This would expand to

```
Couldn't determine a writable folder where to store app configuration file. Specify absolute path manually.
```

This is unfortunately very vague as 1) we don't really store app configurations and 2) it doesn't instruct the user how to specify the absolute path. While I like that we're trying to reuse some of the wording around these exception messages, I think in this case it's better to allow the consumer to specify the complete error message for maximum flexibility. Then we can supply something like:

```
Couldn't determine a writable folder to store app files (such as metadata and Realm files). When constructing the app, set AppConfiguration.BaseFilePath to an absolute path where the app is allowed to write.
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * Improve performance of client reset with automatic recovery and converting top-level tables into embedded tables. (Core upgrade)
 * Flexible sync will now wait for the server to have sent all pending history after a bootstrap before marking a subscription as Complete. (Core upgrade)
 * Slightly improve performance of `Realm.RemoveAll()` which removes all objects from an open Realm database. (Issue [#2233](https://github.com/realm/realm-dotnet/issues/2194))
+* Improve error messages when setting a BaseFilePath for realm or app configuration
```
nirinchev, line 13: This is actually the error message when not setting the `BaseFilePath` - if you set it, you wouldn't have gotten that exception. Also, we need a link to the issue, similarly to the other changelog entries.

LaPeste, line 13: Don't forget to link the gh issue.

Tests/Realm.Tests/Database/InstanceTests.cs
```diff
@@ -763,7 +763,7 @@ public void GetInstance_WhenDynamicAndDoesntExist_ReturnsEmptySchema()
         [TestCase("søren")]
         public void GetInstance_WhenPathContainsNonASCIICharacters_ShouldWork(string path)
         {
```
JacobOscarGunnarsson, line 765: this is only for internal testing so the message might be a bit verbose, but still it addresses the potential issue.. similar issue for the other test files In this PR

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * Improve performance of client reset with automatic recovery and converting top-level tables into embedded tables. (Core upgrade)
 * Flexible sync will now wait for the server to have sent all pending history after a bootstrap before marking a subscription as Complete. (Core upgrade)
 * Slightly improve performance of `Realm.RemoveAll()` which removes all objects from an open Realm database. (Issue [#2233](https://github.com/realm/realm-dotnet/issues/2194))
+* Improve error messages when not setting a BaseFilePath for realm or app configuration (Issue [2863](https://github.com/realm/realm-dotnet/issues/2863))
```
nirinchev, line 13:
```suggestion
* Improve error messages when not setting a BaseFilePath for realm or app configuration. (Issue [2863](https://github.com/realm/realm-dotnet/issues/2863))
```

Tests/Realm.Tests/Database/InstanceTests.cs
```diff
@@ -763,7 +763,7 @@ public void GetInstance_WhenDynamicAndDoesntExist_ReturnsEmptySchema()
...
-            var folder = Path.Combine(InteropConfig.DefaultStorageFolder, path);
+            var folder = Path.Combine(InteropConfig.GetDefaultStorageFolder("Could not determine a writable folder to store the Realm file. When constructing the RealmConfiguration, provide an absolute optionalPath where writes are allowed."), path);
```
nirinchev, line 766: Here we don't expect an error, so the message can be a lot shorter - something like "No error expected here".

Tests/Realm.Tests/RealmTest.cs
```diff
@@ -41,9 +41,9 @@ static RealmTest()
         {
 #if !UNITY
 #pragma warning disable CA1837 // Use Environment.ProcessId instead of Process.GetCurrentProcess().Id
-            InteropConfig.DefaultStorageFolder = Path.Combine(Path.GetTempPath(), $"rt-{System.Diagnostics.Process.GetCurrentProcess().Id}");
+            InteropConfig.SetDefaultStorageFolder(Path.Combine(Path.GetTempPath(), $"rt-{System.Diagnostics.Process.GetCurrentProcess().Id}"));
 #pragma warning restore CA1837 // Use Environment.ProcessId instead of Process.GetCurrentProcess().Id
-            Directory.CreateDirectory(InteropConfig.DefaultStorageFolder);
+            Directory.CreateDirectory(InteropConfig.GetDefaultStorageFolder("Could not determine a writable folder to store the Realm file. When constructing the RealmConfiguration, provide an absolute optionalPath where writes are allowed."));
```
nirinchev, line 46: Same here - no error expected here, let's use a shorter message (and the remaining places)

papafe (approved):
As soon as the CI finishes and everything looks good (apart from the usual suspects), for me it's mergeable

LaPeste (approved):

nirinchev (approved):

PR #3113, opened 2022-11-25
Optimize callAsync

### LINKED ISSUE #3079: Migrate to using app::call_function with string args
Once https://github.com/realm/realm-core/pull/5984 is merged, we should pick it up and remove the extra serialization/deserialization.

JacobOscarGunnarsson:
Removes unnecessary serialisation/deserialisation of args parameter passed to CallAsync

Fixes #3079

### DIFF
--- a/wrappers/src/app_cs.hpp
+++ b/wrappers/src/app_cs.hpp
@@ -126,13 +126,11 @@ namespace binding {
     using UserCallbackT = void(void* tcs_ptr, SharedSyncUser* user, MarshaledAppError err);
     using VoidCallbackT = void(void* tcs_ptr, MarshaledAppError err);
     using BsonCallbackT = void(void* tcs_ptr, realm_value_t response, MarshaledAppError err);
-    using StringCallbackT = void(void* tcs_ptr, const std::string response, MarshaledAppError err);
     using ApiKeysCallbackT = void(void* tcs_ptr, UserApiKey* api_keys, size_t api_keys_len, MarshaledAppError err);
 
     extern std::function<VoidCallbackT> s_void_callback;
     extern std::function<UserCallbackT> s_user_callback;
     extern std::function<BsonCallbackT> s_bson_callback;
-    extern std::function<StringCallbackT> s_string_callback;
     extern std::function<ApiKeysCallbackT> s_api_keys_callback;
 
     inline auto get_string_callback_handler(void* tcs_ptr) {
@@ -141,10 +139,10 @@ namespace binding {
                 std::string error_category = err->error_code.message();
                 MarshaledAppError app_error(err->message, error_category, err->link_to_server_logs, err->http_status_code);
 
-                s_string_callback(tcs_ptr, "", app_error);
+                s_bson_callback(tcs_ptr, realm_value_t{}, app_error);
             }
             else {
-                s_string_callback(tcs_ptr, *response, MarshaledAppError());
+                s_bson_callback(tcs_ptr, to_capi_value(*response), MarshaledAppError());
             }
         };
     }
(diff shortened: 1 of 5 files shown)


JacobOscarGunnarsson:
This code relies on core 13, we should probably wait for [this](https://github.com/realm/realm-dotnet/pull/3103) to be merged before merging this

wrappers/src/app_cs.hpp
```diff
@@ -133,6 +133,22 @@ namespace binding {
     extern std::function<BsonCallbackT> s_bson_callback;
     extern std::function<ApiKeysCallbackT> s_api_keys_callback;
 
+    inline auto get_string_callback_handler(void* tcs_ptr) {
+        return [tcs_ptr](const std::string* response, util::Optional<AppError> err) {
+            if (err) {
+                std::string error_category = err->error_code.message();
+                MarshaledAppError app_error(err->message, error_category, err->link_to_server_logs, err->http_status_code);
+
+                s_bson_callback(tcs_ptr, realm_value_t{}, app_error);
```
JacobOscarGunnarsson, line 142: chose to reuse the `s_bson_callback` as it might be tedious to introduce a `s_string_callback`, but that is definitely an option that is worth discussing here

wrappers/src/marshalling.hpp
```diff
@@ -523,6 +523,11 @@ class Utf16StringAccessor {
     {
         return std::string(m_data.get(), m_size);
     }
+    
+    std::string_view to_string_view() const
+    {
+        return std::string_view(m_data.get(), m_size);
+    }
```
nirinchev, line 530: We could convert that to an operator so that `Utf16StringAccessor` gets implicitly converted.

nirinchev (approved):
Looks reasonable to me

wrappers/src/marshalling.hpp
LaPeste: Is this suggestion going to be picked up?

nirinchev: I've done it in https://github.com/realm/realm-dotnet/pull/3118, so let's merge this as is.

wrappers/src/app_cs.hpp
```diff
@@ -133,6 +133,22 @@ namespace binding {
     extern std::function<BsonCallbackT> s_bson_callback;
     extern std::function<ApiKeysCallbackT> s_api_keys_callback;
 
+    inline auto get_string_callback_handler(void* tcs_ptr) {
+        return [tcs_ptr](const std::string* response, util::Optional<AppError> err) {
```
LaPeste, line 137: `response` is a `const std::string&` in the signature of `call_function`. Here you're passing it as a pointer. It'd be better you stick to have it as a reference for coherence with the original signature.
Additionally, the check at line 146 is unnecessary as response can never be null, given that it's a reference and not a pointer.

LaPeste:
The changelog should be updated. Shouldn't it?

wrappers/src/app_cs.hpp
JacobOscarGunnarsson: image

I think I'm looking at the correct overload and this one takes a `const std::string*` (on core version 13.1.0), are you looking at a different version? It won't build if I change it to const std::string& on your suggested line :/

LaPeste: Never mind. They changed the signature later on. If you look at the [PR](https://github.com/realm/realm-core/pull/5984/files) linked in the original issue you'll see that the parameter has a `const std::string&`
image

But if you look at the [current signature](https://github.com/realm/realm-core/blob/89be4acd41391ce14313ca9d29f4a010696b12d9/src/realm/object-store/sync/app_service_client.hpp#L51), it's `const std::string*`.
I should have checked the current signature and not the PR one. You'd have spotted it right away by compiling the wrappers locally. I did not test before commenting.

JacobOscarGunnarsson: No worries, would have made sense if it was the case! :D I'll revert

Issue #3114, opened 2022-11-28
Implement Remote Mongo API in managed code


nirinchev:
JS implements it in managed code using functions: https://github.com/realm/realm-js/blob/bindgen/packages/realm/src/app-services/MongoClient.ts#L190. It doesn't seem to be a whole lot of code and will reduce the binary size a little.

PR #3118, opened 2022-12-03
Implement Remote Mongo API in managed

nirinchev:
## Description

Based on https://github.com/realm/realm-dotnet/pull/3113. This implements the remote mongo API in managed code. There are 3 main benefits:
1. Removes a bunch of boilerplate code (the managed implementation turned out to be substantially less code than the glue to call into native).
2. Allows the linker to strip the native implementation from Core.
3. The binding generator doesn't support wiring up the native mongo collection API, so this moves us a step closer to being able to use it.

There are some whitespace changes, so recommend to turn them off when reviewing the code.

Fixes https://github.com/realm/realm-dotnet/issues/3114

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


Realm/Realm/Realm.csproj
```diff
@@ -61,12 +61,6 @@
     </None>
     <!-- macOS -->
     <Content Include="wrappers-props\xamarinmac\wrappers.props" Pack="true" PackagePath="build\xamarinmac\Realm.props" />
-    <None Include="..\..\wrappers\build\Darwin\$(Configuration)\librealm-wrappers.dylib">
```
LaPeste, line 64: Is this removed because of Catalyst?

nirinchev: No, this seems like something Visual Studio removed accidentally. Thanks for spotting it!

LaPeste (approved):

papafe (approved):

PR #3121, opened 2022-12-07
Add keypath filtering for Property/CollectionChanged

### LINKED ISSUE #3112: Add keypath filtering for Property/CollectionChanged
"As a first step we should add keypath filtering for our Property/CollectionChanged subscriptions as those are single-level only and it'd be a waste to compute child property diffs. As a follow-up we can expose that publicly and allow developers to define the keypaths they're interested in."

_Originally posted by @nirinchev in https://github.com/realm/realm-dotnet/issues/2917#issuecomment-1135807838_

JacobOscarGunnarsson:
Partially Fixes #3112. initially we will internally filter the notifications in the following manner:

- for objects we won't listen for events happening to collection properties at all. This include backlinks, sets, dictionaries, list and queries/results.
-  For collections, we will only listen to the shallow modifications that are done on the collection itself. this means that modifications to objects within a collection won't fire a notification

This PR needs to wait until [these changes](https://github.com/realm/realm-core/pull/6122) are merged to core


wrappers/src/dictionary_cs.cpp
```diff
@@ -155,11 +155,24 @@ extern "C" {
         delete dictionary;
     }
```
JacobOscarGunnarsson, line 157: Need help to find a good place to put a util function like `get_key_path_array()` that applies for all these cases

fealebenpae:
`computed_properties` contains only backlink/LinkingObjects properties. I think you should also include regular properties where the type is a collection.

Tests/Realm.Tests/Database/NotificationTests.cs
```diff
@@ -102,6 +102,24 @@ public void ListShouldSendNotifications()
...
+            };
+
+            _realm.Write(() =>
+            {
+                person.Friends.Add(new Person());
+                _realm.Refresh();
+                Assert.That(notificationCount, Is.EqualTo(0));
+            });
+        }
+
```
papafe, line 122: You need to call refresh after the write, otherwise there is nothing to refresh here. 
Another test would be to check if you're actually receiving notifications for all the properties of Person

wrappers/src/dictionary_cs.cpp
papafe: Yep. Probably you can generalise this method by having it accept the schema as input

Tests/Realm.Tests/Database/NotificationTests.cs
papafe: Other tests would be also to modify one of the linked persons and see that the notification does not get raised (it's already like this actually)

papafe: Also, definitely take a look at the CI results

Tests/Realm.Tests/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/TestNotificationObject_generated.cs
```diff
@@ -0,0 +1,434 @@
+﻿// <auto-generated />
```
JacobOscarGunnarsson, line 1: Not sure if this generated file should be added

Realm/Realm/DatabaseTypes/Metadata.cs
```diff
@@ -42,5 +42,15 @@ public Metadata(TableKey tableKey, IRealmObjectHelper helper, IDictionary<string
...
         }
+
+        public static IntPtr GetPropertyIndex(string propertyName, Metadata metadata)
+        {
+            if (metadata.PropertyIndices.TryGetValue(propertyName, out var result))
+            {
+                return result;
+            }
+
+            throw new MissingMemberException(metadata.Schema.Name, propertyName);
```
papafe, line 53: Why did you move this?

papafe (commented):
I think the tests make sense! Probably it'll be worth adding also collections of other types, in the same way you did for single link

Tests/Realm.Tests/Database/TestNotificationObject.cs
```diff
@@ -0,0 +1,48 @@
...
+using TestRealmObject = Realms.IRealmObject;
+#endif
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Realms.Tests.Database
+{
+    public partial class TestNotificationObject : TestRealmObject
+    {
+        // Automatically implemented (overridden) properties
```
papafe, line 35: I don't think you need this comment

Realm/Realm/DatabaseTypes/Metadata.cs
JacobOscarGunnarsson: After talking with @fealebenpae we concluded that it made more sense to have the method in the Metadata itself. Please reach out if you disagree. (This might also add some noise to the PR itself as well..)

Realm/Realm/Handles/ObjectHandle.cs
```diff
@@ -351,11 +354,12 @@
...
         }
 
-        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle)
+        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, List<IntPtr> propertyIndices = null)
         {
             EnsureIsOpen();
 
-            var result = NativeMethods.add_notification_callback(this, managedObjectHandle, out var nativeException);
+            IntPtr[] pIArray = propertyIndices.ToArray();
+            var result = NativeMethods.add_notification_callback(this, managedObjectHandle, pIArray, (IntPtr)pIArray.Length, out var nativeException);
```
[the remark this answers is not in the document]
JacobOscarGunnarsson: I assume I can dismiss this alert since the whole idea is to call into unmanaged code?

Realm/Realm/DatabaseTypes/Accessors/ManagedAccessor.cs
```diff
@@ -181,12 +181,15 @@ public void SubscribeForNotifications(Action<string> notifyPropertyChangedDelega
                 throw new RealmFrozenException("It is not possible to add a change listener to a frozen RealmObjectBase since it never changes.");
             }
 
+            PropertyType collectionFlag = PropertyType.Array | PropertyType.Dictionary | PropertyType.Set;
```
papafe, line 184: Probably we also need to check backlinks? I'm not sure we get notifications for those in any case.

Also, it would make sense to calculate those indices only once, they are not going to change

Realm/Realm/Handles/ObjectHandle.cs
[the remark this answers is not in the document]
papafe: Yes, we always get those. We know better than CodeQL in this case 😏

wrappers/src/object_cs.cpp
```diff
@@ -280,11 +281,21 @@ extern "C" {
         });
     }
 
-    REALM_EXPORT ManagedNotificationTokenContext* object_add_notification_callback(Object* object, void* managed_object, NativeException::Marshallable& ex)
+    REALM_EXPORT ManagedNotificationTokenContext* object_add_notification_callback(Object* object, void* managed_object, size_t* property_indices, size_t property_count, NativeException::Marshallable& ex)
     {
+        TableKey tk = object->get_object_schema().table_key;
```
papafe, line 286: As we discussed before, this probably will need to go in his separate method, but for now it's fine.

wrappers/src/object_cs.cpp
```diff
@@ -280,11 +281,21 @@ extern "C" {
...
+        ColKey ck;
+        KeyPathArray keyPathArray;
+        for (int i = 0; i < property_count; i++) {
+            KeyPath keyPath;
+            auto prop = get_property(*object, property_indices[i]);
+            ck = prop.column_key;
+            keyPath.push_back(std::make_pair(tk, ck));
+            keyPathArray.push_back(keyPath);
```
papafe, line 294: Do you need to create `keyPath` at all? Can't you just pass `std::make_pair` to the `push_back` method? Not very good with C++. so it's more of a curiosity 😁

Tests/Realm.Tests/Database/NotificationTests.cs
```diff
@@ -195,6 +195,179 @@ public void DictionaryNotificationToken_KeepsCollectionAlive()
...
+            Assert.That(notificationCount, Is.EqualTo(1));
+        }
+
+        [Test]
+        public void CollectionPropertiesOfDifferentTypeShouldNotFireNotificationsOnChange()
+        {
+            var testObject = _realm.Write(() => _realm.Add(new TestNotificationObject()));
+            testObject.PropertyChanged += (sender, e) =>
+            {
+                Assert.That(e.PropertyName, Is.EqualTo("StringProperty"));
```
papafe, line 222: I don't think this is the most correct approach. For example, the test will pass even if there will be a notification for StringProperty when you modify something else. 
You probably should create a list of strings and just add the `PropertyName` to the list in `PropertyChanged` after every `_realm.Refresh()` you can check if the list contains the elements you expect. Finally, you clear the list so it's empty for next assertion.

Tests/Realm.Tests/Database/NotificationTests.cs
```diff
@@ -195,6 +195,179 @@ public void DictionaryNotificationToken_KeepsCollectionAlive()
...
+                _realm.Remove(set.First());
+                dictionary.Remove("foo");
+            });
+            _realm.Refresh();
+        }
+
+        [Test]
+        public void Link_ShouldOnlyFireNotificationForReassignment()
+        {
+            var testObject = new TestNotificationObject();
```
papafe, line 344: I think you should test if `StringProperty`'s notifications fire properly also here.

Realm/Realm/DatabaseTypes/Accessors/ManagedAccessor.cs
```diff
@@ -181,12 +181,15 @@ public void SubscribeForNotifications(Action<string> notifyPropertyChangedDelega
...
+            List<IntPtr> propertyIndices = (from p in ObjectSchema where (p.Type & collectionFlag) == 0 select Metadata.GetPropertyIndex(p.Name, Metadata)).ToList();
```
fealebenpae, line 185:
```suggestion
            IntPtr[] propertyIndices = (from p in ObjectSchema where (p.Type & collectionFlag) == 0 select Metadata.GetPropertyIndex(p.Name, Metadata)).ToArray();
```
We don't need a mutable list here, an array is enough. And yes, they can be cached like @papafe says.

Realm/Realm/DatabaseTypes/Accessors/ManagedAccessor.cs
fealebenpae: Backlinks are in a separate vector in the Object Store `ObjectSchema` so even if they cause notifications they need to be handled separately. But we need to see if they raise notifications in the first place.

Realm/Realm/DatabaseTypes/Metadata.cs
```diff
@@ -42,5 +42,15 @@ public Metadata(TableKey tableKey, IRealmObjectHelper helper, IDictionary<string
             PropertyIndices = new ReadOnlyDictionary<string, IntPtr>(propertyIndices);
             Schema = schema;
         }
+
+        public static IntPtr GetPropertyIndex(string propertyName, Metadata metadata)
```
fealebenpae, line 46: This should be an instance method, since it's on the `Metadata` class itself, instead of a static method that takes a `Metadata` instance argument.

Realm/Realm/Handles/DictionaryHandle.cs
```diff
@@ -136,7 +136,7 @@ public override void Clear()
             nativeException.ThrowIfNecessary();
         }
 
-        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle)
+        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, List<IntPtr> propertyIndices = null)
```
fealebenpae, line 139:
```suggestion
        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, IntPtr[] propertyIndices = null)
```

Realm/Realm/Handles/ListHandle.cs
```diff
@@ -213,7 +214,7 @@ public void Move(IntPtr sourceIndex, IntPtr targetIndex)
             nativeException.ThrowIfNecessary();
         }
 
-        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle)
+        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, List<IntPtr> propertyIndices = null)
```
fealebenpae, line 217:
```suggestion
        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, IntPtr[] propertyIndices = null)
```

Realm/Realm/Handles/NotifiableObjectHandleBase.cs
```diff
@@ -51,7 +52,7 @@ protected NotifiableObjectHandleBase(SharedRealmHandle root, IntPtr handle) : ba
         {
         }
 
-        public abstract NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle);
+        public abstract NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, List<IntPtr> propertyIndices = null);
```
fealebenpae, line 55:
```suggestion
        public abstract NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, IntPtr[] propertyIndices = null);
```

Realm/Realm/Handles/ResultsHandle.cs
```diff
@@ -159,7 +160,7 @@ public SortDescriptorHandle GetSortDescriptor()
             return new SortDescriptorHandle(Root, result);
         }
 
-        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle)
+        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, List<IntPtr> propertyIndices = null)
```
fealebenpae, line 163:
```suggestion
        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, IntPtr[] propertyIndices = null)
```

Realm/Realm/Handles/SetHandle.cs
```diff
@@ -133,7 +134,7 @@ public override void Clear()
             nativeException.ThrowIfNecessary();
         }
 
-        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle)
+        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, List<IntPtr> propertyIndices = null)
```
fealebenpae, line 137:
```suggestion
        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, IntPtr[] propertyIndices = null)
```

Realm/Realm/Handles/ObjectHandle.cs
```diff
@@ -351,12 +354,13 @@ public override ThreadSafeReferenceHandle GetThreadSafeReference()
             return new ThreadSafeReferenceHandle(result);
         }
 
-        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle)
+        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, List<IntPtr> propertyIndices = null)
```
fealebenpae, line 357:
```suggestion
        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, IntPtr[] propertyIndices = null)
```

fealebenpae (changes requested):

wrappers/src/object_cs.cpp
JacobOscarGunnarsson: If I'm not mistaken we cannot bypass this, `KeyPath` and `KeyPathArray` are defined ass 

`using KeyPath = std::vector<std::pair<TableKey, ColKey>>;`
`using KeyPathArray = std::vector<KeyPath>;`

`add_notification_callback` takes a nested vector since you can provide multiple keyPaths at once if you do it manually (example: "foo.bar.prop1", "foo.bar.prop2". In our case the `keyPath` doesn't really make sense since we only add one element to it (that would be "foo" with the same example). But it's mostly there now to comply with the parameter we need to provide. In the future it will also make more sense once we pass keyPaths ourself

wrappers/src/dictionary_cs.cpp
```diff
@@ -155,11 +155,15 @@ extern "C" {
...
         return handle_errors(ex, [=]() {
-            return subscribe_for_notifications(managed_dict, [dictionary](CollectionChangeCallback callback) {
-                return dictionary->add_notification_callback(callback);
+            return subscribe_for_notifications(managed_dict, [dictionary, property_indices](CollectionChangeCallback callback) {
+                if (!property_indices) {
+                    return dictionary->add_notification_callback(callback);
+                } else {
+                    return dictionary->add_notification_callback(callback, KeyPathArray());
+                }
             });
```
JacobOscarGunnarsson, line 167: this part might look slightly awkward for now since I completely disregard the `property_indices` if it actually contains something. in the future though when we want to pass key paths ourselves it would look something like 

                 if (!property_indices) {
                     return dictionary->add_notification_callback(callback);
                 } else if (property_count == 0){
                     return dictionary->add_notification_callback(callback, KeyPathArray());
                 } else {
                     return dictionary->add_notification_callback(callback, build_key_path_array(property_indices));
                 }

Realm/Realm/DatabaseTypes/Accessors/ManagedAccessor.cs
```diff
@@ -169,6 +171,18 @@ public IRealmObjectBase GetParent()
...
 
+        private IntPtr[] GetPropertyIndices()
+        {
+            if (_propertyIndices != null)
+            {
+                return _propertyIndices;
+            }
+
+            const PropertyType collectionFlag = PropertyType.Array | PropertyType.Dictionary | PropertyType.Set;
+            _propertyIndices = (from p in ObjectSchema where (p.Type & collectionFlag) == 0 select Metadata.GetPropertyIndex(p.Name)).ToArray();
```
fealebenpae, line 182:
```suggestion
            _propertyIndices = ObjectSchema.Where(p => !p.Type.HasFlag(collectionFlag).Select(p => Metadata.GetPropertyIndex(p.Name).ToArray();
```
I think this reads clearer, specifically the usage of [Enum.HasFlag](https://learn.microsoft.com/en-us/dotnet/api/system.enum.hasflag).

fealebenpae (changes requested):
A few small comments. I'd also like to see the PR updated with the Core submodule pointing to the Core branch so I can see the related wrappers changes in full as well.

Realm/Realm/DatabaseTypes/Accessors/ManagedAccessor.cs
JacobOscarGunnarsson: I tried this earlier and didn't seem to succeed, now I also tried to adopt your suggestion but that made tests failed. After reading about it I found this snippet which implies that the semantics are not really what I want. 
image

if I know my logic that would imply that I have to write something in terms of 

`(p.Type.HasFlag(PropertyType.Array) || p.Type.HasFlag(PropertyType.Dictionary || p.Type.HasFlag(PropertyType.Set)))`

which looks slightly bloated IMO..

fealebenpae: You're right, my usage of `HasFlag` is not correct here. Luckily I found out we have an [`IsCollection()` extension method on `PropertyType`](https://github.com/realm/realm-dotnet/blob/f7a10a89d0e8ea3cc8d91056e7f63c572f983d6c/Realm/Realm/Schema/PropertyTypeEx.cs#L205-L224), so you should be able to do something like `ObjectSchema.Where(p => !p.Type.IsCollection(out var _))` (we deliberately discard the out parameter because we don't care for it).

Realm/Realm/DatabaseTypes/Accessors/ManagedAccessor.cs
JacobOscarGunnarsson: Added a test to check this, and they don't in our case

wrappers/src/dictionary_cs.cpp
```diff
@@ -155,11 +155,16 @@ extern "C" {
         delete dictionary;
     }
```
JacobOscarGunnarsson, line 157: I'm starting to believe that these colleciton-methods could be merged into something like

`realm_collection_add_notification_callback(object_store::Collection* collection, void* managed_collection, NativeException::Marshallable& ex)`

PR #3139, opened 2022-12-20
Added diagnostic error for nested classes

### LINKED ISSUE #3130: Add diagnostic for nested classes
At the moment our Source Generator does not support nested classes. We need to emit an error diagnostic for those

papafe:
Fixes #3130 

##  TODO

* [x] Changelog entry
* [x] Tests

### DIFF
--- a/Realm/Realm.SourceGenerator/Diagnostics.cs
+++ b/Realm/Realm.SourceGenerator/Diagnostics.cs
@@ -50,6 +50,7 @@ namespace Realms.SourceGenerator
             TypeNotSupported = 24,
             RealmObjectWithoutAutomaticProperty = 25,
             NotPersistedPropertyWithRealmAttributes = 26,
+            NestedClass = 27,
         }
 
         #region Errors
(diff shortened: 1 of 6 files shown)


fealebenpae (approved):

Realm/Realm.SourceGenerator/Diagnostics.cs
```diff
@@ -271,6 +272,15 @@ public static Diagnostic TypeNotSupported(string className, string propertyName,
                 location);
         }
 
+        public static Diagnostic NestedClass(string className, Location location)
+        {
+            return CreateDiagnosticError(
+                Id.NestedClass,
+                "Nested classes are not supported yet.",
+                $"{className} is a nested class, which is not yet supported.",
```
nirinchev, line 280: Can we emit this as `{parentClass}.{className}`? Not critical, I just feel like it'll make it more obvious that it's nested and where it's nested in.

nirinchev (approved):

Realm/Realm.SourceGenerator/Diagnostics.cs
papafe: Fixed

Issue #2747, opened 2021-12-09
[Unity][Bug] string.Contains() throws NotSupportedException in Unity 2021.2


sandr01d:
**Realm Version**: 10.7.1
**Unity Version**: 2021.2.5f1
**Client OS**: Windows 10 Pro 21H1 19043.1348

I recently upgraded my Unity version from 2020.3 to 2021.2. Since the update, calling string.Contains() in a realm query throws a NotSupportedException. With Unity 2020.3 results get returned as expected and no exception is thrown. My model looks like this:

```cs
public class Tag : RealmObject
{
    [PrimaryKey]
    public int Uid { get; set; }
    [Required]
    public string Name { get; set; }
}
```
and the class executing the query like this:

```cs
using Realms;
using System.IO;
using System.Linq;
using UnityEngine;

public class RealmExample : MonoBehaviour
{
    private static Realm realm;

    private void Awake()
    {
        RealmConfiguration config = new(Path.Combine(Application.dataPath, "database.realm"))
        {
            IsReadOnly = true,
            Schema = new[]
            {
                typeof(Tag),
            },
            SchemaVersion = 11ul
        };
        realm = Realm.GetInstance(config);
    }

    [ContextMenu("QueryTags")]
    public void QueryTags()
    {
        string substring = "T";
        IQueryable<Tag> tags = realm.All<Tag>()
                                    // The line below is the one causing the exception
                                    .Where(t => t.Name.Contains(substring, System.StringComparison.OrdinalIgnoreCase));

        foreach (var t in tags)
        {
            print(t.Name);
        }
    }
}
```

I receive the following exception:

```
NotSupportedException: The left-hand side of the Call operator must be a direct access to a persisted property in Realm.
Unable to process 'value(RealmExample+<>c__DisplayClass2_0).substring'.
Realms.RealmResultsVisitor.GetColumnName (System.Linq.Expressions.MemberExpression memberExpression, System.Nullable`1[T] parentType) (at <5ee71c2d3038455ca732a3febcca3879>:0)
Realms.RealmResultsVisitor.VisitMethodCall (System.Linq.Expressions.MethodCallExpression node) (at <5ee71c2d3038455ca732a3febcca3879>:0)
System.Linq.Expressions.MethodCallExpression.Accept (System.Linq.Expressions.ExpressionVisitor visitor) (at <61774763be294c9f8e2c781f10819224>:0)
[...]

PR #3140, opened 2022-12-20
Fix handling query with string.Compare(string, StringComparison)

### LINKED ISSUE #3134: [Bug]: Queries with string.Contains(string, StringComparison) fails on .NET 6+
### What happened?

In .NET 6 (or 7), `Where` queries which contains `string.Contains(string, StringComparison)` fails to execute by `NotSupportedException`.
(See the code snippet)

I found that it seems to be caused by a workaround introduced by #1854 and this seems not to suit with .NET's string.Compare instance method.

I also wrote a fix of this bug, so I can submit a pull request if it can be accepted.

### Repro steps

1. In a query for Realm, use `string.Compare(string value, StringComparison comparisonType)` instance method (see the code snippet)
    * The first argument (`value`) must not be literal. This issue only happens if the `value` is `MemberExpression` (closure variable satisfies this)
2. Querying fails with `System.NotSupportedException`. (see the stacktrace)

### Version

Realm 10.18.0 on .NET 7.0.101

Local Database only

Console/Server

### Client OS and version

Windows (x64) 10.0.22623.1028

### Code snippets

```csharp
using Realms;
class Klass : RealmObject
{
    public string Text { get; set; }
}

class Program
{
    static void Main()
    {
        var realm = Realm.GetInstance("foo.realm");
        realm.Write(() => realm.Add(new Klass { Text = "foobar" }));
        var str = "FOO"; // for reproduction, the first argument of Compare must not be literal
        foreach (var x in realm.All<Klass>().Where(x => x.Text.Contains(str, StringComparison.OrdinalIgnoreCase)))
                                                     // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <- crash!
            Console.WriteLine(x.Text);
    }
}
```

```shell
Unhandled exception. System.NotSupportedException: The left-hand side of the Call operator must be a direct access to a persisted property in Realm.
Unable to process 'value(Program+<>c__DisplayClass0_0).str'.
   at Realms.RealmResultsVisitor.GetColumnName(MemberExpression memberExpression, Nullable`1 parentType)
   at Realms.RealmResultsVisitor.VisitMethodCall(MethodCallExpression node)
   at Realms.RealmResultsVisitor.VisitMethodCall(MethodCallExpression node)
   at Realms.RealmResults`1.GetOrCreateHandle()
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at Realms.RealmCollectionBase`1.get_IsValid()
   at Realms.RealmCollectionBase`1.Enumerator..ctor(RealmCollectionBase`1 parent)
   at Realms.RealmCollectionBase`1.GetEnumerator()
   at Program.Main() in C:\w\ConsoleApp1\ConsoleApp1\Program.cs:line 15
```
```


### Relevant log output

_No response_

takeshik:
## Description

This fixes #3134.
`RealmResultsVisitor` accidentally parses `CallExpression` with `Compare(string, StringComparison)` as an extension method call, even if it's a true instance method of `string`, not `StringExtensions.Contains(string, string, StringComparison)` call.

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Linq/RealmResultsVisitor.cs
+++ b/Realm/Realm/Linq/RealmResultsVisitor.cs
@@ -58,7 +58,10 @@ namespace Realms
             {
                 internal static readonly LazyMethod Contains = Capture<string>(s => s.Contains(string.Empty));
 
+#if !NETCOREAPP2_1_OR_GREATER
+                // This should indicate StringExtensions.Contains method in Realm, defined only if .NET Core 2.0 or earlier is targeted.
                 internal static readonly LazyMethod ContainsStringComparison = Capture<string>(s => s.Contains(string.Empty, StringComparison.Ordinal));
+#endif
 
                 internal static readonly LazyMethod Like = Capture<string>(s => s.Like(string.Empty, true));
 
@@ -313,10 +316,15 @@ namespace Realms
                 {
                     queryMethod = (q, r, p, v) => q.StringContains(r, p, v, caseSensitive: true);
                 }
-                else if (IsStringContainsWithComparison(node.Method, out var index))
+                else if (IsStringContainsWithComparison(node.Method, out var isExtensionMethod))
                 {
-                    member = node.Arguments[0] as MemberExpression;
-                    stringArgumentIndex = index;
+                    if (isExtensionMethod)
+                    {
+                        // This is an extension method, so the string to compare against is at position 1.
+                        member = node.Arguments[0] as MemberExpression;
+                        stringArgumentIndex = 1;
+                    }
+
                     queryMethod = (q, r, p, v) => q.StringContains(r, p, v, GetComparisonCaseSensitive(node));
                 }
                 else if (AreMethodsSame(node.Method, Methods.String.StartsWith.Value))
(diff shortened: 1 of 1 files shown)


takeshik:
Just now I found an other PR #2753 (or issue #2747).
At a glance, I could not determine if the PR #2753 would solve the issue I reported, anyway this PR may or may not be a duplicate of it.
So please close it if it's a duplicate and not needed.

nirinchev:
Hey, thanks for the PR - it makes sense. I put up https://github.com/realm/realm-dotnet/pull/3154 that's inspired from your PR with slightly fewer changes and added some tests that ensure that reproduced the issue.

nirinchev:
Closing this since #3154 was merged.

PR #3148, opened 2023-01-04
Fix generated usings

nirinchev:
## Description

Fixes the test projects not building due to https://github.com/realm/realm-dotnet/pull/3138


papafe (approved):
Looks good, sorry for the mess

PR #3168, opened 2023-01-09
Updated example project to MAUI and Source Generators

papafe:
Updated example project to use MAUI and the new source generator models. 
I've also slightly modified the way the example work. Before the user had to explicitly press "Save" to save his entry, and we had an open transaction to implement this. This is a pattern we don't recommend, so I've modified this, and all the changes are directly saved in the db. If when returning to the main screen the note that was being modified is empty, then it's removed and a toast appears (as it happens with Google Notes). 
This is implemented using the MessagingCenter. The initial implementation was specifying `Shell.BackButtonBehaviour`, but this was causing a crash due to a bug in MAUI (https://github.com/dotnet/maui/pull/11438). 

Fixes also #3032


examples/QuickJournal/ViewModels/JournalEntriesViewModel.cs
```diff
@@ -0,0 +1,80 @@
...
+using QuickJournal.Models;
+using Realms;
+
+namespace QuickJournal.ViewModels
+{
+    public partial class JournalEntriesViewModel : ObservableObject
+    {
+        private readonly Realm realm;
+
+        [ObservableProperty]
```
nirinchev, line 15: Does this need to be observable property?

examples/QuickJournal/ViewModels/JournalEntriesViewModel.cs
```diff
@@ -0,0 +1,80 @@
...
+        private IQueryable<JournalEntry> entries;
+
+        public JournalEntriesViewModel()
+        {
+            realm = Realm.GetInstance();
+            Entries = realm.All<JournalEntry>();
+
+            WeakReferenceMessenger.Default.Register<EntryModifiedMessage>(this, EntryModifiedHandler);
```
nirinchev, line 23: Does this need the generic arg?

examples/QuickJournal/ViewModels/JournalEntriesViewModel.cs
```diff
@@ -0,0 +1,80 @@
...
+                Metadata = new EntryMetadata
+                {
+                    CreatedDate = DateTimeOffset.Now,
+                }
+            };
+
+            realm.Write(() =>
+            {
+                realm.Add(entry);
+            });
```
nirinchev, line 40:
```suggestion
            var entry = await realm.WriteAsync(() =>
            {
                return realm.Add(new JournalEntry
                {
                    Metadata = new EntryMetadata
                    {
                        CreatedDate = DateTimeOffset.Now,
                    }
                });
            });
```

examples/QuickJournal/ViewModels/JournalEntriesViewModel.cs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        [RelayCommand]
+        public void DeleteEntry(JournalEntry entry)
+        {
+            realm.Write(() =>
+            {
+                realm.Remove(entry);
+            });
+        }
```
nirinchev, line 58:
```suggestion
        public async Task DeleteEntry(JournalEntry entry)
        {
            await realm.WriteAsync(() =>
            {
                realm.Remove(entry);
            });
        }
```

examples/QuickJournal/ViewModels/JournalEntriesViewModel.cs
```diff
@@ -0,0 +1,80 @@
...
+            };
+            await Shell.Current.GoToAsync($"entryDetail", navigationParameter);
+        }
+
+        private async void EntryModifiedHandler(object recipient, EntryModifiedMessage message)
+        {
+            var newEntry = message.Value;
+            if (string.IsNullOrEmpty(newEntry.Body + newEntry.Title))
+            {
+                DeleteEntry(newEntry);
```
nirinchev, line 74:
```suggestion
                await DeleteEntry(newEntry);
```

examples/QuickJournal/ViewModels/JournalEntriesViewModel.cs
```diff
@@ -0,0 +1,80 @@
...
+        private async Task GoToEntry(JournalEntry entry)
+        {
+            var navigationParameter = new Dictionary<string, object>
+            {
+                { "Entry", entry },
+            };
+            await Shell.Current.GoToAsync($"entryDetail", navigationParameter);
+        }
+
+        private async void EntryModifiedHandler(object recipient, EntryModifiedMessage message)
```
nirinchev, line 69: Might be a good idea to add a comment for why we're going through the `WeakReferenceManager` rather than a more intuitive back button handler.

nirinchev (commented):
Looks good, minor comments and suggestions from me.

examples/QuickJournal/ViewModels/JournalEntriesViewModel.cs
papafe: In this case we don't, but I think it's a good pattern nevertheless. If for example we want to use `Realm.GetInstanceAsync` and we want to do it in `OnAppearing`, then we need it.

examples/QuickJournal/ViewModels/JournalEntriesViewModel.cs
papafe: Yes we need it, it can't be inferred

examples/QuickJournal/Platforms/Tizen/Main.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using Microsoft.Maui;
+using Microsoft.Maui.Hosting;
+
+namespace QuickJournal;
+
+class Program : MauiApplication
+{
+    protected override MauiApp CreateMauiApp() => MauiProgram.CreateMauiApp();
```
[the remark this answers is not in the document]
papafe: Didn't you get the memo? 😁

examples/QuickJournal/ViewModels/JournalEntriesViewModel.cs
papafe: Done

examples/QuickJournal/ViewModels/JournalEntriesViewModel.cs
papafe: Done

examples/QuickJournal/ViewModels/JournalEntriesViewModel.cs
papafe: Done

nirinchev (approved):

PR #3170, opened 2023-01-09
Update to Core 13.2.0

nirinchev:
## Description

Fixes #

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


nirinchev:
@fealebenpae I think that's ready to go, but can you double check my cmake/workflow changes as well as the c++ ones and let me know if something looks suspicious.

Issue #3085, opened 2022-11-09
Source generation doesn't support non-nullable object types


peppy:
### What happened?

We are keen to switch to the new source generator, but are blocked by it not supporting non-nullable object types. We have been using realm with non-nullable object types for a good year and it definitely works well, so I would hope that this can be supported by the source generator.

Currently it is disabled via this check:

https://github.com/realm/realm-dotnet/blob/a7b8f3a749d1a863ac617acee656bcf179ed7cd9/Realm/Realm.SourceGenerator/InfoClasses.cs#L250-L254

I'm not sure if this is to be considered a bug or feature request, but it's a blocker for us to use source generation unfortunately.

### Repro steps

- Create a class with non-nullable objects.
- Receive build errors.

### Version

6.x

### What SDK flavour are you using?

Local Database only

### What type of application is this?

Other

### Client OS and version

10.18.0

### Code snippets

An example of a class that would work previously:

```csharp
public partial class RealmNamedFileUsage : IEmbeddedObject
{
    public RealmFile File { get; set; } = null!; // <- problematic line

    public string Filename { get; set; } = null!;

    public RealmNamedFileUsage(RealmFile file, string filename)
    {
        File = file ?? throw new ArgumentNullException(nameof(file));
        Filename = filename ?? throw new ArgumentNullException(nameof(filename));
    }

    private RealmNamedFileUsage()
    {
    }
}

[MapTo("File")]
public partial class RealmFile : IRealmObject
{
    [PrimaryKey]
    public string Hash { get; set; } = string.Empty;
}
```

Our codebase branch with all changes applied that we hope would make source generation work correctly: https://github.com/ppy/osu/compare/master...peppy:osu:realm-source-gen?expand=1

### Stacktrace of the exception/crash you're getting

```shell
Program.cs(19, 5): [RLM010] RealmNamedFileUsage.File has type RealmFile, that does not support the assigned nullability annotiation.
```

nirinchev:
The problem is that properties linking to realm objects are inherently nullable. Since you can always do something like:

```csharp
RealmNamedFileUsage usage = ...;
realm.Remove(usage.File);
```

Now, `usage.File` will be `null`, which makes the nullability annotation incorrect. While your app may not be doing that, there's no way to statically guarantee that, which is why we opted for the safer approach of requiring that all `RealmObject` properties are nullable.

peppy:
[...]

PR #3172, opened 2023-01-10
Add nullability support to source generator

### LINKED ISSUE #2982: [SG] Add support for nullability annotations
We need to support the new nullability annotations. We need to be careful because those annotations can be enabled (at different levels) at the project level, but also via directives for single files

papafe:
This PR adds support for nullability annotations with the source generator. The expectations are that:
-   Nullability will continue to work the same with value types:

	-   If declared as non nullable, it's equivalent to use [Required]. We raise a diagnostic error if [Required] is used instead of nullability

-   For `string` and `byte[]`:

	-   If nullability annotations are off, then it works like now. Developers can use `[Required]` to indicate a required property
	-   If nullability annotations are on. If property is declared as non nullable, it's equivalent to [Required]. We raise a diagnostic error if [Required] is used instead of nullability

-   For objects(links)

	-   If nullability annotations are on, by default we will raise a Diagnostic error if they are not annotated as nullable (this already happens)
	-   This could be too strict for some, so we can add a config options to ignore the nullability annotation on those property

-   For collections:

	-   Collections themselves cannot be annotated as nullable (and shouldn't be), otherwise we raise a diagnostic error.
	-   For the generic types, all the considerations already mentioned apply

-   For RealmValues

	-   Cannot be annotated as nullable

-   For RealmInteger

	-   Can be annotated as nullable, behaves as the underlying type

## Description

Fixes #2982 and #3085

##  TODO

* [x] Changelog entry
* [x] Tests


papafe:
It needs some polish, but the main elements are there

nirinchev (changes requested):
Not sure if I'm missing something or it was already supported before, but it doesn't look like there's any code changes related to string and byte[].

CHANGELOG.md
```diff
@@ -9,6 +9,44 @@
...
+
+      //Backlink
+      [Realms.Backlink("...")]
+      public IQueryable<Dog> MyDogs { get; } //Correct
+
+      [Realms.Backlink("...")]
+      public IQueryable<Dog?> MyDogs { get; } //Error
+  }
+  ```
+  We realise that some developers would still prefer to have more freedom in the nullability annotation of such properties, and it is possible to do so by setting  `realm.ignore_objects_nullability = true` in a global configuration file (more information about global configuration files can be found in the [.NET documentation](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files)). If this is enabled, all the previous properties will be considered valid, and the nullability annotations for properties linking to objects will be ignored.
```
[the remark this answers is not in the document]
papafe: For those the nullability has the same effect of `Required`, so I think it makes sense to not give the possibility of disabling the nullability checks

CHANGELOG.md
```diff
@@ -9,6 +9,44 @@
 * Upgrade OpenSSL from 1.1.1n to 3.0.7. (Core 13.2.0)
 * Converting flexible sync realms to bundled and local realms is now supported (Core 13.2.0)
 * Add support for nested classes for source generated classes. (Issue [#3031](https://github.com/realm/realm-dotnet/issues/3031))
+* Enhanced support for nullable reference types in the model definition. This allows to use realm models as usual when nullable context is active, and removes the need to use of the `Required` attribute to indicate required properties, as this information will be inferred directly from the nullability status. There are some considerations regarding the nullability of properties that link to realm object:
```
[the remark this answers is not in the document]
papafe: Done

Realm/Realm.SourceGenerator/ClassCodeBuilder.cs
```diff
@@ -268,11 +270,11 @@ private string GeneratePartialClass(string interfaceString, string managedAccess
 
 #region {baseInterface} implementation
 
-private {_accessorInterfaceName} _accessor;
+private {_accessorInterfaceName} _accessor = null!;
```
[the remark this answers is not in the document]
papafe: Reasonable

J-Swift:
If I were to use `ignore_objects_nullability` and do something like this
```
public partial class Person: IRealmObject
{
      public Dog MyDog { get; set; }
}
```
but not provide a `Dog` when adding to the realm, would I get a warning (or even better, an error similar to how `Required` worked)? Thats what I would want as a dev who is in favor of having non-nullable references.

nirinchev:
To shine a little light on why that is, it's because at any point you could do something like:

```csharp
var dog = new Dog();
var person = new Person { Dog = new Dog() };

// Here person.Dog is not null
realm.Write(() => realm.Add(person));

// Here we're deleting the Dog from the Realm
realm.Write(() => realm.Remove(dog));

Console.Write(person.Dog); // null
```

If we were to enforce that no `person.Dog` may be null, that would make object deletions quite expensive and notably would introduce extremely complex sync semantics - imagine Client A setting `person1.Dog = dog1` and Client B deleting `dog1` - both these operations are valid in the state they were performed, but when they're merged with one another, the final state would be invalid (`person1.Dog == null`). While we realize not everyone is using Sync, it's the product that is funding development of the local database, so we're careful not to design functionality in a way that will add make sync development overly complex.

J-Swift:
To be clear, I'm not suggesting enforcement as a hard requirement on the edge cases, just that in general I prefer to have the data model reflect intent and leave it up to the developers not to violate that intent when things get ambiguous. There are usually many ways to violate type definitions at a language level (e.g. reflection or typecasting) and/or workflow level (e.g. deserializing JSON off the wire), but that doesn't mean we should loosen type restrictions because of the "technically possible" scenarios.

So for this, I think allowing non-nullable references, and maybe checking on add (and possible update) is good enough in the 95% of scenarios I am likely to run into.

nirinchev:
Gotcha, then yes, I believe that use case should be served by the new config option.

Realm/Realm.SourceGenerator/Parser.cs
```diff
@@ -145,7 +147,7 @@ private static IEnumerable<string> GetUsings(ClassDeclarationSyntax classDeclara
             return usings;
         }
 
-        private static IEnumerable<PropertyInfo> GetProperties(ClassInfo classInfo, IEnumerable<PropertyDeclarationSyntax> propertyDeclarationSyntaxes, SemanticModel model)
```
LaPeste, line 148: Why did you remove the static keyword? It does not look clear to me as the method does not look changed.

Realm/Realm.SourceGenerator/Parser.cs
```diff
@@ -258,7 +260,7 @@ private static IEnumerable<PropertyInfo> GetProperties(ClassInfo classInfo, IEnu
             }
         }
 
-        private static PropertyTypeInfo GetPropertyTypeInfo(ClassInfo classInfo, PropertyInfo propertyInfo, IPropertySymbol propertySymbol, PropertyDeclarationSyntax propertySyntax)
```
LaPeste, line 261: Same question as above.

LaPeste (approved):
It seems to make sense. I just had a super minor change recommendation and a question.

Realm/Realm.SourceGenerator/Parser.cs
papafe: This method at some point uses `_generatorConfig`, that is a private (non static) variable on the class.

Realm/Realm.SourceGenerator/Parser.cs
papafe: This method calls `GetPropertyTypeInfo ` that is not static anymore 😁

Realm/Realm.SourceGenerator/Parser.cs
LaPeste: Argh... sometimes I get a bit lost with the UI of GH when it does that code folding/hiding😫

Tests/Realm.Tests/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/A_generated.cs
```diff
@@ -75,7 +77,7 @@ public void SetManagedAccessor(Realms.IRealmAccessor managedAccessor, Realms.Wea
                 {
                     newAccessor.Realm.Add(oldAccessor.B, update);
                 }
-                newAccessor.B = oldAccessor.B;
+                newAccessor.B = oldAccessor.B!;
```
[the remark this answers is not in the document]
papafe: Our Test assembly doesn't have nullability annotations enabled, so in a sense, yes.

Tests/Realm.Tests/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/A_generated.cs
```diff
@@ -170,12 +172,12 @@ private void UnsubscribeFromNotifications()
 
         public static explicit operator A(Realms.RealmValue val) => val.AsRealmObject<A>();
 
-        public static implicit operator Realms.RealmValue(A val) => Realms.RealmValue.Object(val);
+        public static implicit operator Realms.RealmValue(A? val) => val == null? Realms.RealmValue.Null : Realms.RealmValue.Object(val);
```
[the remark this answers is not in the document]
papafe: It was fixed by a commit suggestion by Andrea, so I didn't re-run the source generation after the fix. I've just done it

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -240,22 +242,101 @@ public bool IsSupportedRequiredType()
...
+            var nullableInternalTypeString = $"{internalTypeString}?";
+
+            if (!isRequired)
+            {
+                if (NullableAnnotation == NullableAnnotation.None && (ScalarType == ScalarType.Data || ScalarType == ScalarType.String))
+                {
+                    return (CompleteFullyQualifiedString + "?", null);
+                }
+
+                if (IsCollection && InternalType.NullableAnnotation == NullableAnnotation.None
```
nirinchev, line 300: This warning is correct - if we're using `InternalType?.CompleteFullyQualifiedString`, it means we're expecting that `InternalType` may be null. If that's the case, we should guard here as well.

Tests/Realm.Tests/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/A_generated.cs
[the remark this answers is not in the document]
nirinchev: Should this be using `!` now? It looks like `B` is nullable in `IAAccessor`.

Tests/Realm.Tests/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/AsymmetricObjectWithAllTypes_generated.cs
```diff
@@ -704,8 +706,8 @@ public string RequiredStringProperty
                 }
             }
 
-            private string _stringProperty;
-            public string StringProperty
+            private string? _stringProperty = null!;
```
nirinchev, line 709: Why are we generating ` = null!` for strings that are optional?

Realm/Realm.SourceGenerator/InfoClasses.cs
nirinchev: Hm... I guess it's not smart enough to figure that out :/ Would have been nice if we have nullable reference types, but 🤷‍♂️

Tests/Realm.Tests/Generated/Realm.SourceGenerator/Realms.SourceGenerator.RealmGenerator/AsymmetricObjectWithAllTypes_generated.cs
papafe: Fixed

nirinchev (approved):

Issue #3161, opened 2023-01-09
Request: Support for Apple TV (tvOS) with Unity


muZZkat:
### Description

We are developing a cross platform game with Apple TV support and it doesn't look like the current version ships with tvOS support. 

We will also be looking at Playstation and Switch support, so maybe I need to look at building from source?

Can anyone maybe point me in the right direction?

Thanks

### How important is this improvement for you?

Dealbreaker

nirinchev:
@muZZkat can you clarify whether you need just the local database or also sync/functions support?

@fealebenpae sounds like building for Apple TV should be simple enough - perhaps we can add preview support for the platform where we build the binaries but we don't setup the testing infrastructure for CI?

muZZkat:
@nirinchev I'm looking for the full setup, local & sync.

I was just looking at the build scripts, it seems to be a pretty simple change to produce a realm-wrappers.framework for tvOS and then drop that into my project.

Playstation/Switch support I'm not sure about yet, although with the realm-core source, I can't really see anything that might stop me adding those targets?

nirinchev:
Yes, just dropping the framework for tvOS should be sufficient, though there were some special casing we needed to do for iOS builds - e.g. [here](https://github.com/realm/realm-dotnet/blob/27b94a20942b286f80b53082e8e0602eab8c7f4a/Realm/Realm.UnityWeaver/UnityWeaver.cs#L266) - I suspect we may need to do something similar for tvOS.

muZZkat:
Correct, the Dll imports need to be "__Internal" on tvOS as well.

muZZkat:
So I managed to hack things together today and got it working,

I Duplicated [build-ios.ps1](https://github.com/realm/realm-dotnet/blob/main/wrappers/build-ios.ps1) and changed it to tvOS, this got me my realm-wrappers.framework for tvOS. Worked out of the box on Mac.

Updated and rebuilt the UnityWeaver dll, this wasn't so easy, I struggled with the developer environment on both Mac and Windows, but in the end, somehow got the `Realm.UnityWeaver.dll` build, although I was sure something was wrong when it was only 98K vs 250K in the latest release. 

Added the new framework to a tvOS folder in downloaded package tarball

Replaced the Realm.UnityWeaver.dll and imported the tarball into Unity, I had errors around the Mono.Cecil reference, I assume this is why my dll was much smaller, does the packaged one include this? I worked around the issue by adding the Mono.Cecil package in unity.

[...]

PR #3174, opened 2023-01-12
Add support for AppleTV

fealebenpae:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/3161

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


wrappers/build-apple-platform.ps1
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env pwsh
```
nirinchev, line 1: Should we remove build-ios.ps1?

.github/templates/wrappers.yml
```diff
@@ -86,15 +86,20 @@ jobs:
...
     strategy:
       matrix:
-        arch: #@ iOSArchs
-    _: #@ template.replace(buildWrappers("pwsh ./wrappers/build-ios.ps1 ${{ matrix.arch }}", "wrappers-ios-${{ matrix.arch }}", enableLto = False, intermediateSteps = [setupXcode()]))
+        platform: #@ applePlatforms
+        target: [ 'Device', 'Simulator' ]
+    _: #@ template.replace(buildWrappers("pwsh ./wrappers/build-apple-platform.ps1 ${{ matrix.platform }} -Targets ${{ matrix.target }}", "wrappers-apple-${{ matrix.platform }}-${{ matrix.target }}", enableLto = False, intermediateSteps = [setupXcode()]))
     #@yaml/map-key-override
-    if: #@ " || ".join([ "needs.check-cache.outputs.wrappers-ios-" + x + " != 'true'" for x in iOSArchs ])
+    if: #@ " || ".join([ "needs.check-cache.outputs.wrappers-apple-" + x + "-Device != 'true'" for x in applePlatforms ])
```
nirinchev, line 102: [nit] Do we need the `-apple` here? Seems like this is producing artifacts like `wrappers-apple-ios-device`, which is kind of verbose.

Realm/Realm.Unity/Runtime/tvOS/Device/realm-wrappers.framework.meta
```diff
@@ -0,0 +1,90 @@
...
+  serializedVersion: 2
+  iconMap: {}
+  executionOrder: {}
+  defineConstraints: []
+  isPreloaded: 0
+  isOverridable: 1
+  isExplicitlyReferenced: 0
+  validateReferences: 1
+  platformData:
+  - first:
```
nirinchev, line 15: This file is actually important even though it's kind of difficult to review - it indicates that the tvOS wrappers are disabled on all platforms. This is because they're enabled on build by the logic in UnityWeaver.cs. This is pretty similar to what we're doing with the iOS wrappers.

Realm/Realm.UnityWeaver/UnityWeaver.cs
```diff
@@ -284,26 +284,34 @@
 
         public void OnPostprocessBuild(BuildReport report)
         {
-            if (report == null || report.summary.platform != BuildTarget.iOS)
+            switch (report?.summary.platform)
             {
-                return;
+                case BuildTarget.iOS:
+                case BuildTarget.tvOS:
+                    UpdateiOSFrameworks(false, false, report.summary.platform);
```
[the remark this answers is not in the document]
nirinchev: This is a false negative - we won't get into this case if `report` is `null`.

Realm/Realm.UnityWeaver/UnityWeaver.cs
```diff
@@ -284,26 +284,34 @@ public void OnPostBuildPlayerScriptDLLs(BuildReport report)
...
+                    break;
             }
-
-            UpdateiOSFrameworks(
-                enableForDevice: false,
-                enableForSimulator: false);
         }
 
         public void OnPreprocessBuild(BuildReport report)
```
nirinchev, line 296: Similarly to iOS, we need to include the correct framework depending on the target we're building.

Realm/Realm.UnityWeaver/UnityWeaver.cs
```diff
@@ -284,26 +284,34 @@
...
+                    enableForSimulator = PlayerSettings.tvOS.sdkVersion == tvOSSdkVersion.Simulator;
+                    break;
+                default:
+                    return;
             }
 
-            UpdateiOSFrameworks(
-                enableForDevice: PlayerSettings.iOS.sdkVersion == iOSSdkVersion.DeviceSDK,
-                enableForSimulator: PlayerSettings.iOS.sdkVersion == iOSSdkVersion.SimulatorSDK);
+            UpdateiOSFrameworks(enableForDevice, enableForSimulator, report.summary.platform);
```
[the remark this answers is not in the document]
nirinchev: False negative again - if `report` is null, we would have returned from the method.

Tests/Realm.Tests/Realm.Tests.csproj
```diff
@@ -109,7 +109,7 @@
         </None>
     </ItemGroup>
 
-    <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND '$(UseRealmNupkgsWithVersion)' == ''">
+    <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND '$(UseRealmNupkgsWithVersion)' == '' AND '$(RealmTestsStandaloneExe)' == 'true'">
```
nirinchev, line 112: Added this to prevent msbuild from complaining when we're building the iOS/tvOS tests without having built the macOS wrappers.

nirinchev (approved):

PR #3179, opened 2023-01-13
Keypath filtering

nirinchev:
## Description

Replaces https://github.com/realm/realm-dotnet/pull/3121. Putting up a different PR to make sure I don't accidentally overlook something.

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


wrappers/src/sync_session_cs.cpp
```diff
@@ -71,6 +71,7 @@ REALM_EXPORT CSharpSessionState realm_syncsession_get_state(const SharedSyncSess
         switch (session->state()) {
         case SyncSession::State::Inactive:
         case SyncSession::State::Dying:
+        case SyncSession::State::Paused:
```
LaPeste, line 74: I suppose changes in this file come from the updates coming from core but have nothing to do with the PR itself. Correct?

wrappers/src/object_cs.cpp
```diff
@@ -35,6 +36,20 @@ using GetNativeSchemaT = void(SchemaForMarshaling schema, void* managed_callback
 namespace realm {
 namespace binding {
     extern std::function<GetNativeSchemaT> s_get_native_schema;
+
+    REALM_FORCEINLINE KeyPathArray construct_key_path_array(const ObjectSchema& object)
+    {
+        KeyPathArray keyPathArray;
+        for (auto& prop : object.persisted_properties) {
+            bool is_primitive = (unsigned short)(prop.type & ~PropertyType::Collection) == (unsigned short)prop.type;
+            if (is_primitive) {
```
LaPeste, line 45: Is this just a temporary limitation or is there more to it?

LaPeste (commented):
I haven't looked at the test yet.
About the implementation I think I understand the overall idea and it mostly makes sense to me but I still have few questions.
I'd like to discuss about this PR this Friday mostly to verify how correct is my understanding.

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -180,21 +180,14 @@ public IRealmCollection<T> Freeze()
...
-            }
-            else if (_deliveredInitialNotification)
-            {
-                callback(this, null, null);
-            }
-
-            _callbacks.Add(callback);
-
-            return NotificationToken.Create(callback, UnsubscribeFromNotifications);
+            return NotificationToken.Create(callback, (c) => UnsubscribeFromNotifications(c, shallow));
```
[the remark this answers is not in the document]
nirinchev: Too much dart 😬

wrappers/src/sync_session_cs.cpp
nirinchev: Correct - was just pulling latest Core to pick up @JacobOscarGunnarsson's changes and those flowed as side effects.

wrappers/src/object_cs.cpp
nirinchev: Not sure what you're asking. Is it about the weird casting above it why we're only adding singular properties to the keypaths array? (Also I realize "primitive" here is probably misleading as it also covers links - the intent was to clarify that it's filtering out collection/computed properties)

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -624,6 +588,100 @@ internal interface IRealmCollectionBase<THandle> : IMetadataObject
...
+        // Shallow here and everywhere else is a bit of a shortcut we're taking until we have proper keypath filtering.
+        // Once we do, we probably want this to be some keypath identifier that we can pass between managed and native.
+        public void Add(NotificationCallbackDelegate<T> callback, bool shallow)
+        {
+            var keyPath = shallow ? KeyPath.Empty : KeyPath.Full;
+            if (_subscriptions.TryGetValue(keyPath, out var subscription))
+            {
+                if (subscription.DeliveredInitialNotification)
+                {
+                    callback(_parent, null, null);
```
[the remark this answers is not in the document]
LaPeste: Unfortunately from `Realm Database emits an initial notification after retrieving the collection.` in our [docs](https://www.mongodb.com/docs/realm/sdk/dotnet/react-to-changes/#register-a-collection-change-listener) it was not clear to me that "retrieving the collection" = "right when subscribing for notification".
To be honest, I am not sure what "retrieving the collection" would really mean:
1. first time you reference the list?
2. first time you create an instance of the class holding that list
3. etc

But thanks for explaining.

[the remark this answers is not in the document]
nirinchev: Typically, it means when the collection is materialized. There are several conditions where this could happen:
1. When you start iterating - this will block the thread and fetch the collection.
2. When you subscribe for notifications - the collection will be created in the background and you'll get the initial notification.

This is mostly relevant for queries which may be expensive to compute if you have many objects to match against. With the second approach, you construct the query, subscribe for notifications, and when it's ready, you'll get a callback with the initial collection that you can use to wire up the UI.

wrappers/src/object_cs.cpp
LaPeste: > the intent was to clarify that it's filtering out collection/computed properties

Of course. For a second I just forgot about the intent. Thanks.

PR #3189, opened 2023-01-19
Set core version in analytics

### LINKED ISSUE #3183: Report what version of core is being used by the current SDK
The current version of core is part of the new metrics.
One way of achieving the goal is to hard-coding the core version in the weaver while doing a new release (on CI).

papafe:
Fixes #3183


Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -61,6 +61,8 @@ namespace RealmWeaver
     // - An anonymized MAC address and assembly name ID to aggregate the other information on.
     internal class Analytics
     {
+        private readonly string _coreVersion = "";
```
nirinchev, line 64: This can be a const

nirinchev: This should also solve the build issue.

papafe: Yep

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -61,6 +61,8 @@ namespace RealmWeaver
     // - An anonymized MAC address and assembly name ID to aggregate the other information on.
     internal class Analytics
     {
+        private const string _coreVersion = "";
```
nirinchev, line 64: We can probably fill in the current Core version since the expectation is that there'll be something here.

nirinchev (approved):

Realm/Realm.Weaver/Analytics/Analytics.cs
papafe: But then we don't fully appreciate the regex that works with an empty string too 😢

LaPeste (approved):

papafe:
@LaPeste I've just merged from your branch because there was a conflict. Given it's approved, feel free to merge it when you think it's appropriate

LaPeste:
> @LaPeste I've just merged from your branch because there was a conflict. Given it's approved, feel free to merge it when you think it's appropriate

The plan is to have a feature branch that's going to be created as soon as my current PR is approved. Any future change about the epic is going to be merged in the feature branch. So, I suppose I'll just wait for the approval of my own.
@nirinchev did I understand correctly?

nirinchev:
Yeah, let's review/merge your branch into a feature branch before we merge more stuff into it to make it easier to review.

PR #3194, opened 2023-01-20
Adopt the exception unification changes

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2796

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


Realm/Realm/Exceptions/RealmExceptionCodes.cs
```diff
@@ -20,49 +20,179 @@ namespace Realms.Exceptions
...
+        RLM_ERR_WEBSOCKET_POLICYVIOLATION = 4407,
+        RLM_ERR_WEBSOCKET_MESSAGETOOBIG = 4408,
+        RLM_ERR_WEBSOCKET_INAVALIDEXTENSION = 4409,
+        RLM_ERR_WEBSOCKET_INTERNALSERVERERROR = 4410,
+        RLM_ERR_WEBSOCKET_TLSHANDSHAKEFAILED = 4411,
+
+        RLM_ERR_CALLBACK = 1000000,
+        RLM_ERR_UNKNOWN = 2000000,
+
+        RowDetached = 1000004000,
```
papafe, line 188: Why this don't have the same casing?

nirinchev: The `RLM_ERR...` ones are coming from Core. The PascalCased ones are custom codes that we define. I'll add a comment to make sure it's clear.

fealebenpae (commented):
LGTM. I like the added round tripping of client reset managed exceptions.

LaPeste (approved):
Overall it makes sense to me. I just left a cosmetic question.

For the tests there is just a timeout on Maui.Android. So all looks good on that front too. Just don't forget to add an entry in the changelog.

As a side note: I know it's not our decision, it's core's. But, I'm not sure `ErrorCodes::Error::OK` is a better choice than `...::NoError` 😅

PR #3203, opened 2023-01-25
Adds stable Id for metrics

### LINKED ISSUE #2706: Stable id in metrics
### Description

Working on https://github.com/realm/realm-dart/pull/134 we "realised" that using the MAC address of en0 (or whatever non-loopback interface is currently active) is not the most stable source for a host id. Modern operating systems go through some trouble to fake the MAC Address, fx switching it depending on the AP you are connected to, etc. As an aside Android, and iOS will always report 02:00:00:00:00:00, unless you have escalated privileges, and it seems likely that something similar will eventually hit the desktop.

On realm-js for node (https://github.com/realm/realm-js/blob/master/lib/submit-analytics.js), and in the above PR, we instead use the approach taken in: https://github.com/automation-stack/node-machine-id and https://github.com/denisbrodbeck/machineid, which is then anonymised by taking a SHA256 hash (in the PR we also add some salt).

This issue serves as a reminder to look into this on realm-dotnet, in particular if/when revamping the metrics gathering later.

_No response_

LaPeste:
Fixes #2706


Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -213,6 +193,74 @@ public static string GetLanguageVersion(ModuleDefinition module, string targetFr
...
+        // obtained from https://github.com/denisbrodbeck/machineid
+        public static string GetAnonymizedUserId()
+        {
+            var id = string.Empty;
+            var currentOs = Environment.OSVersion.Platform;
+
+            try
+            {
+                if (currentOs == PlatformID.Win32S || currentOs == PlatformID.Win32Windows ||
+                    currentOs == PlatformID.Win32NT || currentOs == PlatformID.WinCE)
```
LaPeste, line 206: or should I just check for the only realistic platform that our users should be using (a.k.a. PlatformID.Win32S)

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -213,6 +191,74 @@ public static string GetLanguageVersion(ModuleDefinition module, string targetFr
...
+                    }
+                }
+
+                if (id.Length == 0)
+                {
+                    return "UNKNOWN";
+                }
+                else
+                {
+                    var salt = new byte[] { 5, 67, 101, 45, 99, 239, 51, 111, 205, 174, 76, 16, 85, 158, 29, 8 };
```
LaPeste, line 248: If salt is used in all SDKs, then we must be aligned. This is just a proof of concept.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -213,6 +191,74 @@ public static string GetLanguageVersion(ModuleDefinition module, string targetFr
...
+            var id = string.Empty;
+            var currentOs = Environment.OSVersion.Platform;
+
+            try
+            {
+                if (currentOs == PlatformID.Win32S || currentOs == PlatformID.Win32Windows ||
+                    currentOs == PlatformID.Win32NT || currentOs == PlatformID.WinCE)
+                {
+                    var rk = Registry.LocalMachine;
+                    id = (string)rk.OpenSubKey("SOFTWARE\\Microsoft\\Cryptography").GetValue("MachineGuid");
```
LaPeste, line 207: Change to launch process to extract the registry value, instead of using windows specific libs.

LaPeste:
@nirinchev @papafe this is really ready for review. It's base branch it's now the feature branch.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -190,6 +169,64 @@ public static string GetLanguageVersion(string targetFramework)
...
+                    currentOs == PlatformID.Win32NT || currentOs == PlatformID.WinCE)
+                {
+                    var machineIdToParse = RunProcess("reg", "QUERY HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Cryptography -v MachineGuid");
+                    var regex = new Regex("^\\s+MachineGuid\\s+\\w+\\s+((\\w+-?)+)", RegexOptions.Multiline);
+                    match = regex.Match(machineIdToParse);
+                }
+                else if (currentOs == PlatformID.MacOSX)
+                {
+                    var machineIdToParse = RunProcess("ioreg", "-rd1 -c IOPlatformExpertDevice");
+                    var regex = new Regex("^.*IOPlatformUUID\\\"\\s=\\s\\\"(.+)\\\"", RegexOptions.Multiline);
```
nirinchev, line 190: The RegExp used in dart is `"IOPlatformUUID" = "([^"]*)"`. Any reason not to use the same one?

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -190,6 +169,64 @@ public static string GetLanguageVersion(string targetFramework)
...
+            var id = string.Empty;
+            var currentOs = System.Environment.OSVersion.Platform;
+            Match match = null;
+            try
+            {
+                if (currentOs == PlatformID.Win32S || currentOs == PlatformID.Win32Windows ||
+                    currentOs == PlatformID.Win32NT || currentOs == PlatformID.WinCE)
+                {
+                    var machineIdToParse = RunProcess("reg", "QUERY HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Cryptography -v MachineGuid");
+                    var regex = new Regex("^\\s+MachineGuid\\s+\\w+\\s+((\\w+-?)+)", RegexOptions.Multiline);
```
nirinchev, line 184: Dart doesn't seem to be matching the output with a regex - why are we?

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -190,6 +169,64 @@ public static string GetLanguageVersion(string targetFramework)
...
+                else if (currentOs == PlatformID.Unix)
+                {
+                    try
+                    {
+                        id = File.ReadAllText("/etc/machine-id");
+                    }
+                    catch
+                    {
+                        id = string.Empty;
+                    }
```
nirinchev, line 202: Why do we need this try-catch?

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -190,6 +169,64 @@ public static string GetLanguageVersion(string targetFramework)
             return Unknown();
         }
 
+        // Knowledge on unique machine Ids for different OSes obtained from https://github.com/denisbrodbeck/machineid
+        public static string GetAnonymizedUserId()
+        {
+            var id = string.Empty;
+            var currentOs = System.Environment.OSVersion.Platform;
+            Match match = null;
```
nirinchev, line 177: Rather than creating this `Match` variable here, let's have each case populate the id. We're not really saving any lines of code and the code duplication is rather minimal.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -190,6 +169,64 @@ public static string GetLanguageVersion(string targetFramework)
...
+            try
+            {
+                if (currentOs == PlatformID.Win32S || currentOs == PlatformID.Win32Windows ||
```
nirinchev, line 180: This would probably look better as a switch

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -190,6 +169,64 @@ public static string GetLanguageVersion(string targetFramework)
...
+                    id = match.Groups[1].Value;
+                }
+
+                if (id.Length == 0)
+                {
+                    return Unknown();
+                }
+                else
+                {
+                    var salt = new byte[] { 82, 101, 97, 108, 109, 32, 105, 115, 32, 103, 114, 101, 97, 116 };
```
nirinchev, line 216: Did you get a reply from @nielsenko about why we're salting this? Looking at `node-machine-id` which is used by the js SDK, they're not salting the id.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -190,6 +169,64 @@ public static string GetLanguageVersion(string targetFramework)
...
+                if (match?.Groups.Count > 1)
+                {
+                    id = match.Groups[1].Value;
+                }
+
+                if (id.Length == 0)
+                {
+                    return Unknown();
+                }
+                else
```
nirinchev, line 214: No need for an `else` here.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: Yes I did. As expected it's just some low-effort extra precautions to avoid that the same machine can be easily identified across multiple products (and their metrics)

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: No, I just didn't look at other implementations. But we basically do the same. We look for some `IOPlatformUUID = "` and grab all it's in between the first and last apex.
Do we have to align on these things too? Like even regexes have to be reported in the scope doc?

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: The exact output of the command is
```
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography
    MachineGuid    REG_SZ    xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```
So I just grab the GUID. I don't know why dart doesn't do this.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
nielsenko: Yes, it is just to avoid other unrelated projects to use the same id as us, if they use `node-machine-id`, which I believe is fairly common.

At the time I didn't actually try to have the same id as any other SDK, as it seemed very random what was used.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: For extra security. I know Lubo keeps saying "the file will always be there". But why crashing for a file not found for analytics?

nielsenko: I agree that the metrics must never tear down the build for a customer.

In dart (to catch errors in the metrics) we do run the metrics if REALM_CI is set, but without uploading and also without the error suppression. This is to catch errors in the metrics that would otherwise go unnoticed since we normally always suppress errors, and don't even run them if we detect we are on CI.
```dart
  final skipUpload = isRealmCI || Platform.environment['CI'] != null || Platform.environment['REALM_DISABLE_ANALYTICS'] != null;
  if (skipUpload && !isRealmCI) { // skip early, unless REALM_CI is set
    // skip early
    log.info('Skipping metrics upload');
    return;
  }

// calculate metrics

  if (skipUpload) {
    // skip late
    log.info('Skipping metrics upload (late)');
    return;
  }

// do actual upload
```

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
nirinchev: I would prefer if we do as using different regex-es can mean we could end up parsing different ids for the same machine.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
nirinchev: Same as my previous comment - the reason why we're aligning on commands we use to obtain the id, hashing, and salting is to ensure we are consistent across SDKs.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
nirinchev: Yeah, my point was that we have an outer try-catch wrapping everything and returning `unknown`, so the inner try catch doesn't seem particularly useful.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
nirinchev: Cool, can we add a comment that captures this. For the salt, we probably want to use the same one across our products to make sure we're identifying cases where the same developer uses multiple SDKs. While we believe this is pretty rare right now, we can't really confirm it and aligning on the hashed id would allow us to do so.

nirinchev: Right, but the scope doc is fairly unrelated to the code in the repo. I doubt someone looking at the code in 2-3 years will think of opening the scope doc to understand why we're using a hardcoded value for the salt.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: "Alignment notes" of the scope docs updated.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: "Alignment notes" of the scope docs updated.

nirinchev (approved):

PR #3208, opened 2023-01-30
Add suggestions to the analytics PR

nirinchev:
## Description

Fixes #

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/Realm/Realm.UnityWeaver/UnityWeaver.cs
+++ b/Realm/Realm.UnityWeaver/UnityWeaver.cs
@@ -366,7 +366,7 @@ namespace RealmWeaver
                 BuildTarget.Android => OperatingSystem.Android,
                 BuildTarget.StandaloneLinux64 => OperatingSystem.Linux,
                 BuildTarget.tvOS => OperatingSystem.TvOs,
-                _ => OperatingSystem.Unknown(target.ToString()),
+                _ => Metric.Unknown(target.ToString()),
             };
         }
 
@@ -377,7 +377,7 @@ namespace RealmWeaver
                 RuntimePlatform.WindowsEditor => OperatingSystem.Windows,
                 RuntimePlatform.OSXEditor => OperatingSystem.MacOS,
                 RuntimePlatform.LinuxEditor => OperatingSystem.Linux,
-                _ => OperatingSystem.Unknown(target.ToString()),
+                _ => Metric.Unknown(target.ToString()),
             };
         }
 


Realm/Realm.Weaver/Analytics/Metric.cs
```diff
@@ -55,83 +66,132 @@ public static class Framework
             public const string Xamarin = "Xamarin";
         }
 
-        public static readonly Dictionary<string, string> UserEnvironment = new Dictionary<string, string>()
+        public static class Environment
```
nirinchev, line 69: Replaced the dict with a class of constants since we weren't automating anything about the dictionary population like we do with the features.

Realm/Realm.Weaver/Analytics/Metric.cs
```diff
@@ -55,83 +66,132 @@ public static class Framework
...
+            public const string FrameworkUsedInConjunction = "Framework"; // this refers to UI frameworks and similar Realm is used together with
+            public const string FrameworkUsedInConjunctionVersion = "Framework Version";
+            public const string SdkInstallationMethod = "Installation Method";
+            public const string IdeUsed = "IDE";
+            public const string IdeUsedVersion = "IDE Version"; // this holds info about the msbuild version
+            public const string NetFramework = "Net Framework";
+            public const string NetFrameworkVersion = "Net Framework Version";
+        }
+
+        public static class Feature
```
nirinchev, line 95: After some deliberation, I decided it's a good idea to add the features as constants to make sure we don't introduce bugs via typos. We still have the dictionary and there are some features where we use the strings (e.g. ``List`1``).

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -80,126 +84,131 @@ internal Analytics(Config config, ImportedReferences references, ILogger logger)
...
+                [Feature.GetInstanceAsync] = instruction => AnalyzeRealmApi(instruction, Feature.GetInstanceAsync),
+                [Feature.GetInstance] = instruction => AnalyzeRealmApi(instruction, Feature.GetInstance),
+                [Feature.Find] = instruction => AnalyzeRealmApi(instruction, Feature.Find),
+                [Feature.WriteAsync] = instruction => AnalyzeRealmApi(instruction, Feature.WriteAsync),
+                [Feature.ThreadSafeReference] = instruction => AnalyzeRealmApi(instruction, Feature.ThreadSafeReference),
 
                 // check if it's the right signature, that is 2 params in total of which
                 // the second a bool and that it's set to true.
-                ["Add"] = instruction =>
+                // TODO - Nikola: why?
```
nirinchev, line 115: Per comment - not sure why we're looking for an `Add` that takes two params where the second is true?

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -80,126 +84,131 @@ internal Analytics(Config config, ImportedReferences references, ILogger logger)
...
+                [Feature.JWT] = instruction => AnalyzeRealmApi(instruction, Feature.JWT),
+                [Feature.ApiKey] = instruction => AnalyzeRealmApi(instruction, Feature.ApiKey),
+                [Feature.ServerApiKey] = instruction => AnalyzeRealmApi(instruction, Feature.ServerApiKey),
+                [Feature.Function] = instruction => AnalyzeRealmApi(instruction, Feature.Function),
+                [Feature.CallAsync] = instruction => AnalyzeRealmApi(instruction, Feature.CallAsync),
+                [Feature.GetMongoClient] = instruction => (true, Feature.GetMongoClient),
+                [Feature.DynamicApi] = instruction => (true, Feature.DynamicApi)
             };
+
+            (bool ShouldDelete, string DictKey) AnalyzeCollectionProperty(IMemberDefinition member, string primitiveKey, string referenceKey, int genericArgIndex = 0)
```
nirinchev, line 173: These are a bunch of helper methods that simplify some of the code.

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -352,13 +354,15 @@ private void AnalyzeClassMethods(TypeDefinition type)
...
+                        key = key.Substring(prefix.Length);
                     }
 
-                    if (!_apiAnalysisSetters.ContainsKey(key) && cil.Operand is MethodReference methodReference &&
+                    if (!_apiAnalysisSetters.ContainsKey(key) &&
+                        cil.Operand is MethodReference methodReference &&
                         methodReference.ReturnType.DeclaringType != null)
                     {
                         // when dealing with ThreadSafeReference
+                        // TODO nikola: why?
```
nirinchev, line 365: Per comment - not sure I understand why we're special casing TSR like this?

LaPeste: This should have simply been
`key = methodReference.ReturnType.DeclaringType.Name;`
I shouldn't have done `.ToString()`. It was at the beginning when I was not as comfortable moving around in Mono.Cecil.

Realm/Realm.Weaver/Analytics/Analytics.cs
nirinchev: Hm... I guess we expect that everyone is doing insertions, so there's no point in tracking that?

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -338,12 +343,9 @@ private void AnalyzeClassMethods(TypeDefinition type)
...
 
-                    var index = new int[]
-                    {
-                        key.IndexOf("get_", StringComparison.Ordinal),
-                        key.IndexOf("set_", StringComparison.Ordinal),
-                        key.IndexOf("add_", StringComparison.Ordinal)
-                    }.Max();
+                    var (prefix, index) = prefixes.Select(p => (p, key.IndexOf(p, StringComparison.Ordinal)))
+                        .OrderByDescending(p => p.Item2)
+                        .First();
```
LaPeste, line 348: Is this change for performance or readability?

LaPeste (approved):
Makes sense to me and I agree with the points you brought up. I just left a bunch of minor questions.
I'm also gonna fix the issues with the tests. It's small stuff.

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -209,33 +218,34 @@ internal void AnalyzeUserAssembly(ModuleDefinition module)
...
+                _realmEnvMetrics[UserEnvironment.HostCpuArch] = GetHostCpuArchitecture;
+                _realmEnvMetrics[UserEnvironment.TargetOsType] = _config.TargetOSName;
+                _realmEnvMetrics[UserEnvironment.TargetCpuArch] = GetTargetCpuArchitecture(module);
+                _realmEnvMetrics[UserEnvironment.FrameworkUsedInConjunction] = frameworkInfo.Name;
+                _realmEnvMetrics[UserEnvironment.FrameworkUsedInConjunctionVersion] = frameworkInfo.Version;
+                _realmEnvMetrics[UserEnvironment.LanguageVersion] = GetLanguageVersion(_config.TargetFramework);
+                _realmEnvMetrics[UserEnvironment.RealmSdkVersion] = module.FindReference("Realm").Version.ToString();
+                _realmEnvMetrics[UserEnvironment.CoreVersion] = "FILL ME";
+                _realmEnvMetrics[UserEnvironment.SdkInstallationMethod] = "FILL ME";
+                _realmEnvMetrics[UserEnvironment.IdeUsed] = "FILL ME";
```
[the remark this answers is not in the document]
nirinchev: I thought we'll report the Unity editor for the IDE unless we're capturing that information somewhere else? Also, since it's not clear how we'll capture the msbuild version, I thought it's better to leave it blank for now.

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -209,33 +218,34 @@ internal void AnalyzeUserAssembly(ModuleDefinition module)
...
                     InternalAnalyzeSdkApi(type);
                 }
 
-                _realmEnvMetrics.TryGetValue("PartitionSyncConfiguration", out var isPbsUsed);
-                _realmEnvMetrics.TryGetValue("FlexibleSyncConfiguration", out var isFlxSUsed);
-                _realmEnvMetrics["IsSyncEnabled"] = ((!string.IsNullOrEmpty(isPbsUsed) && isPbsUsed == "1") || (!string.IsNullOrEmpty(isFlxSUsed) && isFlxSUsed == "1")).ToString().ToLower();
+                // We need to first analyze the features before we can set `IsSyncEnabled`.
+                _realmFeaturesToAnalyze.TryGetValue(Feature.PartitionSyncConfiguration, out var isPbsUsed);
+                _realmFeaturesToAnalyze.TryGetValue(Feature.FlexibleSyncConfiguration, out var isFlxSUsed);
+                _realmEnvMetrics[UserEnvironment.IsSyncEnabled] = (isPbsUsed == 1 || isFlxSUsed == 1).ToString().ToLower();
```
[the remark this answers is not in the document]
nirinchev: This is checking them in `_realmFeaturesToAnalyze` which is a dictionary of string->int. If the key is not present or the value is 0, `isPbsUsed` will be 0.

Realm/Realm.Weaver/Analytics/Analytics.cs
nirinchev: This is fairly nitty and possibly unnecessary, but it was triggering my OCD as the previous code was relying on all prefixes being exactly 4 characters long. I really dislike hardcoding values like that as it obscures the connection between the pieces of code using them (the index lookup and then the substring call).

Realm/Realm.Weaver/Analytics/Analytics.cs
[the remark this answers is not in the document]
LaPeste: > I thought we'll report the Unity editor for the IDE
No no, we're reporting Unity under frameworks as it's "the framework" that Realm is being used with. Same as we do for xamarin, maui etc.
Then I'd just leave the field empty. I changed to this "empty" instead of "FILL ME" in my PR

Realm/Realm.Weaver/Analytics/Analytics.cs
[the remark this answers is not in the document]
LaPeste: My bad, it's not a string. I misread.

PR #3211, opened 2023-02-01
Don't wait for analytics submission

nirinchev:
## Description

Fixes #

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -80,18 +80,24 @@ internal class Analytics
...
+
+            _references = references;
             _logger = logger;
 
             _realmFeaturesToAnalyze = Metric.SdkFeatures.Keys.ToDictionary(c => c, _ => (byte)0);
 
             _classAnalysisSetters = new()
             {
-                [Feature.IEmbeddedObject] = member => (true, Feature.IEmbeddedObject),
-                [Feature.IAsymmetricObject] = member => (true, Feature.IAsymmetricObject),
```
LaPeste, line 94: Good catch

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -239,7 +250,7 @@ internal void AnalyzeUserAssembly(ModuleDefinition module)
                 _realmEnvMetrics[UserEnvironment.NetFramework] = _config.TargetFramework;
                 _realmEnvMetrics[UserEnvironment.NetFrameworkVersion] = _config.TargetFrameworkVersion;
 
-                foreach (var type in module.Types)
+                foreach (var type in module.Types.ToArray())
```
LaPeste, line 253: Why is this necessary?

LaPeste (commented):
The main idea seems good. We need to investigate how fast it's now VS before. Mostly curiosity as this improvement was almost for free.
I'd be interesting to investigate what happens in a console build, as you mentioned.

Realm/Realm.Weaver/Analytics/Analytics.cs
nirinchev: This is executing while the weaver is running, which means that if the weaver adds new types (and it does for the schema), iteration will break.

LaPeste: Makes sense. Thank you.

Issue #3223, opened 2023-02-13
Expose sync client timeout options


nirinchev:
Should roughly mirror the same options in Swift: https://github.com/realm/realm-swift/blob/3eaf6f9f172f1722d8b0994563b80e2bacd5a86d/Realm/RLMSyncManager.h#L156

Issue #3222, opened 2023-02-13
Expose SyncConfiguration.CancelAsyncOperationsOnNonFatalErrors


nirinchev:
This allows users to get meaningful exception information whenever the connection timeouts or there are other transient session errors which may otherwise keep the async operation waiting for a long time.

PR #3224, opened 2023-02-13
Expose options to control sync timeouts

nirinchev:
## Description

Adds a few sync timeout options to AppConfiguration as well as an option to cancel async operations on non-fatal session errors.

Fixes https://github.com/realm/realm-dotnet/issues/3222
Fixes https://github.com/realm/realm-dotnet/issues/3223

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Sync/SyncTimeoutOptions.cs
+++ b/Realm/Realm/Sync/SyncTimeoutOptions.cs
@@ -1,0 +1,108 @@
+﻿////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2023 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License")
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+using System;
+
+namespace Realms.Sync
+{
+    /// <summary>
+    /// Options for configuring timeouts and intervals used by the sync client.
+    /// </summary>
+    public class SyncTimeoutOptions
+    {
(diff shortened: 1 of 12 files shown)


Realm/Realm/Sync/SyncTimeoutOptions.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// <remarks>
+        /// Shorter values will make connection state change notifications more
+        /// responsive as it will only change to `disconnected` after this much time has
+        /// elapsed, but overly short values may result in spurious disconnection
+        /// notifications when the server is simply taking a long time to respond.
+        /// <br/>
+        /// Defaults to 2 minutes.
+        /// </remarks>
+        /// <value>The pong timeout.</value>
+        public TimeSpan PongKeepAliveTimeout { get; set; } = TimeSpan.FromMinutes(2);
```
fealebenpae, line 84: I don’t think we should expose this, as with Platform Networking we won’t have any control of the websocket ping-pong frames.

Realm/Realm/Sync/SyncTimeoutOptions.cs
```diff
@@ -0,0 +1,108 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+using System;
+
+namespace Realms.Sync
+{
+    /// <summary>
+    /// Options for configuring timeouts and intervals used by the sync client.
+    /// </summary>
```
fealebenpae, line 25: Wouldn’t a value type be better here?

fealebenpae (changes requested):

Realm/Realm/Sync/SyncTimeoutOptions.cs
nirinchev: I believe those are the sync ping-pong messages, not the websocket ones.

Realm/Realm/Sync/SyncTimeoutOptions.cs
nirinchev: Performance is not critical and structs are a little annoying to work with. I wanted to allow people to modify a single default, like:

```csharp
config.SyncTimeoutOptions.ConnectTimeout = TimeSpan.FromSeconds(15);
```

With a struct, you would have to replace the entire thing, making it a fair bit more verbose.

fealebenpae (approved):

Issue #3225, opened 2023-02-13
Don't indiscriminately register bson serializers


nirinchev:
There may be cases where users would like to override the way bson serializers are registered - we should allow them to do so by registering a provider rather than serializers.

PR #3226, opened 2023-02-13
Use BsonSerializer.RegisterProvider

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/3225

##  TODO

* [x] Changelog entry

### DIFF
--- a/Realm/Realm/Native/NativeCommon.cs
+++ b/Realm/Realm/Native/NativeCommon.cs
@@ -22,6 +22,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


Realm/Realm/Native/NativeCommon.cs
```diff
@@ -75,6 +76,8 @@ internal static void Initialize()
                 HttpClientTransport.Initialize();
                 AppHandle.Initialize();
                 SubscriptionSetHandle.Initialize();
+
+                SerializationHelper.Initialize();
```
fealebenpae, line 80: I’m beginning to think we should use a module initializer soon

fealebenpae (approved):

PR #3228, opened 2023-02-14
Fix unity hanging while in its sort-of-headless mode

LaPeste:


Realm/Realm.UnityWeaver/UnityWeaver.cs
```diff
@@ -82,8 +82,12 @@ public static void Initialize()
...
             {
                 _listRequest = Client.List();
-                _installMethodTask = new TaskCompletionSource<string>();
-                EditorApplication.update += OnEditorApplicationUpdate;
+
+                if (!Application.isBatchMode)
+                {
+                    _installMethodTask = new TaskCompletionSource<string>();
+                    EditorApplication.update += OnEditorApplicationUpdate;
+                }
```
nirinchev, line 90: Can we simplify things a little bit by doing something like:

```suggestion
                _installMethodTask = new TaskCompletionSource<string>();

                if (Application.isBatchMode)
                {
                    // In batch mode, `update` won't get called until compilation is complete,
                    // which means we'll deadlock when we block compilation on the tcs completing
                    _installMethodTask.TrySetResult(Metric.Unknown());
                }
                else
                {
                    EditorApplication.update += OnEditorApplicationUpdate;
                }
```

That way we can avoid checking for `null` later and instead do something like:

```csharp
InstallationMethod = _installMethodTask.Task.Wait(1000) ? _installMethodTask.Task.Result : Metric.Unknown();
```

nirinchev (approved):
Minor suggestion for reworking the logic a little bit, otherwise looks good.

PR #3231, opened 2023-02-15
Adds a `RealmObservableObject`

AndreasReitberger:
## Description
This adds a `RealmObservableObject` which inherits from the `RealmObservableObjectBase` which inherits from the `ObservableObject`. This allows to use a `RealmObject` in combination to the `ObservableObject` and its features.

Fixes #
This allows to use the `[ObservableProperty]` which creates all needed source code for you.

```cs
[ObservableProperty]
[property: PrimaryKey]
Guid id = Guid.Empty;
```

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/Realm/Realm/DatabaseTypes/RealmObservableObject.cs
+++ b/Realm/Realm/DatabaseTypes/RealmObservableObject.cs
@@ -1,0 +1,27 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2020 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms
+{
+    /// <summary>
+    /// Base for any object that can be persisted in a <see cref="Realm"/>.
+    /// </summary>
+    public class RealmObservableObject : RealmObservableObjectBase, IRealmObject
+    {
+    }
+}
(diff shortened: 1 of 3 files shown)


nirinchev:
Hey, thanks for your contribution. Unfortunately, we cannot accept the PR in its current form. One of the issues is that it takes a dependency on `CommunityToolkit.Mvvm`, which is not something we want to burden our consumers with, especially on platforms that don't need it, such as Unity.

Can you clarify a little bit about your use case - why is it important that your `RealmObject` models inherit from `ObservableObject`? They already implement `INotifyPropertyChanged` which should get you most of the benefits you get from `ObservableObject`.

AndreasReitberger:
@nirinchev 
I already thought that. Most of my libraries are using `SQLite` at the moment and working with the `[ObservableProperty]`. In order to use this attribute, the object must inherit from the `ObservableObject`.
https://learn.microsoft.com/en-us/dotnet/communitytoolkit/mvvm/generators/observableproperty?source=recommendations

However, `Realm` needs to inherit the class from the `RealmObject`, so it can't inherit from the `ObservableObject`. Either `Realm` would support this, like I added it in the PR, or I need to replace my properties and remove the `[ObservableProperty]`

nirinchev:
The question is why do you need `[ObservableProperty]` - Realm will automatically raise property change notifications, similarly to what you would gain from `[ObservableProperty]`. E.g. this code:

```csharp
public class MyModel : ObservableObject
{
    [ObservableProperty]
    private string name;
}
```

Would be roughly equivalent to

```csharp
public class MyModel : RealmObject
{
}
```

I.e. both will raise notifications when the property `Name` changes. The only difference is that Realm will not raise `PropertyChanging` notifications prior to the update, so if you depend on those, they won't be available.

AndreasReitberger:
The benefit is, that the `[ObserableProperty]` creates all needed source code automatically. I assumed that I need to do this manually here. However if the `RealmObject` also notifies with a single line in code, there is no benifit anymore.

nirinchev:
Yeah, that's not immediately obvious, but what'll happen under the hood is that we'll replace the getter/setter of that property with something that raises notifications. If you'd prefer to see the actual code that gets generated, you could instead use our newly released source generators and define your models like:

```csharp
public partial class MyModel : IRealmObject
{
}
```

Note the `partial` keyword and the `IRealmObject` interface that replaces the `RealmObject` type. That will generate a bunch of supporting classes which you can read through to see how Realm works.

PR #3239, opened 2023-02-16
Fix UWP not building on CI

LaPeste:


Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -530,5 +530,18 @@ public enum AnalyticsCollection
             DryRun,
             Full,
         }
+
+        private class ActionDriver
+        {
+            public bool IsToDelete { get; }
```
nirinchev, line 536: This reads kinda weird - can we name it `ShouldDelete`?

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -530,5 +530,18 @@ public enum AnalyticsCollection
             DryRun,
             Full,
         }
+
+        private class ActionDriver
```
nirinchev, line 534: This could be a struct and we can probably name it something more concrete - like `FeatureAnalysisResult` or something similar.

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -102,13 +102,13 @@ public Analytics(Config config, ImportedReferences references, ILogger logger, M
                     member is PropertyDefinition property &&
                     (property.PropertyType.IsIRealmObjectBaseImplementor(_references) ||
                     property.PropertyType.IsRealmObjectDescendant(_references)) ?
-                    (true, Feature.RealmObjectReference) : default,
-                [Feature.RealmValue] = member => (true, Feature.RealmValue),
+                    new ActionDriver(true, Feature.RealmObjectReference) : new ActionDriver(),
```
nirinchev, line 105: These (and everywhere else) can be simplified by the new target-typed new expressions.

nirinchev (commented):
Mostly nits from me - otherwise looks good.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -63,14 +63,16 @@ public static string GetTargetOsName(FrameworkName frameworkName)
...
+            if (targetOs.ContainsIgnoreCase("win") || targetOs == ".NETFramework")
             {
                 return OperatingSystem.Windows;
             }
 
+            if (targetOs.ContainsIgnoreCase("uap"))
+            {
+                return OperatingSystem.Uwp;
+            }
+
```
LaPeste, line 75: I meant to do this, and this seemed a small enough PR

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -304,14 +304,14 @@ void AnalyzeClassFeature(string key, PropertyDefinition property)
             {
                 if (_classAnalysisSetters.TryGetValue(key, out var featureFunc))
                 {
-                    var (shouldDelete, keyToSet) = featureFunc(property);
+                    var actionDriver = featureFunc(property);
```
nirinchev, line 307: This var name is bugging me a little - can we rename to `analysisResult` or something simlar?

PR #3244, opened 2023-02-20
Adds minor fixes of bugs discovered while manually testing

LaPeste:


Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -93,7 +93,7 @@ public Analytics(Config config, ImportedReferences references, ILogger logger, M
 
             _references = references;
             _logger = logger;
-
+            _realmEnvMetrics = new();
```
nirinchev, line 96: Why was that needed?

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -233,19 +233,22 @@ private void AnalyzeUserAssembly(ModuleDefinition module)
                 _realmEnvMetrics[UserEnvironment.UserId] = GetAnonymizedUserId();
                 _realmEnvMetrics[UserEnvironment.ProjectId] = SHA256Hash(Encoding.UTF8.GetBytes(module.Name));
                 _realmEnvMetrics[UserEnvironment.RealmSdk] = ".NET";
+                _realmEnvMetrics[UserEnvironment.RealmSdkVersion] = module.FindReference("Realm")?.Version.ToString();
```
nirinchev, line 236: The weaver version and the Realm version are the same. So rather than going through FindReference, you can give it the version of the weaver assembly.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -63,16 +64,11 @@ public static string GetTargetOsName(FrameworkName frameworkName)
                 return OperatingSystem.Linux;
             }
 
-            if (targetOs.ContainsIgnoreCase("win") || targetOs == ".NETFramework")
+            if (targetOs.ContainsIgnoreCase("win") || targetOs.ContainsIgnoreCase("net"))
```
nirinchev, line 67: Wouldn't that also catch things like .NET Core, .NET standard and so on? We don't want to report those as windows anymore.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -105,13 +101,13 @@ public static (string Name, string Version) GetFrameworkAndVersion(ModuleDefinit
             }
             else
             {
-                // TODO andrea: the correctness of these names need to be verified in projects that use each of the packages
-                // I didn't have any handy one.
-                var possibleFrameworks = new string[] { "Xamarin.Forms", "Mono.Android", "Xamarin.iOS", "Microsoft.Maui.Sdk" };
+                // the order is important as both, xamarin android and maui android use Mono.Android.
+                // So, in order to report Maui it has to be before Mono.Android
+                var possibleFrameworks = new string[] { "Xamarin.Forms.Platform.UAP", "Microsoft.Maui", "Mono.Android", "Xamarin.iOS", "Xamarin.Mac" };
```
nirinchev, line 106: What about xamarin.tvos?

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -136,34 +137,39 @@ public static (string Name, string Version) GetFrameworkAndVersion(ModuleDefinit
...
+        public static string GetLanguageVersion(string netFramework, string netFrameworkVersion)
         {
             // We don't have a reliable way to get the version in the weaver so we're using the default version
             // associated with the framework.
             // Values taken from https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version
-            if (targetFramework.ContainsIgnoreCase("netcoreapp2") ||
-                targetFramework.ContainsIgnoreCase("netframework"))
+            if (!netFramework.ContainsIgnoreCase("net"))
+            {
+                return Unknown($"Likely not the model assembly, but the platform specific one: {netFramework}");
```
nirinchev, line 147: This is going to be a very long string to send. Should we move this to a comment and instead use `Unknown(netFramework)`?

nirinchev (commented):
Mostly questions from me. Also, do you have any idea why those zero-length array warning have suddenly appeared? Those look like false positives, but it's weird that we're seeing them.

Realm/Realm.Weaver/Analytics/Analytics.cs
LaPeste: All other fields were initialized in the constructor (because they have to), so just for consistency I brought this there to have all initialization in 1 place.
If we have rules on style about this, I'll revert.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
nirinchev: For targetOS, we're currently reporting .NET Standard as Windows, which is misleadingly inflating the windows numbers we're seeing. Since those projects are inherently cross platform, we should instead report that to allow comparing builder numbers between Windows and the other supported OS-s.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: > What about xamarin.tvos?

I am not sure how to test this since the our .sln doesn't have a tvOS project. So I'll add `Xamarin.tvOS` hoping for the best. If you have suggestions on how to test this, I'd appreciate.

> Also, do we want to differentiate between Xamarin.Forms and regular xamarin?

I don't have an answer, but I'll differentiate since I guess it's good info.

Realm/Realm.Weaver/Analytics/Analytics.cs
nirinchev: No, we don't have strict rules on that - I generally prefer to initialize fields wherever they are declared if possible - it saves a line of code and makes it more obvious, especially with the target-typed ctor (since you have the type on the same line). I'm fine either way, was just surprised to see this and thought I'm missing something.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
nirinchev: We do have a tvOS project in the Realm.sln. I didn't bother adding it to the -Windows solution because you wouldn't be able to build on windows.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
LaPeste: Ahhh, now I remember the conversation. Thank you for clarifying this.

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -97,21 +106,26 @@ System.Environment.OSVersion.Platform switch
...
+                    "Xamarin.Forms.Platform.Android", "Xamarin.Forms.Platform.iOS",
+                    "Xamarin.Forms.Platform.tvOS", "Xamarin.Mac",
+                    "Xamarin.Forms.Platform"
+                };
+
                 AssemblyNameReference frameworkUsedInConjunction = null;
                 foreach (var toSearch in possibleFrameworks)
                 {
-                    frameworkUsedInConjunction = module.FindReference(toSearch);
+                    frameworkUsedInConjunction = module.AssemblyReferences.Where(a => a.Name == toSearch).SingleOrDefault();
```
nirinchev, line 128: Okay, I don't think this is going to work as intended. We have 3 frameworks we want to detect (in this order):

1. MAUI
2. Xamarin.Forms
3. Xamarin (native)
4. Catalyst
5. UWP

I think to achieve that we want something like:

```csharp
var possibleFrameworks = new[]
{
    "Microsoft.Maui", // MAUI
    "Xamarin.Forms.Core", // Xamarin.Forms
    "Xamarin.iOS", // Xamarin
    "Xamarin.tvOS", // Xamarin
    "Xamarin.Mac" // Xamarin
    "Mono.Android", // Xamarin
    "Microsoft.MacCatalyst", // Catalyst
    "Windows.Foundation.UniversalApiContract" // UWP
}
```

Realm/Realm.Weaver/Analytics/Analytics.cs
```diff
@@ -233,21 +232,24 @@ private void AnalyzeUserAssembly(ModuleDefinition module)
                 _realmEnvMetrics[UserEnvironment.UserId] = GetAnonymizedUserId();
                 _realmEnvMetrics[UserEnvironment.ProjectId] = SHA256Hash(Encoding.UTF8.GetBytes(module.Name));
                 _realmEnvMetrics[UserEnvironment.RealmSdk] = ".NET";
+                _realmEnvMetrics[UserEnvironment.RealmSdkVersion] = Assembly.GetExecutingAssembly().GetName().Version.ToString();//module.FindReference("Realm")?.Version.ToString();
```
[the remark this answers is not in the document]
LaPeste: 🤦🏻

Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -63,14 +64,14 @@ public static string GetTargetOsName(FrameworkName frameworkName)
                 return OperatingSystem.Linux;
             }
 
-            if (targetOs.ContainsIgnoreCase("win") || targetOs == ".NETFramework")
+            if (targetOs.ContainsIgnoreCase("win") || targetOs.ContainsIgnoreCase("netframework"))
```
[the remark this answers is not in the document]
LaPeste: That framework name comes from the [ModuleWeaver](https://github.com/realm/realm-dotnet/blob/9510c2c7f2d4d93c1c7654b5e67bd1f26b9649c1/Realm/Realm.Fody/ModuleWeaver.cs#L39).
I think you're right, it should be the TFM.
So I'll make this condition just like 
```suggestion
            if (targetOs.ContainsIgnoreCase("net4"))
```

PR #3246, opened 2023-02-20
Adds hostOS tests to the weaver

LaPeste:


Tests/Weaver/Realm.Fody.Tests/AnalyticsTests.cs
```diff
@@ -104,19 +104,45 @@ public void Submit_WhenDisabled_PayloadIsEmpty()
...
+        private static string ConvertToMetricOS(PlatformID platformID) =>
+            platformID switch
+            {
+                PlatformID.Win32NT or PlatformID.Win32S or PlatformID.Win32Windows or PlatformID.WinCE => Metric.OperatingSystem.Windows,
+                PlatformID.MacOSX => Metric.OperatingSystem.MacOS,
+                PlatformID.Unix => Metric.OperatingSystem.Linux,
+                _ => platformID.ToString()
+            };
+
+        private void ValidateAnalyticsPayloadAllFrameworks<T>((string featureName, T expectedValue)[] payloadFeatures)
```
nirinchev, line 133: Rather than have this be generic, can we make `expectedValue` a string and have the caller pass in the correct value?

Tests/Weaver/Realm.Fody.Tests/AnalyticsTests.cs
```diff
@@ -104,19 +104,45 @@ public void Submit_WhenDisabled_PayloadIsEmpty()
...
-                var payload = BsonSerializer.Deserialize<BsonDocument>(response).AsBsonDocument;
-                Assert.That(payload[featureName].AsString, Is.EqualTo(expectedResult.ToString()), $"Feature {featureName} was not reported as used: {expectedResult} in {framework}");
+                CompileAnalyticsProject();
+                ValidateAnalyticsPayloadAllFrameworks(new[] {
+                    (featureName: Metric.Environment.HostOsType, expectedValue: ConvertToMetricOS(currentOs.Platform)),
+                    (featureName: Metric.Environment.HostOsVersion, expectedValue: currentOs.Version.ToString()) });
+            }
+            catch (Exception e)
+            {
+                Assert.Fail($"{e.Message}");
```
nirinchev, line 120: Why do we need this? An uncaught exception would fail the test anyway.

nirinchev (changes requested):

Tests/Weaver/Realm.Fody.Tests/AnalyticsTests.cs
LaPeste: Because in the past I wanted to given some extra contextual info, which then I stopped doing and forgot to remove the try/catch altogether. Remove it.

nirinchev (approved):

Issue #2854, opened 2022-03-15
[Bug]:  NotifyCollectionChangedAction.Replace and Reset is not fired on RealmCollection


PawKanarek:
### What happened?

Hi, I've noticed a bug where RealmCollection doesn't behave in the same way as ObservableCollection.
From what I've observed, RealmCollection doesn't fire `Replace` & `Reset` from [INotifyCollectionChanged](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.notifycollectionchangedaction?view=net-6.0) Interface, thus my app is not updating the UI correctly when I switched from ObservableCollection to RealmDb in my ViewModels. I've made a very simple example to prove my point.

### Repro steps

Launch provided code snippet in console application.

### Version

net5.0 for console app & Xamarin for my main project issue

### What SDK flavour are you using?

Local Database only

### What type of application is this?

Xamarin

### Client OS and version

Android, iOS, ConsoleApp in .net core Console app for macOS 

### Code snippets

```c#
using System;
using System.Collections.Generic;
using System.Collections.ObjectModel;
using System.Collections.Specialized;
using System.IO;
using System.Threading.Tasks;
using Nito.AsyncEx;
using Realms;

namespace ConsoleApp3
{
    public class Program
    {
        public static async Task Main(string[] args)
        {
            ObservableCollectionTest();
            AsyncContext.Run(RealmCollectionTest);
        }

        private static void ObservableCollectionTest()
        {
            var item0 = new RealmItem { Id = 0 };
            var item1 = new RealmItem { Id = 1 };
            var item2 = new RealmItem { Id = 2 };
            // add items to collection 
            var colllection = new MyCollection(new[] { item0, item1 });

            // listen for notification
            var actions = new List<NotifyCollectionChangedAction>(5);
            colllection.CollectionChanged += (sender, eventArgs) =>
            {
                actions.Add(eventArgs.Action);
            };

            colllection.Add(item2); // Add items
            colllection.MoveItem(1, 2); // Move items
            colllection.Remove(item0); // Remove items
            colllection[1] = item2; // Replace items
            colllection.Clear(); // Reset items
            
            Console.WriteLine($"Actions for ObservableCollection: {string.Join(", ", actions)}");
        }

        private static void RealmCollectionTest()
        {
            var path = Path.Combine(Environment.CurrentDirectory, "realm.realm");
            var realm = Realm.GetInstance(new RealmConfiguration(path));

[...]

PR #3254, opened 2023-02-22
[v11] 12. Raise Replace events for collection notifications

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/2854

This doesn't technically need v11, but it would have conflicted, so I decided to do it on top of the nullable annotations branch.

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
+++ b/Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
@@ -312,28 +312,37 @@ namespace Realms
 
                 var raiseAdded = TryGetConsecutive(change.InsertedIndices, i => this[i], out var addedItems, out var addedStartIndex);
 
-                if (raiseAdded || raiseRemoved)
-                {
-                    if ((raiseAdded && raiseRemoved) ||
-                        (raiseAdded && addedItems == null) ||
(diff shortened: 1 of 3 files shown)


Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -311,28 +311,37 @@ private void OnChange(IRealmCollection<T> sender, ChangeSet? change)
...
-                        (raiseAdded && addedItems == null) ||
-                        (raiseRemoved && removedItems == null))
-                    {
-                        RaiseCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
-                        RaisePropertyChanged();
-                        return;
-                    }
+                var raiseReplaced = TryGetConsecutive(change.NewModifiedIndices, i => this[i], out var replacedItems, out var replacedStartIndex);
 
+                // Only raise specialized notifications if we have exactly one change type to report
```
papafe, line 316: Could we compute the replaced notifications after added and removed so we don't need to raise Reset in case we have added + replaced or removed + replaced? 
With replaced the index should be stable, so I suppose it could work?

PR #3256, opened 2023-02-27
Extract CPU architecture from Unity

nirinchev:
## Description

Fixes #

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/Realm/Realm.UnityWeaver/UnityWeaver.cs
+++ b/Realm/Realm.UnityWeaver/UnityWeaver.cs
@@ -31,6 +31,7 @@ using UnityEditor.PackageManager;
 using UnityEditor.PackageManager.Requests;
 using UnityEngine;
 using static RealmWeaver.Analytics;
+using CpuArchitecture = RealmWeaver.Metric.CpuArchitecture;
 using OperatingSystem = RealmWeaver.Metric.OperatingSystem;
 
 namespace RealmWeaver
@@ -120,7 +121,7 @@ namespace RealmWeaver
 
                 WeaveAssemblyCore(assemblyPath, assembly.allReferences,
                     Metric.Framework.UnityEditor, Application.unityVersion,
-                    GetTargetOSName(Application.platform), netFrameworkInfo.TargetArchitecture,
+                    GetTargetOSName(Application.platform), GetHostCpuArchitecture(),
                     netFrameworkInfo.Name, netFrameworkInfo.Version);
             };
         }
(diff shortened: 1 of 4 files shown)


Realm/Realm.Weaver/Analytics/AnalyticsUtils.cs
```diff
@@ -272,45 +269,6 @@ public static string GetAnonymizedUserId()
...
-                0 => "None",
-                1 => CpuArchitecture.Arm64,
-                2 => "Unity_Universal",
-                _ => Unknown(System.Environment.OSVersion.Platform.ToString())
-            };
-
-            return Unknown(convertedValue);
-        }
-
-        private static string ConvertArchitectureToMetricsVersion(string arch)
```
LaPeste, line 289: If I recall correctly the reason for this to exist was that we trusted what Mono.Cecil was telling us about the assembly architecture and it uses i386 or amd64. Since we don't use that info anymore I guess it's good for this to go.

Realm/Realm.UnityWeaver/UnityWeaver.cs
```diff
@@ -120,7 +121,7 @@ public static void Initialize()
 
                 WeaveAssemblyCore(assemblyPath, assembly.allReferences,
                     Metric.Framework.UnityEditor, Application.unityVersion,
-                    GetTargetOSName(Application.platform), netFrameworkInfo.TargetArchitecture,
+                    GetTargetOSName(Application.platform), GetHostCpuArchitecture(),
```
LaPeste, line 124: With this you want to say that the target arch is the same as the host's arch as it's mono building for the editor which is effectively running on the host/dev's machine?
Same question for the same change below.

Realm/Realm.UnityWeaver/UnityWeaver.cs
```diff
@@ -467,12 +468,50 @@ private static string GetTargetOSName(RuntimePlatform target)
...
+                    AndroidArchitecture.ARMv7 => CpuArchitecture.Arm,
+                    AndroidArchitecture.ARM64 => CpuArchitecture.Arm64,
+
+                    // These two don't have enum values in our Unity reference dll, but exist in newer versions
+                    // See https://github.com/Unity-Technologies/UnityCsReference/blob/70abf502c521c169ee8a302aa48c5600fc7c39fc/Editor/Mono/PlayerSettingsAndroid.bindings.cs#L14
+                    (AndroidArchitecture)(1 << 2) => CpuArchitecture.X86,
+                    (AndroidArchitecture)(1 << 3) => CpuArchitecture.X64,
+                    _ => CpuArchitecture.Universal,
+                },
+                BuildTarget.StandaloneWindows64 or BuildTarget.StandaloneLinux64 or BuildTarget.XboxOne => CpuArchitecture.X64,
```
LaPeste, line 501: How do you know and how have you tested that `BuildTarget.XboxOne` is 64 bit?

LaPeste (commented):
The structure looks fine, I just have some questions.

Realm/Realm.UnityWeaver/UnityWeaver.cs
nirinchev: I read on Wikipedia that Xbox One and newer consoles are 64 bit. Haven't tested it and only added it because I'm thinking we could setup the uwp wrappers to deploy on Xbox builds as experimental support.

Realm/Realm.UnityWeaver/UnityWeaver.cs
nirinchev: Yes, since the editor build is going to be deployed on the same machine it's running on, it's safe to assume the editor architecture is the same as the app architecture.

PR #3259, opened 2023-03-02
Add support for compensating writes

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/3258

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm/Exceptions/CompensatingWriteException.cs
```diff
@@ -0,0 +1,80 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Realms.Sync.Exceptions;
+
+namespace Realms.Exceptions
```
fealebenpae, line 23: Why not add this to `Realm.Sync.Exceptions`?

Realm/Realm/Native/SyncError.cs
```diff
@@ -0,0 +1,49 @@
...
+        public MarshaledVector<StringStringPair> user_info_pairs;
+
+        public MarshaledVector<CompensatingWriteInfo> compensating_writes;
+    }
+
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct CompensatingWriteInfo
+    {
+        public PrimitiveValue reason;
+        public PrimitiveValue object_name;
```
fealebenpae, line 46: I really don't like we use a whole `PrimitiveValue` when we just need a string 😛
I've been working on a new `String` struct for native marshaling for the websocket code we could use instead.

wrappers/src/marshalling.hpp
```diff
@@ -105,6 +105,29 @@ typedef struct realm_value {
...
+typedef struct realm_sync_error {
+    int32_t error_code;
+    realm_value_t message;
+    realm_value_t log_url;
+    bool is_client_reset;
+
+    struct {
+        std::pair<char*, char*>* data;
+        size_t count;
+    } user_info_pairs;
```
fealebenpae, line 123: Looks like we should make a native `marshaled_vector<T>` struct.

wrappers/src/marshalling.hpp
```diff
@@ -105,6 +105,29 @@ typedef struct realm_value {
     }
 } realm_value_t;
 
+typedef struct realm_sync_error_compensating_write_info {
```
fealebenpae, line 108: The `typedef struct foobar { ... } foobar` syntax is unnecessary in C++.
Simply declaring
```c++
struct foobar {
   ...
};
```
is sufficient. `realm_value_t` was copied from a C header, where the declaration needs to be more verbose.

fealebenpae (changes requested):

Realm/Realm/Native/SyncError.cs
nirinchev: Isn't a "whole" `PrimitiveValue` like 17 bytes instead of the 16 (on 64 bit platforms) you would otherwise use 😄 We also have a [`StringValue`](https://github.com/realm/realm-dotnet/blob/547ba96b29613fbe2e525bc0c974835efafa87da/Realm/Realm/Native/PrimitiveValue.cs#L285-L290) type which I could have used instead, but was too lazy to move some code around. As far as I can tell, it's the same as the `String` struct you're proposing in your PR.

fealebenpae: Oh, you're right - I missed `StringValue`.

I don't think using `PrimitiveValue` here is a good idea, because I suspect marshaling a union is more expensive as .NET doesn't have "unions" the way C understands them.

nirinchev: Okay, it was bugging me anyway that we're not more explicit about the field types. I replaced usages with `StringValue`.

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -28,7 +28,7 @@
 using Realms.Sync.ErrorHandling;
 using Realms.Sync.Exceptions;
 using Realms.Sync.Native;
-using CompensatingWriteInfo = Realms.Exceptions.CompensatingWriteInfo;
+using CompensatingWriteInfo = Realms.Sync.Exceptions.CompensatingWriteInfo;
```
fealebenpae, line 31:
```suggestion
```
You don't need this anymore as `Realms.Sync.Exceptions` is already used above.

fealebenpae (approved):

Realm/Realm/Handles/SessionHandle.cs
nirinchev: We do because there's a native `CompensatingWriteInfo` struct as well as the public-facing object of the same name we return to users. Without this line, I'd need to fully qualify usages of `CompensatingWriteInfo`.

fealebenpae: Derp. I missed that. Makes sense.

nirinchev: Moved the native struct to a nested struct and now we really don't need it.

PR #3263, opened 2023-03-15
Create simple project with Avalonia

papafe:
This PR adds an example project that uses Realm and [Avalonia](https://github.com/AvaloniaUI/Avalonia).

Differently from the example that we already have this application is mainly targeting desktop platforms, that is why the main view is a split view, with the todo lists collection on the left, and the current opened todo list on the right. For now I've tried it only on MacOS, and it looks good on it. 

##  TODO

* [x] README.md


examples/SimpleToDoAvalonia/SimpleToDo.csproj
```diff
@@ -0,0 +1,30 @@
...
+        <AvaloniaUseCompiledBindingsByDefault>true</AvaloniaUseCompiledBindingsByDefault>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <Folder Include="Models\" />
+        <AvaloniaResource Include="Assets\**" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <TrimmerRootAssembly Include="Avalonia.Themes.Fluent" />
```
nirinchev, line 17: This doesn't seem to be committed.

nirinchev (approved):
Looks good - I added a couple of very minor suggestions in https://github.com/realm/realm-dotnet/pull/3266 that are up to you whether you want to incorporate or not.

examples/SimpleToDoAvalonia/SimpleToDo.csproj
papafe: Good catch! I didn't look too much at the default project file

papafe: The assembly?

papafe:
@nirinchev I've merged your changes (thanks!), added the readme file for the example, as well as updated the example part of the main readme file too. 
Once we resolve the `TrimmerRootAssembly` discussion this is mergeable for me.

examples/SimpleToDoAvalonia/SimpleToDo.csproj
papafe: Aaah I see. I've actually been using Rider with this, since it was also what Avalonia recommended, and that doesn't show anything in the solution explorer. I suspect that is a VS Mac bug. 
I would prefer not to remove that line from the project file, given that it was given by the Avalonia template. They could be doing something dynamic with that assembly and that's why it was supposed to not be trimmed, so I wouldn't risk it 😁

nirinchev: Sounds good!

PR #3274, opened 2023-03-24
Annotate the test assemblies for nullability

nirinchev:
## Description

Part of https://github.com/realm/realm-dotnet/issues/3275

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


Realm/Realm.SourceGenerator/ClassCodeBuilder.cs
```diff
@@ -200,13 +200,16 @@ private string GeneratePartialClass(string interfaceString, string managedAccess
...
                     var shouldSetAlways = property.IsRequired ||
-                        property.TypeInfo.NullableAnnotation == NullableAnnotation.Annotated ||
-                        property.TypeInfo.IsRealmInteger ||
-                        property.TypeInfo.ScalarType == ScalarType.Date ||
-                        property.TypeInfo.ScalarType == ScalarType.Decimal ||
-                        property.TypeInfo.ScalarType == ScalarType.ObjectId ||
-                        property.TypeInfo.ScalarType == ScalarType.Guid;
+                        (property.TypeInfo.ScalarType == ScalarType.String && property.TypeInfo.NullableAnnotation != NullableAnnotation.Annotated) ||
+                        (property.TypeInfo.ScalarType == ScalarType.Data && property.TypeInfo.NullableAnnotation != NullableAnnotation.Annotated) ||
+                        property.TypeInfo.ScalarType == ScalarType.Date;
```
nirinchev, line 212: This is interesting - not sure why the previous rules were like that, but they didn't make a whole lot of sense to me. @papafe if you can remember why you had that condition and if you think I've broken something, let me know!

papafe: @nirinchev  So, this was taken almost verbatim from the weaver, so I think I didn't give much thought about it when I wrote it.
I suppose that if the default values for decimal, objectId, guid, and nullable properties is the same in core and .NET then your change is fine. 
Maybe we were being too "defensive" in the weaver?

nirinchev: Oh yeah, I should have thought of that. The reason why this is like that in the weaver is because those are complex structures and checking against the default is more involved. For ints, bools, and doubles we can check for default using `Brfalse_S` but we can't do that for structs.

Now that you mention it though, I can see we have this comment in the weaver:

https://github.com/realm/realm-dotnet/blob/c54b3d21125e738cd32257d38d2692d7e3354a3a/Realm/Realm.Weaver/RealmWeaver.cs#L1074

I wish I had added a more meaningful explanation for why this is the case 🫥

nirinchev: Okay, looks like I left this comment in the [PR description](https://github.com/realm/realm-dotnet/pull/857):

> It doesn't support nullable value types because the IL is getting out of hand as is and didn't seem like it would add a lot of value.

So I guess it's the same story as structs and now that we have SG, we can also handle this case fairly cheaply.

PR #3282, opened 2023-03-31
Add minimal FTS support

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/3281

Blocked on https://github.com/realm/realm-core/issues/6465

##  TODO

* [x] Changelog entry
* [x] Weaver support
* [x] FullTextMatch extension method
* [x] Schema tests
* [x] Tests


wrappers/src/utf8.hpp
```diff
@@ -0,0 +1,380 @@
+/*************************************************************************
```
papafe, line 1: What is this doing?

papafe (commented):
Makes sense!

wrappers/src/utf8.hpp
nirinchev: This was a file that does utf8<->utf16 conversion - it was previously in Core but got deleted.

Realm/Realm.SourceGenerator/InfoClasses.cs
```diff
@@ -236,6 +236,11 @@ public bool IsSupportedIndexType()
             return _indexableTypes.Contains(ScalarType);
         }
 
+        public bool IsSupportedFullTextType()
+        {
+            return CollectionType == CollectionType.None && ScalarType == ScalarType.String;
```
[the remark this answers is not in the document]
nirinchev: Oh, so we would get ScalarType == None when CollectionType != None?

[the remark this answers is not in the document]
papafe: Correct. `InternalType` has information on the type of the collection elements

Realm/Realm/Attributes/IndexMode.cs
```diff
@@ -0,0 +1,50 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+namespace Realms
+{
+#if PRIVATE_INDEXMODE
+    internal enum IndexMode
```
nirinchev, line 22: Rename to IndexType and set a base type.

Tests/Realm.Tests/Database/TestObjects.cs
```diff
@@ -1126,6 +1126,34 @@ public partial class RealmValueObject : TestRealmObject
...
 
+    public partial class ObjectWithFtsIndex : TestRealmObject
+    {
+        [PrimaryKey]
+        public string Title { get; set; }
+
+#if TEST_WEAVER
+        [Required]
+#endif
+        [Indexed(IndexMode.FullText)]
```
cmelchior, line 1137: Design question: Doesn't this API prevent further customization of Full-Text? I.e I would assume it is very possible that we would want to be able to configure various attributes on Full-Text once/when we increase its capabilities. I.e. this is the full-text annotation in Android: https://developer.android.com/reference/androidx/room/Fts4

nirinchev: It kinda does, yes. We could add more arguments to the `Indexed` attribute that relate only to FTS if we wanted to or we can introduce another attribute in the future. I opted against going the another attribute route right now because we don't have short/medium term plans to extend our FTS support and didn't want to bake too  many assumptions about the future of the feature into the initial API. It's also the API we agreed on in the Scope, so figured everyone is on board with it.

cmelchior: The scope mention that the API for other SDKs should be fleshed out in the design doc, which never seemed to happen, so not sure I agree this was the API we wanted to ship with. I read it as just an example of how it could look if we exposed the Core API.

Maybe I'm biased because other frameworks on Android use a dedicated annotation for this, but it seems counterproductive to ship an API that we know for sure we most likely have to break if/when we add more features, even if they are not planned right now. Especially since this API doesn't really provide any meaningful type-safety anyway. Extending the current `[Indexed]` with more attributes also seems a bit weird as they would be meaningless for the normal index type.

So I would strongly encourage a direction with its own annotation, In Kotlin I exposed it with a `@FullText` annotation here: https://github.com/realm/realm-kotlin/pull/1368/files#diff-327b30a309e034a752eb56ac993d64f6d291cd9e7af0ddd9018947bf6fddacf2R88

nirinchev: There's nothing preventing us from shipping a new attribute in the future in a backwards compatible way. I'd rather not prematurely optimize for a case when more configuration options may come in before we know what these will look like and if an attribute would be the best way to go about them - e.g. if we want to allow declaring an FTS index over multiple properties, doing so via attributes could end up being clunky and we may choose a different strategy. So rather than trying to design an API for the future, I'd rather extend the existing one we have that has a meaningful name and is well positioned to support our current needs. If our needs change beyond what is currently supported, we'll figure something out, but as I said, I don't envision this happening in the short/medium term - if ever.

cmelchior: I wouldn't call this a premature optimization, as semantically you are getting the same API, it is just more flexible. While I agree that an FTS index over multiple properties would be slightly awkward using annotations (but not impossible), most modifications like tokenizer strategy, casing, mixing normal and FTS index would all be supported easily by using a dedicated annotation.

cmelchior: I wouldn't call this a premature optimization, as semantically you are getting the same API, it is just more flexible. While I agree that an FTS index over multiple properties would be slightly awkward using annotations (but not impossible), most modifications like tokenizer strategy, casing, mixing normal and FTS index would all be supported easily by using a dedicated annotation.

PR #3300, opened 2023-04-24
Add support for geospatial queries

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/3299

## Summary

Adds initial support for geo queries. In this first version, we're not introducing new datatypes, so users will need to make sure their own types conform to the shape expected for geo queries. We're providing a few helper shapes to write queries against, though those can only be used for querying and cannot be stored in the database.

### Example point definition

```csharp
public partial class CustomGeoPoint : TestEmbeddedObject
{
    [MapTo("coordinates")]
    private IList<double> Coordinates { get; } = null!;

    [MapTo("type")]
    private string Type { get; set; } = "Point";

    public double Latitude => Coordinates.Count > 1 ? Coordinates[1] : throw new Exception($"Invalid coordinate array. Expected at least 2 elements, but got: {Coordinates.Count}");

    public double Longitude => Coordinates.Count > 1 ? Coordinates[0] : throw new Exception($"Invalid coordinate array. Expected at least 2 elements, but got: {Coordinates.Count}");

    public CustomGeoPoint(double latitude, double longitude)
    {
        Coordinates.Add(longitude);
        Coordinates.Add(latitude);
    }
}
```

### Example query

```csharp
var circle = new GeoCircle((23.459, 49.983), Distance.FromKilometers(10));

// Above example uses implicit conversion from a tuple to GeoPoint, you can also use the explicit one
var circle = new GeoCircle(new GeoPoint(latitude: 23.459, longitude: 49.983), Distance.FromKilometers(10));

// Construct a polygon from a collection of points. First and last one need to be the same
// in order to close the polygon.
var polygon = new GeoPolygon((10, 10), (10, 20), (0, 20), (10, 10));

// Construct a polygon with holes
var outerRing = new GeoPoint[] { (10, 10), (10, 20), (0, 20), (10, 10));
var hole1 = new GeoPoint[] { (1, 1), (1, 2), (0, 2), (1, 1));
var hole2 = new GeoPoint[] { (5, 5), (5, 6), (4, 6), (5, 5));
var polygon = new GeoPolygon(outerRing, hole1, hole2);

// Construct a box from bottom left and top right corners
var box = new GeoBox(bottomLeftCorner: (0, 0), topRightCorner: (10, 10));

// Query using LINQ
var query = realm.all<Company>().Where(c => QueryMethods.GeoWithin(c.Location, circle));

// Query using RQL
var query = realm.All<Company>().Filter("Location geoWithin $0", circle);
```

##  TODO

* [ ] Changelog entry
* [x] Tests (if applicable)


Realm/Realm/Native/Queries/NativeQueryArgument.cs
```diff
@@ -0,0 +1,89 @@
...
+        private NativeGeoBox _box;
+
+        [FieldOffset(0)]
+        private NativeGeoPolygon _polygon;
+
+        [FieldOffset(0)]
+        private NativeGeoSphere _sphere;
+
+        [FieldOffset(0)]
+        private NativeGeoPoint _point;
```
nirinchev, line 39: Remove for now

Realm/Realm/DatabaseTypes/Geospatial/GeoPolygon.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.Linq;
+using System.Runtime.InteropServices;
+using Realms.Native;
+
+namespace Realms
+{
+    public class GeoPolygon : GeoShapeBase
+    {
+        public GeoPoint[] Points { get; }
```
nirinchev, line 28: Use ReadonlyList or IEnumerable

wrappers/src/filter.hpp
```diff
@@ -46,12 +46,45 @@ inline Results* get_filtered_results(const SharedRealm& realm, const ConstTableR
...
+        case query_argument_type::POLYGON:
+        case query_argument_type::SPHERE:
+            ++geo_store_size;
+            break;
         }
-        else {
-            mixed_args.push_back(from_capi(arguments[i].link.object, true));
+    }
+
+    geo_store.reserve(geo_store_size);
```
nirinchev, line 63: Consider allocating `args_count` for the geo_store, even though they're not likely to be all geo - it simplifies the code.

wrappers/src/marshalling.hpp
```diff
@@ -45,6 +45,13 @@ enum class realm_value_type : unsigned char {
     RLM_TYPE_UUID,
 };
 
+enum class query_argument_type : unsigned char {
```
nirinchev, line 48: uint8_t

wrappers/src/marshalling.hpp
```diff
@@ -105,6 +112,37 @@ typedef struct realm_value {
     }
 } realm_value_t;
 
+typedef struct geo_point {
+    double latitude;
+    double longitude;
+} geo_point_t;
```
nirinchev, line 118:
```suggestion
struct geo_point {
    double latitude;
    double longitude;
};
```

CHANGELOG.md
```diff
@@ -6,6 +6,71 @@
...
 * Support sort/distinct based on values from a dictionary e.g. `.Filter("TRUEPREDICATE SORT(meta['age'])")`. (Core 13.14.0)
+* Added initial support for geospatial queries on points. (Issue [#3299](https://github.com/realm/realm-dotnet/issues/3299))
+  * In this version, only queries of the form "is this point contained in this shape" (equivalent to [$geoWithin](https://www.mongodb.com/docs/manual/reference/operator/query/geoWithin/) in MongoDB) are supported.
+  * There is no index support right now.
+  * There is no dedicated type for persisted geospatial points. Instead, points should be stored as GeoJson-shaped embedded object and queries will use duck-typing to check if the shape contains the object. For convenience, here's an example embedded object that you can use in liew of a Realm-provided dedicated type:
```
papafe, line 12:
```suggestion
  * There is no dedicated type for persisted geospatial points. Instead, points should be stored as GeoJson-shaped embedded object and queries will use duck-typing to check if the shape contains the object. For convenience, here's an example embedded object that you can use in lieu of a Realm-provided dedicated type:
```

Realm/Realm/DatabaseTypes/Geospatial/GeoPolygon.cs
```diff
@@ -0,0 +1,159 @@
...
+        public GeoPolygon(IEnumerable<GeoPoint> outerRing, params IEnumerable<GeoPoint>[] holes)
+            : this(new[] { outerRing.ToList().AsReadOnly() }.Concat(holes.Select(h => h.ToList().AsReadOnly())).ToArray())
+        {
+        }
+
+        private GeoPolygon(IReadOnlyList<GeoPoint>[] linearRings)
+        {
+            foreach (var polygon in linearRings)
+            {
+                Argument.Ensure(polygon.Count > 3, $"Each linear ring (both the outer one and any holes) must have at least 4 points, but {LinearRingToString(polygon)} only had {polygon.Count}.", nameof(linearRings));
```
papafe, line 109: In the docs for the class it says that each ring needs to have at least 3 points, not 4

Realm/Realm/DatabaseTypes/Geospatial/GeoPolygon.cs
```diff
@@ -0,0 +1,159 @@
...
+                Argument.Ensure(polygon[0] == polygon[polygon.Count - 1], $"The first and the last points of the polygon {LinearRingToString(polygon)} must be the same.", nameof(linearRings));
```
papafe, line 110: Same here, I thought in this case the polygon would be considered close anyways

papafe (approved):
Apart from those few notes it looks fine

Realm/Realm/DatabaseTypes/Geospatial/GeoPolygon.cs
nirinchev: The docs say 3 distinct points and the first and the last must match, implying at least 4 points.

Realm/Realm/DatabaseTypes/Geospatial/GeoPolygon.cs
nirinchev: That was the case at some point, but Core changed it, so we now need to explicitly close the polygon. Will update the docs.

nirinchev: Actually, I can see that MongoDB is more relaxed than GeoJSON, so might be worth doing it for the SDKs as well.

PR #3321, opened 2023-05-11
[WIP] Fix sourcelink

### LINKED ISSUE #3320: Fix support for SourceLink
We should fix our support for [SourceLink](https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink) as it would allow developers (and us) to step into Realm source code. 
At the moment we already have the necessary tags in the project file (https://github.com/realm/realm-dotnet/blob/39d9b02cd780c56e429bacc6d50cab4cce3ace47/Realm/AssemblyInfo.props), but it's not working on either Windows or MacOS. We should investigate why.

In the online nuget explorer (https://nuget.info/packages/Realm/11.0.0), it says that SourceLink it's not working because we don't include Symbols for the UWP wrapper in our package, but we do. It seems that SkiaSharp has a similar issue (https://nuget.info/packages/SkiaSharp.NativeAssets.UWP/2.88.3).
Obviously the cause could be something completely different.

papafe:
Just a test to see if updating the sourcelink package fixes our issue

Fixes #3320 

##  TODO

* [ ] Changelog entry

### DIFF
--- a/Realm/AssemblyInfo.props
+++ b/Realm/AssemblyInfo.props
@@ -21,7 +21,7 @@
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


papafe:
Didn't try installing the nuget package locally, but the nuget package analyzer still reports the same issues

PR #3394, opened 2023-07-25
Added offline login article

### LINKED ISSUE #3326: Example: Offline login/ realm access

papafe:
This PR shows a draft of the article we could publish as a guide for offline login with Realm in the .NET SDK. 
Probably we are not going to add this in our repo, but I have created a PR to get an initial technical review.

Fixes #3326

### DIFF
--- a/examples/OfflineLogin.md
+++ b/examples/OfflineLogin.md
@@ -1,0 +1,55 @@
+# Offline synced realm in the .NET Realm SDK #
+
+With MongoDB Realm offline-first capabilities, it is always possible to read and write to the database even when using Device Sync, independently from the connection status of the app. In this small article, we are going to see how to open a synced realm depending on the situation in the .NET Realm SDK
+
+There are two methods in the .NET Realm SDK that can be used to open a Realm: `Realm.GetInstance` (synchronously) and `Realm.GetInstanceAsync` (asynchronously). The two methods not only differ in asynchronicity, but have a different behavior when used to open a synced realm.
+
+### GetInstance ###
+
+This method will return as soon as the realm has been opened. For this reason, this method is the recommended one to use in the majority of application, as it works independently from the connection status. Once the realm has been opened, it will continue to synchronize in the background. 
+
+
+### GetInstanceAsync ###
+
+This method will complete after the realm has been opened and is fully synchronized. For this reason, this method is recommended to use only when it is essential to work with a fully synchronized realm, for example just after the first login of the user on a freshly installed application. 
+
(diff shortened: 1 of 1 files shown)


nirinchev (commented):
I think we should include a complete code snippet for the recommended flow. Something like:

```csharp
Realm realm;
var user = app.CurrentUser;
if (user == null)
{
    // We don't have a user - show login screen and wait until the user has logged in
    await PresentLoginScreen();

    // Since we just logged in, we're likely to have connectivity
    // so open the Realm async
    var cts = new CancellationTokenSource(5000);
    try
    {
        realm = Realm.GetInstanceAsync(GetConfig(), cts.Token); 
    }
    catch (TaskCanceledException)
    {
        realm = Realm.GetInstance(GetConfig());
    }
}
else
{
    // We've already logged in, so immediately open the Realm with
    // whatever data there is. If your app absolutely needs the latest
    // data, then swap that out for GetInstanceAsync.
    realm = Realm.GetInstance(GetConfig());
}
```

examples/OfflineLogin.md
```diff
@@ -0,0 +1,90 @@
+# Offline synced realm in the .NET Realm SDK
+
+With MongoDB Realm offline-first capabilities, it is always possible to read and write to the database even when using Device Sync, independently from the connection status of the app. In this small article, we are going to see how to open a synced realm depending on the situation in the .NET Realm SDK
+
+There are two methods in the .NET Realm SDK that can be used to open a Realm: `Realm.GetInstance` (synchronously) and `Realm.GetInstanceAsync` (asynchronously). The two methods not only differ in asynchronicity, but have a different behavior when used to open a synced realm. We will first take a look at those two methods on their own, then show a recommended flow for opening a realm. 
+
+#### GetInstance
```
nirinchev, line 7: [q] Why are these h4? Seems kind of weird going from h1 to h4 with nothing in between.

examples/OfflineLogin.md
```diff
@@ -0,0 +1,90 @@
...
+    ```
+    In most cases this approach could be an overkill, as the exception would be raised for any kind of non-fatal errors, even for just a brief connection issue. For this reason it is recommended to use the previous approach, as it allows to retry in the specified time frame.
+
+### Recommended flow
+
+Given the differences between `GetInstance` and `GetInstanceAsync` that we have presented, here is an example of a recommended flow that uses both methods to open a realm, and that can be used in most applications:
+
+```csharp
+var user = app.CurrentUser;
+var realmConfig = GetRealmConfig();
```
nirinchev, line 59: You can't create a config without the user, so this will need to be moved further down.

nirinchev (approved):
Final nits from me

papafe:
This has been added to the QuickJournalSync example, so I'm closing this

PR #3403, opened 2023-07-31
Add new subscribe API

nirinchev:
## Description

Fixes RNET-865

Adds `IQueryable.SubscribeAsync` syntax sugar.
Adds `cancellationToken` argument to `Session.WaitForDownloadAsync`, `Session.WaitForUploadAsync`, and `SubscriptionSet.WaitForSynchronization`. This doesn't cancel the underlying Core wait operation, just the managed task. While it's not ideal, the native wait operation is not really expensive and it will eventually complete anyway, so we are not leaking resources.

There are some whitespace changes, so it might be a good idea to review with the "Ignore whitespace" option.

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -209,22 +210,36 @@ public void UnsubscribeNotifications()
             }
         }
 
-        public async Task WaitAsync(ProgressDirection direction)
+        public async Task WaitAsync(ProgressDirection direction, CancellationToken? cancellationToken)
         {
             var tcs = new TaskCompletionSource();
+            if (cancellationToken?.IsCancellationRequested == true)
+            {
+                tcs.TrySetCanceled(cancellationToken.Value);
```
papafe, line 218: I am not sure I understand why we need this and the following line. Can we just return?

papafe (approved):
Looks good. 
By the way, it was a little bit difficult to see the body of `SubscribeAsync` together  with the removal of the namespace block 😁

Tests/Realm.Tests/Database/APITests.cs
```diff
@@ -19,6 +19,7 @@
 using System;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
```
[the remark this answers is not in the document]
nirinchev: We don't - had some tests there, that I then removed and this was a leftover.

Realm/Realm/Handles/SessionHandle.cs
nirinchev: If cancellation is requested, the expected behavior would be to throw a `TaskCancelledException`, even if we didn't start doing the work. Regarding why we're cancelling the tcs vs throwing a `TaskCancelledException`, I guess both are roughly the same and don't have strong feelings about one or the other.

nirinchev:
> By the way, it was a little bit difficult to see the body of SubscribeAsync together with the removal of the namespace block

That's why you're supposed to read the PR description 😜

papafe (approved):

Issue #3149, opened 2023-01-05
Public name not set when creating native schema


nirinchev:
To allow RQL queries that use the managed property name rather than the internal one, we need to configure `public_name` in the Core schema. See https://github.com/realm/realm-dart/pull/1078 for inspiration. A minimal test would be something like:

```csharp
partial class Foo : IRealmObject
{
    [MapTo("bar")]
    public string Value { get; set; }
}

var foos = realm.All<Foo>().Filter("Value == $0", "abc");
```

PR #3409, opened 2023-08-02
Pass public_name down to Core

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/3149

##  TODO

* [ ] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Schema/Property.cs
+++ b/Realm/Realm/Schema/Property.cs
@@ -181,7 +181,7 @@ namespace Realms.Schema
         internal SchemaProperty ToNative(Arena arena) => new()
         {
             name = StringValue.AllocateFrom(Name, arena),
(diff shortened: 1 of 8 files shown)


Realm/Realm/Schema/Property.cs
```diff
@@ -180,6 +181,7 @@ internal Property(in SchemaProperty nativeProperty)
         internal SchemaProperty ToNative(Arena arena) => new()
         {
             name = StringValue.AllocateFrom(Name, arena),
+            managed_name = StringValue.AllocateFrom(ManagedName != Name ? ManagedName : string.Empty, arena),
```
fealebenpae, line 184: You can initialize this with `null` instead of `string.Empty` - `capi_to_std(realm_string_t)` knows to return an empty `std::string` if the marshaled string buffer is null. This shaves off an empty, unnecessary allocation.

fealebenpae (approved):

PR #3414, opened 2023-08-09
Improvements related to Remote MongoDB access

nirinchev:
## Description

This is still WIP, but explores a SG-based approach for using RealmObject classes for remote mongodb access.

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)
* [ ] Link properties
* [ ] Asymmetric objects
* [ ] Embedded properties


Tests/Realm.Tests/Sync/MongoClientTests.cs
```diff
@@ -2173,6 +2261,47 @@ private async Task<MongoClient.Collection<Foo>> GetCollection()
...
+
+            // Use sync to create the schema/rules
+            SyncConfigurationBase config = appConfigType == AppConfigType.FlexibleSync ? GetFLXIntegrationConfig(user) : GetIntegrationConfig(user);
+
+            config.Schema = GetTransitiveSchema(typeof(T), new()).Values.ToArray();
+
+            using var realm = await GetRealmAsync(config);
+            await WaitForUploadAsync(realm);
+
+            var client = user.GetMongoClient(ServiceName); // TODO: this should be provided by Sync
```
nirinchev, line 2278: No longer relevant.

Tests/Realm.Tests/Sync/MongoClientTests.cs
```diff
@@ -2173,6 +2261,47 @@ private async Task<MongoClient.Collection<Foo>> GetCollection()
...
+        private async Task<MongoClient.Collection<T>> GetCollection<T>(string appConfigType = AppConfigType.Default)
+            where T : class, IRealmObject
+        {
+            var app = App.Create(SyncTestHelpers.GetAppConfig(appConfigType));
+            var user = await GetUserAsync(app);
+
+            // Use sync to create the schema/rules
+            SyncConfigurationBase config = appConfigType == AppConfigType.FlexibleSync ? GetFLXIntegrationConfig(user) : GetIntegrationConfig(user);
+
+            config.Schema = GetTransitiveSchema(typeof(T), new()).Values.ToArray();
```
papafe, line 2273: How come we're trying to limit the schema here? Is it to improve the speed of the test?

papafe:
I've just merged my branch into this one. 
It seems the only thing left to do is decide how to deal with the change in the default serialization of types.

papafe:
@nirinchev As we discussed:
- I've added some tests to document the issue with `DateTimeOffset`
- I've added the change in the breaking change section of the changelog
- I've added the static method `Realm.SetLegacySerialization` so that developers can use the previous serialization.
- I've added a test to verify that the static method works, but unfortunately I've put the `Ignore` tag on it as it should only be tested manually. Serializers get cached once used, so if we change to the legacy serialization we can't switch back. 

If the CI is green (or almost) this is mergeable

PR #3416, opened 2023-08-10
Avoiding GC of session when there are subscribers

papafe:
We are trying to avoid that `realm.SyncSession` gets garbage collected, at least until there are subscribers to `realm.SyncSession.PropertyChanged`

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Realm/Realm/Realm.cs
+++ b/Realm/Realm/Realm.cs
@@ -171,7 +171,7 @@ namespace Realms
         private WeakReference<SubscriptionSet>? _subscriptionRef;
 
         private State _state;
-        private WeakReference<Session>? _sessionRef;
+        private SessionProvider? _sessionProvider;
         private Transaction? _activeTransaction;
 
         internal readonly SharedRealmHandle SharedRealmHandle;
@@ -247,22 +247,9 @@ namespace Realms
                     throw new NotSupportedException("Realm.SyncSession is only valid for synchronized Realms (i.e. ones that are opened with FlexibleSyncConfiguration or PartitionSyncConfiguration).");
                 }
 
-                if (_sessionRef is null || !_sessionRef.TryGetTarget(out var session) || session.IsClosed)
-                {
-                    var sessionHandle = SharedRealmHandle.GetSession();
-                    session = new Session(sessionHandle);
+                _sessionProvider ??= new SessionProvider();
 
-                    if (_sessionRef is null)
-                    {
-                        _sessionRef = new WeakReference<Session>(session);
-                    }
-                    else
-                    {
-                        _sessionRef.SetTarget(session);
-                    }
-                }
-
-                return session;
+                return _sessionProvider.GetSession(SharedRealmHandle);
             }
         }
 
@@ -1356,6 +1343,54 @@ namespace Realms
         }
 
         #endregion Transactions
+
+        internal class SessionProvider
+        {
+            private WeakReference<Session>? _weakSessionRef;
+            private Session? _strongSessionRef;
+
+            public Session GetSession(SharedRealmHandle handle)
+            {
+                if(_strongSessionRef?.IsClosed == false)
+                {
+                    return _strongSessionRef;
+                }
+                else if (_weakSessionRef?.TryGetTarget(out var targetSession) == true && !targetSession.IsClosed)
+                {
+                    return targetSession;
+                }
+
+                var session = new Session(handle.GetSession());
+                session.Subscribed += Session_Subscribed;
+                session.Unsubscribed += Session_Unsubscribed;
+
+                if (_weakSessionRef is null)
+                {
+                    _weakSessionRef = new WeakReference<Session>(session);
+                }
+                else
+                {
+                    _weakSessionRef.SetTarget(session);
+                }
+
+                return session;
+            }
+
+            private void Session_Subscribed(object? sender, EventArgs e)
+            {
+                if (sender is not Session session)
+                {
+                    return;
+                }
+
+                _strongSessionRef = session;
+            }
+
+            private void Session_Unsubscribed(object? sender, EventArgs e)
+            {
+                _strongSessionRef = null;
+            }
+        }
 
         internal class RealmMetadata
         {
--- a/Realm/Realm/Realm.cs
+++ b/Realm/Realm/Realm.cs
@@ -247,9 +247,9 @@ namespace Realms
                     throw new NotSupportedException("Realm.SyncSession is only valid for synchronized Realms (i.e. ones that are opened with FlexibleSyncConfiguration or PartitionSyncConfiguration).");
                 }
 
-                _sessionProvider ??= new SessionProvider();
+                _sessionProvider ??= new SessionProvider(SharedRealmHandle);
 
-                return _sessionProvider.GetSession(SharedRealmHandle);
+                return _sessionProvider.GetSession();
             }
         }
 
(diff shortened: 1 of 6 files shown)


Realm/Realm/Realm.cs
```diff
@@ -247,22 +247,9 @@ public Session SyncSession
...
-                    if (_sessionRef is null)
-                    {
-                        _sessionRef = new WeakReference<Session>(session);
-                    }
-                    else
-                    {
-                        _sessionRef.SetTarget(session);
-                    }
-                }
+                _sessionProvider ??= new SessionProvider();
```
nirinchev, line 250: You can pass the handle in the provider ctor as it doesn't appear like the provider is shared between different realms.

Realm/Realm/Realm.cs
```diff
@@ -1357,6 +1344,54 @@ internal void ExecuteOutsideTransaction(Action action)
...
+            private WeakReference<Session>? _weakSessionRef;
+            private Session? _strongSessionRef;
+
+            public Session GetSession(SharedRealmHandle handle)
+            {
+                if(_strongSessionRef?.IsClosed == false)
+                {
+                    return _strongSessionRef;
+                }
+                else if (_weakSessionRef?.TryGetTarget(out var targetSession) == true && !targetSession.IsClosed)
```
nirinchev, line 1358: No need for an `else`.

Realm/Realm/Realm.cs
```diff
@@ -1357,6 +1344,54 @@ internal void ExecuteOutsideTransaction(Action action)
...
+                session.Unsubscribed += Session_Unsubscribed;
+
+                if (_weakSessionRef is null)
+                {
+                    _weakSessionRef = new WeakReference<Session>(session);
+                }
+                else
+                {
+                    _weakSessionRef.SetTarget(session);
+                }
```
nirinchev, line 1374: Why not allocate a new reference in both cases? Are we winning a lot by calling `SetTarget`?

Realm/Realm/Realm.cs
```diff
@@ -1357,6 +1344,54 @@ internal void ExecuteOutsideTransaction(Action action)
...
+                    return;
+                }
+
+                _strongSessionRef = session;
+            }
+
+            private void Session_Unsubscribed(object? sender, EventArgs e)
+            {
+                _strongSessionRef = null;
+            }
```
nirinchev, line 1392: Rather than use events, we can pass the provider to the session constructor and have the session call methods on the provider. I know it's roughly the same thing, but I feel it's more straightforward.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -1016,13 +1016,73 @@ public void Session_Free_Instance_Even_With_PropertyChanged_Subscribers()
...
+
+                realm.SyncSession.PropertyChanged += HandlePropertyChanged;
+
+                GC.Collect();
+                Assert.That(weakSessionRef.IsAlive, Is.True);
+
+                realm.SyncSession.PropertyChanged -= HandlePropertyChanged;
+
+                GC.Collect();
+                Assert.That(weakSessionRef.IsAlive, Is.False);  //TODO This fails
```
nirinchev, line 1078: We do have a `WaitUntilReferencesAreCollected` method in TestHelpers that runs GC.Collect with retries - not sure why this test is failing but may be worth trying to use that instead.

Realm/Realm/Realm.cs
papafe: I took the code the code that was already there, and I thought it made sense, so we avoid having the GC collecting all those "old" `WeakReference`?

Realm/Realm/Realm.cs
papafe: The constructor is used also here : https://github.com/realm/realm-dotnet/blob/68febca6905095940b22bcee1d7f58e57aa2910b/Realm/Realm/Handles/SessionHandle.cs#L293

So I think it would be awkward to pass a nullable provider. 
Even not considering that, I am not totally convinced about passing the provider as an argument to the constructor, because it means that now `Session` would have a dependency on `SessionProvider`, that is not necessary.

nirinchev: Fair enough. I do feel the session already has a dependency on the `SessionProvider`, but we're trying to be roundabout about it. These events don't make a ton of sense by themselves and we're only raising them because we know there's going to be some very special class that cares about whether the session has subscribers or not. Since we're modifying the Session API to cater to the new provider class we're introducing and we don't expect these events to ever be used by anything else, I'd say we have a pretty strong coupling between those two types. That being said, it's up to you to decide on the design - I'd be fine with either approach.

Realm/Realm/Realm.cs
nirinchev: I guess whoever wrote the old code must have been very smart 🤓 Perhaps a bit too smart - I'd be surprised if that ever made a difference as it'd only trigger if the user is accessing the session property multiple times and letting it get GC-ed in between, so clearly not a hot path. In any case, since it's moved code, you don't need to simplify it unless you want to.

Realm/Realm/Realm.cs
papafe: Done

Realm/Realm/Realm.cs
papafe: Done

Realm/Realm/Realm.cs
papafe: Yeah, I agree, premature optimisation probably, I've removed it 😁

Tests/Realm.Tests/Sync/SessionTests.cs
papafe: That worked nicely 😄

Realm/Realm/Realm.cs
papafe: I see your point. I changed to pass a delegate to the constructor, it feels like a middle-ground, so we don't need to make the methods on the provider public.

papafe:
I fixed the PR according to comments. Apart from the discussion about coupling, this is ready to merge for me.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -1016,13 +1016,76 @@ public void Session_Free_Instance_Even_With_PropertyChanged_Subscribers()
...
+
+                // We want to have more assurance that the reference is not being collected at a later time
+                await WaitUntilReferencesAreCollected(500, weakSessionRef);
+                GC.Collect();
+                Assert.That(weakSessionRef.IsAlive, Is.True);
+
+                realm.SyncSession.PropertyChanged -= HandlePropertyChanged;
+
+                await WaitUntilReferencesAreCollected(500, weakSessionRef);
+                GC.Collect();
```
nirinchev, line 1080: This is no longer necessary - `WaitUntilReferencesAreCollected` will call it internally.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -1016,13 +1016,76 @@ public void Session_Free_Instance_Even_With_PropertyChanged_Subscribers()
...
+                    return new WeakReference(realm.SyncSession);
+                })();
+
+                Assert.That(weakSessionRef.IsAlive, Is.True);
+
+                realm.SyncSession.PropertyChanged += HandlePropertyChanged;
+
+                // We want to have more assurance that the reference is not being collected at a later time
+                await WaitUntilReferencesAreCollected(500, weakSessionRef);
+                GC.Collect();
```
nirinchev, line 1074: Same here - probably not needed.

nirinchev (approved):
Two minor nits + need to fix the js-style brace alignment, otherwise good to go.

PR #3421, opened 2023-08-14
PR suggestions for yg/websocket

nirinchev:
## Description

Fixes #

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


Realm/Realm/Native/SyncSocketProvider.WebSocket.cs
```diff
@@ -232,30 +225,28 @@ protected unsafe override void Execute(IntPtr observer)
...
                 {
                     NativeMethods.observer_error_handler(observer);
                 }
 
-                var bytes = Encoding.UTF8.GetBytes(_description);
-                fixed(byte* data = bytes)
-                {
-                    NativeMethods.observer_closed_handler(observer, _clean, _status, new() { data = data, size = bytes.Length });
-                }
+                // TODO: should we execute the closed handler if the error handler is called?
```
fealebenpae, line 247: AFAIK yes, at least that's the behavior of browser websockets we emulate.

PR #3422, opened 2023-08-14
Fix incorrect assertion and cleanup CollectionTests.cs

nirinchev:
## Description

Fixes an incorrect assertion where we were asserting on the wrong collection and cleans up the code a little. Make sure to review with whitespace off since I moved the namespace to a file-scoped syntax.

### DIFF
--- a/Tests/Realm.Tests/Database/CollectionTests.cs
+++ b/Tests/Realm.Tests/Database/CollectionTests.cs
@@ -30,2153 +30,2154 @@ using MongoDB.Bson;
 using NUnit.Framework;
 using Realms.Exceptions;
 
-namespace Realms.Tests.Database;
-
-[TestFixture, Preserve(AllMembers = true)]
-public class CollectionTests : RealmInstanceTest
+namespace Realms.Tests.Database
 {
-    private const string AnimalFarm = "Animal Farm";
-    private const string LordOfTheRings = "The Lord of the Rings";
-    private const string LordOfTheFlies = "Lord of the Flies";
-    private const string WheelOfTime = "The Wheel of Time";
-    private const string Silmarillion = "The Silmarillion";
-
-    [Test]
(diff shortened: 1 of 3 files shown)


Tests/Realm.Tests/Database/CollectionTests.cs
```diff
@@ -24,2180 +29,2154 @@
...
-            var summaryMatches = _realm.All<ObjectWithFtsIndex>().Filter($"{nameof(ObjectWithFtsIndex.Summary)} TEXT $0", testData.Query).ToArray().Select(o => o.Title);
-            Assert.That(summaryMatches, Is.EquivalentTo(testData.ExpectedResults));
+        var summaryMatches = _realm.All<ObjectWithFtsIndex>().Filter($"{nameof(ObjectWithFtsIndex.Summary)} TEXT $0", testData.Query).ToArray().Select(o => o.Title);
+        Assert.That(summaryMatches, Is.EquivalentTo(testData.ExpectedResults));
 
-            var nullableSummaryMatches = _realm.All<ObjectWithFtsIndex>().Filter($"{nameof(ObjectWithFtsIndex.NullableSummary)} TEXT $0", testData.Query).ToArray().Select(o => o.Title);
-            Assert.That(summaryMatches, Is.EquivalentTo(testData.ExpectedResults));
-        }
+        var nullableSummaryMatches = _realm.All<ObjectWithFtsIndex>().Filter($"{nameof(ObjectWithFtsIndex.NullableSummary)} TEXT $0", testData.Query).ToArray().Select(o => o.Title);
+        Assert.That(nullableSummaryMatches, Is.EquivalentTo(testData.ExpectedResults));
```
nirinchev, line 1936: This was the wrong assertion - even though we were running the query on the nullable field, we weren't validating the results against the expected results.

papafe (approved):

PR #3426, opened 2023-08-17
Starting one cluster per platform on CI

### LINKED ISSUE #3423: Fix timeouts on CI
We are experiencing timeouts on the `main` action because we have multiple atlas app congesting the cluster. In order to solve this we have decided to opt for creating a new cluster for each platform.

papafe:
Fixes #3423


Tools/DeployApps/BaasClient.cs
```diff
@@ -158,7 +159,7 @@ private string _shortDifferentiator
             }
         }
 
-        private string _appSuffix => $"-{_shortDifferentiator}-{_clusterName}";
+        private string _appSuffix => $"-{_shortSuffix}";
```
papafe, line 162: If we're running all tests on different clusters, do we even need to have the suffix?

papafe:
It seems that this is working properly. I'm not sure about Android though, because it seems the timeout is not respected, I suppose there is a timeout inside the device farm?

.github/workflows/deploy-baas.yml
```diff
@@ -26,12 +26,20 @@ env:
...
     outputs:
-      clusterName: ${{ steps.deploy-cluster.outputs.clusterName }}
+      clusterName: ${{ steps.generate-cluster-name.outputs.clusterName }}
     timeout-minutes: 20
     steps:
+    - name: Generate cluster name
+      id: generate-cluster-name
+      run: |
+        name=$( echo ${{ github.run_id }}-${{ github.run_attempt }} | md5sum | cut -c1-7)
+        echo "clusterName=$name" >> "$GITHUB_OUTPUT"
```
nirinchev, line 42: Why not output the cluster name + the differentiator at this point already? Then we don't need to remember to append the differentiator everywhere else.

nirinchev: Oh.. is it because we can't create dynamic job outputs for matrix jobs?

.github/workflows/test-macos.yml
```diff
@@ -28,7 +28,7 @@ jobs:
   test-xamarin:
     runs-on: macos-latest
     name: Xamarin.macOS
-    timeout-minutes: 45
+    timeout-minutes: 50
```
nirinchev, line 31: We can bump that to 60 to be on the safe side if 45 was too little.

Tools/DeployApps/BaasClient.cs
nirinchev: We probably still want to differentiate the apps (so that we don't have a gazillion apps with the same name in the UI). If we never look at the UI, then you're right that it doesn't matter.

Tools/DeployApps/BaasClient.cs
```diff
@@ -535,7 +536,7 @@ private async Task<BaasApp[]> GetApps()
                 {
                     var name = doc["name"].AsString;
 
-                    if (!name.EndsWith(_appSuffix))
+                    if (!name.EndsWith(_appSuffix))  //Don't we check this in the Where?
```
nirinchev, line 539: We do - we can remove it.

.github/templates/build.lib.yml
```diff
@@ -43,17 +43,23 @@ deploy-cluster:
     AtlasPrivateKey: #@ secret_AtlasPrivateKey
 #@ end
 
-#@ def cleanupBaas(dependencies = []):
-#@ needs = []
+#@ def cleanupBaas(dependencies = [], targets = []):
```
nirinchev, line 46: Can we unify the naming here so that both dependencies and targets are named the same way?

nirinchev (commented):
Looks good, I have a few minor questions/suggestions where I was too lazy to check this for myself 😄

.github/workflows/deploy-baas.yml
papafe: Unfortunately yes. At the moment the outputs overwrite each other, so the last one that runs is the one that wins. 
They are working on it though: https://github.com/actions/runner/pull/2477

nirinchev: Merged: March, no timeframe for when it'll be actually available, lovely 🥲

papafe: Yeah, I wouldn't hold my breath 😁

.github/templates/build.lib.yml
papafe: The "problem" here is that MacOS has one test file (test-macos.yml), but two different tests (xamarin and catalyst), so there's not a 1-to-1 correspondence between dependencies and targets. 
The thing we could do is to create a function that gets the targets given the file dependency, but I'm not sure how much of an advantage we'll get. I agree that now it's kinda ugly

nirinchev: Yeah, we can keep it as is, maybe just add a note for why this is needed so that we don't forget.

Issue #3429, opened 2023-08-30
Expose user auth change notifications


nirinchev:
Should be similar to what we have in Kotlin: https://www.mongodb.com/docs/realm/sdk/kotlin/users/authenticate-users/#authentication-changes-as-a-flow

papafe:
After we do this, we should update https://github.com/realm/realm-dotnet-samples/tree/main/QuickJournalSync too (both code and docs preferably)

PR #3433, opened 2023-08-31
Add User.Changed event

nirinchev:
## Description

Addresses https://github.com/realm/realm-dotnet/issues/3429, but is not a real fix as it only notifies that the user changed, but doesn't tell us what exactly changed. We need https://github.com/realm/realm-core/issues/6516 to achieve more precise notifications.

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Tests/Realm.Tests/Sync/UserManagementTests.cs
+++ b/Tests/Realm.Tests/Sync/UserManagementTests.cs
@@ -20,6 +20,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
+using System.Threading.Tasks;
 using MongoDB.Bson;
 using MongoDB.Bson.Serialization.Attributes;
 using NUnit.Framework;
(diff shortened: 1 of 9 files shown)


fealebenpae (approved):

Tests/Realm.Tests/Sync/UserManagementTests.cs
```diff
@@ -983,6 +984,107 @@ public void UserToStringOverride()
...
+            SyncTestHelpers.RunBaasTestAsync(async () =>
+            {
+                var user = await GetUserAsync();
+
+                var tcs = new TaskCompletionSource();
+                user.Changed += (s, _) =>
+                {
+                    try
+                    {
+                        Assert.That(s, Is.EqualTo(user));
```
papafe, line 999: Shouldn't we check here that the user state is already correct?

Tests/Realm.Tests/Sync/UserManagementTests.cs
```diff
@@ -983,6 +984,107 @@ public void UserToStringOverride()
...
+
+                await collection.UpdateOneAsync(filter, update);
+
+                customUserData = await user.RefreshCustomDataAsync();
+                Assert.That(customUserData!["age"].AsInt32, Is.EqualTo(199));
+
+                await TestHelpers.AssertThrows<TimeoutException>(() => tcs.Task.Timeout(2000));
+            });
+
+            void OnUserChanged(object sender, EventArgs e)
```
papafe, line 1082:
```suggestion
            void OnUserChanged(object? sender, EventArgs e)
```

papafe (approved):

papafe (approved):
Looks good, I just left some minor comments. And obviously it needs the changelog

PR #3451, opened 2023-09-20
Allow customizing json ignore attribute for serialized classes

### LINKED ISSUE #2579: Ignore base properties when serializing RealmObjects
### Description

RealmObject inheritors should only serialize their own properties when using Json.NET/System.Text.Json. Currently, a bunch of base properties will be added to the json string without offering any value. While users can customize serialization, it's a bit of a hassle and we should do our best to have the SDK just work.

Fairly niche, but nice to have anyway

nirinchev:
## Description

Fixes #2579

##  TODO

* [x] Changelog entry
* [ ] Tests (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## vNext (TBD)
 
 ### Enhancements
-* Added support for customizing the ignore attribute applied on certain generated properties of Realm models. The configuration option is called `realm.custom_ignore_attribute` and can be set in a global configuration file (more information about global configuration files can be found in the [.NET documentation](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files)). The Realm generator will treat this as an opaque string, that will be appended to the `[IgnoreDataMember]` and `XmlIgnore` attributes already applied on these members. The attributes must be fully qualified unless the namespace they reside in is added to a global usings file. For example, this is how you would add `JsonIgnore` from `System.Text.Json`:
+* Added support for customizing the ignore attribute applied on certain generated properties of Realm models. The configuration option is called `realm.custom_ignore_attribute` and can be set in a global configuration file (more information about global configuration files can be found in the [.NET documentation](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files)). The Realm generator will treat this as an opaque string, that will be appended to the `IgnoreDataMember` and `XmlIgnore` attributes already applied on these members. The attributes must be fully qualified unless the namespace they reside in is added to a global usings file. For example, this is how you would add `JsonIgnore` from `System.Text.Json`:
 
   ```
   realm.custom_ignore_attribute = [System.Text.Json.Serialization.JsonIgnore]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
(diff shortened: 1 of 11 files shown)


Realm/Realm.SourceGenerator/ClassCodeBuilder.cs
```diff
@@ -43,16 +43,29 @@ internal class ClassCodeBuilder
...
         {
             _classInfo = classInfo;
 
+            _ignoreFieldAttribute = new(() =>
+            {
+                var result = "[IgnoreDataMember, XmlIgnore]";
+                var customAttribute = generatorConfig.CustomIgnoreAttribute;
+                if (!string.IsNullOrEmpty(customAttribute))
+                {
+                    result += customAttribute;
```
papafe, line 63: Maybe we should put a newline between the two?

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -93,27 +93,45 @@ public abstract class RealmCollectionBase<T>
             }
         }
 
-        [IgnoreDataMember]
+        [IgnoreDataMember, XmlIgnore]
+#if NET6_0
```
papafe, line 97: Shouldn't we use `NET6_0_OR_GREATER`?

CHANGELOG.md
```diff
@@ -1,7 +1,12 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+* Added support for customizing the ignore attribute applied on certain generated properties of Realm models. The configuration option is called `realm.custom_ignore_attribute` and can be set in a global configuration file (more information about global configuration files can be found in the [.NET documentation](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files)). The Realm generator will treat this as an opaque string, that will be appended to the `[IgnoreDataMember]` and `XmlIgnore` attributes already applied on these members. The attributes must be fully qualified unless the namespace they reside in is added to a global usings file. For example, this is how you would add `JsonIgnore` from `System.Text.Json`:
```
papafe, line 4:
```suggestion
* Added support for customizing the ignore attribute applied on certain generated properties of Realm models. The configuration option is called `realm.custom_ignore_attribute` and can be set in a global configuration file (more information about global configuration files can be found in the [.NET documentation](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files)). The Realm generator will treat this as an opaque string, that will be appended to the `IgnoreDataMember` and `XmlIgnore` attributes already applied on these members. The attributes must be fully qualified unless the namespace they reside in is added to a global usings file. For example, this is how you would add `JsonIgnore` from `System.Text.Json`:
```

papafe (approved):
Makes sense to me!

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
nirinchev: Oh, yeah, good point. Technically, we only target 6, but it's good to be safe in case we add more tfms on the Realm package.

Realm/Realm.SourceGenerator/ClassCodeBuilder.cs
nirinchev: I don't have strong opinion about it - both are valid C# and was thinking putting everything related on one line would make it easier to ignore if you're just reading the code.

papafe: Yeah, maybe it makes sense. I'm also not strongly opinionated on this.

PR #3460, opened 2023-10-13
Rework some tests

nirinchev:
## Description

This is some prep work for running the .NET tests on evergreen - some tests need to be reworked slightly to account for the different timings. We're also replacing `[Explicit]` with `[Ignore]` to avoid accidentally running the explicit tests when a filter is applied.

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -56,22 +56,22 @@ public class SessionTests : SyncTestBase
         [Preserve]
         static SessionTests()
         {
-            var preserveRecoverHandler = new RecoverUnsyncedChangesHandler
+            _ = new RecoverUnsyncedChangesHandler
```
papafe, line 59: What is this test doing?

papafe (approved):
Looks good! I just have a question

Tests/Realm.Tests/Sync/SessionTests.cs
nirinchev: Not sure - I think it may have been added because we create some things with reflection, but it's possible it's no longer needed. Can't check now, so will merge as is, but will keep it in mind to remove at some point.

PR #3468, opened 2023-11-03
Added RetryCustomConfirmationAsync

### LINKED ISSUE #3463: Expose `App.EmailPassword.retryCustomConfirmation`
Should behave like what we have for the web sdk: https://www.mongodb.com/docs/realm/web/manage-email-password-users/#retry-a-user-confirmation-function.

papafe:
Fixes #3463

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/Tests/Realm.Tests/Sync/SyncTestHelpers.cs
+++ b/Tests/Realm.Tests/Sync/SyncTestHelpers.cs
@@ -92,6 +92,8 @@ namespace Realms.Tests.Sync
 
         public static string GetVerifiedUsername() => $"realm_tests_do_autoverify-{Guid.NewGuid()}";
 
+        public static string GetUnconfirmedUsername() => $"realm_tests_do_not_confirm-{Guid.NewGuid()}@g.it";
+
         public static async Task TriggerClientResetOnServer(SyncConfigurationBase config)
         {
             var userId = config.User.Id;
--- a/Tests/Realm.Tests/Sync/UserManagementTests.cs
+++ b/Tests/Realm.Tests/Sync/UserManagementTests.cs
@@ -401,6 +401,42 @@ namespace Realms.Tests.Sync
         }
 
         [Test]
+        public void User_RetryCustomConfirmationAsync_WorksInAllScenarios()
+        {
+            SyncTestHelpers.RunBaasTestAsync(async () =>
+            {
+                // Standard case
+                var unconfirmedMail = SyncTestHelpers.GetUnconfirmedUsername();
+                var credentials = Credentials.EmailPassword(unconfirmedMail, SyncTestHelpers.DefaultPassword);
+
+                // The first time the confirmation function is called we return "pending", so the user needs to be confirmed.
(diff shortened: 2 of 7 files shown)


Tests/Realm.Tests/Sync/SyncTestHelpers.cs
```diff
@@ -92,6 +92,8 @@ public static void RunBaasTestAsync(Func<Task> testFunc, int timeout = 30000)
 
         public static string GetVerifiedUsername() => $"realm_tests_do_autoverify-{Guid.NewGuid()}";
 
+        public static string GetUnconfirmedUsername() => $"realm_tests_do_not_confirm-{Guid.NewGuid()}@g.it";
```
papafe, line 95: `RetryCustomConfirmation` will fail if the input doesn't look like a proper email. We could add the same to `GetVerifiedUsername` for consistency

Tools/DeployApps/BaasClient.cs
```diff
@@ -88,12 +88,34 @@ public class FunctionReturn
...
+                  if (username.includes(""realm_tests_do_not_confirm"")) {
+                    const mongodb = context.services.get('BackingDB');
+                    let collection = mongodb.db('test_db').collection('not_confirmed');
+                    let result = await collection.findOne({'email': username});
+
+                    if(result === null)
+                    {
+                        let newVal = {
+                            'email': username,
+                            'token': token,
```
papafe, line 106: Saving the token and tokenId is not necessary now, but it could be useful if we want to write tests for `ConfirmAsync`

Tests/Realm.Tests/Sync/UserManagementTests.cs
```diff
@@ -400,6 +400,42 @@ public void User_LinkCredentials_WhenInUse_Throws()
...
+            {
+                // Standard case
+                var unconfirmedMail = SyncTestHelpers.GetUnconfirmedUsername();
+                var credentials = Credentials.EmailPassword(unconfirmedMail, SyncTestHelpers.DefaultPassword);
+
+                // The first time the confirmation function is called we return "pending", so the user needs to be confirmed.
+                // At the same time we save the user email in a collection.
+                await DefaultApp.EmailPasswordAuth.RegisterUserAsync(unconfirmedMail, SyncTestHelpers.DefaultPassword).Timeout(10_000, detail: "Failed to register user");
+
+                var ex3 = await TestHelpers.AssertThrows<AppException>(() => DefaultApp.LogInAsync(credentials));
```
nirinchev, line 416: Not super important, but why is this `ex3` rather than just `ex`?

nirinchev (approved):
Looks good.

Tests/Realm.Tests/Sync/UserManagementTests.cs
papafe: Because before I had the sections in the opposite order and I forgot to rename the exceptions 😁

PR #3479, opened 2023-11-15
RNET-1083: Add support for progress estimate on progress notifications

### LINKED ISSUE #3478: Add support for progress notifications
Depends on https://github.com/realm/realm-core/pull/7124

papafe:
This PR:
- Adds `ProgressEstimate` to `SyncProgress`. It's a float value between 0.0 and 1.0 representing the estimate of the current progress.
- Deprecates `TransferableBytes` and `TransferredBytes` in `SyncProgress` as these values are not accurate (especially for flexible sync). 
- Enable using `GetProgressObservable` with flexible sync. 

Fixes #3478 

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


papafe:
This is just a draft PR in case we have some discussing to do. It's also built on top of a PR in core.

papafe:
I'll park this PR for now, as there is still some discussion about the kind of values that should be reported in the various cases.

papafe:
I'm going to park this for now, as the core implementation is not yet finished, and we can't write reliable tests without it.

Realm/Realm/Sync/ProgressNotifications/SyncProgress.cs
```diff
@@ -36,14 +39,24 @@ public struct SyncProgress
         /// successfully transferred.
         /// </summary>
         /// <value>The number of transferable bytes.</value>
+        [Obsolete("Not accurate, use ProgressEstimate instead.")]
         public ulong TransferableBytes { get; }
```
nirinchev, line 43: Since this is going into next-major, we don't need to obsolete them and can just remove the fields instead.

nirinchev (commented):
Looks good to me - I'd just prefer that we clean up the obsolete stuff in this PR already rather than merge it and have a follow-up PR that removes all the obsolete things as we prepare for 12.0

Realm/Realm/Sync/ProgressNotifications/SyncProgress.cs
papafe: Right!

Realm/Realm/Sync/ProgressNotifications/SyncProgress.cs
```diff
@@ -16,34 +16,26 @@
...
-        internal SyncProgress(ulong transferred, ulong transferable)
+        internal SyncProgress(double progressEstimate)
         {
-            TransferredBytes = transferred;
-            TransferableBytes = transferable;
+            ProgressEstimate = progressEstimate;
         }
 
-        internal bool IsComplete => TransferableBytes == TransferredBytes;
+        internal readonly bool IsComplete => ProgressEstimate >= 1.0;
```
nirinchev, line 39: This readonly seems redundant here since the struct is already readonly.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -750,18 +756,33 @@ public void Session_OnSessionError()
             });
         }
 
-        [TestCase(ProgressMode.ForCurrentlyOutstandingWork)]
-        [TestCase(ProgressMode.ReportIndefinitely)]
-        public void SessionIntegrationTest_ProgressObservable(ProgressMode mode)
+#pragma warning disable CS0618 // Type or member is obsolete
```
nirinchev, line 759: This pragma shouldn't be necessary anymore

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -782,29 +803,24 @@ public void SessionIntegrationTest_ProgressObservable(ProgressMode mode)
...
                         completionTcs.TrySetException(e);
                     }
 
-                    if (p.TransferredBytes >= p.TransferableBytes)
+                    if (p.IsComplete)
                     {
-                        completionTcs.TrySetResult(p.TransferredBytes);
+                        if (p.ProgressEstimate != 1.0)
+                        {
+                            throw new Exception($"Expected progress estimate to be complete if and only if ProgressEstimate == 1.0");
```
nirinchev, line 820: This is outside the try-catch, so if we get here, we won't set the completionTcs exception and instead will throw it into the void.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -750,18 +756,33 @@ public void Session_OnSessionError()
             });
         }
 
-        [TestCase(ProgressMode.ForCurrentlyOutstandingWork)]
-        [TestCase(ProgressMode.ReportIndefinitely)]
-        public void SessionIntegrationTest_ProgressObservable(ProgressMode mode)
+#pragma warning disable CS0618 // Type or member is obsolete
+        [Test]
+        public void SessionIntegrationTest_ProgressObservable(
```
nirinchev, line 761: Previously this test was only testing for callbacks being invoked because the notifications were unreliable. We should improve the test to now validate that meaningful progress is being reported and not that it was just invoked.

nirinchev (commented):
Looks good, though it'd be nice to add a test that is a bit more strict about the notifications. Particularly, we should have a test for the download direction, which we don't seem to at the moment.

papafe:
This is going to get parked until the core PR on which it depends it has been merged. 
What we need to do then:
- Add tests
- Remove bytes from progress notification

papafe:
@nirinchev this is mergeable for me (I can fix the lint warning after your review 😁 )

Realm/Realm/Sync/ProgressNotifications/ProgressMode.cs
```diff
@@ -25,15 +25,12 @@ public enum ProgressMode
...
-        /// Notifications will always report the latest number of transferred bytes, and the most up-to-date number of
-        /// total transferable bytes.
         /// </summary>
         ReportIndefinitely,
 
         /// <summary>
-        /// The callback will, upon registration, store the total number of bytes to be transferred. When invoked, it will
-        /// always report the most up-to-date number of transferable bytes out of that original number of transferable bytes.
-        /// When the number of transferred bytes reaches or exceeds the number of transferable bytes, the callback will
+        /// The callback will be active until the current transferable bytes are transferred. When invoked, it will always report the
```
kiburtse, line 32: it's a bit weird now to refer to transferred/transferable after these fields were removed. Also, it makes no sense any more for download direction. Probably, a good idea is to rephrase it in a way that it refers to snapshot committed version for upload on callback registration (that is essentially transferable bytes), and for download it is whatever server decides for the outgoing or next batch of changesets to catch up the client to the latest version.

papafe: @kiburtse I agree with you, it's kinda imprecise. 
Would you think the following will make sense?

```
The callback will be active:
- For uploads, until all the unsynced changesets available at the moment of the registration of the callback are sent to the server.
- For downloads, until the client catches up to the current data (available at the moment of registration) or the next batch of changesets sent from the server.
```

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -405,10 +405,12 @@ private static IntPtr NotifyAfterClientReset(IntPtr beforeFrozen, IntPtr after,
         }
 
         [MonoPInvokeCallback(typeof(NativeMethods.SessionProgressCallback))]
-        private static void HandleSessionProgress(IntPtr tokenPtr, ulong transferredBytes, ulong transferableBytes)
+        private static void HandleSessionProgress(IntPtr tokenPtr, ulong transferredBytes, ulong transferableBytes, double progressEstimate)
         {
             var token = (ProgressNotificationToken?)GCHandle.FromIntPtr(tokenPtr).Target;
-            token?.Notify(transferredBytes, transferableBytes);
+            // This is used to provide a reasonable progress estimate until the core work is done
```
nirinchev, line 411: Single-line comment should be preceded by blank line ([`https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1515.md`](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1515.md))

😛

Realm/Realm/Handles/SessionHandle.cs
```diff
@@ -405,10 +405,12 @@ private static IntPtr NotifyAfterClientReset(IntPtr beforeFrozen, IntPtr after,
...
+            double managedProgressEstimate = transferableBytes > 0.0 ? transferredBytes / transferableBytes : 0.0;
```
nirinchev, line 412: If transferable is 0, shouldn't progress be 1?

nirinchev (approved):
Looks good - there's a few lint warnings you could drive-by fix.

Realm/Realm/Sync/ProgressNotifications/ProgressMode.cs
papafe: @nirinchev I'll fix the other two small things. I suppose this is the latest thing to agree on. Do you have input on this?

PR #3487, opened 2023-12-08
Static queries build on

papafe:
This is here just to make easier to understand the changes that have been made compared to the base branch. 

TODO

- [x] Add tests for embedded objects
- [x] Add tests for embedded objects in collections
- [x] Add tests for objects in `RealmValue`
- [x] Add tests for collections of `RealmValue` (with objects too)
- [x] Add tests with extra fields
- [x] Add tests with missing fields
- [x] Add tests with mismatched type fields
- [x] Clean up
- [x] Add documentation
- [x] Changelog


papafe:
Hey @nirinchev, I've put up this PR to make it easier to see the changes I've made on top of your branch. This isn't complete, but after today I'll be working again on Wednesday, so if you want you can give a look in the meantime. 
There's some stuff missing, but I'm planning to keep the TODO list here in the PR descriptions updated, and let me know if I'm forgetting something

papafe:
Most of the main functionalities are being tested. There are some tests that are not passing now due to some bugs that have been reported. 

Only some smaller things are necessary for this to be merged in my opinion, and I have them in the TODO checklist at the top.

Realm/Realm/Helpers/RealmObjectSerializer.cs
```diff
@@ -28,49 +28,68 @@
 namespace Realms.Serialization;
 
 #pragma warning disable SA1600 // Elements should be documented
-
 [EditorBrowsable(EditorBrowsableState.Never)]
-public abstract class RealmObjectSerializer : IBsonSerializer
+public static class RealmObjectSerializer
```
papafe, line 32: I thought it was a little confusing that `RealmObjectSerializer` before was both a base class for serializers and contain a collection of serializers, so I've divided these two responsibilities in two different classes.

Realm/Realm/Sync/MongoClient.cs
```diff
@@ -162,10 +170,17 @@ public class Collection<TDocument>
 
             internal Collection(Database database, string name)
             {
+                _client = database.Client;
                 Database = database;
                 Name = name;
             }
 
+            internal Collection(MongoClient client, string name)
```
papafe, line 178: I kept the same collection type for both `client.GetCollection` and `client.Database.GetCollection`, as the property names (and descriptions) seems to be compatible enough.

Tests/Realm.Tests/Database/TestObjects.cs
```diff
@@ -981,14 +981,15 @@ public partial class ObjectWithRequiredStringList : TestRealmObject
     public partial class ObjectWithEmbeddedProperties : TestRealmObject
     {
         [PrimaryKey]
-        public int PrimaryKey { get; set; }
+        [MapTo("_id")]
```
papafe, line 984: I've slightly modified some classes for them to be compatible with sync, in order to not create new classes just for that.

Realm/Realm/Sync/MongoClient.cs
```diff
@@ -44,12 +48,39 @@ public class MongoClient
...
+        {
+            if (!SchemaByName.TryGetValue(typeof(TRealmObject), out var schema))
+            {
+                throw new NotSupportedException("This method is only supported for source-generated classes - i.e. ones that inherit from IRealmObject rather than RealmObject.");
+            }
+
+            return new Collection<TRealmObject>(this, schema.Name);
+        }
+
+        //TODO This has been done to avoid the reflection in the GetCollection method. Need to look at this again
```
papafe, line 78: I'm not very convinced by this. I think we talked about this at some point but I don't remember, could we use something similar to create the default schema for realm?

.github/templates/build-packages.yml
```diff
@@ -87,8 +87,7 @@ jobs:
     steps:
       - #@ template.replace(checkoutCode())
       - #@ template.replace(setupAndroid())
-      - _: #@ template.replace(setupDotnet("6.0.417"))
-      - #@ setupWorkloads('tvos ios maccatalyst android')
+      - #@ template.replace(setupWorkloads('tvos ios maccatalyst android'))
```
nirinchev, line 90: Are those changes part of this PR or something that got picked up from main? If the latter, can we also update the base PR so that the only changes left are the ones that need to be reviewed.

papafe: Yes, this is coming from main, I'll merge that into the base PR

papafe: Just done

nirinchev (approved):
Looks reasonable, I have a few minor comments.

Realm/Realm/Sync/MongoClient.cs
```diff
@@ -44,12 +47,41 @@ public class MongoClient
...
+        public static void RegisterSchema(Type type, ObjectSchema schema)
```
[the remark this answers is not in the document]
papafe: Yes, I'm not convinced by this either, there's another comment about it. One difference is that here we get only the source generated classes, and ignore the woven ones that don't have the serializer. 
Apart from this difference, in the other comment I was wondering if we could we use something similar to create the default schema for realm and avoid the woven initializer. I think we discussed about this at some point but I don't remember what we said about it.

Realm/Realm/Helpers/SerializationHelper.cs
```diff
@@ -226,7 +246,7 @@ public RealmValue Deserialize(BsonDeserializationContext context, BsonDeserializ
 
                         return value;
                     case BsonType.Document:
-                        throw new NotImplementedException("TODO");
+                        return DeseriealizeDBRef(context, args);
```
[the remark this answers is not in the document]
papafe: Yes, let's go step by step 👀

Realm/Realm/Sync/MongoClient.cs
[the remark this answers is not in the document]
nirinchev: One drawback of doing it in the static class constructor is that this is only invoked when the class is referenced. It is fine here since passing the type as a generic argument will invoke the static ctor, but it wouldn't be fine for the Realm schema, since where we open the Realm, we rarely enumerate all the types that can go in there.

We could generate a module initializer, but that'll only work for C# 9 projects, whereas the woven initializer works for all projects, since it's a .net feature that has existed for a very long time.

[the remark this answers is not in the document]
papafe: Makes sense.

Regarding our current issue here. We could use `RealmSchema.Default`, but:
- As this contains all types, we'd need to manually check if they are source generated (checking if they implement `IRealmObjectBase`)
- We need to add methods to the `RealmSchema` to get `ObjectSchema` given a type. Not a biggie, but we don't have such method at the moment
- It won't work for classes with `Explicit` attribute. I suppose it's debatable if this is an issue or not. 

What do you think?

papafe:
I fixed some comments, and I've moved the tests to a new fixture, to improve readability.

Realm/Realm/Sync/MongoClient.cs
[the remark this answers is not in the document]
nirinchev: Hm... a bit annoying with the `[Explicit]` - I am split between keeping the current approach and just doing the reflection - considering it's an http call, I am not particularly worried about the cost of reflection.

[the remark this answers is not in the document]
papafe: I think you're right, let's keep it with reflection for now, it's the cleanest solution.

Realm/Realm/Sync/MongoClient.cs
```diff
@@ -44,12 +48,34 @@ public class MongoClient
         /// <value>The name of the remote MongoDB service.</value>
         public string ServiceName { get; }
 
+        private static readonly ConcurrentDictionary<Type, ObjectSchema> SchemaByType = new();
```
nirinchev, line 51: This isn't used.

nirinchev (approved):
There a few build warnings that need to be fixed, but otherwise looks good. Also, not sure if you want to take a look at the tests as part of this PR or once you merge it, but there are quite a few failures which seem legit.

Tests/Realm.Tests/Database/TestObjects.cs
```diff
@@ -1141,6 +1143,10 @@ public partial class LinksObject : TestRealmObject
 
         public IDictionary<string, LinksObject?> Dictionary { get; } = null!;
 
+        public LinksObject()
```
nirinchev, line 1146: Why do we need this ctor?

Tests/Realm.Tests/Sync/StaticQueriesTests.cs
```diff
@@ -0,0 +1,1257 @@
...
+                    {
+                        ["dict_1"] = new IntPropertyObject { Int = 7 },
+                        ["dict_2"] = new IntPropertyObject { Int = 8 },
+                        ["dict_null"] = RealmValue.Null
+                    }
+                }),
+            }
+        };
+
+        // TODO This is going to be fixed with https://jira.mongodb.org/browse/BAAS-27410
```
nirinchev, line 633: That ticket seems to be resolved, but perhaps it's not deployed to QA yet?

Tests/Realm.Tests/Database/TestObjects.cs
papafe: It is necessary when this is a woven class. I'll put preprocessor directives around it to make it clearer.

Tests/Realm.Tests/Sync/StaticQueriesTests.cs
papafe: Exactly.

Tests/Realm.Tests/Database/TestObjects.cs
nirinchev: Ugh... might need to suppress the id warning then.

papafe: Yes, I'll fix that too

PR #3499, opened 2024-01-18
Bump System.Net.Security to 4.3.2

nirinchev:
## Description

4.3.0 has been marked as vulnerable. While users can manually upgrade to 4.3.1, it's easy to miss, so it's better to just pull in the latest non-vulnerable version. This is a transitive dependency we take on through System.Net.WebSockets.Client, but unfortunately the latest version of that package is 4.3.2 and it still depends on Security 4.3.0.


papafe (approved):

papafe:
Looks good, but it seems we need to fix the unity build

papafe:
Should we merge this so can make a release?

PR #3501, opened 2024-01-19
Add keypath filtering to notifications

papafe:
This PR adds support for keypath filtering in notifications

## Description

Fixes ##1398

##  TODO

* [x] Changelog entry
* [x] Correct documentation
* [x] Tests for lists/sets/dictionaries notifications
* [x] Cleaning up internals
* [x] Try to unify subscriptions path (keypath/full/shallow)


papafe:
This works, and most of the tests are there and passing. It requires still some love though, so this is still a draft. I'll keep the TODO list updated in the PR description.

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
```diff
@@ -0,0 +1,114 @@
...
+public class KeyPathsCollection : IEnumerable<KeyPath>
+{
+    private IEnumerable<KeyPath> _collection;
+
+    private static readonly KeyPathsCollection _shallow = new KeyPathsCollection(KeyPathsCollectionType.Shallow);
+    private static readonly KeyPathsCollection _default = new KeyPathsCollection(KeyPathsCollectionType.Default);
+
+    internal KeyPathsCollectionType Type { get; set; }
+
+    private KeyPathsCollection(KeyPathsCollectionType type, IEnumerable<KeyPath>? collection = null)
```
nirinchev, line 37: I think taking an enumerable here is not ideal - it's not wrong, but makes the API semantics a bit awkward - i.e. we enumerate it multiple times and if it was a generator-based enumerable or something, we'd get weird results. Perhaps using `ICollection` is a better choice here and both array and list conform to the interface.

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /// <inheritdoc/>
+    IEnumerator IEnumerable.GetEnumerator()
+    {
+        return GetEnumerator();
+    }
+}
+
+public class KeyPath
```
nirinchev, line 97: Should we override `ToString`, `Equals`, etc. here? Also, should we make it a struct rather than a class?

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
```diff
@@ -0,0 +1,114 @@
...
+{
+    internal string Path { get; private set; }
```
nirinchev, line 99: This doesn't need a setter, does it?

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    private KeyPath(string path)
+    {
+        Path = path;
+    }
+
+    public static implicit operator KeyPath(string s) => new(s);
```
nirinchev, line 106: Should we add an implicit string(KeyPath) operator?

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -189,15 +190,41 @@ public IRealmCollection<T> Freeze()
...
 
-        public IDisposable SubscribeForNotifications(NotificationCallbackDelegate<T> callback)
-            => SubscribeForNotificationsImpl(callback, shallow: false);
+        public IDisposable SubscribeForNotifications(NotificationCallbackDelegate<T> callback, KeyPathsCollection? keyPathCollection = null)
+        {
+            keyPathCollection ??= KeyPathsCollection.Default;
+
+            if (keyPathCollection.Type == KeyPathsCollectionType.Full && !ContainsRealmObjects())
+            {
+                    throw new InvalidOperationException("Key paths can be used only with collections of Realm objects");
```
nirinchev, line 199:
```suggestion
                throw new InvalidOperationException("Key paths can be used only with collections of Realm objects");
```

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -189,15 +190,41 @@ public IRealmCollection<T> Freeze()
             return CreateCollection(frozenRealm, frozenHandle);
         }
 
-        public IDisposable SubscribeForNotifications(NotificationCallbackDelegate<T> callback)
-            => SubscribeForNotificationsImpl(callback, shallow: false);
+        public IDisposable SubscribeForNotifications(NotificationCallbackDelegate<T> callback, KeyPathsCollection? keyPathCollection = null)
```
nirinchev, line 193:
```suggestion
        public IDisposable SubscribeForNotifications(NotificationCallbackDelegate<T> callback, KeyPathsCollection? keyPathsCollection = null)
```

No strong opinion about this, just bugs me that it's different between methods.

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -396,18 +423,31 @@ private void UpdateCollectionChangedSubscriptionIfNecessary(bool isSubscribed)
...
+            {
+                callback(this, GetFromNative(changes));
+                return;
+            }
+
+            var changeset = GetFromNative(changes);
+            _notificationCallbacks.Value.Notify(changeset, type);
+        }
+
+        private ChangeSet? GetFromNative(CollectionChangeSet? changes)
```
nirinchev, line 450: I don't think this needs a separate method - we still need the changeset only in `NotifyCallbacks` and regardless of what the type is.

Realm/Realm/Handles/DictionaryHandle.cs
```diff
@@ -136,12 +138,17 @@ public override void Clear()
             nativeException.ThrowIfNecessary();
         }
 
-        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, bool shallow)
+        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle,
+            KeyPathsCollection keyPathsCollection, IntPtr callback = default)
```
nirinchev, line 142: Do we need this default?

Realm/Realm/Handles/ListHandle.cs
```diff
@@ -213,11 +215,17 @@ public void Move(IntPtr sourceIndex, IntPtr targetIndex)
...
         }
 
-        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, bool shallow)
+        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle,
+            KeyPathsCollection keyPathsCollection, IntPtr callback = default)
         {
             EnsureIsOpen();
 
-            var result = NativeMethods.add_notification_callback(this, managedObjectHandle, shallow, out var nativeException);
+            using Arena arena = new Arena();
```
nirinchev, line 223:
```suggestion
            using Arena arena = new();
```

Realm/Realm/Handles/DictionaryHandle.cs
```diff
@@ -136,12 +138,17 @@ public override void Clear()
...
         {
             EnsureIsOpen();
 
-            var result = NativeMethods.add_notification_callback(this, managedObjectHandle, shallow, out var nativeException);
-            nativeException.ThrowIfNecessary();
+            using Arena arena = new Arena();
```
nirinchev, line 146:
```suggestion
            using Arena arena = new();
```

Realm/Realm/Handles/ResultsHandle.cs
```diff
@@ -155,12 +160,20 @@ public SortDescriptorHandle GetSortDescriptor()
...
         }
 
-        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, bool shallow)
+        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle,
+            KeyPathsCollection keyPathsCollection, IntPtr callback = default)
         {
             EnsureIsOpen();
 
-            var result = NativeMethods.add_notification_callback(this, managedObjectHandle, shallow, out var nativeException);
+            using Arena arena = new Arena();
```
nirinchev, line 168:
```suggestion
            using Arena arena = new();
```

Realm/Realm/Handles/SetHandle.cs
```diff
@@ -136,11 +138,17 @@ public override void Clear()
...
         }
 
-        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle, bool shallow)
+        public override NotificationTokenHandle AddNotificationCallback(IntPtr managedObjectHandle,
+            KeyPathsCollection keyPathsCollection, IntPtr callback = default)
         {
             EnsureIsOpen();
 
-            var result = NativeMethods.add_notification_callback(this, managedObjectHandle, shallow, out var nativeException);
+            using Arena arena = new Arena();
```
nirinchev, line 146:
```suggestion
            using Arena arena = new();
```

wrappers/src/notifications_cs.hpp
```diff
@@ -105,9 +106,13 @@ static inline void handle_changes(ManagedNotificationTokenContext* context, Coll
...
-            if (!pair.second.empty()) {
-                properties.emplace_back(get_property_index(context->schema, ColKey(pair.first)));
+        //Property names are necessary only for shallow notifications (PropertyChanged/CollectionChanged)
+        if (type == key_path_collection_type::SHALLOW)
+        {
+            for (auto& pair : changes.columns) {
+                if (!pair.second.empty()) {
+                    properties.emplace_back(get_property_index(context->schema, ColKey(pair.first)));
+                }
             }
```
nirinchev, line 116: Not sure I understand why

wrappers/src/notifications_cs.hpp
```diff
@@ -121,23 +126,59 @@ static inline void handle_changes(ManagedNotificationTokenContext* context, Coll
...
+        handle_changes(context, changes, type, callback);
     });
 
     return context;
 }
+
+static inline std::optional<KeyPathArray> build_keypath_array_impl(const SharedRealm& realm, StringData class_name, key_path_collection_type type, MarshaledVector<realm_string_t> keypaths) {
+    std::optional<KeyPathArray> keypath_array;
+
+    if (type == key_path_collection_type::FULL) {
```
nirinchev, line 150: Can this be a switch

nirinchev (commented):
Haven't looked at the tests yet, but figured I'd add my comments so far and review the rest another time.

Realm/Realm/DatabaseTypes/INotifiable.cs
```diff
@@ -33,8 +33,9 @@ internal interface INotifiable<TChangeset>
         /// Method called when there are changes to report for that object.
         /// </summary>
         /// <param name="changes">The changes that occurred.</param>
-        /// <param name="shallow">Whether the changes are coming from a shallow notifier or not.</param>
-        void NotifyCallbacks(TChangeset? changes, bool shallow);
+        /// <param name="type">The type of the key paths collection related to the notification.</param>
+        /// <param name="callbackNative">The eventual callback to call for the notification (if type == Full) </param>
+        void NotifyCallbacks(TChangeset? changes, KeyPathsCollectionType type, IntPtr callbackNative = default);
```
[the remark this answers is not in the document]
papafe: We use the same callback for dictionary key notifications, so either we leave it like this or we pass any value when called. 
Regarding getting nicer type, I agree. The issue here is that the actual type of the callback is ` NotificationCallbackDelegate<T> `, so we don't know T yet. In this case I'd need to add a new generic parameter to `INotifiable` for that

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
```diff
@@ -0,0 +1,114 @@
...
+
+//TODO Add docs for this and the following types
+public class KeyPathsCollection : IEnumerable<KeyPath>
+{
+    private IEnumerable<KeyPath> _collection;
+
+    private static readonly KeyPathsCollection _shallow = new KeyPathsCollection(KeyPathsCollectionType.Shallow);
+    private static readonly KeyPathsCollection _default = new KeyPathsCollection(KeyPathsCollectionType.Default);
+
+    internal KeyPathsCollectionType Type { get; set; }
```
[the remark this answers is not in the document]
papafe: Nope

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
papafe: It doesn't

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public static implicit operator KeyPath(string s) => new(s);
+}
+
+internal enum KeyPathsCollectionType
+{
+    Default,
+    Shallow,
+    Full
```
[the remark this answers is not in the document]
papafe: Yeah, the naming is up for debate 😁 
I agree with changing `Full` to `Explicit`, but I'd keep `Default`. Default gives up to 4 levels depth, but it's possible to go deeper by using wildcards, so it's not really full maybe?

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
papafe: I didn't add it because I didn't think it would be as useful as the opposite

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
papafe: Yes to all

wrappers/src/notifications_cs.hpp
papafe: The comment is not totally correct, it should only be for PropertyChanged. At the moment, we need the property names only to raise `PropertyChanged`, while it's not used for `CollectionChanged` or `SubscribeForNotifcations`

wrappers/src/notifications_cs.hpp
papafe: Yes

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
papafe: Reasonable

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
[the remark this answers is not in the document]
nirinchev: Fair enough, I guess I'm a bit allergic to using names like "default" because they don't really give you much information unless you know what the default is. And yes, `Full` is not really "full", but `Wildcard4LevelsDeep` is a bit of a mouthful 😆 I'm fine with a different name, but I think in a lot of places where the type is used, it'd be useful to know that `Default` actually means deep change notifications.

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
nirinchev: I agree - it's not clear to me that it's very useful either, just figured it's kind of weird that you construct an object but then can't get the info out of it. Again, I don't see any practical application to it, just satisfying my OCD 😅

wrappers/src/notifications_cs.hpp
nirinchev: Gotcha. Still, I think it's a bit awkward that we're using a seemingly unrelated piece of information to decide to extract the property change info. Yes, property change notifications use shallow, but that's incidental and prone to change.

Realm/Realm/DatabaseTypes/KeyPathCollection.cs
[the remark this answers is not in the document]
papafe: I agree on this. Well, we could either go with "Full" or "Deep" (in contrast with "Shallow"). I don't have a strong opinion on those either 😁

wrappers/src/notifications_cs.hpp
papafe: I agree with this, I'll remove the if. When we get to object notifications, we can think about how to make the properties calculations only for objects and not for collections.

Realm/Realm/Handles/DictionaryHandle.cs
papafe: Nope

Realm/Realm/DatabaseTypes/INotifiable.cs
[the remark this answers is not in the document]
papafe: I've been giving a thought about the nicer type and didn't manage to find a reasonable approach to do it. Do you have any suggestions regarding this?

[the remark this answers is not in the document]
nirinchev: I pushed https://github.com/realm/realm-dotnet/pull/3512 as a suggestion - it's not the greatest in the world, but I feel it's slightly better than passing raw pointers.

CHANGELOG.md
```diff
@@ -1,7 +1,27 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+* Add support for passing a key paths collection (`KeyPathsCollection`) when using `IRealmCollection.SubscribeForNotifications`. Passing a `KeyPathsCollection` allows to specify which changes in properties should raise a notification.
+
+  A `KeyPathsCollection` can be obtained by:
+  - building it explicitly by using the method `KeyPathsCollection`;
```
papafe, line 7:
```suggestion
  - building it explicitly by using the method `KeyPathsCollection.Of`;
```

Realm/Realm/DatabaseTypes/RealmCollectionBase.cs
```diff
@@ -189,15 +190,41 @@ public IRealmCollection<T> Freeze()
...
+                var managedResultsHandle = GCHandle.Alloc(this, GCHandleType.Weak);
+                var callbackHandle = GCHandle.Alloc(callback, GCHandleType.Weak);
+
+                var token = Handle.Value.AddNotificationCallback(GCHandle.ToIntPtr(managedResultsHandle), keyPathsCollection,
+                    GCHandle.ToIntPtr(callbackHandle));
+
+                return NotificationToken.Create(callback, _ => token.Dispose());
+            }
+
+            // For notifications with type Default or Shallow we cache the callbacks on the managed level, to avoid creating multiple notifications in core
```
papafe, line 220:
```suggestion
            // For notifications with type Full or Shallow we cache the callbacks on the managed level, to avoid creating multiple notifications in core
```

papafe:
@nirinchev Fixed the final things and added the docs, this should be ready to merge. 
I tried not to repeat myself in the docs, hopefully it makes sense overall

nirinchev (approved):

PR #3502, opened 2024-01-23
Updated core to 13.26.0

### LINKED ISSUE #3482: RealmException: posix_fallocate() failed
### What happened?

Hello!

I've been using Realm (.Net) 11.5.0 in Unity for months now with no issues (Windows Editor & Android Builds)

After upgrading to 11.6.1 it's still working in the Editor, but the attached exception is thrown when retrieving a Realm instance on Android.

The lock file is created, but the Realm (and other associated files) are not. This occurs whether opening an existing Realm or creating a new.

Unity Version: 2022.3.13f1
Android Version: 10 (ARM64 \ IL2CPP)

### Repro steps

See attached code.

### Version

11.6.1

### What Atlas Services are you using?

Local Database only

Unity

### Client OS and version

Android 10 (ARM64 \ IL2CPP)

### Code snippets

RealmConfiguration config = new RealmConfiguration([file]);
config.EncryptionKey = [key];
config.Schema = new Type[] { typeof([type]) };
config.FallbackPipePath = [path];

Realm realm = Realm.GetInstance(config);

```shell
RealmException: posix_fallocate() failed: Operation not supported on transport endpoint: Operation not supported on transport endpoint (95)
Realms.NativeException.ThrowIfNecessary ()
Realms.SharedRealmHandle.Open (Realms.Native.Configuration configuration)
Realms.RealmConfiguration.CreateHandle (Realms.Native.Configuration& configuration)
Realms.RealmConfigurationBase.CreateRealm ()
```

### Relevant log output

_No response_

papafe:
## Description

Fixes #3482

### DIFF
--- a/Tests/Realm.Tests/Sync/SyncTestBase.cs
+++ b/Tests/Realm.Tests/Sync/SyncTestBase.cs
@@ -138,6 +138,13 @@ namespace Realms.Tests.Sync
             return new User(handle, app);
         }
 
+        // This could be useful when opening a "fake" sync realm locally
+        protected void SetFakeSyncRoute(App? app)
+        {
+            app ??= DefaultApp;
+            app.Handle.SetFakeSyncRouteForTesting();
+        }
+
         protected async Task<Realm> GetIntegrationRealmAsync(string? partition = null, App? app = null, int timeout = 10000)
         {
             var config = await GetIntegrationConfigAsync(partition, app);
@@ -256,15 +263,29 @@ namespace Realms.Tests.Sync
             return config;
         }
 
-        protected PartitionSyncConfiguration GetFakeConfig(App? app = null, string? userId = null, string? optionalPath = null)
+        protected PartitionSyncConfiguration GetFakeConfig(App? app = null, string? userId = null,
+            string? optionalPath = null, bool setFakeSyncRoute = false)
         {
             var user = GetFakeUser(app, userId);
+
+            if (setFakeSyncRoute)
+            {
+                SetFakeSyncRoute(app);
+            }
+
             return UpdateConfig(new PartitionSyncConfiguration(Guid.NewGuid().ToString(), user, optionalPath));
         }
 
-        protected FlexibleSyncConfiguration GetFakeFLXConfig(App? app = null, string? userId = null, string? optionalPath = null)
+        protected FlexibleSyncConfiguration GetFakeFLXConfig(App? app = null, string? userId = null, 
+            string? optionalPath = null, bool setFakeSyncRoute = false)
         {
             var user = GetFakeUser(app, userId);
+
+            if (setFakeSyncRoute)
+            {
+                SetFakeSyncRoute(app);
+            }
+
             return UpdateConfig(new FlexibleSyncConfiguration(user, optionalPath));
         }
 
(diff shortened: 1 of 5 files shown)


papafe:
@nirinchev this should fix both errors that came out from CI
- The core method `get_base_url` now returns a copy of the string, and not a reference as before
- There's now an additional check done by core, that was crashing our app because the `sync_route` (the route to the server) was null. Added a method to set a fake route, that seems to be enough for our tests in which we create a fake local sync server. I've decided to keep it a separate method from the "GetSyncUserForTesting", but it's up for debate

Tests/Realm.Tests/Sync/SyncTestBase.cs
```diff
@@ -138,6 +138,13 @@ protected User GetFakeUser(App? app = null, string? id = null, string? refreshTo
             return new User(handle, app);
         }
 
+        // This could be useful when opening a "fake" sync realm locally
+        protected void SetFakeSyncRoute(App? app)
+        {
+            app ??= DefaultApp;
+            app.Handle.SetFakeSyncRouteForTesting();
```
nirinchev, line 145: I'm not sure how I feel about this - I know it's just for tests, but if this is called with one of our integration apps, it would sets its sync route to a fake value, even though a real value should exist. I don't know if sync will refresh that route at some point, but I worry that as it stands, it's possible that running this test first, then running some integration tests that use the app, those will fail due to the route being misconfigured.

Tests/Realm.Tests/Sync/SyncTestBase.cs
```diff
@@ -256,15 +263,29 @@ private static T UpdateConfig<T>(T config)
...
+            if (setFakeSyncRoute)
+            {
+                SetFakeSyncRoute(app);
+            }
+
             return UpdateConfig(new PartitionSyncConfiguration(Guid.NewGuid().ToString(), user, optionalPath));
         }
 
-        protected FlexibleSyncConfiguration GetFakeFLXConfig(App? app = null, string? userId = null, string? optionalPath = null)
+        protected FlexibleSyncConfiguration GetFakeFLXConfig(App? app = null, string? userId = null,
```
nirinchev, line 279:
```suggestion
        protected FlexibleSyncConfiguration GetFakeFLXConfig(App? app = null, string? userId = null,
```

Tests/Realm.Tests/Sync/SyncTestBase.cs
papafe: Mmmh, I understand. Good news is that all the tests are passing now, including sync tests, so I suppose it's not a problem "yet".
If we want to be more robust for the tests we could:
- Set the fake value only if the sync route is currently empty/null
- Set it to null/empty again at the end of the test (if modified)
- Maybe create infrastructure in core to create those fake synced realm more "organically"?
I am not sure if it's worth to do this though.

papafe: Given that this is the only holdup and it's only for tests, what do you think about merging this and creating a follow up ticket so we come back to it when the core investigation is finished?

nirinchev: Sure

PR #3523, opened 2024-02-15
Add support for running tests on BaaSaas locally

papafe:
This PR adds support for running tests locally using BaaSaas. This is a step towards running the CI tests on BaaSaas too


papafe:
It requires some cleaning, but most of the work is done.

Tests/Realm.Tests/Sync/SyncTestHelpers.cs
```diff
@@ -41,21 +41,18 @@ public static class SyncTestHelpers
...
-            try
-            {
-                var uri = ConfigHelpers.GetSetting("BaasUrl");
-                if (uri != null)
-                {
-                    BaasUri = new Uri(uri);
-                }
-            }
-            catch
+            var uri = ConfigHelpers.GetSetting("BaasUrl");
```
papafe, line 49: I'm not sure why we had a try catch around it

Tests/Realm.Tests/Sync/SyncTestHelpers.cs
```diff
@@ -167,28 +164,36 @@ public static (string[] RemainingArgs, IDisposable? Logger) SetLoggerFromArgs(st
...
 
 #if !UNITY
-            try
-            {
-                var cluster = ConfigHelpers.GetSetting("Cluster")!;
-                var apiKey = ConfigHelpers.GetSetting("ApiKey")!;
-                var privateApiKey = ConfigHelpers.GetSetting("PrivateApiKey")!;
-                var groupId = ConfigHelpers.GetSetting("GroupId")!;
-                var differentiator = ConfigHelpers.GetSetting("Differentiator") ?? "local";
+            var cluster = ConfigHelpers.GetSetting("Cluster")!;
```
papafe, line 173: I'm a little bit conflicted if it makes sense to add a "Mode" to the config file, instead of relying on the fields in the config file to implicitly deduce the kind of baas client we need.

Tools/DeployApps/BaasClient.cs
```diff
@@ -863,6 +881,154 @@ private static class Schemas
...
+                        return container;
+                    }
+
+                    await Task.Delay(3000);
+                }
+
+                throw new Exception($"Container with id={containerId} was not found or ready after {maxRetries} retrues");
+            }
+
+            private async Task<T?> CallEndpointAsync<T>(HttpMethod method, string relativePath, object? payload = null)
```
papafe, line 980: This is almost the same as `BaasClient.SendAsync`, without the refresh token capabilities. We could remove this

Tools/DeployApps/BaasClient.cs
```diff
@@ -863,6 +881,180 @@ private static class Schemas
             },
             GenericBaasRule(differentiator, "foos"));
         }
+
+        private class BaaSaasClient
```
nirinchev, line 885: I would probably call it `Baasaas` - capitalizing the `S` makes it super weird, especially in camelcase, where it looks like the splitting is `baa saa s` 😅

Tools/DeployApps/BaasClient.cs
```diff
@@ -863,6 +881,180 @@ private static class Schemas
...
+                    {
+                    }
+
+                    await Task.Delay(2000);
+                }
+
+                throw new Exception($"Container with id={containerId} was not found or ready after {maxRetries} retrues");
+            }
+
+            private async Task<T?> CallEndpointAsync<T>(HttpMethod method, string relativePath, object? payload = null)
```
nirinchev, line 1006: Should this return `T?` - I feel the caller should control the return type through the generic.

nirinchev (approved):
Looks good; just nits from me. I know it's kind of unrelated, but can we fix the warnings as well.

Tools/DeployApps/BaasClient.cs
```diff
@@ -739,6 +699,64 @@ private async Task RefreshAccessTokenAsync()
...
+            }
+
+            var content = new StringContent(jsonContent);
+
+            content.Headers.ContentType = new MediaTypeHeaderValue("application/json");
+
+            return content;
+        }
+
+        public static (Dictionary<string, string> Extracted, string[] RemainingArgs) ExtractArguments(string[] args, params string[] toExtract)
```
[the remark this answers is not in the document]
papafe: This is also used by the sync test helpers to setup the logger from the command line arguments. Maybe this shouldn't be the place for this method at all 😁

PR #3532, opened 2024-02-21
Increasing timeouts on Windows/UWP/Android

papafe:
Tests on CI are timing out since we added new sync tests, so for now we're increasing the timeouts.

### DIFF
--- a/.github/actions/run-android-device-farm-test/action.yml
+++ b/.github/actions/run-android-device-farm-test/action.yml
@@ -22,6 +22,7 @@ runs:
       with:
         project_arn: ${{ inputs.project-arn}}
         device_pool_arn: ${{ inputs.device-pool-arn }}
+        timeout: 2700
         app_file: ${{ inputs.apk-path }}
         app_type: ANDROID_APP
         test_type: APPIUM_PYTHON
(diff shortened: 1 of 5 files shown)


.github/actions/run-android-device-farm-test/action.yml
```diff
@@ -22,6 +22,7 @@ runs:
       with:
         project_arn: ${{ inputs.project-arn}}
         device_pool_arn: ${{ inputs.device-pool-arn }}
+        timeout: 2700
```
nirinchev, line 25: Will that be enough - the others got bumped from 45 to 50.

.github/templates/test-net-framework.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   run-tests:
     runs-on: windows-latest
     name: .NET Framework
-    timeout-minutes: 45
```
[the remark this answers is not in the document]
papafe: Yeah, that's what I did before, I didn't want to be too hungry 😁

.github/actions/run-android-device-farm-test/action.yml
papafe: The tests themselves usually take a little more than 40 min, that usually was going after the limit of 45 minutes including the time to build and some other work. This timeout here is only for the tests, so I thought that 45 minutes (2700 seconds) would be enough. I just increased it a little bit more to be on the safe side.

PR #3535, opened 2024-02-23
Fix various failing CI tests

### LINKED ISSUE #3526: Investigate flaky tests
It seems that some of the tests for static queries merged with https://github.com/realm/realm-dotnet/pull/3414 are flaky

papafe:
Fixes #3526

### DIFF
--- a/Tests/Realm.Tests/Sync/SessionTests.cs
+++ b/Tests/Realm.Tests/Sync/SessionTests.cs
@@ -1328,7 +1328,7 @@ namespace Realms.Tests.Sync
 
                 sw.Stop();
 
-                Assert.That(sw.ElapsedMilliseconds, Is.LessThan(20));
+                Assert.That(sw.ElapsedMilliseconds, Is.LessThan(30));
             });
         }
 
@@ -1377,7 +1377,7 @@ namespace Realms.Tests.Sync
 
                 sw.Stop();
 
-                Assert.That(sw.ElapsedMilliseconds, Is.LessThan(20));
+                Assert.That(sw.ElapsedMilliseconds, Is.LessThan(30));
                 Assert.That(readerRealm.All<SyncAllTypesObject>().Count(), Is.Zero);
 
                 await GetSession(readerRealm).WaitForDownloadAsync();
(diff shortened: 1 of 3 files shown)


Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -1328,7 +1328,7 @@ public void Session_WaitForUpload_CanBeCancelled()
 
                 sw.Stop();
 
-                Assert.That(sw.ElapsedMilliseconds, Is.LessThan(5));
```
papafe, line 1331: Even locally these two tests don't pass all the time

Tests/Realm.Tests/Sync/StaticQueriesTests.cs
```diff
@@ -1257,7 +1257,7 @@ private async Task<MongoClient.Collection<T>> GetCollection<T>(string appConfigT
 
             SyncConfigurationBase config = appConfigType == AppConfigType.FlexibleSync ? GetFLXIntegrationConfig(user) : GetIntegrationConfig(user);
 
-            using var realm = await GetRealmAsync(config);
```
papafe, line 1260: We need to wait for the schema to be synchronized, otherwise the call to the backends without the database name could raise an exception

Tests/Realm.Tests/Sync/StaticQueriesTests.cs
```diff
@@ -1257,7 +1257,7 @@ private async Task<MongoClient.Collection<T>> GetCollection<T>(string appConfigT
...
+            using var realm = await GetRealmAsync(config, true);
```
papafe, line 1260: We need to wait for the schema to be synchronized, otherwise the call to the backends without the database name could raise an exception

Tests/Realm.Tests/Sync/UserManagementTests.cs
```diff
@@ -1166,11 +1166,11 @@ public void UserCustomDataChange_RaisesChanged()
 
                 await tcs.Task;
 
-                tcs = new();
-
                 // Unsubscribe and verify that it no longer raises user changed
                 user.Changed -= OnUserChanged;
 
+                tcs = new();
```
papafe, line 1172: Sometimes no `TimeoutException` was raised at the end of the method on CI. I verified that actually we get two `Changed` notifications, one for the custom data, and one for the user token refresh. 
I suspect that sometimes the second notification was being raised after the new token completion source was created but before the unsubscription, that's why the timeout had no effect.

nirinchev (approved):

PR #3536, opened 2024-02-27
Update to core 14

papafe:
##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)
* [x] Wait for core to fix issue regarding queries with decimals.  https://github.com/realm/realm-core/commit/b1d32d3d20fcb80c747aad6b37b4a9d9179e0730
* [x] Fix issue regarding "fake" sync tests
* [x] Fix issue regarding opening "guid.realm" realm in read only mode. https://github.com/realm/realm-core/pull/7389
* [ ] Check which realm studio version is correct


papafe:
Most of the changes are here, but we still need to wait for the fixes to the problems found during testing and the correct realm studio version. They are all tracked in the PR description

PR #3539, opened 2024-03-04
Update prepare-release.yml to use `GITHUB_TOKEN`

kraenhansen:
## Description

Fixes an issue caused by SSO being enforced.

##  TODO

* [ ] Changelog entry
* [ ] Tests (if applicable)


nirinchev (approved):

kraenhansen:
Drafting as it seems the `GITHUB_TOKEN` cannot trigger workflows: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

> If you do want to trigger a workflow from within a workflow run, you can use a GitHub App installation access token or a personal access token instead of GITHUB_TOKEN to trigger events that require a token.

PR #3555, opened 2024-03-12
[RNET-1121] Removed collection equality in RealmValue

### LINKED ISSUE #3552: Remove equality for collections inside RealmValue

papafe:
Fixes #3552

##  TODO

* [x] Tests

### DIFF
--- a/Tests/Realm.Tests/Database/RealmValueWithCollections.cs
+++ b/Tests/Realm.Tests/Database/RealmValueWithCollections.cs
@@ -27,8 +27,10 @@ namespace Realms.Tests.Database
     [TestFixture, Preserve(AllMembers = true)]
     internal class RealmValueWithCollections : RealmInstanceTest
     {
-        public static Func<int, List<RealmValue>> ListGenerator = i => { return new List<RealmValue> { $"inner{i}", i }; };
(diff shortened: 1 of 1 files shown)


Realm/Realm/DatabaseTypes/RealmValue.cs
```diff
@@ -1533,8 +1533,6 @@ public bool Equals(RealmValue other)
                 RealmValueType.ObjectId => AsObjectId() == other.AsObjectId(),
                 RealmValueType.Guid => AsGuid() == other.AsGuid(),
                 RealmValueType.Object => AsIRealmObject().Equals(other.AsIRealmObject()),
-                RealmValueType.List => AsList().SequenceEqual(other.AsList()),
-                RealmValueType.Dictionary => AsDictionary().DictionaryEquals(other.AsDictionary()),
                 RealmValueType.Null => true,
```
nirinchev, line 1536: This looks a bit suspicious 🤨

nirinchev (approved):

Realm/Realm/DatabaseTypes/RealmValue.cs
papafe: The `Null` part? It got me for a second too 😁 
Before the switch we check that the type of the two RealmValues is the same. So if they're both null they're automatically equal.

nirinchev: Ah, makes sense

PR #3570, opened 2024-04-09
RNET-1132: Add tests for sync schema migration

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/3569


Tools/DeployApps/BaasClient.cs
```diff
@@ -473,9 +475,9 @@ private async Task<BaasApp> CreateFlxApp(string name)
                 {
                     state = "enabled",
                     database_name = GetSyncDatabaseName(name),
-                    queryable_fields_names = new[] { "Int64Property", "GuidProperty", "DoubleProperty", "Int", "Guid", "Id", "PartitionLike" },
+                    queryable_fields_names = new[] { "Int64Property", "GuidProperty", "DoubleProperty", "Int", "Guid", "Id", "PartitionLike", "Differentiator" },
```
papafe, line 478: Not a biggie, but maybe we should find another name for "differentiator"? We already use it for differentiating the apps, maybe it could create confusion

Tests/Realm.Tests/Sync/SyncMigrationTests.cs
```diff
@@ -0,0 +1,365 @@
...
+                Assert.That(objv1.BoolValue, Is.EqualTo(false));
+                Assert.That(objv1.DateValue, Is.EqualTo(new DateTimeOffset(1, 1, 1, 0, 0, 0, TimeSpan.Zero)));
+                Assert.That(objv1.DecimalValue, Is.EqualTo(Decimal128.Zero));
+                Assert.That(objv1.DoubleValue, Is.EqualTo(0));
+                Assert.That(objv1.IntValue, Is.EqualTo(0));
+                Assert.That(objv1.ObjectIdValue, Is.EqualTo(ObjectId.Empty));
+                Assert.That(objv1.StringValue, Is.EqualTo(string.Empty));
+                Assert.That(objv1.UuidValue, Is.EqualTo(Guid.Empty));
+                Assert.That(objv1.BinaryValue, Is.EqualTo(Array.Empty<byte>()));
+                Assert.That(objv1.WillBeRemoved, Is.EqualTo(string.Empty));
```
papafe, line 196: So the server is keeping different property values depending on the schema version?

papafe: Makes sense, I read it wrong

Tools/DeployApps/BaasClient.cs
nielsenko: We do the same in dart.

nielsenko (approved):

Tests/Realm.Tests/Sync/SyncMigrationTests.cs
```diff
@@ -0,0 +1,365 @@
...
+
+        public ObjectId? ObjectIdValue { get; set; }
+
+        public Guid? UuidValue { get; set; }
+
+        public byte[]? BinaryValue { get; set; }
+    }
+
+    [MapTo("Nullables"), Explicit]
+    public partial class NullablesV1 : IRealmObject
```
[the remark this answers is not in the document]
nirinchev: But that's the entire plot of this test - it used to have nullables, then some developer forgot that in json schema the nullability is specified outside of the type declaration, so accidentally removed all the nullables and when it was caught, they reverted to `NullablesV0`, which has all the nullables 😄 It's high stakes drama with plot twists and character development.

Tools/DeployApps/BaasClient.cs
nirinchev: I feel it's a reasonable name given that it differentiates things based on some criteria. Is it the best name - probably not, but I liked it better than `PartitionLike` 😄

Tests/Realm.Tests/Sync/SyncMigrationTests.cs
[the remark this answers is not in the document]
nielsenko: Well, when you put it like that ..

PR #3575, opened 2024-04-16
Hide collections in mixed from public API

papafe:


nielsenko (approved):
LGTM.

Consider prefixing out-commented doc comments with "// " instead of "//" and out-commented code with "////" to avoid the warnings from [StyleCop SA1005](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1005.md)

papafe:
> LGTM.
> 
> Consider prefixing out-commented doc comments with "// " instead of "//" and out-commented code with "////" to avoid the warnings from [StyleCop SA1005](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1005.md?rgh-link-date=2024-04-17T07%3A05%3A13Z)

Yes, I just used the VS keyboard shortcut to comment out everything to be fast 😁

PR #3585, opened 2024-04-25
Update to latest Core

nirinchev:
## Description

The GHA workflow changes are related to macos-14 recently becoming macos-latest, which broke our assumptions about the architecture.

This depends on https://github.com/realm/realm-core/pull/7632 - once it's merged, we should retarget master.

##  TODO

* [ ] Changelog entry
* [ ] Tests

### DIFF
--- a/.github/templates/test-ios.yml
+++ b/.github/templates/test-ios.yml
@@ -7,7 +7,7 @@ name: test-ios
 _: #@ template.replace(testDefinition())
 jobs:
   test-xamarin:
-    runs-on: macos-13
+    runs-on: macos-12
     name: Xamarin.iOS
     timeout-minutes: 90
     steps:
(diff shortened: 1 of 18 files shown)


.github/templates/test-ios.yml
```diff
@@ -7,7 +7,7 @@ name: test-ios
 _: #@ template.replace(testDefinition())
 jobs:
   test-xamarin:
-    runs-on: macos-latest
+    runs-on: macos-12
```
papafe, line 10: Shouldn't we this and the others to 13?

papafe (approved):
Looks reasonable

.github/templates/test-ios.yml
nirinchev: 13 doesn't have the xamarin tooling. We could manually install it, but I'd rather just use 12 while possible and then either deprecate Xamarin or add a step to manually install at that point.

PR #3587, opened 2024-04-29
RNET-1109: Add App.UpdateBaseUri

nirinchev:
## Description

Fixes https://github.com/realm/realm-dotnet/issues/3521

##  TODO

* [x] Changelog entry
* [x] Tests

### DIFF
--- a/Realm/Realm/Handles/AppHandle.cs
+++ b/Realm/Realm/Handles/AppHandle.cs
@@ -349,13 +349,13 @@ namespace Realms.Sync
             return new Uri(uriString);
         }
 
-        public async Task UpdateBaseUriAsync(Uri newUri)
+        public async Task UpdateBaseUriAsync(Uri? newUri)
         {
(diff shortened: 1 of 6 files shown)


Realm/Realm/Handles/AppHandle.cs
```diff
@@ -343,6 +349,23 @@ public Uri GetBaseUri()
             return new Uri(uriString);
         }
 
+        public async Task UpdateBaseUriAsync(Uri? newUri)
+        {
+            var tcs = new TaskCompletionSource();
+            var tcsHandle = GCHandle.Alloc(tcs);
+            try
+            {
+                var url = newUri?.ToString().TrimEnd('/') ?? string.Empty;
```
papafe, line 358: What is the effect of passing a nullable Uri? Having it set to the default value?

papafe (approved):
Looks good, I suppose it just needs the changelog

PR #3588, opened 2024-04-30
RNET-1119: Add privacy manifest

### LINKED ISSUE #3541: Privacy Manifest support would be needed in realm-wrappers
### Problem

[Like realm-swift did](https://github.com/realm/realm-swift/issues/8428), it seems that realm-dotnet also need to address Apple's new [Privacy Manifest](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files) reguration.

Apple has [announced](https://developer.apple.com/news/?id=3d8a9yyh) that new publication and updates must support Privacy Manifests starting May 1.

### Solution

To compliant the new regulation, it would be needed to include a privacy manifest file into `realm-wrappers.framework`.

As an experiment, I wrote [some patch](https://github.com/takeshik/realm-dotnet/commit/cf0d9c4ab6b8731160dedae538982403853ab376) to include a manifest into `realm-wrappers.framework`.
In this patch, `PrivacyInfo.xcprivacy` is brought from realm/realm-swift#8455. I think that the declaration in the manifest can be appropriated as is, but I'm not certain.

### Alternatives

_No response_

Would be a major improvement

### Feature would mainly be used with

Local Database only

papafe:
Fixes #3541


papafe:
I'm closing this as probably we don't need this for MAUI, as the minimum privacy manifest already covers our cases. 
The PR workflow was modified to create packages faster too. I'll keep the branch in case we need to revisit this at some point.

PR #3595, opened 2024-05-03
RNET-1126: Fix flaky tests on CI

### LINKED ISSUE #3560: Fix flaky static queries tests
We should fix flaky static queries tests, for example in:
- https://github.com/realm/realm-dotnet/runs/22894980207#r0s1
- https://github.com/realm/realm-dotnet/runs/22612387507

It seems most of the time it's the tests `RealmObjectAPI_RealmValueLinks_AtlasToRealm` and `RealmObjectAPI_Links_AtlasToRealm` that are giving issues. I think this could be due to some timing issues, for example if objects are created but the properties are not set yet, or if clearing the collection at the beginning is not yet propagated to realm before continuing.

papafe:
Fixes #3560

### DIFF
--- a/Tests/Realm.Tests/Sync/StaticQueriesTests.cs
+++ b/Tests/Realm.Tests/Sync/StaticQueriesTests.cs
@@ -760,7 +760,8 @@ namespace Realms.Tests.Sync
 
                 var realmValObj = realm.Find<RealmValueObject>(obj.Id)!;
 
-                await WaitForConditionAsync(() => {
+                await WaitForConditionAsync(() => 
+                {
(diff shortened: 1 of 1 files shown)


papafe:
The cause of the flakyness seems to be due to some various timing issues:
- Reusing the same object for different tests led to realm exceptions when adding an object with the same primary key
- When objects are added directly to Atlas, the synchronization of objects does not respect the boundaries of changesets, that can be broken down in smaller pieces (especially when working with links). This means that sometimes the elements of collections of links are not added all together, and this did not work with our expectations. This is the reason I've added much more `WaitForConditions`.

Issue #3597, opened 2024-05-06
Error when trying to migrate IList data type with DynamicApi.GetList


Scorter:
### What happened?

I'm migrating data that was previously "float?" and IList<float?" to the type "string" and "IList<string>". I am attaching the code that performs this migration, and everything goes well when I give a DynamicApi.GetList in an IEmbeddedObject, however, when I call a DynamicApi.GetList in a "float?" gives the error "KeyNotFoundException: The given key '' was not present in the dictionary". him being the "float?" type. The error only happens when I try to call a GetList in the "MeasuredValues" field, the rest of the code works fine and performs the correct conversion. Does anyone have an idea what I might be doing wrong? See the error that appears:

![image](https://github.com/realm/realm-dotnet/assets/88690064/dfd51fc1-ad39-4650-9be1-aa90c356db29)

See how MeasuredValues ​​is in the database, it appears in the scope as List<float> but it is a List<float?>, I believe this is a bug in the realm, but I guarantee that in the code we are saving it as IList<float?>.

![image](https://github.com/realm/realm-dotnet/assets/88690064/1b358767-8dd8-4916-bb6d-172775b2ec38)

I'm attaching a drive with the old bank (the bank that tries to get "MeasuredValues") in case it helps with anything. Link drive :
[LinkDrive](https://drive.google.com/file/d/1rdPrO4Ahk91n5N_7t4nedUfR9h63CQuO/view?usp=sharing)

### Repro steps

1. Save an object in the database with the field type "IList<float?>
2. Perform a database migration by calling GetList from an "Ilist<float?">

The error occurs exactly on this line:
![image](https://github.com/realm/realm-dotnet/assets/88690064/08454e1e-735e-41c4-b3d6-84f48d71fb99)

### Version

.NET Framework

### What Atlas Services are you using?

Local Database only

### What type of application is this?

Unity

### Client OS and version

Unity

### Code snippets

```
if (oldSchemaVersion < 2)
{
    var oldOrders = migration.OldRealm.DynamicApi.All("OrderServiceEntity");
    var newOrders = migration.NewRealm.All<OrderServiceEntity>();

    for (int i = 0; i < newOrders.Count(); i++)
    {
        var oldOrder = oldOrders.ElementAt(i);
        var newOrder = newOrders.ElementAt(i);

        var oldCalibration = oldOrder.DynamicApi.Get<IEmbeddedObject>("Calibration");
        var newCalibration = newOrder.Calibration;

[...]

PR #3601, opened 2024-05-07
RNET-1145, RNET-1143: Fix reading schema from native

nirinchev:
## Description

This fixes how we're reading Realm schema from native. It seems that for a few string properties, the default value in Core is empty string, whereas in .NET those are nullable, so our checks for `object_type == null` were failing and we were trying to lookup metadata for primitive properties.

Additionally, this adds `Migration.FindInNewRealm` since I saw we're missing that, which made the test easier to write.

Fixes https://github.com/realm/realm-dotnet/issues/3600
Fixes https://github.com/realm/realm-dotnet/issues/3597

##  TODO

* [x] Changelog entry
* [x] Tests


papafe (approved):
Looks good. Just left a small comment and maybe we should fix those warnings about the non-nullable property

PR #3603, opened 2024-05-13
Use pkl to generate gha workflows

nirinchev:


.github/pkl-workflows/helpers/Lint.pkl
```diff
@@ -0,0 +1,69 @@
...
+import "Common.pkl"
+
+const actionVerifyTODO = "nirinchev/verify-todo@ffa352ac028b4cdc8bc626471d33aa341b3ab7c9"
+
+function lint(): gha.Job = new {
+  name = "Verify TODOs"
+  `runs-on` = new gha.UbuntuLatest{}
+  steps {
+    Steps.checkoutWithoutMatchers(false)
+    new gha.Step {
```
papafe, line 14: I think we can leave the `gha.Step` here out, right? Mostly a question more than a comment.
I can see we leave it out in `verifyNamespaces`

Tools/DeployApps/BaasClient.cs
```diff
@@ -1125,7 +1125,7 @@ private async Task<ContainerInfo> WaitForContainer(string containerId, int maxRe
                     await Task.Delay(2000);
                 }
 
-                throw new Exception($"Container with id={containerId} was not found or ready after {maxRetries} retrues");
```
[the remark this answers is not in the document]
nirinchev: Because we're actually decrementing `maxRetries`, so if we get here, `maxRetries` will always be 0.

.github/pkl-workflows/helpers/Common.pkl
```diff
@@ -0,0 +1,91 @@
...
+
+const wrapperBinaryNames: List<String> =
+  List("macos", "catalyst")
+  + linuxArchs.map((arch) -> "linux-\(arch)")
+  + androidABIs.map((abi) -> "android-\(abi)")
+  + windowsArchs.map((arch) -> "windows-\(arch)")
+  + uwpArchs.map((arch) -> "windows-uwp-\(arch)")
+  + applePlatformTargets((platform, target) -> "\(platform)-\(target)")
+
+const defaultEnv: Mapping<String, String | Boolean> = new {
```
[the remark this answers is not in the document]
nirinchev: That's matching the type of `env`.

Tools/DeployApps/BaasClient.cs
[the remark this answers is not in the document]
papafe: So, as we said, maybe it makes sense to save the original maxRetries value ;)

.github/workflows/wrappers.yml
```diff
@@ -36,607 +29,618 @@ jobs:
...
+      if: matrix.os == 'ubuntu' && steps.check-cache-tvOS-Simulator.outputs.cache-hit == 'true'
+      uses: actions/upload-artifact@v4
+      with:
+        name: wrappers-tvOS-Simulator
+        path: wrappers/build/**
+        retention-days: ${{ github.event_name != 'pull_request' && 30 || 1 }}
+        if-no-files-found: error
+    - if: matrix.os == 'ubuntu'
+      run: git clean -fdx
+    strategy:
```
papafe, line 383: Is there a way of putting this before all the other steps as it was before? I thought this was missing at a first glance

.github/pkl-workflows/GithubAction/GithubAction.pkl
```diff
@@ -0,0 +1,690 @@
...
+/// Runs your workflow anytime when someone forks a repository, which triggers the fork event.
+/// For information about the REST API, see https://developer.github.com/v3/repos/forks/#create-a-fork
+///
+/// https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#fork
+class Fork extends Trigger
+
+/// Runs your workflow when someone creates or updates a Wiki page, which triggers the gollum event.
+///
+/// https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#gollum
+class Gollum extends Trigger
```
papafe, line 127: 👀

papafe (approved):
I have given a look around and it mostly made sense, even though it's a dense PR.
I think I'm just a little bit torn on the fact that we know have fewer but huge workflows instead of having them split in smaller ones. Hopefully with pkl it should be easier to manage those though

papafe (approved):
I have given a look around and it mostly made sense, even though it's a dense PR.
I think I'm just a little bit torn on the fact that we know have fewer but huge workflows instead of having them split in smaller ones. Hopefully with pkl it should be easier to manage those though

.github/pkl-workflows/GithubAction/GithubAction.pkl
nirinchev: Yeah 🤷‍♂️

PR #3606, opened 2024-05-21
Revert "Hide collections in mixed from public API"

nirinchev:
Reverts realm/realm-dotnet#3575


nirinchev:
Failing tests are due to a change in schema migrations on the server. I've asked here: https://mongodb.slack.com/archives/C04NACGT7J7/p1716329105823099.

PR #3615, opened 2024-06-03
RNET-1137: Fix progress notification

### LINKED ISSUE #3580: Follow up on progress notification
This is a reminder to follow up on the progress notification when the core work is done to:
- Get the progress notification value from core
- Fix the tests

papafe:
This PR uses the correct core value for the progress notification, and adds some basic tests

Fixes #3580

##  TODO

* [x] Changelog entry
* [x] Tests

### DIFF
--- a/Tests/Realm.Tests/Sync/SessionTests.cs
+++ b/Tests/Realm.Tests/Sync/SessionTests.cs
@@ -756,7 +756,6 @@ namespace Realms.Tests.Sync
             });
         }
 
-        // This test needs to be revisited when the work on progress notification is finished.
         [Test]
         public void SessionIntegrationTest_ProgressObservable(
             [ValueSource(nameof(AppTypes))] string appType,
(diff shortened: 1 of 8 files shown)


papafe:
@nirinchev I've added a basic test, not sure how extensive we want to be with testing for this.

Tests/Realm.Tests/Sync/SessionTests.cs
```diff
@@ -828,25 +837,18 @@ public void Session_OnSessionError()
                             }
 
                             completionTcs.TrySetResult();
+                            lastReportedProgress = p.ProgressEstimate;
```
nirinchev, line 840: Why is this happening after we complete the tcs?

papafe: It shouldn't, it should happen all the time, I'll fix it 🤦

nirinchev (approved):

PR #3630, opened 2024-06-18
RNET-1159: Flexible-schema POC

### LINKED ISSUE #3629: Flexible schema POC

papafe:
This PR is a POC for the flexible schema

Fixes #3629


papafe:
@nirinchev @elle-j This is not complete, but I thought it would be good to have an initial feedback. 
What I think it's missing:
- Try to reduce the internal API methods
- Retrieve the complete schema from realm (not only the model and extra properties)
- Tests for queries/keypath/notifications (these requires the core implementation too)

PR #3634, opened 2024-06-26
RNET-1161: Implement support for using a log level for a specific log category

### LINKED ISSUE #3633: Support using a log level for a specific log category

elle-j:
## Description

Deprecates `Logger` in favor of its new base class `RealmLogger` (see notes below on avoiding breaking changes).

Primarily, enables more control over which category of messages should be logged and at which criticality level, by setting and getting a `LogLevel` for a given `LogCategory`. The hierarchy of categories starts at `LogCategory.Realm`.

```csharp
RealmLogger.SetLogLevel(LogLevel.Warn, LogCategory.Realm.Sync);
RealmLogger.GetLogLevel(LogCategory.Realm.Sync.Client.Session); // LogLevel.Warn
```

Adds a function logger that accepts a callback that will receive the `LogLevel`, `LogCategory`, and the message when invoked.

```csharp
RealmLogger.Default = RealmLogger.Function((level, category, message) => /* custom implementation */);
```

Adds a `RealmLogger.Log()` overload taking a category.
```csharp
RealmLogger.Default.Log(LogLevel.Warn, LogCategory.Realm, "A warning message");
```

### Would-be breaking changes

Users have been able to provide their own `Logger` subclass via the `Logger.Default` `set` accessor. Custom loggers need to override and implement `Logger.LogImpl()`, and with the addition of log categories in this PR, the `LogImpl()` signature would require breaking changes.

### Solution

Essentially, the previous `Logger` class has been renamed `RealmLogger`, and another class called `Logger` now derives from `RealmLogger`, with the only own member being the `LogImpl()` with an abstract method identical to before to avoid the breaking change.

```csharp
public abstract class Logger : RealmLogger
{
    protected abstract void LogImpl(LogLevel level, string message);

    protected override void LogImpl(LogLevel level, LogCategory category, string message)
    {
        LogImpl(level, message);
    }
}
```

Fixes #3633

@realm/devdocs 

##  TODO

* [x] Changelog entry
* [x] Tests
* [x] API docs

### DIFF
--- a/Realm/Realm/Handles/SharedRealmHandle.cs
+++ b/Realm/Realm/Handles/SharedRealmHandle.cs
@@ -851,7 +851,7 @@ namespace Realms
         [MonoPInvokeCallback(typeof(NativeMethods.LogMessageCallback))]
         private static void LogMessage(StringValue message, LogLevel level, StringValue categoryName)
         {
-            RealmLogger.LogDefault(level, LogCategory.FromName(categoryName!), message!);
+            RealmLogger.CoreLogDefault(level, LogCategory.FromName(categoryName!), message!);
         }
 
         [MonoPInvokeCallback(typeof(NativeMethods.MigrationCallback))]
--- a/Realm/Realm/Handles/SharedRealmHandle.cs
+++ b/Realm/Realm/Handles/SharedRealmHandle.cs
@@ -49,6 +49,17 @@ namespace Realms
 
         private static class NativeMethods
         {
+            // This is a wrapper struct around MarshaledVector since P/Invoke doesn't like it
+            // when the MarshaledVector is returned as the top-level return value from a native
+            // function. This only manifests in .NET Framework and is not an issue with Mono/.NET.
+            // The native return value is MarshaledVector without the wrapper because they are binary
+            // compatible.
+            [StructLayout(LayoutKind.Sequential)]
+            public struct CategoryNamesContainer
+            {
+                public MarshaledVector<StringValue> CategoryNames;
+            }
+
 #pragma warning disable IDE0049 // Use built-in type alias
 #pragma warning disable SA1121 // Use built-in type alias
 
@@ -232,7 +243,7 @@ namespace Realms
             public static extern void set_log_level(LogLevel level, [MarshalAs(UnmanagedType.LPWStr)] string category_name, IntPtr category_name_len);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_get_log_category_names", CallingConvention = CallingConvention.Cdecl)]
-            public static extern MarshaledVector<StringValue> get_log_category_names();
+            public static extern CategoryNamesContainer get_log_category_names();
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_get_operating_system", CallingConvention = CallingConvention.Cdecl)]
             public static extern IntPtr get_operating_system(IntPtr buffer, IntPtr buffer_length);
@@ -284,6 +295,7 @@ namespace Realms
         public static void SetLogLevel(LogLevel level, LogCategory category) => NativeMethods.set_log_level(level, category.Name, (IntPtr)category.Name.Length);
 
         public static string[] GetLogCategoryNames() => NativeMethods.get_log_category_names()
+            .CategoryNames
             .ToEnumerable()
             .Select(name => name.ToDotnetString()!)
             .ToArray();
--- a/Realm/Realm/Handles/SharedRealmHandle.cs
+++ b/Realm/Realm/Handles/SharedRealmHandle.cs
@@ -839,7 +839,7 @@ namespace Realms
         [MonoPInvokeCallback(typeof(NativeMethods.LogMessageCallback))]
         private static void LogMessage(StringValue message, LogLevel level, StringValue categoryName)
         {
-            Logger.LogDefault(level, LogCategory.FromName(categoryName!), message!);
+            Logger.LogDefault(level, message!, LogCategory.FromName(categoryName!));
         }
 
         [MonoPInvokeCallback(typeof(NativeMethods.MigrationCallback))]
--- a/Realm/Realm/Handles/SharedRealmHandle.cs
+++ b/Realm/Realm/Handles/SharedRealmHandle.cs
@@ -75,9 +75,8 @@ namespace Realms
             [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
             public delegate void DisposeGCHandleCallback(IntPtr handle);
 
-            // TODO(lj): Update arg order to be the same across the SDK.
             [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
-            public delegate void LogMessageCallback(StringValue message, LogLevel level, StringValue categoryName);
+            public delegate void LogMessageCallback(LogLevel level, StringValue categoryName, StringValue message);
 
             [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
             public delegate void HandleTaskCompletionCallback(IntPtr tcs_ptr, [MarshalAs(UnmanagedType.U1)] bool invoke_async, NativeException ex);
@@ -849,7 +848,7 @@ namespace Realms
         }
 
         [MonoPInvokeCallback(typeof(NativeMethods.LogMessageCallback))]
-        private static void LogMessage(StringValue message, LogLevel level, StringValue categoryName)
+        private static void LogMessage(LogLevel level, StringValue categoryName, StringValue message)
         {
             RealmLogger.CoreLogDefault(level, LogCategory.FromName(categoryName!), message!);
         }
--- a/Realm/Realm/Handles/SharedRealmHandle.cs
+++ b/Realm/Realm/Handles/SharedRealmHandle.cs
@@ -224,6 +224,9 @@ namespace Realms
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_refresh_async", CallingConvention = CallingConvention.Cdecl)]
             public static extern bool refresh_async(SharedRealmHandle realm, IntPtr tcs_handle, out NativeException ex);
 
+            [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_get_log_level", CallingConvention = CallingConvention.Cdecl)]
+            public static extern LogLevel get_log_level([MarshalAs(UnmanagedType.LPWStr)] string category_name, IntPtr category_name_len);
+
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_set_log_level", CallingConvention = CallingConvention.Cdecl)]
             public static extern void set_log_level(LogLevel level, [MarshalAs(UnmanagedType.LPWStr)] string category_name, IntPtr category_name_len);
 
@@ -271,6 +274,8 @@ namespace Realms
             NativeMethods.install_callbacks(notifyRealm, getNativeSchema, openRealm, disposeGCHandle, logMessage,
                 notifyObject, notifyDictionary, onMigration, shouldCompact, handleTaskCompletion, onInitialization);
         }
+
+        public static LogLevel GetLogLevel(LogCategory category) => NativeMethods.get_log_level(category.Name, (IntPtr)category.Name.Length);
 
         public static void SetLogLevel(LogLevel level, LogCategory category) => NativeMethods.set_log_level(level, category.Name, (IntPtr)category.Name.Length);
 
--- a/Realm/Realm/Handles/SharedRealmHandle.cs
+++ b/Realm/Realm/Handles/SharedRealmHandle.cs
@@ -850,7 +850,7 @@ namespace Realms
         [MonoPInvokeCallback(typeof(NativeMethods.LogMessageCallback))]
         private static void LogMessage(LogLevel level, StringValue categoryName, StringValue message)
         {
-            RealmLogger.CoreLogDefault(level, LogCategory.FromName(categoryName!), message!);
+            RealmLogger.Default.LogAnyLevel(level, LogCategory.FromName(categoryName!), message!);
         }
 
         [MonoPInvokeCallback(typeof(NativeMethods.MigrationCallback))]
--- a/Realm/Realm/Handles/SharedRealmHandle.cs
+++ b/Realm/Realm/Handles/SharedRealmHandle.cs
@@ -851,7 +851,7 @@ namespace Realms
         [MonoPInvokeCallback(typeof(NativeMethods.LogMessageCallback))]
         private static void LogMessage(StringValue message, LogLevel level, StringValue categoryName)
         {
-            Logger.LogDefault(level, message!, LogCategory.FromName(categoryName!));
+            RealmLogger.LogDefault(level, LogCategory.FromName(categoryName!), message!);
         }
 
         [MonoPInvokeCallback(typeof(NativeMethods.MigrationCallback))]
--- a/Realm/Realm/Handles/SharedRealmHandle.cs
+++ b/Realm/Realm/Handles/SharedRealmHandle.cs
@@ -63,8 +63,9 @@ namespace Realms
             [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
             public delegate void DisposeGCHandleCallback(IntPtr handle);
 
+            // TODO(lj): Update arg order to be the same across the SDK.
             [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
-            public delegate void LogMessageCallback(StringValue message, LogLevel level);
+            public delegate void LogMessageCallback(StringValue message, LogLevel level, StringValue categoryName);
 
             [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
             public delegate void HandleTaskCompletionCallback(IntPtr tcs_ptr, [MarshalAs(UnmanagedType.U1)] bool invoke_async, NativeException ex);
@@ -224,7 +225,7 @@ namespace Realms
             public static extern bool refresh_async(SharedRealmHandle realm, IntPtr tcs_handle, out NativeException ex);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_set_log_level", CallingConvention = CallingConvention.Cdecl)]
-            public static extern bool set_log_level(LogLevel level);
+            public static extern void set_log_level(LogLevel level, [MarshalAs(UnmanagedType.LPWStr)] string category_name, IntPtr category_name_len);
 
             [DllImport(InteropConfig.DLL_NAME, EntryPoint = "shared_realm_get_operating_system", CallingConvention = CallingConvention.Cdecl)]
             public static extern IntPtr get_operating_system(IntPtr buffer, IntPtr buffer_length);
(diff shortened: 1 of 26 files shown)


Realm/Realm/Logging/LogCategory.cs
```diff
@@ -0,0 +1,132 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+using System;
+using System.Collections.Generic;
+using Realms.Helpers;
+
+namespace Realms.Logging
+{
+    // TODO(lj): Remove 'abstract' and use this class instead of 'LeafLogCategory'?
```
elle-j, line 25: @nirinchev, I could use the `LogCategory` instead of the `LeafLogCategory` but wouldn't it be clearer having the designated leaf class? I don't have a strong opinion on this but the allowed type may indicated more clearly, e.g. compare:
* `public LeafLogCategory Session { get; } = /* can only be a LeafLogCategory here */`
* `public LogCategory Session { get; } = /* can technically be any LogCategory here */`

Realm/Realm/Logging/Logger.cs
```diff
@@ -35,8 +35,11 @@ public abstract class Logger
     {
         private readonly Lazy<GCHandle> _gcHandle;
 
+        // TODO(lj): Use this default log level?
+        private static readonly LogLevel _defaultLogLevel = LogLevel.Info;
+        private static readonly LogCategory _defaultLogCategory = LogCategory.Realm;
+        private static LogCategory _logCategory = _defaultLogCategory;
```
nirinchev, line 41: Why do we need those fields?

Realm/Realm/Logging/Logger.cs
```diff
@@ -64,39 +67,73 @@ public abstract class Logger
...
+        /// Gets a <see cref="FunctionLogger"/> that proxies Log calls to the supplied function. The message will
+        /// already be formatted with the default message formatting that includes a timestamp.
+        /// </summary>
+        /// <param name="logFunction">Function to proxy log calls to.</param>
+        /// <returns>
+        /// A <see cref="Logger"/> instance that will invoke <paramref name="logFunction"/> for each message.
+        /// </returns>
+        public static Logger Function(Action<string> logFunction) => new FunctionLogger((level, category, message) => logFunction(FormatLog(level, category, message)));
+
+        // TODO(lj): Deprecate
```
nirinchev, line 80: Yes

Realm/Realm/Logging/Logger.cs
```diff
@@ -64,39 +67,73 @@ public abstract class Logger
...
+        public static Logger Function(Action<LogLevel, LogCategory, string> logFunction) => new FunctionLogger(logFunction);
 
         /// <summary>
         /// Gets or sets the verbosity of log messages.
         /// </summary>
         /// <value>The log level for Realm-originating messages.</value>
         public static LogLevel LogLevel
         {
-            get => _logLevel;
+            // TODO(lj): Do we want to deprecate the getter as well?
```
nirinchev, line 105: Yes - the entire property should be deprecated.

Realm/Realm/Logging/Logger.cs
```diff
@@ -64,39 +67,73 @@ public abstract class Logger
...
-                _logLevel = value;
-                SharedRealmHandle.SetLogLevel(value);
+                SetLogLevel(value);
             }
         }
 
+        public static LogCategory LogCategory
+        {
+            get => _logCategory;
+        }
```
nirinchev, line 117: I don't think this makes sense

Realm/Realm/Logging/Logger.cs
```diff
@@ -64,39 +67,73 @@ public abstract class Logger
...
+
+        public static LogLevel GetLogLevel(LogCategory? category = null)
+        {
+            // TODO(lj): Perhaps we should grab the current category (`_logCategory`)
+            //           instead of the default here? If there hasn't been a category
+            //           explicitly set, it will still be the default.
```
nirinchev, line 123: I think we should remove the current category altogether.

Realm/Realm/Logging/Logger.cs
```diff
@@ -64,39 +67,73 @@ public abstract class Logger
...
+
+        public static void SetLogLevel(LogLevel level, LogCategory? category = null)
+        {
+            category ??= _defaultLogCategory;
+            SharedRealmHandle.SetLogLevel(level, category);
+            _logCategory = category;
+        }
+
+        // TODO(lj): Would it make sense to also deprecate the Default setter
+        //           and provide e.g. `SetDefaultLogger()`?
```
nirinchev, line 136: Why would we deprecate  the logger setter?

Realm/Realm/Logging/Logger.cs
```diff
@@ -120,12 +157,19 @@ protected Logger()
 
         internal static void LogDefault(LogLevel level, string message) => Default?.Log(level, message);
```
nirinchev, line 158: This should be removed

Realm/Realm/Logging/Logger.cs
```diff
@@ -120,12 +157,19 @@ protected Logger()
...
 
+        internal static void LogDefault(LogLevel level, LogCategory category, string message) => Default?.Log(level, category, message);
+
         /// <summary>
         /// Log a message at the supplied level.
         /// </summary>
         /// <param name="level">The criticality level for the message.</param>
         /// <param name="message">The message to log.</param>
         public void Log(LogLevel level, string message)
```
nirinchev, line 167: This should be deprecated.

Realm/Realm/Logging/Logger.cs
```diff
@@ -120,12 +157,19 @@ protected Logger()
...
+        {
+            Log(level, LogCategory, message);
```
nirinchev, line 169: This should probably pass the `Sdk` category.

wrappers/src/shared_realm_cs.cpp
```diff
@@ -295,8 +296,19 @@ REALM_EXPORT void shared_realm_install_callbacks(
...
-REALM_EXPORT void shared_realm_set_log_level(Logger::Level level) {
-    LogCategory::realm.set_default_level_threshold(level);
+REALM_EXPORT Logger::Level shared_realm_get_log_level(uint16_t* category_name_buf, size_t category_name_len) {
+    Utf16StringAccessor category_name(category_name_buf, category_name_len);
+    // TODO(lj): Usage in Core:
+    auto& category = LogCategory::get_category(category_name);
+    return Logger::get_default_logger()->get_level_threshold(category);
+
+    // TODO(lj): But this seems to work as well:
+    // return LogCategory::get_category(category_name).get_default_level_threshold();
```
nirinchev, line 306: This looks nicer as a one-liner.

Realm/Realm/Logging/Logger.cs
elle-j: * The `_defaultLogLevel` is currently unused, was keeping that as a reminder to myself to see if we needed it to conform with some prior behavior.
* `_defaultLogCategory` is here to set it at one place (currently used in 2 places)
* `_logCategory` was initially used for getting the current log category set but I see now that it can be removed 👍

Realm/Realm/Logging/Logger.cs
elle-j: With the removal of `_logCategory`, this field can also be removed

Realm/Realm/Logging/Logger.cs
elle-j: To potentially have more consistency with `SetLogLevel()` and the deprecation if the log level setter.

nirinchev: The reason `SetLogLevel` is a method is because it needs to take an extra argument now, but we should still prefer to use properties over `GetXyz/SetXyz` methods.

elle-j: Okay good to know 🙂

Realm/Realm/Logging/Logger.cs
elle-j: We could deprecate it, but then they'd always need to provide the category if using the `public void Log(LogLevel level, LogCategory category, string message)` signature. Kotlin and Swift uses `category, level, message`, so it's somewhat inconsistent with the `SetLogLevel()` args, I believe Dart uses `level, message, optional category` (so also inconsistent but provides optionality).

Realm/Realm/Logging/Logger.cs
```diff
@@ -120,12 +145,20 @@ protected Logger()
...
         /// <param name="message">The message to log.</param>
         public void Log(LogLevel level, string message)
+        {
+            // TODO(lj): See if `LogCategory.Realm.SDK` should be preferred.
+            Log(level, _defaultLogCategory, message);
+        }
+
+        public void Log(LogLevel level, LogCategory category, string message)
         {
             if (level < LogLevel)
```
nirinchev, line 163: This should now check for `level < GetLogLevel(category)`

Realm/Realm/Logging/Logger.cs
nirinchev: We can take the dart approach and add an optional category to the log method.

elle-j: 🤟

elle-j: It's been [updated](https://github.com/realm/realm-dotnet/pull/3634/commits/ab3ee611393e570d3ac020d417e68f4f89c27330) to use the category as an optional 3rd param. I think it feels slightly off logging with the message in between the level and category, but could be okay.

Realm/Realm/Logging/Logger.cs
elle-j: We currently have a bunch of references to this signature, and I think it could make sense to keep it in order to only have the `Logger` pass the `LogCategory.Realm.SDK` when logging from the SDK (instead of at all the calling sites).

Realm/Realm/Logging/Logger.cs
```diff
@@ -134,34 +178,35 @@ public void Log(LogLevel level, string message)
...
         }
 
         /// <summary>
         /// The internal implementation being called from <see cref="Log"/>.
         /// </summary>
         /// <param name="level">The criticality level for the message.</param>
+        /// <param name="category">TODO(lj).</param>
         /// <param name="message">The message to log.</param>
-        protected abstract void LogImpl(LogLevel level, string message);
+        protected abstract void LogImpl(LogLevel level, LogCategory category, string message);
```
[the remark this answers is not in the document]
elle-j: Rearranged args in [this commit](https://github.com/realm/realm-dotnet/pull/3634/commits/ab3ee611393e570d3ac020d417e68f4f89c27330).

wrappers/src/shared_realm_cs.cpp
```diff
@@ -48,7 +48,8 @@ using namespace realm::util;
 using OpenRealmCallbackT = void(void* task_completion_source, ThreadSafeReference* ref, NativeException::Marshallable ex);
 using RealmChangedT = void(void* managed_state_handle);
 using ReleaseGCHandleT = void(void* managed_handle);
-using LogMessageT = void(realm_string_t message, util::Logger::Level level);
+// TODO(lj): Update arg order to be the same across the SDK.
+using LogMessageT = void(realm_string_t message, util::Logger::Level level, realm_string_t category_name);
```
elle-j, line 52: Will likely update this to use the same arg order as the SDK's `Logger.Log()` for consistency (level, message, category), even though having the message in between looks somewhat odd.

Realm/Realm/Native/SyncSocketProvider.WebSocket.cs
```diff
@@ -29,6 +29,7 @@
 
 namespace Realms.Native;
 
+// TODO(lj): For the logs happening in this class, should we log for `LogCategory.Realm.Sync.Client`?
```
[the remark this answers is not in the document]
clementetb: I think it should be the SDK category as this is a SDK component.

Realm/Realm/Logging/Logger.cs
```diff
@@ -118,27 +152,29 @@ protected Logger()
...
 
         /// <summary>
-        /// Log a message at the supplied level.
+        /// Log a message at the supplied level and category.
         /// </summary>
         /// <param name="level">The criticality level for the message.</param>
         /// <param name="message">The message to log.</param>
-        public void Log(LogLevel level, string message)
+        /// <param name="category">The category for the message. Defaults to <see cref="LogCategory.RealmLogCategory.SDK"/> if not specified.</param>
+        public void Log(LogLevel level, string message, LogCategory? category = null)
```
clementetb, line 163: Have you explored overloading this fuction instead of modifying the signature to keep compatibility?

We could have two functions one with the old signature, and the new one. It would be easier later on to deprecate the old behavior.

Realm/Realm/Handles/SharedRealmHandle.cs
```diff
@@ -19,6 +19,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
```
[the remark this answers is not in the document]
elle-j: This is used for the `Select` API [here](https://github.com/realm/realm-dotnet/blob/e5a6a8e4ab217cc48a51c98a506a5bd0a70ee67f/Realm/Realm/Handles/SharedRealmHandle.cs#L288) 👍

Realm/Realm/Logging/Logger.cs
elle-j: This started off as:

```csharp
public void Log(LogLevel level, string message);
public void Log(LogLevel level, LogCategory category, string message);
```

But we updated it to have a default last param briefly discussed in [this thread](https://github.com/realm/realm-dotnet/pull/3634#discussion_r1655062184). My current understanding is that we don't want to deprecate `Log(level, message)` but to keep providing the option not to provide an explicit category.

Although if this was to change I agree that it'd be clearer to have overloads. Personally, I think the DX feels better having `level, category, message` over a `level, message, category` alternative if we can provide this w/o breaking APIs (including for `LogImpl()`).

Realm/Realm/Logging/Logger.cs
```diff
@@ -147,21 +183,24 @@ public void Log(LogLevel level, string message)
         /// </summary>
         /// <param name="level">The criticality level for the message.</param>
         /// <param name="message">The message to log.</param>
-        protected abstract void LogImpl(LogLevel level, string message);
+        /// <param name="category">The category for the message.</param>
+        protected abstract void LogImpl(LogLevel level, string message, LogCategory? category = null);
```
nirinchev, line 187: This is still a breaking change - the addition of the argument is the breaking part, the nullability or the default value are inconsequential (you can verify that by trying to keep the old implementation in any of the built-in loggers). Therefore, we should not make the category nullable as it's not really helping and we always provide category anyway.

wrappers/src/marshalling.hpp
```diff
@@ -28,6 +28,15 @@
 
 namespace realm::binding {
 
+/// A struct used when marshaling of `MarshaledVector` cannot be
+/// compiled, e.g. for MSVC when returning a `MarshaledVector` from
+/// CPP directly, as compared to when nested within another struct.
+struct AnyMarshaledVector
```
nirinchev, line 34: I'd probably call this `TypeErasedMarshaledVector` - it's a bit more verbose, but also more descriptive for the intended usage.

wrappers/src/marshalling.hpp
```diff
@@ -86,6 +95,12 @@ struct MarshaledVector
 
     iterator begin() const noexcept { return {items}; }
     iterator end() const noexcept { return {items + count}; }
+
+    /// Needed for MSVC when returning a `MarshaledVector` from CPP
+    /// directly, as compared to when nested within another struct.
+    AnyMarshaledVector for_any_type_marshalling() const {
+        return {items, count};
+    }
```
nirinchev, line 103: Note sure this brings in a lot of value vs just creating `AnyMarshaledVector` directly. I feel 

```cpp
REALM_EXPORT AnyMarshaledVector shared_realm_get_log_category_names() {
  // ...
  return { result.data(), result.size() };
}
```

reads better than

```cpp
REALM_EXPORT AnyMarshaledVector shared_realm_get_log_category_names() {
  // ...
  return MarshaledVector(result).for_any_type_marshalling();
}
```

wrappers/src/marshalling.hpp
elle-j: Good idea 🙂

wrappers/src/marshalling.hpp
elle-j: It was mainly to follow the pattern used in e.g. [error_handling.hpp](https://github.com/realm/realm-dotnet/blob/4bc30579d5d11ecfab52cb5452423af4f31fc686/wrappers/src/error_handling.hpp#L75), although that code has additional logic prior to returning the struct.

Realm/Realm/Logging/Logger.cs
```diff
@@ -64,39 +64,73 @@ public abstract class Logger
...
+
         /// <summary>
         /// Gets a <see cref="FunctionLogger"/> that proxies Log calls to the supplied function.
         /// </summary>
         /// <param name="logFunction">Function to proxy log calls to.</param>
         /// <returns>
         /// A <see cref="Logger"/> instance that will invoke <paramref name="logFunction"/> for each message.
         /// </returns>
-        public static Logger Function(Action<LogLevel, string> logFunction) => new FunctionLogger(logFunction);
+        [Obsolete("Use Function(Action<LogLevel, LogCategory, string> logFunction).")]
```
papafe, line 84: Should we suggest the other new method too?

elle-j: There should only have bene one new function introduced here. Are you referring to `Function(Action<string> logFunction)`? If so, that one already existed but moved it up 👍

papafe: My bad 😁

Realm/Realm/Logging/Logger.cs
elle-j: As per our discussion the latest [updates](https://github.com/realm/realm-dotnet/pull/3634/commits/55f074d0624d37de1be54c18d3c7f31b97f8d57c) are now using the `Logger` as the derived class, avoiding a breaking change. (The PR description has also been updated to highlight the reasoning for this.)

Realm/Realm/Logging/Logger.cs
elle-j: Update: Overloads are now being used, and notes on avoiding a breaking change has been added to the PR description.

Realm/Realm/Helpers/Argument.cs
```diff
@@ -95,7 +95,7 @@ public static T ValidateNotNull<T>(T value, string paramName)
 
         public static void AssertDebug(string message)
         {
-            Logger.LogDefault(LogLevel.Error, $"{message} {OpenIssueText}");
+            RealmLogger.LogDefault(LogLevel.Error, $"{message} {OpenIssueText}");
```
nirinchev, line 98: We should settle on either `RealmLogger.LogDefault` or `RealmLogger.Default.Log` - having both is confusing and opens us up for slight inconsistencies.

Realm/Realm/Logging/Logger.cs
```diff
@@ -111,57 +163,89 @@ public static Logger Default
...
+        public void Log(LogLevel level, LogCategory category, string message)
+        {
+            if (level < GetLogLevel(category))
             {
                 return;
             }
 
             try
             {
-                LogImpl(level, message);
```
nirinchev, line 137: This can probably call `CoreLog` after the category check.

nirinchev (approved):

Realm/Realm/Helpers/Argument.cs
elle-j: I agree, I can incorporate that into this PR 👍  Since the `LogDefault` is internal, I'll update the usage to the public-facing `Default.Log()`.

PR #3643, opened 2024-07-09
Add the to-be-generated Flexible Schema models

elle-j:
## Description

Adds to-be-generated Flexible Schema models to the POC.

##  TODO

* [ ] Uncomment the user-defined fields while still allowing the accessors to be defined in the generated models (see PR comment).


Tests/Realm.Tests/Database/FlexibleSchemaPocTests.cs
```diff
@@ -179,16 +228,42 @@ public FlexibleSchemaPocContainer(string containedObjectType)
     // User-defined
     public partial class Dog : IMappedObject
     {
-        public string Name { get; set; }
+        // TODO(lj): Make these user-defined fields work together with
+        //           generated partial class defining the same fields.
```
elle-j, line 232: TODO

PR #3654, opened 2024-07-25
RNET-1168: Make mongodb connection string used for tests configurable

### LINKED ISSUE #3653: Allow configuring mongodb conn string used for integration tests

kmorkos:
## Description

Fixes #3653

##  TODO

* [x] Changelog entry
* [x] Tests


Tests/Realm.Tests/Sync/SyncTestHelpers.cs
```diff
@@ -194,12 +194,8 @@ private static async Task CreateBaasAppsAsync()
             {
                 _baasClient = await BaasClient.Atlas(BaasUri!, differentiator, TestHelpers.Output, cluster, apiKey, privateApiKey, groupId);
             }
-            else
-            {
-                _baasClient = await BaasClient.Docker(BaasUri!, "local", TestHelpers.Output);
-            }
```
kmorkos, line 200: From what I can tell, this client would always match the one initialized by `ExtractBaasSettingsAsync` so avoiding threading the connection string here, but lmk if there's some other workflow that relies on this

kmorkos:
@nirinchev bumping this

Tests/Realm.Tests/Sync/SyncTestHelpers.cs
nirinchev: This is invoked when testing against a local baas docker image, so we should probably bring it back, even if most of the time people test against baasaas.

Issue #3418, opened 2023-08-11
AOT analysis warning when using backlinks on iOS


charlesroddie:
### What happened?

When trying to [build an iOS project with NativeAOT](https://devblogs.microsoft.com/dotnet/announcing-dotnet-8-preview-6/#support-for-targeting-ios-platforms-with-nativeaot) with Realm using backlinks we get AOT analysis warnings.

Previously there were many more warnings but updating to use the source generator cleared up most of them.

### Repro steps

Use a [backlink](https://www.mongodb.com/docs/realm/sdk/dotnet/model-data/relationships/) in a realm object

Compile for iOS with NativeAOT

### Version

11.2.0

### What Atlas Services are you using?

-- select --

### What type of application is this?

Xamarin

### Client OS and version

iOS

### Code snippets

```csharp
[Backlink(nameof(R_GPanel.Tags))]
public IQueryable<R_GPanel> GPanels { get; } = null!;
```

### Relevant log output

```shell
ILC : AOT analysis warning IL3050: ContentRealmTypes.R_Word.GPanels.get: Using member 'System.Linq.Queryable.AsQueryable<R_GPanel>(IEnumerable`1<R_GPanel>)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. Enumerating in-memory collections as IQueryable can require creating new generic types or methods, which requires creating code at runtime. This may not work when AOT compiling. [/Users/runner/work/1/s/Platforms/iOS/50iOS.csproj]
```

nirinchev:
Are you certain that the class that contains `GPanels` is a source-generated class? Asking because those should not be using `AsQueryable` for backlinks getters.

charlesroddie:
Ah sorry some confusion here about what is in what branch - we'll check again.

charlesroddie:
Yes you are right. Sorry for our confusion here. Closing as invalid.

Issue #3417, opened 2023-08-11
[ NativeAot ] Improve support


FoggyFinder:
### Problem

Currently attempt to use [NativeAot](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/) leads to "Unhandled Exception: System.InvalidOperationException: No properties in Item, has linker stripped it?" exception. Expectedly, adding `Preserve` attribute has no effect in this case.

_**MCVE**_ (`RealmAot1`)

csproj:
```xml
<Project Sdk="Microsoft.NET.Sdk">

	<PropertyGroup>
		<OutputType>Exe</OutputType>
		<TargetFramework>net7.0</TargetFramework>
		<ImplicitUsings>enable</ImplicitUsings>
		<Nullable>enable</Nullable>
		<PublishAot>true</PublishAot>
	</PropertyGroup>

	<ItemGroup>
		<PackageReference Include="Realm" Version="11.3.0" />
	</ItemGroup>
</Project>
```
cs:
```cs
using MongoDB.Bson;
using Realms;

var config = new RealmConfiguration("Realm1.realm")
{
    Schema = new[]
    {
        typeof(Item)
    }
};
var realm = Realm.GetInstance(config);
var id = new ObjectId("64bb9508e033d8cfa91e66d3");
var item = realm.Find<Item>(id);
if (item is null)
{
    var testItem = new Item
    {
        Id = id,
        Name = "Do this thing",
        Status = "Open",
        Assignee = "Aimee"
    };
    realm.Write(() =>
    {
        item = realm.Add(testItem);
        Console.WriteLine($"[ Added ] {item.Name} : {item.Status} > {item.Assignee}");
    });
}
else
{
    Console.WriteLine($"{item.Name} : {item.Status} > {item.Assignee}");
}

public partial class Item : IRealmObject
{
    [PrimaryKey]
    [MapTo("_id")]
    public ObjectId Id { get; set; } = ObjectId.GenerateNewId();
    [MapTo("assignee")]
    public string? Assignee { get; set; }
    [MapTo("name")]
    public string? Name { get; set; }
    [MapTo("status")]
    public string? Status { get; set; }
}
```

_**Workaround**_: add the assembly to custom rd

add to csproj
```xml
	<ItemGroup>
		<RdXmlFile Include="Custom.rd.xml" />
	</ItemGroup>
```

Custom.rd.xml

```xml
<Directives>
	<Application>
[...]

PR #3727, opened 2026-09-10
Annotate reflection in Realm.dll for trimming and AOT

charlesroddie:
`Realm.dll` ships with no trimming annotations, so a trimmed or NativeAOT app sees warnings originating in Realm's own reflection. This annotates the reflection that can be described statically and suppresses the sites whose target lives outside the trimmed app closure.

### Changes

- `Argument.Ensure<T>` and `WovenAttribute.HelperType` declare the constructor requirement they rely on. The latter also lets the trimmer preserve the generated object helper that `Realm.CreateRealmObjectMetadata` activates, so this closes a real trimming hazard rather than only silencing a warning.
- `RealmObjectTypeDelegator.GetPropertyImpl` matches the annotation on the `TypeDelegator` method it overrides.
- `SerializationHelper` builds `RealmIntegerSerializer<T>` from a switch over the four supported integer types instead of `MakeGenericType`. This removes an AOT warning as well as a trim warning, and makes the path work under NativeAOT.
- `WovenGetterMethodInfo.Invoke`, the Unity and UWP probes in `InteropConfig`, and the Unity IL2CPP module-initializer fallback in `RealmSchema` are suppressed with justifications. The fallback moves into a named method, because a suppression on a field initializer does not reach the lambda the compiler generates from it.

### Measured effect

A console app with a source-generated model, opening a Realm, writing an object and running a LINQ query, published with `PublishTrimmed`, `TrimMode=full` and `TrimmerSingleWarn=false`:

| | Before | After |
|---|---|---|
| Trim warnings in Realm code | 12 | 3 |
| AOT warnings in Realm code | 8 | 7 |

With the whole assembly rooted via `TrimmerRootAssembly`, warning sites in Realm code drop from 20 to 10. No new warnings appear anywhere in the assembly.

### What is deliberately left

Three warnings remain, in `ObjectSchema.Builder(Type)` and `Property.FromPropertyInfo`. Both reflect over model types to read the generated static `RealmSchema` field and the woven properties. Nothing currently guarantees the trimmer preserves those members, so this is a genuine hazard, not a spurious warning, and it is the same failure the "No properties in class" troubleshooting note describes. Suppressing it would hide that. The fix belongs in the source generator, emitting a `DynamicDependency` so the members survive trimming, which is a larger change and better reviewed on its own.

The dynamic API in `Realms.Dynamic` also still warns. It is reflection-based by nature and wants `RequiresUnreferencedCode` and `RequiresDynamicCode`, which would surface new warnings for every caller. That is a deliberate API decision and is left out of this PR.

Warnings attributed to `MongoDB.Bson` are untouched here. They have a separate cause and a separate fix.

Relates to #3417. That issue reports NativeAOT failing with "No properties in Item, has linker stripped it?", which is the failure behind the two `ObjectSchema.Builder` warnings left above, so this PR does not close it. Worth noting that #3418, closed as completed, describes the backlink AOT warning that `Property.FromPropertyInfo` still produces today.

### Testing

The managed assembly builds clean with no new compiler or StyleCop warnings. I could not run the test suite, because the native wrappers are not built in my checkout. The change is annotation-only apart from the `RealmIntegerSerializer` switch, which covers exactly the four type arguments `RealmInteger<T>` supports and returns null otherwise, matching the previous behaviour for unsupported types.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Realm/Realm/Attributes/WovenAttribute.cs
```diff
@@ -1,4 +1,4 @@
-﻿////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////
```
charlesroddie: spurious diff

charlesroddie: Good catch, thanks. My editing script read with utf-8-sig and wrote plain utf-8, so it silently dropped the byte order mark on five files. Fixed in d4c30c0, and every file in this PR now matches its original BOM state.

The same flaw hit #3728 in both directions, adding a BOM to 18 files that never had one as well as stripping it elsewhere. Fixed there too.

charlesroddie:
@cla-bot check

PR #3728, opened 2026-09-10
Remove BSON serialization, completing Atlas removal and removing trim warnings from MongoDB.Bson

charlesroddie:
A MongoDB.Bson dependency was added in https://github.com/realm/realm-dotnet/commit/7b77c15e8f14fc68333d3fd482e9a0d5468ec314 to support Atlas Device Sync. This had the unfortunate side effect of adding a lot of trim warnings to Realm, so that the publishing process for NativeAOT apps cannot guarantee correctness.

Atlas was removed in https://github.com/realm/realm-dotnet/commit/384b9f2057e85df59914870ebc7670e9dc3ffdac , which "Removed all functionality related to App Services/Atlas Device Sync", but this did not remove the MongoDB.Bson serialization dependencies. So this PR finishes this removal.

### Changes

ObjectId is copied into Realms so that its IConvertible.ToType can drop the conversion to BsonObjectId, the single reference that pulls MongoDB.Bson's entire reflective serializer stack into any app that compiles a LINQ expression tree, which every Realm query does. It is taken from MongoDB.Bson (under Apache 2.0) with the `BsonUtils` and `BsonConstants` helpers inlined and the `IConvertible` conversions to `BsonObjectId` and `BsonString` dropped. Those two conversions are the entry point into the BSON object model.

`Decimal128` is untouched. It has no reference to any BSON document type, and it produces no warnings even alongside expression trees, so it still comes from `MongoDB.Bson`.

The BSON serialization stack is removed: `SerializationHelper`, `RealmObjectSerializer`, `RealmObjectSerializerBase`, and the per-model serializers the source generator emitted, along with their tests and the now-dead BSON assertion helpers in `TestHelpers`. Nine generated files for sync-only model classes are also deleted, since their source classes went with sync and were never cleaned up.

### On removing the serialization API

The changelog entry that introduced it says it added "automatic serialization and deserialization of Realm classes when using methods on `MongoClient.Collection`", and the entry before it introduced `MongoClient.GetCollection<T>`. Both were removed with Atlas Device Sync. The capability appears nowhere in `Docs/`. Nothing in the SDK has called it since sync was removed, and Extended JSON is MongoDB's wire encoding rather than a general interchange format, so callers wanting JSON already reach for `System.Text.Json`, which works on Realm models today.

### Measured effect

A console app with a source-generated model, opening a Realm, writing an object and running a LINQ query, published with `PublishTrimmed` and `TrimMode=full`:

| | Before | After |
|---|---|---|
| MongoDB.Bson warnings | 22 | 0 |
| Total warnings | 34 | 11 |

The 11 that remain are all in Realm's own code and are addressed separately in #3727, which is independent of this one and touches `SerializationHelper.cs`, so whichever merges second needs a trivial rebase.

Relates to #3417. Clearing this noise makes the warnings that matter visible, but it does not fix the failure that issue reports.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


charlesroddie:
@cla-bot check

Realm/Realm/DatabaseTypes/ObjectId.cs
```diff
@@ -0,0 +1,823 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the ObjectId class.
+        /// </summary>
+        /// <param name="timestamp">The timestamp (expressed as a DateTime).</param>
+        /// <param name="machine">The machine hash.</param>
+        /// <param name="pid">The PID.</param>
+        /// <param name="increment">The increment.</param>
+        [Obsolete("This constructor will be removed in a later release.")]
```
charlesroddie: obsolete things can be removed where possible

charlesroddie: Done, all seven are gone: the two four-argument constructors, the `Machine`, `Pid` and `Increment` properties, and `Pack`/`Unpack`. Upstream marks each of them "will be removed in a later release", and none is used inside the struct, by Realm, or by the tests, so the copy starts without them. That is 126 lines.

`GetTimestampFromDateTime` stays, since `GenerateNewId` still needs it, which is also what keeps the inlined `UnixEpoch` and `ToUniversalTime` helpers alive.

Rechecked after the change: Realm and the test projects build clean, all 34 source generator tests pass, and the trim measurement is unchanged at 0 MongoDB.Bson warnings.

